PR #5, opened 2014-07-29
Generic module to compute the stock quantity available to promise

clonedagain:
# Commit 1: Main module + option to exclude incoming deliveries

The "quantity available to promise" is the quantity of a product that you can commit to deliver to a customer.
Depending on the business, it may depend on expected deliveries, manufacturing, expected sales, or many other parameters.
To handle this concept in a modular and configurable way, @lbellier and I add a generic module to compute the stock quantity available to promise, and a 3 pluggable implementations that users can choose from. All implementations can be used at the same time.
## Code trick: function field modularity

By default function fields are not very modular (you need to redefine the whole field to override the method). stock_available takes care of this by making the function fields call the pool instead, so that stock_available_immediately and other future implementations need only override the function _product_available.
## Changes to stock_available_immediately

The existing module stock_available_immediately by @guewen and @sebastienbeau is turned into the first optional implementation.
stock_available_immediately is rewritten to compute "virtual - incoming" instead of "real - outgoing", which should be mostly the same except for rounding.
The field name "immediately_usable_qty" is unchanged for compatibility, but the field is now called "Available to promise" in the views and help texts (this wording seems more widespread).
# Commit 2: option to include products that can be readily manufactured

The module stock_available_mrp takes immediate manufacturing capability into account in the stock quantity available to promise.
We consider a product can be readily manufactured if all its direct components are available.
# Commit 3: option to exclude products already proposed in sale quotations

The module stock_available_sale takes sale quotations into account in the stock quantity available to promise, so that sales persons don't include the same products in 2 quotations.

### DIFF
--- a/stock_available/product.py
+++ b/stock_available/product.py
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    This module is copyright (C) 2014 Numérigraphe SARL. All Rights Reserved.
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm, fields
+import openerp.addons.decimal_precision as dp
+
+
+# Expose the method as a function, like when the fields are defined,
+# and use the pool to call the method from the other modules too.
+def _product_available_fnct(self, cr, uid, ids, field_names=None, arg=False,
+                            context=None):
+    return self.pool['product.product']._product_available(
+        cr, uid, ids, field_names=field_names, arg=arg, context=context)
+
+
+class ProductProduct(orm.Model):
+    """Add a field for the stock available to promise.
+
+    Useful implementations need to be installed through the Settings menu or by
+    installing one of the modules stock_available_*
+    """
+    _inherit = 'product.product'
+
+    def __init__(self, pool, cr):
+        """Use _product_available_fnct to compute all the quantities."""
+        # Doing this lets us change the function and not redefine fields
+        super(ProductProduct, self).__init__(pool, cr)
+        for coldef in self._columns.values():
+            if (isinstance(coldef, fields.function)
+                    and coldef._multi == 'qty_available'):
+                coldef._fnct = _product_available_fnct
+
+    def _product_available(self, cr, uid, ids, field_names=None, arg=False,
+                           context=None):
+        """No-op implementation of the stock available to promise.
+
+        Must be overridden by another module that actually implement
+        computations.
+        The sub-modules MUST call super()._product_available BEFORE their own
+                computations
+
+        Side-effect warning: This method may change the list passed as the
+            field_names parameter, which will then alter the caller's state."""
+        # If we didn't get a field_names list, there's nothing to do
+        if field_names is None:
+            return super(ProductProduct, self)._product_available(
+                cr, uid, ids, field_names=field_names, arg=arg,
+                context=context)
+
+        if context is None:
+            context = {}
+
+        # Load virtual_available if it's not already asked for
+        # We need it to compute immediately_usable_qty
+        # We DO want to change the caller's list so we're NOT going to
+        # work on a copy of field_names.
+        if ('virtual_available' not in field_names
+                and 'immediately_usable_qty' in field_names):
+            field_names.append('virtual_available')
+
+        # Compute the core quantities
+        res = super(ProductProduct, self)._product_available(
+            cr, uid, ids, field_names=field_names, arg=arg, context=context)
+
+        # By default, available to promise = forecasted quantity
+        if ('immediately_usable_qty' in field_names):
+            for stock_qty in res.itervalues():
+                stock_qty['immediately_usable_qty'] = \
+                    stock_qty['virtual_available']
+
+        return res
+
+    _columns = {
+        'immediately_usable_qty': fields.function(
+            _product_available_fnct, multi='qty_available',
+            type='float',
+            digits_compute=dp.get_precision('Product Unit of Measure'),
+            string='Available to promise',
+            help="Stock for this Product that can be safely proposed "
+                 "for sale to Customers.\n"
+                 "The definition of this value can be configured to suit "
+                 "your needs"),
+    }
--- a/stock_available/product.py
+++ b/stock_available/product.py
@@ -56,8 +56,10 @@ class ProductProduct(orm.Model):
         The sub-modules MUST call super()._product_available BEFORE their own
                 computations
 
-        Side-effect warning: This method may change the list passed as the
-            field_names parameter, which will then alter the caller's state."""
+        Side-effect warning: By design, we want to change the behavior of the
+            caller (make it aware that an extra field is being computed).
+            For this, this method MAY change the list passed as the parameter
+            `field_names`."""
         # If we didn't get a field_names list, there's nothing to do
         if field_names is None:
             return super(ProductProduct, self)._product_available(
--- a/stock_available/product.py
+++ b/stock_available/product.py
@@ -43,8 +43,8 @@ class ProductProduct(orm.Model):
         # Doing this lets us change the function and not redefine fields
         super(ProductProduct, self).__init__(pool, cr)
         for coldef in self._columns.values():
-            if (isinstance(coldef, fields.function)
-                    and coldef._multi == 'qty_available'):
+            if (isinstance(coldef, fields.function) and
+                    coldef._multi == 'qty_available'):
                 coldef._fnct = _product_available_fnct
 
     def _product_available(self, cr, uid, ids, field_names=None, arg=False,
@@ -73,8 +73,8 @@ class ProductProduct(orm.Model):
         # We need it to compute immediately_usable_qty
         # We DO want to change the caller's list so we're NOT going to
         # work on a copy of field_names.
-        if ('virtual_available' not in field_names
-                and 'immediately_usable_qty' in field_names):
+        if ('virtual_available' not in field_names and
+                'immediately_usable_qty' in field_names):
             field_names.append('virtual_available')
 
         # Compute the core quantities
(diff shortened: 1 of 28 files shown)


clonedagain:
There is a small problem with the "context trick", some fields are not computed. I'll fix it and rebase the branch but please keep reviewing in the meantime.

clonedagain:
Fixed the context trick. It does have a small side effect (adds values to the caller's field_names) that I documented, and I think it's safe.

guewen:
Wow, impressive. I will try to take some time soon for a review.
Meantime, can you rebase from the 7.0 branch so Travis will check your PR please?

clonedagain:
The tests pass, so it's ready for you to review.
Rebased and re-flake8'ed.
I also fixed a small bug (the sale module was not considering the production capability).

clonedagain:
There was a small bug that became visible because of last fix  (used to product wrong warning messages, now triggered an endless recursion).
It's fixed now and rebased.

gurneyalex:
@clonedagain what is your opinion regarding this PR and #2559: is it blocking? Something we can fix in OCB if official is buggy?

clonedagain:
Fixing this in OCB would not be fair to users who stick to the official branch. Our users were pretty unhappy when the onchange broke, and it was rather hard to track down.

gurneyalex:
OK, thanks (and apologies for the long delay on considering this PR...). 

I've set things as work in progress, and asked for an update on the official addon issue.

clonedagain:
Rebased to avoid conflicts due to recent flake8ification.

clonedagain:
I've split this PR in 2, so now the blocking part is isolated in https://github.com/OCA/stock-logistics-warehouse/pull/25.
Travis build passed (the errors in the log seem unrelated): can you please tag "needs review"?

stock_available/product.py
```diff
@@ -0,0 +1,100 @@
...
+                           context=None):
+        """No-op implementation of the stock available to promise.
+
+        Must be overridden by another module that actually implement
+        computations.
+        The sub-modules MUST call super()._product_available BEFORE their own
+                computations
+
+        Side-effect warning: This method may change the list passed as the
+            field_names parameter, which will then alter the caller's state."""
```
guewen: Is the side-effect intended? Couldn't `field_names` be copied before?

guewen: I have my answer below. Yes it is intended.

stock_available_mrp/product.py
```diff
@@ -0,0 +1,124 @@
...
+
+                    if 'potential_qty' in field_names:
+                        res[product.id]['potential_qty'] += min_qty
+                    if 'immediately_usable_qty' in field_names:
+                        res[product.id]['immediately_usable_qty'] += min_qty
+
+        return res
+
+    def _compute_potential_qty_from_bom(self, cr, uid, bom_id, to_uom,
+                                        context):
```
guewen: Having a keyword argument on the context would be better to respect the convention, btw you call it with the keyword just above

stock_available_mrp/product.py
```diff
@@ -0,0 +1,124 @@
...
+        """Compute the potential qty from BoMs with components available"""
+        bom_obj = self.pool['mrp.bom']
+        uom_obj = self.pool['product.uom']
+        if 'uom' in context:
```
guewen, line 82: If you change the argument to a keyword arg, need to check if context is not None... (but it's correct currently)

stock_available_sale/product.py
```diff
@@ -0,0 +1,232 @@
...
+        """Find the shops matching the current context
+
+        See the helptext for the field quoted_qty for details"""
+        shop_ids = []
+        # Account for one or several locations in the context
+        # Take any shop using any warehouse that has these locations as stock
+        # location
+        if context.get('location', False):
+            # Either a single or multiple locations can be in the context
+            if not isinstance(context['location'], list):
```
guewen: Could it be a tuple? Just asking, I have no idea

guewen:
I am a bit troubled by the side effect on the caller's arguments, but I have no alternative implementation to propose. 
You did an excellent and well built work here, I say :+1: and hope that you will have another positive review.

stock_available_sale/product.py
clonedagain: I didn't come across such a case, but maybe it could be a set(). I could test for "iterable" instead.

clonedagain:
Thanks, I've taken your remarks into account in the latest patch.

clonedagain:
@gdgellatly @pedrobaeza maybe you would be interested in these modules, would you care to review them please? Or else, could you please notify someone who could make a review?

pedrobaeza:
Wow, what a great contribution! It's hard to follow, but indeed it's very well thought. I have only missed one visual hint for the user of which things are being used for computing the stock available to promise when you plug the different modules. Maybe a label next to the field that is built from the concatenation of a string that it's got from each module?

The rest LGTM :+1:

gdgellatly:
Hi,

I haven't fully reviewed the code but looks good.  I'd make the following comments IRT mrp in particular where I have already implemented this functionality (although nowhere near as nicely).

In the function _compute_potential_qty_from_bom I really feel you should use bom_explode.  This will do 3 very important things for you.
1. Allow you to forget about uom problems as bom_explode manages that.
2. Allow for efficiency factors, roundings etc as again bom_explode manages that
3. Reduce calls by batch calling the available functions

I'm not sure if it helps or not but here is my equivalent code for mrp (designed to work with OCB and bom_variant_multi on v7)

``` python
class ProductProduct(orm.Model):
    _inherit = 'product.product'

    # noinspection PyUnusedLocal
    def _able_to_make(self, cr, uid, ids, field_names=None, args=False,
                      context=None):
        if context is None:
            context = {}
        bom_obj = self.pool['mrp.bom']
        res = {prod_id: {'able_to_make_real': 0.0,
                         'able_to_make_virtual': 0.0} for
               prod_id in ids}
        for product in self.browse(cr, uid, ids, context):
            bom_id = bom_obj._bom_find(cr, uid, product.id, product.uom_id.id)
            if bom_id is False:
                continue
            bom_id = bom_obj.browse(cr, uid, bom_id, context)
            bom_product_details = {}
            bom_ctx = context.copy()
            bom_ctx.update({'product_id': product.id, 'prior_product_id': product.id})
            bom_explode_details = bom_obj._bom_explode(
                cr, uid, bom_id, 1, context=bom_ctx)[0]
            for bom_product in bom_explode_details:
                bom_product_details[bom_product.get('product_id')] = bom_product
            bom_stock = self._product_available(cr, uid, bom_product_details.keys(),
                                                field_names=['qty_available', 'virtual_available'],
                                                context=context)
            max_real = []
            max_virt = []
            for bom_product in bom_stock.keys():
                max_real.append(bom_stock[bom_product].get('qty_available') /
                                bom_product_details[bom_product].get('product_qty'))
                max_virt.append(bom_stock[bom_product].get('virtual_available') /
                                bom_product_details[bom_product].get('product_qty'))
            res[product.id] = {'able_to_make_real': min(max_real),
                               'able_to_make_virtual': min(max_virt)}

        return res

    _columns = {
        'able_to_make_real': fields.function(
            _able_to_make, type='float', string='Real Able To Make',
            help='Based on current stock levels the maximum amount that '
                 'could be produced', multi='qty_atm',
            digits_compute=dp.get_precision('Product UoM')),
        'able_to_make_virtual': fields.function(
            _able_to_make, type='float', string='Virtual Able To Make',
            help='Based on predicted stock levels the maximum amount that '
                 'could be produced', multi='qty_atm',
            digits_compute=dp.get_precision('Product UoM')),
    }
```

I will take a closer look next week.

clonedagain:
Thanks @gdgellatly, @lbellier is going to tackle the migration of these modules to v8, I'll ask him to check your suggestion before.

clonedagain:
@gdgellatly your suggestion to use  `bom_explode` is interesting but I'm worried about one point: doesn't it "explode" all the levels of BoM recursively?
For the moment we've taken care to only check the first level of BoM, mainly because delivery date and manufacturing delays are not taken into account.
Our assumption is that if we have semi-finished goods, we can manufacture finished goods "fast". On the contrary, if we have raw materials only and several levels of BoM, it will "probably" be long so we can't promise it to customers.

clonedagain:
@lepistone, @gfcapalbo, would you care to have a look at this PR please?
I read you're migrating stock_available_immediately to v8, and here we're trying to extend it and make it modular.

gdgellatly:
@clonedagain Well the answer to that is largely a matter of assumption. Yes it will explode recursively.
If your assumptions are:

All manufacturers always hold stock of sub assemblies.
All products have sub assemblies that take a long time to manufacture, while the parent product does not.
The business or user only wants stock that is instantly available.

Then of course you are right and I am sure this is true for some manufacturers.

When I wrote my code I was under a different set of requirements.
Sub assemblies are a matter of convenience and built to order (and truthfully pretty rare).
In the specific case the sub assembly typically took less than 10% of the time of the parent or the parent was really just a matching of two subassemblies.
The user primarily wanted to know whether they could fill an order with existing materials without having to purchase any raw materials (typically on a 9 week delay). 

But in any case, this discussion should not block the merge, as it is both hard to elaborate and hard to implement.  Future enhancement, could adequately address both cases, although bom_explode is the method I believe can address both requirements.

That said the bom_explode method does have limitations, particularly when it comes to aggregation errors due to efficiency, uom and rounding as it scales up and down.  In some cases these could be an order of magnitude or more.

clonedagain:
Thanks @gdgellatly, I hope we have an opportunity to improve upon this indeed.

Since you wrote it's not a blocker and @pedrobaeza and @guewen gave a :+1:, can this be merged please? It would help us focus on v8.

When it's merged, I'll be happy to file TODO issues:
- add a feature to explain how the results were computed (so users can know which raw material is missing for example)
- extend the computation to unfold all BoM levels, possibly taking dates into account

lepistone:
Hi,

I didn't look into the details, but on general terms I'd say if it is good (as people say here), even if it can be improved or changed later, I'm happy to merge.

Maybe before merging you could add a known issues / TODO section in the readme or something like that.

Thanks!

clonedagain:
OK @lepistone I'll do that.
EDIT: done, but flake8 still complains, I'm looking into it.

clonedagain:
Fixed, the remaining PEP warnings are in existing modules. OK to merge?

pedrobaeza:
Consider to use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

lepistone:
Almost :smile: I see some warnings which are related to these modules.

clonedagain:
@pedrobaeza i wanted to play the big boy here, but now I realize README files do not replace `description` in v7, right?

pedrobaeza:
Yeah, that's right. In v7, both are needed.

clonedagain:
OK I hope this one is good :-)

pedrobaeza:
Yeah, very good! I merge when Travis finishes.

clonedagain:
Oh well, of course including the OCA template broke pep8 compliance :-) One more try!

pedrobaeza:
There are some flake8 issues, but from other modules, so I merge.

Thanks for the great job, Lionel.

PR #8, opened 2014-09-03
8.0 port module stock_reserve

yvaucher:


pedrobaeza:
Can't you do this with reserved quants?

yvaucher:
@pedrobaeza Here we want to be able to reserve a quantity of products manually and release them.
It also adds an expiration date on those reservations.

Here we want to be able to reserve some products from a specific location for a specific duration.

gurneyalex:
I've given the addon a good shake, added some tests and fixed an issue found on the way

LGTM :+1:

stock_reserve/model/product.py
```diff
@@ -0,0 +1,85 @@
...
+        product_ids = self._get_products()
+        domain = [('product_id', 'in', product_ids),
+                  ('state', 'in', ['draft', 'assigned'])]
+        reservations = StockReservation.search(domain)
+        self.reservation_count = sum(reserv.product_uom_qty
+                                     for reserv in reservations)
+
+    @api.multi
+    def action_view_reservations(self):
+        assert len(self._ids) == 1, "Expected 1 ID, got %r" % self._ids
```
guewen, line 44: Is the assertion necessary with the new api?
Also, if `self._ids` is a tuple, it will fail the string substitution so it should be put in a tuple (I don't know what `self._ids` is supposed to be so do not take my remark into account if it is irrelevant).

stock_reserve/model/product.py
```diff
@@ -0,0 +1,85 @@
...
+        ref = 'stock_reserve.action_stock_reservation_tree'
+        product_ids = self._get_products()
+        action_dict = self._get_act_window_dict(ref)
+        action_dict['domain'] = ("[('product_id','in',[" +
+                                 ','.join(map(str, product_ids)) + "])]")
```
guewen:
```
action_dict['domain'] = [('product_id', 'in', product_ids)]
```

works too

stock_reserve/model/product.py
pedrobaeza, line 44: Please use the official `self.ids` instead.

stock_reserve/model/product.py
```diff
@@ -0,0 +1,85 @@
...
+        product_id = self._ids[0]
+        domain = [('product_id', '=', product_id),
+                  ('state', 'in', ['draft', 'assigned'])]
+        reservations = StockReservation.search(domain)
+        self.reservation_count = sum(reserv.product_uom_qty
+                                     for reserv in reservations)
+
+    @api.multi
+    def action_view_reservations(self):
+        assert len(self._ids) == 1, "Expected 1 ID, got %r" % self._ids
```
guewen, line 75: Same comment than above

stock_reserve/model/product.py
guewen, line 44: in fact shouldn't it be `@api.one`?

pedrobaeza, line 44: Yeah, it can be, although the meaning maybe it's not the same if he wants to limit the operation to one record each time.

yvaucher, line 44: @api.one would return a list of returned results I'll check if it works for an action

pedrobaeza, line 44: Yeah, that's true, so the behaviour changes. It must be this way then

yvaucher, line 44: Just tested it, button action calling a method with decorator @api.one does nothing.

stock_reserve/model/stock_reserve.py
```diff
@@ -0,0 +1,196 @@
...
+        ref_form2 = 'stock.action_move_form2'
+        action = data_obj.xmlid_to_object(ref_form2)
+        action_dict = action.read()
+        action_dict['name'] = _('Reservation Move')
+        # open directly in the form view
+        ref_form = 'stock.view_move_form'
+        view_id = data_obj.xmlid_to_res_id(ref_form)
+        action['views'] = [(view_id, 'form')]
+        action['res_id'] = reserv['move_id']
+        return action
```
yvaucher: wrong object is returned.

Should return action_dict

stock_reserve/model/product.py
gurneyalex, line 44: maybe this should be reported as an odoo bug, don't you think?

nbessi:
Effectivly  it looks like a bug:

```
Decorate a record-style method where `self` is expected to be a 
singleton instance. The decorated method automatically loops on records,
and makes a list with the results. In case the method is decorated 
with  @returns, it concatenates the resulting instances
```

 Maybe action button are explicitly decortated with multi and adding one has no effect. AFAIK there is not mutex test on those decorators.

lepistone:
The discussion on the decorator doesn't seem me to be this module's fault. So :+1:, thanks!

lepistone:
Hi @pedrobaeza do you agree it's a problem in odoo and we can merge that one? thanks a lot!

pedrobaeza:
Yeah, sure. :+1:

lepistone:
thanks!

PR #18, opened 2014-10-20
[FIX] Decimal precision parameter

clonedagain:
Decimal precision parameter for products is 'Product Unit of Measure' in v7

### DIFF
--- a/__unported__/configurable_stock_level/configurable_stock_level/product.py
+++ b/__unported__/configurable_stock_level/configurable_stock_level/product.py
@@ -48,7 +48,7 @@ class product_product(Model):
     
     _columns = {'configurable_stock_level': fields.function(_compute_configurable_level,
                                                             type='float',
-                                                            digits_compute=dp.get_precision('Product UoM'),
+                                                            digits_compute=dp.get_precision('Product Unit of Measure'),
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why fixing an unported module?

clonedagain:
There is a part that is ported: stock_available_immediately/product.py
The fix dates back to before the module was marked "unported": https://code.launchpad.net/~numerigraphe-team/stock-logistic-warehouse/7.0-fix-decimal-precision/+merge/223881
I hadn't managed to push it to github until today.

clonedagain:
Looks like this fails on flake8 for an unchanged file.

pedrobaeza:
OK, we can patch files waiting for a migration.

:+1:

pedrobaeza:
That is unrelated, but if you want, go ahead to fix the branch.

clonedagain:
Sorry @pedrobaeza I thought it was related, unfortunately I won't have time to flakeify the branch.

pedrobaeza:
It's only one error, because the branch was already greeinify, but now there is another condition that it's being checked, but don't worry, we can do it in another PR.

lmignon:
:+1:  LGTM. Thank you

PR #19, opened 2014-10-20
Record product valuation history

clonedagain:
In v7 the product cost prices are not kept in a history, so this module simply proposes a wizard to let users record them at key moments (ie. at annual inventory), or on regular basis using a "cron" job.

### DIFF
--- a/stock_valuation_history/__init__.py
+++ b/stock_valuation_history/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    This module is copyright (C) 2013 Numérigraphe SARL. All Rights Reserved.
+#
+#    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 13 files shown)


clonedagain:
@jgrandguillaume and @yvaucher, maybe this overlaps with https://github.com/OCA/margin-analysis/tree/7.0/product_price_history: would you be kind enough to review this and share your opinion please?

yvaucher:
@clonedagain Sorry for the late reply. indeed I think it has the same purpose

You also have a wizard to show inventory valuation in https://github.com/OCA/margin-analysis/blob/7.0/product_price_history/wizard/historic_prices.py

So I'm :-1: for this PR 

PR are welcome on product_price_history

clonedagain:
No big deal, we need to reconsider this anyway as we're migrating to v8.
Thanks for your review.

PR #23, opened 2014-11-27
[ADD] stock skid

max3903:


stock_skid/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+    'website': 'http://www.savoirfairelinux.com',
+    'depends': [
+        'product',
+        'stock',
+        'mrp',
+    ],
+    'category': 'Warehouse',
+    'description': """
+This module allow you to manage skids in your warehouse including
+partial ones.
```
pedrobaeza, line 44: An explanation about what is a skid would be good (which one of this: http://en.wikipedia.org/wiki/Skid?)

pedrobaeza, line 44: Seeing code, it seems to be only the management of skid labels, doesn't it?

stock_skid/report/skid_label.py
```diff
@@ -0,0 +1,111 @@
...
+
+import time
+from collections import namedtuple
+
+from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
+from openerp.report import report_sxw
+
+
+class SkidLabel(namedtuple("Label", "wizard label_data")):
+    @property
```
clonedagain, line 28: What does `@property` do in v7?

clonedagain:
Oh! skids are platforms alternative to pallets right? http://recycling.about.com/od/Pallet_Recycling_Glossary/ss/What-Is-A-Skid.htm

Then @max3903 would you please provide a proper README with more details on what this module does and how it's used in your warehouse?
We also have a module to print stock labels that we stick on pallets, so we may be able to make a propre review, and maybe converge to common code.

PR #25, opened 2014-11-28
[IMP] Warn WRT stock available to promise

clonedagain:
Hook into sale_stock to make it warn salespersons when the stock available
to promise is insufficient.
This is an extension of https://github.com/OCA/stock-logistics-warehouse/pull/5 .
## Code trick: context to override the virtual stock

The new module also has a context trick that lets it return the value of the stock available to promise in the field "virtual stock". This lets future sub-modules change the behavior of modules that were written to use the virtual stock, to use the new value instead with minimal impact.
## Merge blockers

This cannot be merged just yet because:
- [x] https://github.com/odoo/odoo/issues/2559 must be fixed
- [ ] I must add a test to check whether this actually works (must fail now, must succeed when previous bug is fixed in Odoo)


clonedagain:
Please tag this PR "blocked" until odoo/odoo#2559 is fixed

clonedagain:
The blocking bug seems fixed, we still need to prove it and then we can merge.

clonedagain:
I've just read about https://github.com/OCA/sale-workflow/tree/7.0/sale_exceptions, maybe we should remix this patch to reuse this framework.
I'm undecided, opinions welcome.

gurneyalex:
@clonedagain I agree that sale_exceptions is a good way to achieve what you want.

clonedagain:
OK closing then, I'll look into it again this and sale_exceptions this summer as we migrate to v8.
Thanks for the advice.

clonedagain:
For the records, it's indeed very easy to implement using sale_exceptions, and directions have been added in https://github.com/OCA/stock-logistics-warehouse/pull/81

PR #28, opened 2014-12-12
Add new modules for simplified transfer of goods between stock locations...

codingforfun:
Add two new modules to ease creation of pickings and corresponding stock moves for transfer of goods between stock locations.
1. Wizard to select source and destination module and goods to transfer
2. Customization: Model the 
transfer of goods via a dedicated transit stock location (ship, truck etc.). This creates 2 separate pickings per transfer. One to move goods from source location to transit location and another one to receive goods from transit location into destination.

bildschirmfoto vom 2014-12-12 17 40 37


stock_internal_transfer/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{
+    'name': 'Stock internal transfer',
+    'version': '0.1',
```
pedrobaeza: You can put 1.0

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,150 @@
...
+##############################################################################
+
+
+from openerp.osv import orm, fields
+from openerp.netsvc import LocalService
+import openerp.addons.decimal_precision as dp
+from openerp.tools.translate import _
+
+
+class stock_internal_transfer_line(orm.TransientModel):
```
pedrobaeza: Use CamelCase naming in classes.

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,150 @@
...
+        'product_id': fields.many2one('product.product', string="Product", required=True, ondelete='CASCADE'),
+        'quantity': fields.float("Quantity", digits_compute=dp.get_precision('Product Unit of Measure'), required=True),
+        'product_uom': fields.many2one('product.uom', 'Unit of Measure', required=True, ondelete='CASCADE'),
+        'wizard_id': fields.many2one('stock.internal.transfer', string="Wizard", ondelete='CASCADE'),
+    }
+
+    def onchange_product_id(self, cr, uid, ids, product_id, context=None):
+        uom_id = False
+        if product_id:
+            product = self.pool.get('product.product').browse(cr, uid, product_id, context=context)
```
pedrobaeza: You can use `self.pool['product.product']`

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,150 @@
...
+            'location_dest_id': wiz.location_dest_id.id,
+            'partner_id': wiz.delivery_partner_id.id,
+            # we require all products to be available for shipment
+            'move_type': 'one',
+        }
+        return picking_data
+
+    def _create_move_lines(self, cr, uid, ids, picking_id, context=None):
+        assert len(ids) == 1, 'This function should only be used for a single id at a time.'
+        move_obj = self.pool.get('stock.move')
```
pedrobaeza: Use self.pool[]

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,150 @@
...
+            move_data = {
+                'product_id': line.product_id.id,
+                'product_qty': line.quantity,
+                'product_uom': line.product_uom.id,
+                'picking_id': picking.id,
+                'location_id': picking.location_id.id,
+                'location_dest_id': picking.location_dest_id.id,
+                'name': line.product_id.name_get()[0][1],
+            }
+            move_obj.create(cr, uid, move_data)
```
pedrobaeza: Propagate context.

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,150 @@
...
+
+        picking_data = self._prepare_picking(cr, uid, ids, context=context)
+        picking_id = picking_obj.create(cr, uid, picking_data)
+        self._create_move_lines(cr, uid, ids, picking_id, context=context)
+
+        wf_service = LocalService("workflow")
+        wf_service.trg_validate(uid, 'stock.picking',
+                                picking_id, 'button_confirm', cr)
+        picking_id = [picking_id]
+        picking_obj.action_assign(cr, uid, picking_id)
```
pedrobaeza, line 129: Propagate context.

pedrobaeza:
Very interesting module!

Please take care of the flake8 and pylint errors.

OSguard:
Test fails according rounding bug.

codingforfun:
The OCA CLA issues has been already discussed [here](https://github.com/OCA/web/pull/117) and is has been concluded that the Initos company CLA applies to me.

stock_internal_transfer_transit/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{
+    'name': 'Stock internal transfer transit',
+    'version': '1.0',
+    'category': '',
+    'summary': "Process internal transfers using a dedicated "
+               "transfer stock location.",
```
laetitia-gangloff: It seems to long to get a right display in the interface :

image

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,171 @@
...
+
+    _columns = {
+        'location_id': fields.many2one(
+            'stock.location', string="Location", required=True,
+            ondelete='CASCADE', domain=[('usage', '<>', 'view')]),
+        'location_dest_id': fields.many2one(
+            'stock.location', string="Dest. Location", required=True,
+            ondelete='CASCADE', domain=[('usage', '<>', 'view')]),
+        'delivery_partner_id': fields.many2one(
+            'res.partner', string="Delivery Address", required=True),
```
laetitia-gangloff: Why delivery_partner_id is required ?
In stock.picking, partner_id is not required.

stock_internal_transfer/stock_internal_transfer.py
```diff
@@ -0,0 +1,171 @@
...
+
+        picking_data = self._prepare_picking(cr, uid, ids, context=context)
+        picking_id = picking_obj.create(cr, uid, picking_data, context=context)
+        self._create_move_lines(cr, uid, ids, picking_id, context=context)
+
+        wf_service = LocalService("workflow")
+        wf_service.trg_validate(uid, 'stock.picking',
+                                picking_id, 'button_confirm', cr)
+        picking_id = [picking_id]
+        picking_obj.action_assign(cr, uid, picking_id, context=context)
```
laetitia-gangloff: I get this error message : TypeError: action_assign() got an unexpected keyword argument 'context'
And if I look action_assign, it seems the definition is : https://github.com/odoo/odoo/blob/7.0/addons/stock/stock.py#L768 ->  def action_assign(self, cr, uid, ids, *args)

so I think it is not possible to set **kwargs.

stock_internal_transfer_transit/stock_internal_transfer.py
```diff
@@ -0,0 +1,112 @@
...
+            context.get('stock_internal_transfer_no_transit', False)
+        if wiz.mode == 'force' or no_transit:
+            return picking_data
+
+        # update destination location to transfer using
+        # intermediate transit stock location
+        model_data_obj = self.pool['ir.model.data']
+        stock_loc_transit_id = model_data_obj.\
+            get_object_reference(cr, uid, 'stock_internal_transfer_transit',
+                                 'stock_location_transit')[1]
```
laetitia-gangloff, line 45: It doesn't work in multi-company context !

stock_internal_transfer_transit/stock_internal_transfer.py
```diff
@@ -0,0 +1,112 @@
...
+    def _prepare_picking(self, cr, uid, ids, context=None):
+        assert len(ids) == 1,\
+            'This function should only be used for a single id at a time.'
+
+        picking_data = super(StockInternalTransfer, self).\
+            _prepare_picking(cr, uid, ids, context=context)
+        wiz = self.browse(cr, uid, ids[0], context=context)
+
+        no_transit = context and \
+            context.get('stock_internal_transfer_no_transit', False)
```
laetitia-gangloff: I didn't see the set of this context.
Maybe, the transit location, can be set as an optional parameter ?
If the transit location is set, you create 2 pickings, else, you create only on picking.

What do you think ?

stock_internal_transfer_transit/stock_internal_transfer.py
```diff
@@ -0,0 +1,112 @@
...
+
+        picking_data.update({'location_dest_id': stock_loc_transit_id, })
+        return picking_data
+
+    def _show_picking(self, picking_ids):
+        act_data = super(StockInternalTransfer, self).\
+            _show_picking(picking_ids)
+        act_data.update({
+            'domain': [('id', 'in', picking_ids)],
+            'view_mode': 'tree,form',
```
laetitia-gangloff: I think, it should depends of the length of picking_ids.

stock_internal_transfer_transit/stock_internal_transfer.py
```diff
@@ -0,0 +1,112 @@
...
+                # chain moves
+                move.write({
+                    'move_dest_id': new_id,
+                    'move_history_ids': [(4, new_id)]
+                })
+
+            wf_service = LocalService("workflow")
+            wf_service.trg_validate(uid, 'stock.picking',
+                                    transit2dst_id, 'button_confirm', cr)
+            # picking_model.action_assign(cr, uid, [transit2dst_id])
```
laetitia-gangloff: If it is in comment, maybe can you remove it ?

laetitia-gangloff:
Nice modules !

gurneyalex:
I'm sorry I missed this PR when I worked on `stock_route_transit` (https://github.com/OCA/stock-logistics-transport/tree/8.0/stock_route_transit). `stock_route_transit` is made for v8 but shares the concept of a transit location with these addons and the focus is different (internal transfer here vs. supplier / customer transfer).

stock_internal_transfer/__openerp__.py
```diff
@@ -0,0 +1,52 @@
...
+    'version': '1.0',
+    'category': '',
+    'summary': 'Wizard to simplify creation of internal stock transfers',
+    'description': """
+Wizard to simplify creation of internal transfers between stock locations.
+Creates stock moves and pickings to transfer selected goods between two
+stock locations.
+""",
+    'author': 'initOS GmbH & Co. KG',
+    'website': 'http://www.initos.com',
```
gurneyalex: Could you please add "Odoo Community Association (OCA)" to the authors?

stock_internal_transfer_transit/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+    'description': """
+Customize wizard from 'stock_internal_transfer' module to process internal
+transfers between stock locations using a dedicated 'transit' stock location.
+Each transfer creates two pickings and corresponding moves.
+First picking is to move goods from their source stock location into the
+transit stock location (truck, ship etc) for transport.
+Second picking is to move goods from transit stock location to
+destination stock location on arrival.
+""",
+    'author': 'initOS GmbH & Co. KG',
```
gurneyalex: could you please add 'Odoo Community Association (OCA)' to the list of authors?

stock_internal_transfer_transit/stock_internal_transfer.py
gurneyalex: I agree, you should add:

    if context is None:
        context = {} 

at the begining of the method.

OSguard:
should we add to the readme that you check "stock_route_transit" before migrating this one?

stock_internal_transfer_transit/stock_internal_transfer.py
laetitia-gangloff: @gurneyalex sorry for the mistake, my comment was about : stock_internal_transfer_no_transit.
I can not see the use of it in the rest of these modules.

gurneyalex:
@OSguard would be a good idea, yes. Do you think it would make sense to move this addon to the 7.0 branch of OCA/stock-logistics-transport ?

OSguard:
yes, but we should fix the code here and move after. The stock-logistics-transport did not exist as we create the pull request

stock_internal_transfer/stock_internal_transfer.py
codingforfun, line 129: Which one you mean? 'picking_obj.action_assign()' ? 
As far as I can see it is defined as 'def action_assign(self, cr, uid, ids, *args)'. The 'args' parameter seams to be unused and it's unclear from the code what would be expected here. Or did I miss something?

stock_internal_transfer_transit/stock_internal_transfer.py
gurneyalex: my bad, you're correct. Could not find it either.

stock_internal_transfer_transit/stock_internal_transfer.py
codingforfun, line 45: Ok. I'm not experienced with multi-company setup. Why do you think it won't work and would be the problem?
Of course it uses the same transfer location for all companies in this case but I think the basic behaviour should work. Any idea how I could solve this in a more clean way i.e. on a by company basis?

stock_internal_transfer_transit/stock_internal_transfer.py
codingforfun: good point

stock_internal_transfer_transit/stock_internal_transfer.py
laetitia-gangloff, line 45: You define stock_location_transit as a child of stock_location_company. And on stock_location_company there is a company_id defined. And on your location there is also a company_id defined (by the _default).

I test it, in a multi-company environment, and I get access error :

Access Denied
The requested operation cannot be completed due to security restrictions. Please contact your system administrator.
(Document type: Location, Operation: read)

And it is normal, there is some record rules.

A solution is maybe, to remove the company on the record stock_location_transit,  I am not sure it is the best solution but it works.

stock_internal_transfer/stock_internal_transfer.py
codingforfun: The initial idea here was to have the final destination on all delivery slips. E.g. in the transit version or other inherited versions of this module the intermediate locations could have completly different address or even no address being a truck or ship. But you are right it should get set internally automatically and shouldn't be required. It shouldn't be in the wizard at all I think. 
I will remove this from the wizard and get the partner from the destination location.

laetitia-gangloff: I think it can stay in the wizard, maybe just not mandatory ?

stock_internal_transfer_transit/stock_internal_transfer.py
codingforfun, line 45: Ok. I changed it like you suggested. This is an acceptable fix I think. 
To do it right every company would need to have its own transit location and goods should get transferred to the transit location belonging to the source company to reflect that goods are still owned by the sending company normally during transfer. 
But this would need more complex changes to create a proper transit location for every company during install of the module or on company creation.

stock_internal_transfer/stock_internal_transfer.py
pedrobaeza, line 129: You're right. You can leave it as is.

codingforfun:
Ok. I think I fixed all open issues. Anything I missed? 
Next steps would be: 
- Move code to `stock-logistics-transport`, open a new MR there and close this one
- Add note to `README.md` about checking `stock_route_transit` before migrating this one 

Right?

jgrandguillaume:
Hi,

I would say yes :)

regards,

laetitia-gangloff:
It is ok for me !
Thank you :)

yvaucher:
@codingforfun Commits are not linked to your github account have you configured author's email on your local config ?

codingforfun:
I opened a new pull request https://github.com/OCA/stock-logistics-transport/pull/15 according to the previous discussion.

dek-odoo:
Hello Guys, is the module **stock-internal-transfer** dropped for future versions or merged with **stock_route_transit** ?

Thanks.

@OSguard @codingforfun @pedrobaeza

PR #33, opened 2015-02-13
Stock available immediately2

gfcapalbo:
Port to 8.0


stock_available_immediately/product.py
```diff
@@ -26,11 +26,9 @@
 
 class product_immediately_usable(orm.Model):
```
lepistone: please rename the class to `Product` (or `ProductTemplate`)

stock_available_immediately/product.py
```diff
@@ -55,7 +53,7 @@ def _product_available(self, cr, uid, ids, field_names=None,
         if 'immediately_usable_qty' in field_names:
             for product_id, stock_qty in res.iteritems():
                 res[product_id]['immediately_usable_qty'] = \
-                    stock_qty['qty_available'] + stock_qty['outgoing_qty']
+                    stock_qty['qty_available'] - stock_qty['outgoing_qty']
```
lepistone: This seems correct, but it scares me. Was it this wrong on the original 7.0 version?

lepistone:
Another remark: the modification of `field_names` with the long commentary seems unnecessary: I gave a look to the original `_product_available` method in the `stock` module and it doesn't use `field_names` at all.

lepistone:
The first commit has been done with an incorrect author. That can complicate logs and attribution. Can you fix that with an interactive rebase and `git commit --amend --author` ? (`git filter-branch` should also work). Thanks!

stock_reserve/view/product.xml
```diff
@@ -1,25 +1,10 @@
...
-            <field name="arch" type="xml">
-                <xpath expr="//div[@name='buttons']" position="inside">
-                    <button class="oe_inline oe_stat_button" name="action_view_reservations"
-                        type="object" groups="base.group_sale_salesman" icon="fa-lock">
-                        <field string="Stock Reservations" name="reservation_count" widget="statinfo" />
-                    </button>
-                </xpath>
-            </field>
-        </record>
-
```
StefanRijnhart, line 18: As this stanza was deleted, does this mean that you can not inspect the number of reserved product variants?

StefanRijnhart, line 18: In fact, maybe the changes to stock_reserve are unintended as it was already ported in revision c6d897dadbe208c774b9274e568b003c713b950c and the rest of this branch concerns another module?

stock_available_immediately/product.py
gfcapalbo: In 8.0 outgoing qty is calculated as an absolute quantity.
https://github.com/odoo/odoo/blob/8.0/addons/stock/product.py#L380-L381
https://github.com/odoo/odoo/blob/8.0/addons/stock/product.py#L112-L152

In 7.0 that wasn't the case amounts are subtracted, and returned as a negative value:
https://github.com/odoo/odoo/blob/7.0/addons/stock/product.py#L334-338

therefore our interpretation of immediately_usable_qty needed to be changed to keep uniformity in user experience between versions.

stock_reserve/view/product.xml
gfcapalbo, line 18: it's been made useless by the modified:
https://github.com/Therp/stock-logistics-warehouse/blob/stock_available_immediately2/stock_reserve/view/product.xml#L4-L16

that serves the same function.

StefanRijnhart, line 18: Thanks for the clarification. Can you please move this change to a separate PR so that we can deal with both issues separately?

gfcapalbo:
Pull request closed, moved to #35, sorry for the inconvenience.

gfcapalbo:
nevermind , we will continue on #33, the original pull request.

clonedagain:
Why is there a change in stock_reserve?

StefanRijnhart:
Setting to Needs fixing as per @clonedagain's comment given the earlier discussion about changes in stock_reserve.

gfcapalbo:
Stock_reserve modifications have been reverted.

clonedagain:
@gfcapalbo I'll try to push an updated branch for you to merge, to match the changes committed yesterday in v7.
I'll to it this morning if I can.

clonedagain:
Easier said than done...
The cleanest way would have been to merge the latest v7 commit in your branch, but `7.0` has never been forward-ported to `8.0` yet.
I'll push a branch with just the matching patch. I'm sure we can do better but what?

stock_available_immediately/product.py
```diff
@@ -20,42 +20,24 @@
...
 
-class product_immediately_usable(orm.Model):
+class ProductTemplate(orm.Model):
     """
-    Inherit Product in order to add an "immediately usable quantity"
-    stock field
     Immediately usable quantity is : real stock - outgoing qty
     """
-    _inherit = 'product.product'
+    _inherit = 'product.template'
```
clonedagain: Why is it `product.template` now?

stock_available_immediately/product.py
clonedagain: In the newest v7 this was changed to available immediately = available - incomming.
I hope it's ok for you?

clonedagain:
@gfcapalbo, I've pushed https://github.com/Therp/stock-logistics-warehouse/pull/1 to help you integrate the latest changes from v7. Please review it and merge it into your branch.

stock_available_immediately/product.py
gfcapalbo: Not strictly necessary, following the tendency of odoo to put most product.product fields in the template.

StefanRijnhart:
@clonedagain It is probably too late now, but I don't understand why this PR should include stock_available_mrp and stock_available_sale now all of a sudden. What is the reason? I would like to see PRs to deal with as little unrelated changes as possible.

StefanRijnhart:
I don't think that inheriting product.template instead of product.product is cutting it for this module's migration to 8.0. I am testing with a 'product' (i.e. product.template) with two 'variants' (i.e. product.product). There are no incoming or outgoing goods. I'm setting the stock of one variant to two items on stock, the other to three. This results in an immediately available quantity of 5 on both!

This confusion seems to be caused by counting products instead of variants. If you look at addons/stock/product.py, you see that fields like outgoing_qty are now defined both on product.template as on product.product, each with their own function. You need to do the same for these modules.

clonedagain:
@StefanRijnhart for consistency I've cherry-picked all the commits of the (admittedly massive) PR merged in v7, that's why *_mrp and *_sale are included.
I thought it would be difficult to bring in some modules now and others later, given that all impact the config file of the generic module "stock_available".

Maybe it would be simpler if we could just merge 7.0 into 8.0 once in a while?
If others are willing to help with this, I could contribute PRs to review the conflict resolutions.

StefanRijnhart:
@clonedagain thanks for clarifying. I guess I can still easily review them by diffing each of these modules with the 7.0 branch.

Forward merging would be really nice. Unfortunately, because of $#@!^ git's lack of support for tracking moved files and directories, merging into a repo that still keeps modules in **unported** will probably be too much work.

gfcapalbo:
Pushed review after @StefanRijnhart  observation about product.template, it seems that now  that _immediately_usable_qty is computed correctly for product.template and product.product/variants.

StefanRijnhart:
Thanks. Tests are failing as indicated by Travis CI but these are in unrelated modules. 

However, there are still some leftovers from the merge-in of the 7.0 branch by @clonedagain: both stock_available_sale and stock_available_mrp appear in **unported** as well as in the repo root. The latter contain only the README.rsts. @clonedagain, is this on purpose maybe related to the recent **unported** discussion or do these files need to be moved?
There is another test that fails. It is in stock_available_sale.

clonedagain:
Not on purpose, I must have failed somewhere.
I'll try to make this right tomorrow if no-one can fix it before.

gfcapalbo:
@clonedagain I'll do it now.

gfcapalbo:
ready for review.

gurneyalex:
:+1: although a small automated test would be very welcome.

gurneyalex:
@clonedagain could  you rebase the branch to trigger a runbot build?

clonedagain:
@gurneyalex sorry I can't, I don't own this branch. @gfcapalbo can you do it please?

gurneyalex:
@clonedagain and @gfcapalbo I started #57 to finalize this PR -> runbot will be available there.

StefanRijnhart:
@gurneyalex we are working today on tests for this module. Can you give us another day?

gurneyalex:
@StefanRijnhart @gfcapalbo sure, I'm always happy to not do work. And I've fixed a bug which was preventing the runbot from starting builds this morning, so we should be shortly green.

Thanks for your time.

gurneyalex:
:+1:

stock_available_immediately/tests/test_stock_available_immediately.py
```diff
@@ -0,0 +1,120 @@
...
+#
+##############################################################################
+from openerp.tests.common import TransactionCase
+
+
+class testStockLogisticsWarehouse(TransactionCase):
+
+    def test01_stock_levels(self):
+        """checking that immediately_usable_qty actually reflects \
+the variatins in stock, both on product and template"""
```
StefanRijnhart: Typo in 'variations'

gfcapalbo:
Fixed typo.

StefanRijnhart:
:+1:

stock_available/product.py
```diff
@@ -0,0 +1,66 @@
+#-*- coding: utf-8 -*-
```
gurneyalex, line 1: adding a space after `#` will make flake8 happy and travis :green_heart: :

```
./stock_available/product.py:1:1: E265 block comment should start with '# '
```

PR #37, opened 2015-03-01
[IMP][7.0] stock available reserved quantity

llacroix:


stock_available_reserved_quantity/models/stock_production_lot.py
```diff
@@ -0,0 +1,60 @@
...
+        for lot in self.browse(cr, uid, ids, context=context):
+            if lot.retained_stock < 0:
+                return False
+
+        return True
+
+    _inherit = "stock.production.lot"
+
+    _columns = {
+        'retained_stock': fields.float(string='Retained stock'),
```
joaoalf: @llacroix For float fields it is good to add a precision.

llacroix: Which precision should we add?

joaoalf: @llacroix 
Ex: 

``` python
import openerp.addons.decimal_precision as dp
...
'product_qty': fields.float('Quantity', digits_compute=dp.get_precision('Product Unit of Measure'), ...)
```

llacroix:
It would be nice to see code coverage. Any idea what happened? I don't see coverage on PRs anymore.

yvaucher:
I replayed last green travis jobs to see if we got coverage

jgrandguillaume:
Hi,

Could I ask you to give a bit more explanation in the manifest ? I do not understand the exact features and purpose of this work, what is "Retained stock" ?

Thanks for your clarification.

Regards,

llacroix:
@jgrandguillaume I'll update the manifest later today.

clonedagain:
Hi, I hadn't noticed your PR but the idea is interesting in v7 : set a quantity on each lot and report the total reserved.

Indeed you must give more explanation in the `README`.

I'd also like you to integrate with the module  `stock_available`,which is a generic module to compute the stock available for sale : 
- add an option to install you module in  `stock_available`'s config
- inherit the compute method of the field "stock_available_immediately" to subtract the reserved quantity from the quantity available for sale.
  You can use `stock_available_immediately`, `stock_available_sale` and `stock_available_mrp` as examples.

PR #43, opened 2015-03-17
fix #42: reserve location should be outside WH

### LINKED ISSUE #42: tests of stock_reserve_sale fail with recent odoo/odoo. it works with older odoo
I'm investigating this now.

lepistone:
This has always been wrong on v8, but because of the very obscure
odoo/odoo#5797, this seemed to work every time we had single-step
reception.

Please note that this block of XML is noupdate, so for existing
installations you need to change the parent of the reservation location
to something outside your warehouses manually.

Incidentally, this makes the branch green independently of odoo/odoo#5797.

### DIFF
--- a/stock_reserve/__openerp__.py
+++ b/stock_reserve/__openerp__.py
@@ -21,7 +21,7 @@
 
 {'name': 'Stock Reservation',
  'summary': 'Stock reservations on products',
- 'version': '0.1',
+ 'version': '0.2',
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'category': 'Warehouse',
  'license': 'AGPL-3',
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

guewen:
:+1:

StefanRijnhart:
Could you write the simple migration script to fix the data if it still refers to stock.stock_location_company?

StefanRijnhart:
Thanks. Note that the combination of name and model in ir_model_data is not necessarily unique. Can you replace the model criterium by a select by module?

lepistone:
@StefanRijnhart done.

StefanRijnhart:
Thanks! :+1:

PR #45, opened 2015-04-15
[7.0] quantity warning on lot with negative number

llacroix:


quantity_warn_on_lot/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+        'stock',
+    ],
+    'external_dependencies': {
+        'python': [],
+    },
+    'data': [
+        'views/stock_move_split.xml',
+    ],
+    'installable': True,
+    'application': True,
```
joaoalf: @llacroix I don't think that this module qualify as an application.

quantity_warn_on_lot/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
```
joaoalf, line 1: @llacroix Copyright?

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,185 @@
+# -*- coding: utf-8 -*-
```
joaoalf, line 1: @llacroix Copyright?

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,185 @@
...
+        catch this error and instead open a new view that requires
+        confirmation to continue the split or keep editing.
+
+        Before returning this method, we have to rollback changes to
+        be certain that none of the change made before raising the
+        error will get saved to the database.
+        """
+        base_func = super(StockMoveSplit, self).split_lot
+
+        try:
```
joaoalf: @llacroix Why not use a orm_except and let odoo handle the roolback?

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,185 @@
...
+        :params product_id: The product in question
+        :params prodlot_id: The prodlot in question
+        :params quantity: The quantity of products available
+        """
+        self.product_id = product_id
+        self.prodlot_id = prodlot_id
+        self.quantity = quantity
+
+
+class StockMoveSplit(orm.TransientModel):
```
joaoalf, line 30: @llacroix In version 7.0 is more common to put wizards (models and views) in the wizard namespace (folder).

quantity_warn_on_lot/__openerp__.py
llacroix: Good point, fixed

quantity_warn_on_lot/models/__init__.py
llacroix, line 1: Done

quantity_warn_on_lot/models/stock_move_split.py
llacroix, line 1: Done

quantity_warn_on_lot/models/stock_move_split.py
llacroix: It's not an orm exception. I had in mind to do that at first but this will cause other problems. If a real orm exception is raised. It will get caught there and it's possible that the other exceptions will get silenced by a weird wizard that pops up.

In this case, we want to handle the exception without raising errors.

quantity_warn_on_lot/models/stock_move_split.py
llacroix, line 30: I'll try to move that in a different folder with the according views.

quantity_warn_on_lot/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
```
joaoalf: Copyright?

quantity_warn_on_lot/models/stock_move_split.py
joaoalf: @llacroix I'm ok with the exception part. The thing is that resort to manual transaction control usually means that the developper is trying to accomplish something in a non-odoo way. maybe what you need is a validation step (a 'verify stock' button or something like that) to check if the split can be performed before trying to perform it.

llacroix: @joaoalf I understand. The goal of the module is to add a check before the split. Which means that the user has to click on the split and if something wrong happens, it will show the popup. Otherwise it does as it always did. Though, as you say, it seems a bit strange and the wizard part isn't very elegant though. The other problem is that it is tracking which button are pressed.

A validation button wouldn't speak as much as a dialog box saying "You're doing something wrong. But you can keep going with it or correct it." That said, the check is just a warning and doesn't override the real validation that the split button does.

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,73 @@
...
+        if context is None:
+            context = {}
+        base_func = super(StockMoveSplit, self).split
+        res = base_func(cr, uid, ids, move_ids, context)
+        move_obj = self.pool.get('stock.move')
+        uom_obj = self.pool.get('product.uom')
+        product_obj = self.pool.get('product.product')
+        for data in self.browse(cr, uid, ids, context=context):
+            for move in move_obj.browse(cr, uid, move_ids, context=context):
+                product_uom = product_obj.browse(
```
joaoalf: @ehdem Why not `move.product_uom` ?

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,71 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import logging
+
+from openerp.osv import orm, osv
```
joaoalf: @ehdem Why import osv?

quantity_warn_on_lot/models/stock_move_split.py
```diff
@@ -0,0 +1,71 @@
...
+                for line in lines:
+                    quantity = line.quantity
+                    total_move_qty += quantity
+                    amount_actual = uom_obj._compute_qty_obj(
+                        cr, uid, move.product_uom,
+                        line.prodlot_id.stock_available,
+                        product_uom, context=context)
+                    if (data.location_id.usage == 'internal') and \
+                            (quantity > (amount_actual or 0.0)):
+                        raise osv.except_osv(
```
joaoalf: @ehdem Why not `raise orm.except_orm()`?

yvaucher:
To rebase

.travis.yml
```diff
@@ -1,3 +1,4 @@
+sudo: required
```
gurneyalex, line 1: to remove.

Cedric-Pigeon:
@llacroix Hello, do you plan to achieve this PR ? Is it still usefull ? Thanks.

llacroix:
@Cedric-Pigeon not planning anything. Can't say if it's useful. Though not for me.

PR #46, opened 2015-04-21
Add partner_area_data and partner_area_management

Ehtaga:
This module was written to enter a location's area based on different units of measure
The information is dependant on the different locations entered


bwrsandman:
:+1:

pedrobaeza:
Some things:
- Why partner_area? You are mentioning stock locations, so at least stock prefix should be in the name.
- First module doesn't contain models, but only translations.

Ehtaga:
Ok for the name.

The first module adds surface units of measure (m² and feet²). They are necessary to set a default value in the second module. This way, the second module depends on the first one and the data is loaded before the model.

bwrsandman:
Can we merge?

pedrobaeza:
It's OK for me.

PR #47, opened 2015-04-23
ADD module stock dynamic mto

florian-dacosta:
Hello,

With products configured with the make to order route, it can happen that we end with some stock in the warehouse.
For example, we have a sale order, so a purchase order is made, then we cancel the sale order and we have un-wanted stock because the purchase order was already validated by the supplier.

This stock will stay forever since everytime a new sale order is made for this product, a purchase order will be created, no matter if we have stock or not.

So the goal of this module is to take product from the stock if any, even if the route configured is make to order.
To do so, on each pull rule belonging to the make to orde route, we can configure an equivalent make to stock rule.
Thanks for reviews


jgrandguillaume:
Hi,

Thanks for the contrib ! A few remarks / Questions:
- In the description, you say :
  The purpose of the module is to give the possibility to a pull rule make to order to act like a make to stock rule depending of the virtual stock of a product.
  => Can you explain what is the behavior of the system here ? Like : If the virtual stock is XY then the behavior is blablabla, etc...
- Can I ask you to take the OCA manifest model for **openerp**.py (see: https://github.com/OCA/maintainer-tools/tree/master/template/module) ?
- I found the feature interesting, but I won't change the standard MTO route for this. What about creating a new procurement method for it ? Like we've made here : https://github.com/OCA/purchase-workflow/tree/8.0/vendor_consignment_stock/model

The idea would be to have a new route called for example "MTO if not stock". What do you think ? This way, we do not interfer too much with the standard and people can use this new route. 

Regards,

Joël

florian-dacosta:
Hey,

Thanks for your comment.
I will update the description : there are 3 case : 
My product has the MTO route.
1) There is no virtual stock (as it should since it is MTO product)
=> nothing different happens
2) I have a virtual stock of 1 product
=> A sale order of 1 product is made. Since the virtual stock is 1 we do not need to buy more products, the procurement's pull rule is changed from make to order rule to a make to stock rule. => No purchase is made.

3) I have a I have a virtual stock of 1 product
=> A sale order of 3 products is made. Since the virtual stock is 1, I still need 2 product to be able to deliver the sale order.
When the procurement order of the 3 products is confirmed, we split it. The quantity of the initial procurement will change to 2. And the rule will still be a make to order rule.
The splited procurement will be the same but its quantity will change to 1 and the pull rule will be forced to be a make to stock pull rule.
=> Only 2 products will be purchased, from the first splited procurement.

I'll change the manifest's model also.

For the last comment, I guess we could do it as you say, but I am not sure it is better, it seems more complicated to me. 
Because on each product, we will have 2 MTO route, which will do the exact same thing for most cases and if we check both routes, only one would work. The user could be mistaken, etc...
I do not really see any case where somenone would like to choose the classic mto rule for some products and the MTO if not stock rule for others. So, I prefer to have to configure this option on the warehouse level (as we usually have 1 mto rule per warehouse) than at the product level.

The way it is made today, is that you have to configure the make to stock rule on your classic MTO rule. So if you do not do anything, the system won't be changed or else, you want it to happen.

Do you think of any case we would want this option per product? Even if there are no real cases, you still prefer make an additional routes on the product?
I am not really opposed to it, I am just not sure it is better, but I will change it if needed ;)

I will make all the modifications as soon have I have your confirmation on this last point.

Regards,
Florian

pedrobaeza:
Isn't this a MTS+MTO route?

florian-dacosta:
Yes, depending of the virtual stock, it will act like a MTS route or a MTO route, or both (when it splits the procurement)
I am not sure to understand the question

pedrobaeza:
It's popular warehouse management nomenclature to speak about MTS+MTO like the rule that act like this. Many people demand this on the new WMS.

florian-dacosta:
Oh ok, yes. But this notion does not exist yet in Odoo right? 
The name of the module  can change of course.
So what do you think of the PR @pedrobaeza ?

pedrobaeza:
Apart from the comments from @jgrandguillaume, rename the module to _stock_mts_mto_rule_ and it's OK for me.

pedrobaeza:
Sorry, I push close by mistake.

florian-dacosta:
@jgrandguillaume 
Here, I tried to take into consideration your advices.
I have created a new action procurement and new route, pull rule, etc...

At the end, it is cleaner this way because, if the mts+mto route is chosen, the procurement with this rule will create 1 or 2 others procurement (no stock move). And these new procurements will create the stock moves.
So it adds a step and at least it keeps a historic of what happened.

On the bad side, it duplicates mto and mts pull rule.
I was thinking of re-using the existing mts and mto rule since they should be exactly the same, but I am not sure it is better.

Let me know what you think, I still can change the way it is done of course

jgrandguillaume:
Hi @florian-dacosta 

Thanks a lot to take my comments into account, you did a great job ! I review it again, and I have a few question / demand:
- Can you explain me better the "bad side" above, I'm not getting your point here. Why duplicating the existing rules ? As you're using standard MTO or MTS once the split is done, you should not need other routes right ? Or what am I missing ? If I'm right, please adapt that to use the standard route once the split is done.
- I would add a tooltip in the warehouse screen where you put the "Use MTO+MTS rules " field to explain what it does to tick it (user need to know here).
- During my tests, I tried on the runbot to check that box on the "Your company" warehouse, save and try to open the routes using the button "Route" at top right corner  and and raise an error [1]
- Since this module bring an important feature in the WMS, people will like to rely on it. For such an important module, we definitely need tests. One for testing that the standard MTO or MTS is not broken, one to test that your feature is working properly according to the described usage.

A part from those points, it really a good contributions, thanks a lot. Once those points are ok, you'll get my :+1: 

Regards,

Joël

[1]  Error:

 File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/addons/web/controllers/main.py", line 928, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/addons/stock/stock.py", line 3630, in view_all_routes_for_wh
    all_routes += self.get_all_routes_for_wh(cr, uid, wh, context=context)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/api.py", line 336, in old_api
    result = method(recs, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946491-47-fe31fa/openerp/addons/stock_mts_mto_rule/model/warehouse.py", line 120, in get_all_routes_for_wh
    warehouse.mts_mto_mto_rule_id.route_id
AttributeError: 'stock.warehouse' object has no attribute 'mts_mto_mto_rule_id'

florian-dacosta:
@jgrandguillaume Thanks for your reply
Standards MTO and MTS rules are already linked to a route. That is why I did not want to use it for the new mts+mto rule, which depend of another route...But finally it does not make any difference since we don'tnot use routes to split the procurement.

So I did add tests, fix the bug, add the tooltip and use the generics mto and mts rules.

I just have one doubt left: 
If the MTS+MTO route is used, we will have one more step (one more procurement for each sale order line).
Since this new procurement's goal is just to create other procurements (no stock.move nor purchase order line...) it won't ever change its state from 'running' to 'done'.
I handled it overwriting the 'def run' method from procurement, but I am not sure it is the right way to do it.

falinwa:
Hello florian-dacosta,

Thanks for your module,
actually we are developing a module to achieve this function too.
But the strange thing is we have a different way of code to achieve this function, and your module seems more complex than mine.

We also have test your module, because I think if your module is better we just focused to make this module great.

But for functionality side your module seems lack on some side especially in manufacture.

Here is our feedback:
1. Case for procurement generate PO automatically work fine as the expected behavior.
The strange thing is, when we already received the product, in the delivery order we must click button "recheck availibity" first, its not automatically check, and the line split into 2.
1. When we sell the product (MTO+MTS, and manufacture) and the product has BoM with product (MTO+MTS, and buy), the system will generate MO as the expected behavior,
   but it will not generate the purchase order; even its not generate the procurement. So I think its a bug.
2. When I create manufacture order manually with product (MTO+MTS, and manufacture) and the product has BoM with product (MTO+MTS, and buy), the system not generate the purchase order ; even its not generate the procurement,

So I think you don't test on the manufacture side?

There's another question too:
Why we need to check the checkbox MTO+MTS rule in warehouse side?
and we also need to check he checkbox MTO+MTS in product too.

why we don't just configure it on product level? what is the purpose on the warehouse level?

and I want to ask your feedback about our module, you can get it on:
https://www.odoo.com/apps/modules/master/fal_procurement_check_stock/

florian-dacosta:
Hello @falinwa 
Thanks for your interest in this module.

First, I have no preference between my module, or yours, or another, as long as it works fine.

I have tested your module, I see some problems : 
1) If you order 5 products, you have 2 products in stock : a purchase order is created with 3 products, it is fine.
But you have only one stock.move, because you do not split it.The state of this stock.move is 'waiting for another move' because it comes form the MTO rule.
When you receive the 3 products, the stock.move is partially available, but it won't reserve the 2 products from stock, because it is waiting another move...The stock move won't ever be fully reserved.

That is the main problem.

2) there is incoherence between uom_qty and uos_qty. 
In the same case than 1) you will have a procurement with uom_qty=3 and uos_qty = 5.

Beside that, it works.
But I am not sure it is great to create a procurement imediatly canceled, it is kind of useless.
Also, the procure_method of the move will always be 'make_to_order', even if you take all the products from stock. It is not problem, but it is not very coherent.

About my module : 
For make to stock product, I have tested without my module and the stock_move is not available automatically, we still need to check avaibility. 
Or you can wait for the mrp scheduler to run, it will check avaibility. I believe it is not a bug since it is done this way in native odoo.

The goal of splitting the stock move into 2 is to have coherent datas.
It only happens if the ordered_qty > qty available and qty_available > 0.
One move is make to stock, the other is make to order. Only this second one will generate the procurement which will be used to create the purchase order.

About the manufacture side, I had not tested since I believe it does not impact it.
I have just tested now, and I had the same result with or without my module.
I can't reproduce the bug you have described to me.

Finally, why the checkbox on the warehouse form : 
It is the samebehavior when you change the outgoing shippings for example. It will automatically create/delete the rules needed.
I guess the main goal is that if you create a new warehouse, it will generate automatically the rules needed for mts+mto management.
This way, it becomes very easy to configure, since the user does not need to create any procurement rules, it is created automatically when the box is checked.

Anyway, I do not know on which conditions you have tested the manufacture side of this module, but if you do it with the demo datas and a clean database, it should work fine.
Else, you can explain to me how to reproduce the bugs you have found, I would be glad to correct it.

maybe @jgrandguillaume can give a look to your module and tell what he thinks is best for OCA.
Thanks again for your interest!

falinwa:
Hi Florian,

Thanks for your test for our module,

For problem 1, I test on my side and when we receive the product, the delivery order will automatically be ready to transfer. So I don’t know why its not worked on your side.

For problem 2, I agree with you and we have fixed this.

To split the stock move, and consider the first stock move is for make to stock and the second is make to order; actually its good for technical point of view.

And actually I agree to have this approach.

But when I have discussed this with my boss and my client, and they don’t like this approach, 

They said it will be very messy on the view and they prefer to keep on 1 stock move with the checking of the stock.

So I think we can’t contribute further for this module.

At least I will explain the bug that I found in your module, here to produce:
1. ```
     Create product A (with MTS+MTO, manufacture), and add the stock to be 1
   ```
2. ```
     Create product B (with MTS+MTO, buy), don’t forget to add the supplier too, and add the stock to be 1.
   ```
3. ```
     Create BoM for product A, with bom line of product B, with the quantity of 1.
   ```
4. ```
     Make a sale quotation for product A, with qty of 3.
   ```
5. ```
     Confirm the quotation, and the system will generate the procurement
   ```
6. ```
     Run the scheduler
   ```
7. ```
     The system will generate manufacture to produce product A with qty of 2, but the system not generate the purchase order for product B with qty of 1
   ```

Expected behavior:
1.       The system will generate manufacture to produce product A with qty of 2, and the system generate the purchase order for product B with qty of 1

Thank You.

Best Regards,

Hans Yonathan

Odoo Support

FALINWA Limited

Expert in Finance, Information System & Business Intelligence

Odoo Silver Partner

Website:  http://www.falinwa.com/ www.falinwa.com

Tel. 13670137019 (China - Shenzhen)

From: Florian [mailto:notifications@github.com] 
Sent: Tuesday, May 19, 2015 7:08 PM
To: OCA/stock-logistics-warehouse
Cc: falinwa 1
Subject: Re: [stock-logistics-warehouse] ADD module stock dynamic mto (#47)

Hello @falinwa https://github.com/falinwa  
Thanks for your interest in this module.

First, I have no preference between my module, or yours, or another, as long as it works fine.

I have tested your module, I see some problems : 
1) If you order 5 products, you have 2 products in stock : a purchase order is created with 3 products, it is fine.
But you have only one stock.move, because you do not split it.The state of this stock.move is 'waiting for another move' because it comes form the MTO rule.
When you receive the 3 products, the stock.move is partially available, but it won't reserve the 2 products from stock, because it is waiting another move...The stock move won't ever be fully reserved.

That is the main problem.

2) there is incoherence between uom_qty and uos_qty. 
In the same case than 1) you will have a procurement with uom_qty=3 and uos_qty = 5.

Beside that, it works.
But I am not sure it is great to create a procurement imediatly canceled, it is kind of useless.
Also, the procure_method of the move will always be 'make_to_order', even if you take all the products from stock. It is not problem, but it is not very coherent.

About my module : 
For make to stock product, I have tested without my module and the stock_move is not available automatically, we still need to check avaibility. 
Or you can wait for the mrp scheduler to run, it will check avaibility. I believe it is not a bug since it is done this way in native odoo.

The goal of splitting the stock move into 2 is to have coherent datas.
It only happens if the ordered_qty > qty available and qty_available > 0.
One move is make to stock, the other is make to order. Only this second one will generate the procurement which will be used to create the purchase order.

About the manufacture side, I had not tested since I believe it does not impact it.
I have just tested now, and I had the same result with or without my module.
I can't reproduce the bug you have described to me.

Finally, why the checkbox on the warehouse form : 
It is the samebehavior when you change the outgoing shippings for example. It will automatically create/delete the rules needed.
I guess the main goal is that if you create a new warehouse, it will generate automatically the rules needed for mts+mto management.
This way, it becomes very easy to configure, since the user does not need to create any procurement rules, it is created automatically when the box is checked.

Anyway, I do not know on which conditions you have tested the manufacture side of this module, but if you do it with the demo datas and a clean database, it should work fine.
Else, you can explain to me how to reproduce the bugs you have found, I would be glad to correct it.

maybe @jgrandguillaume https://github.com/jgrandguillaume  can give a look to your module and tell what he thinks is best for OCA.
Thanks again for your interest!

—
Reply to this email directly or view it on GitHub https://github.com/OCA/stock-logistics-warehouse/pull/47#issuecomment-103441807 .

mylbco:
@florian-dacosta
@jgrandguillaume 

Hi how is the status on this pull request ?
I think I can use this module ...

Peter

florian-dacosta:
Hi,
Well, feel free to test and give your feedback.
This PR needs reviews ;)

mylbco:
cloned it already ;-)

Peter Langenberg
Zaakvoerder

Bezoek onze website :  http://goog_220292044/www.bubbles-it.be
Volg op twitter via : @mylbco
Adres : Luikstraat 67, 9160 Lokeren
https://www.google.be/maps/place/Luikstraat,+9160+Lokeren

 “If you think a professional is expensive, wait till you try an amateur.”

Paul "Red" Adair

2015-06-19 15:02 GMT+02:00 Florian notifications@github.com:

michaeljohn32:
I've tested this with some "real" data for about two weeks, and it works as advertised.  Code looks clear and concise.  Great work @florian-dacosta!
+1

mylbco:
Florian

I've done some testing.

First test with scenario's described in the readme work like a charm.
But I got a scenario where the system throws a traceback. 
Scenario
1. create a demodatabase (sale - purchase) with demo data
2.  install the module dropshipment,
3. Create a sale order (eg) agrolait
4. In this sales order create a new product (eg) abc / product type stockable / routes dropshipping + buy / add supplier info.
5. Confirm the order

I get this traceback : 

  File "/home/peterl/odoo/openerp/http.py", line 537, in _handle_exception
  File "/home/peterl/odoo/openerp/http.py", line 574, in dispatch
    result = self._call_function(*_self.params)
  File "/home/peterl/odoo/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/home/peterl/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/http.py", line 307, in checked_call
    return self.endpoint(_a, *_kw)
  File "/home/peterl/odoo/openerp/http.py", line 803, in __call__
    return self.method(_args, *_kw)
  File "/home/peterl/odoo/openerp/http.py", line 403, in response_wrap
    response = f(_args, *_kw)
  File "/home/peterl/odoo/addons/web/controllers/main.py", line 948, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/peterl/odoo/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/addons/portal_sale/portal_sale.py", line 67, in action_button_confirm
    return super(sale_order, self).action_button_confirm(cr, uid, ids, context=context)
  File "/home/peterl/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/addons/sale/sale.py", line 602, in action_button_confirm
    self.signal_workflow(cr, uid, ids, 'order_confirm')
  File "/home/peterl/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/models.py", line 3557, in signal_workflow
    result[res_id] = workflow.trg_validate(uid, self._name, res_id, signal, cr)
  File "/home/peterl/odoo/openerp/workflow/__init__.py", line 85, in trg_validate
    return WorkflowService.new(cr, uid, res_type, res_id).validate(signal)
  File "/home/peterl/odoo/openerp/workflow/service.py", line 91, in validate
    res2 = wi.validate(signal)
  File "/home/peterl/odoo/openerp/workflow/instance.py", line 75, in validate
    wi.process(signal=signal, force_running=force_running, stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 120, in process
    ok = self._split_test(activity['split_mode'], signal, stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 248, in _split_test
    self._join_test(t[0], t[1], stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 257, in _join_test
    WorkflowItem.create(self.session, self.record, activity, inst_id, stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 95, in create
    workflow_item.process(stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 120, in process
    ok = self._split_test(activity['split_mode'], signal, stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 248, in _split_test
    self._join_test(t[0], t[1], stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 257, in _join_test
    WorkflowItem.create(self.session, self.record, activity, inst_id, stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 95, in create
    workflow_item.process(stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 120, in process
    ok = self._split_test(activity['split_mode'], signal, stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 248, in _split_test
    self._join_test(t[0], t[1], stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 257, in _join_test
    WorkflowItem.create(self.session, self.record, activity, inst_id, stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 95, in create
    workflow_item.process(stack=stack)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 116, in process
    if not self._execute(activity, stack):
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 162, in _execute
    returned_action = self.wkf_expr_execute(activity)
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 313, in wkf_expr_execute
    return self.wkf_expr_eval_expr(activity['action'])
  File "/home/peterl/odoo/openerp/workflow/workitem.py", line 291, in wkf_expr_eval_expr
    result = eval(line, env, nocopy=True)
  File "/home/peterl/odoo/openerp/tools/safe_eval.py", line 314, in safe_eval
    return eval(c, globals_dict, locals_dict)
  File "", line 1, in <module>
  File "/home/peterl/odoo/openerp/api.py", line 239, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/api.py", line 546, in new_api
    result = method(self._model, cr, uid, self.ids, _args, *_kwargs)
  File "/home/peterl/odoo/addons/sale/sale.py", line 735, in action_ship_create
    procurement_obj.run(cr, uid, proc_ids, context=context)
  File "/home/peterl/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/api.py", line 363, in old_api
    result = method(recs, _args, *_kwargs)
  File "/home/peterl/stock-logistics-warehouse-mylbco/stock_mts_mto_rule/model/procurement.py", line 67, in run
    res = super(ProcurementOrder, self).run(autocommit=autocommit)
  File "/home/peterl/odoo/openerp/api.py", line 239, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/home/peterl/odoo/openerp/api.py", line 546, in new_api
    result = method(self._model, cr, uid, self.ids, _args, *_kwargs)
ValueError: "run() got an unexpected keyword argument 'autocommit'" while evaluating
u'action_ship_create()'

pedrobaeza:
This can be due to https://github.com/odoo/odoo/pull/6775. In OCB is already merged, so you can try if you don't get the error with it.

mylbco:
Thanks Pedro, I'll check it out.

Peter Langenberg
Zaakvoerder

Bezoek onze website :  http://goog_220292044/www.bubbles-it.be
Volg op twitter via : @mylbco
Adres : Luikstraat 67, 9160 Lokeren
https://www.google.be/maps/place/Luikstraat,+9160+Lokeren

 “If you think a professional is expensive, wait till you try an amateur.”

Paul "Red" Adair

2015-06-29 11:08 GMT+02:00 Pedro M. Baeza notifications@github.com:

mylbco:
@pedrobaeza  
I did the test with OCB as you asked (odoo/odoo#6775)
Then it works, thank you very much ... nice to know.
@florian-dacosta 
blamed the wrong person ;-)

mistotebe:
Nice module. What is the use case for the two extra rules (mts_rule_id, mto_rule_id)?

Can it not just insert itself as another choice in 'procure_method', check if it needs to split itself, set procure_method to make_to_stock and/or make_to_order and then let the original implementation act as appropriate (running self.run() again)? Would make it even simpler I believe?

florian-dacosta:
@mistotebe 
sorry for the delay, I was in holidays,
Sometimes we need to split the procurement into 2 procurements : 
Example of use case : We have 1 product in stock.
We have a sale order with 2 products
=> We need to create 2 differents procurements.
1 make to stock and 1 make to order.

In the code, it is simpler to add an action rather a procure method. Also, I think it is not the goal of the procure_method.

I think it is simple enought. Yes it adds a step in the procurement's process, but it is automatic, and then it creates the right procurement(s), which use the original rules/implementation.

mistotebe:
@florian-dacosta: I've tried drafting how I'd change your code to do what I suggested and now I see why that would not work. Action seems to be the sanest way to go.

LGTM

nhomar:
+1 for me.

Can we merge this module? we tested in our side and it works as expected.

pedrobaeza:
Merging!

Issue #50, opened 2015-04-25
Unable to release reservation or confirm sales quotation when > 1 quotation lines


mlaitinen:
I'm unable to confirm a sales quotation or release all order line reservations if there are more than 1 `sale.order.line` with reservations.

The error message is `('ValueError', 'Expected singleton: stock.reservation(34, 35)')`

Steps to reproduce:
1. Create a sales quotation with more than one reservable product lines
2. Reserve all products, separately or all at once
3. Confirm the sales quotation or release all lines at once (not separately. that works.)

mlaitinen:
I have a commit for this, but I'm not sure how to create a pull request without including the translations, which are already committed in pull request #49.

PR #60, opened 2015-07-11
Allow releasing multiple reservations at once

mlaitinen:
Fixes issue #50


stock_reserve/model/stock_reserve.py
```diff
@@ -132,8 +132,9 @@ def release(self):
         """
         Release moves from reservation
         """
-        move_recs = self.move_id
-        move_recs.action_cancel()
+        for rec in self:
```
pedrobaeza: You can do better: `self.mapped('move_id').action_cancel()`

mlaitinen: Awesome, thanks! I learned about `mapped()` only after I had originally made this change.

guewen:
:+1:

pedrobaeza:
:+1:

PR #61, opened 2015-07-22
[ADD] add modules to use product packaging

laetitia-gangloff:
packaging_extended: to use uom in packaging
purchase_packaging: to improve supplierinfo with purchase uom and packaging and use them in purchase order
sale_packaging: to use packaging uom in sale order line


chrisdec:
Functional tests : 

It works very well regarding configuration and creation of manual purchase orders.

But I noticed a problem with the scheduler. The procurement generated by the scheduler (based on reordering rules) is correct but the linked PO is not. The supplier is correct but the packaging is not taken into account and thus unit of measures in the PO are not correct.

laetitia-gangloff:
@chrisdec I make some corrections following your test.

packaging_extended/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+from . import (
+    models,
+)
```
nhomar: one line pliz.

nhomar:
Hello,

Can you change the name of packaging_extended I think name packaging_uom fit better to what it is doing (you can use the one you decide is just an example.

The words *extended *extra *plus do not say too much about what a module does.

Regards.

purchase_packaging/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+from . import (
+    models,
```
nhomar: one line pliz.

nhomar:
I do not like too much the rule, but there it is, can you split the PR is a PR per module pliz.

You can follow this steps.:

https://github.com/odoomrp/odoomrp-wip/wiki/Migration-to-OCA

laetitia-gangloff:
@nhomar Thank you for your review ! I renamed packaging_extended in packaging_uom as you suggested.
I am not sure it is a good idea to split these 3 (little) modules in 3 PR.
Indeed, puchase_packaging and sale_packaging are dependent of packaging_uom, and can not be tested without it.
Is it possible to make a PR dependent of another PR to make test pass ?

pedrobaeza:
You can make copy branches that include the original commit of packaging_uom, and add commits for other module, and make the PR, so dependency is included. When the first PR will be merged, only rebasing you will get again a correct PR for the second and third module.

purchase_packaging/tests/test_packaging.py
```diff
@@ -0,0 +1,321 @@
...
+        po.signal_workflow('purchase_confirm')
+        sm = po.picking_ids[0].move_lines[0]
+        self.assertEqual(sm.product_packaging.id,
+                         self.product_packaging_34.id)
+        self.assertEqual(sm.product_uom.id,
+                         self.env.ref('product.product_uom_dozen').id)
+        self.assertAlmostEqual(sm.product_uom_qty, 16)
+
+    def test_procurement(self):
+        """ On supplierinfo set min_qty as 0
```
lmignon: IMO it's better to describe the steps inside the code before each step

chrisdec:
:+1: (functional tests)

sbidoul:
@laetitia-gangloff @chrisdec can you improve the README's? The current proposed description explain the fields that are added, but for people not familiar with the intented use case, it's difficult to understand why the modules are useful.

A few sentences explaining what the module do that cannot be done in standard Odoo would help. And maybe a specific example.

purchase_packaging/models/product.py
```diff
@@ -0,0 +1,49 @@
...
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    @api.model
+    def _default_min_qty_uom_id(self):
+        return self.env.ref('product.product_uom_unit')
+
+    packaging_id = fields.Many2one('product.packaging', 'Logisitical Units')
+    product_uom = fields.Many2one(compute='_compute_product_uom',
```
Cedric-Pigeon, line 34: better to call it product_uom_id, isn't it ?

laetitia-gangloff, line 34: It is an override of the existing field, so I cannot rename it.

gurneyalex:

laetitia-gangloff:
It is rebased. Thank you.

purchase_packaging/models/purchase.py
```diff
@@ -0,0 +1,238 @@
...
+        When packaging_id is set, uom_id is readonly,
+        so we need to reset the uom value in the vals dict
+        """
+        if vals.get('packaging_id'):
+            vals['product_uom'] = self.env['product.packaging'].browse(
+                vals['packaging_id']).uom_id.id
+        return vals
+
+    @api.model
+    def create(self, vals):
```
gurneyalex, line 188: you need to add `@api.return('self', lambda rec: rec.id)`

sale_packaging/models/sale_order_line.py
```diff
@@ -0,0 +1,90 @@
...
+        When product_packaging is set, product_uom is readonly,
+        so we need to reset the uom value in the vals dict
+        """
+        if vals.get('product_packaging'):
+            vals['product_uom'] = self.env['product.packaging'].browse(
+                vals['product_packaging']).uom_id.id
+        return vals
+
+    @api.model
+    def create(self, vals):
```
gurneyalex, line 86: you need to add @api.return here too

gurneyalex:
@laetitia-gangloff sorry for taking so long to review. A couple of missing decorators and this will be ok for me.

laetitia-gangloff:
@gurneyalex Thank you very much for your review !
I corrected missing decorators.
I need to improve documentation ( :( ) for example Purchase packaging can be used to order by multiple.

And I hesitate to set product_qty on purchase order editable, and recompute the value according to minimal uom... Maybe later.

lmignon:
@laetitia-gangloff IMO, what you propose should be done in an other PR. :+1: (Code + Functional)

gurneyalex:
:+1:

rousseldenis:
:+1: @gurneyalex Seems to be good for a merge ?
Plan to migrate in v10

lmignon:
I'll start the migration to v10

gurneyalex:
it's weird that travis is red... @Cedric-Pigeon could you try rebasing (again)?

rousseldenis:
@gurneyalex This is due to stock_change_qty_reason module error.

rousseldenis:
I will rebase it before merging

rousseldenis:
Closing this

PR #72, opened 2015-08-12
fix stock_mts_mto procurement check method

florian-dacosta:
Hello,

In stock_mts_mto module, we have a new type of procurement rule action : 'split procurement'.
This rule is used to create one or 2 procurements.

I had define the def _check method in a way that the procurement having this type of rule would be 'done' once it had created the other procurements.

But I am not sure it is a good idea. For example it is not compatible with the module sale_order_back2draft, because when you cancel your order and put it to draft state again, Odoo try to delete the procurements. But it won't be able to delete a done procurement.
That is why, I think it may be better to change the procurement's state to done, only when its 'children procurements' are done or cancel.

What do you think?
Maybe there is a better solution...

### DIFF
--- a/stock_mts_mto_rule/model/procurement.py
+++ b/stock_mts_mto_rule/model/procurement.py
@@ -19,11 +19,20 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ###############################################################################
-from openerp import models, api
+from openerp import models, api, fields
 
 
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
+
+    mts_mto_procurement_id = fields.Many2one(
+        'procurement.order',
+        string="Mto+Mts Procurement",
+        copy=False)
+    mts_mto_procurement_ids = fields.One2many(
+        'procurement.order',
+        'mts_mto_procurement_id',
+        string="Procurements")
 
     @api.multi
     def get_mto_qty_to_order(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't better to check procurements from the same group instead of creating new fields?

florian-dacosta:
@pedrobaeza I thought about it, because I also would like to avoid to add new fields.
But it didn't seem very logical to put the state to done if any procurement of the group is also done... It could be a procurement regarding another product, it does make much sense.

For example, if I have 4 products with route mts+mto and one product make to stock.
If the procurement of the product make to stock goes to 'done', all 4 procurement with rule mts+mto will also go to done at the same time. I am not sure it is better.

florian-dacosta:
Another problem if you depend on all the other procurements of the group, is that you do not know when to do the check of the mto+mts procurement.
So it would be done by the cron mrp, but we do not want to delay the check of the procurements, because it would also delay the sale's workflow (the shipped flag wound not be True before the mrp cron runs).

And for more complicated use cases, it can add problems (like for cases when we group/split sale orders).

So I did one more changes in order to check the mto+mts procurement's state when its 'child' procurement is checked. (to avoid the delay I was talking about.)
Now it seems ok. The only problem I see is adding the fields...but I guess it can be useful to have a link between these procurements.

Of course, I would be happy change this logic if we find a better solution!

stock_mts_mto_rule/model/procurement.py
```diff
@@ -19,12 +19,21 @@
...
 #
 ###############################################################################
-from openerp import models, api
+from openerp import models, api, fields
 
 
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
 
+    procurement_id = fields.Many2one(
```
pedrobaeza: Call it `mts_mto_procurement` for identifying its use

pedrobaeza:
OK, I understand your reasoning, so it seems there's no better solution, so :+1: with a little remark.

florian-dacosta:
Thanks for your feedback, I have change the field name now.

stock_mts_mto_rule/model/procurement.py
```diff
@@ -19,12 +19,21 @@
...
 
 
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
 
+    mts_mto_procurement_id = fields.Many2one(
+        'procurement.order',
+        string="Procurement",
+        copy=False)
+    procurement_ids = fields.One2many(
```
pedrobaeza: Sorry for not mentioning, but this implies to change this one also because it follow the same logic: mts_mto_procurement_ids

bguillot:
Yes it's better to set the procurement done only if its children are done. :+1:

gurneyalex:

florian-dacosta:
Since there are 2 thumbs up, is it possible to merge?
Thanks

rvalyi:
done

PR #74, opened 2015-08-20
[FIX] Added a type check to product_id_change of sale.order.line in stock_r…

mlaitinen:
…eserve_sale. Sometimes result['warning'] can be present, but False and the old code fails to deal with it and raises an error.

If you go through the original [product_id_change](https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L1088) of `sale.order.line`, you'll notice that `result['warning']` can very well be `False`. There's already `result.setdefault('warning', {})` that sets `result['warning']` to `{}`, but it will not touch the value if `warning` exists in `result`. Despite having `False` as the value.

### DIFF
--- a/stock_reserve_sale/model/sale.py
+++ b/stock_reserve_sale/model/sale.py
@@ -180,7 +180,8 @@ class SaleOrderLine(models.Model):
                     "the quantity of the stock reservation will "
                     "be automatically adjusted to %.2f.") % qty
             msg += "\n\n"
-            result.setdefault('warning', {})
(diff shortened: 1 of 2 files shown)


gurneyalex:

mlaitinen:
Rebased.

I honestly don't know what to do about the decreased coverage. Or does it even matter in this context?

mlaitinen:
Checks pass :v:

gurneyalex:
:+1:

Cedric-Pigeon (approved):

pedrobaeza (approved):

Issue #76, opened 2015-08-24
stock_inventory_preparation_filter: Test on runbot from OCA was not successful


ybofr:
Hi,

I wanted to select on product category that contains two products. I was able to select the product category, but nothing happened related to the filtering. No product came listed below. And when I tried to add a product manually, it came on the top, where I should only select the categories.

It looks to me as if something is wrong in the module and it did not work at all as I was expecting.

Basically, there is no filtering. The module at the moment allows to indicate a filter, but nothing else is happening. Looks to me like a bug, but I'm not sure the way, it was meant to work.

I have retested it, I can for instance select products, they will appear on the upper screen part, but not as line detail for the inventory. So, it really does not seem to work.

Picture was added.

<img width="1005" alt="inventory_adjustments_-_odoo" src="https://cloud.githubusercontent.com/assets/7601880/9450308/d51c87c4-4a9f-11e5-9f5d-667189931566.png">

pedrobaeza:
Hi, Eva, you're misunderstanding inventory (in general) behaviour: the only lines added are the ones who already contains some stock. This is the standard behaviour with the existing filters, and the expanded ones only restrict the initial selection to look for.

For these case, I see that you have added two products (eva test 1 raw and eva test 2 raw), but without stock, there will be no line in this inventory. I have added 5 units for the product 1, and make another inventory, and you can see that there's a line now with this 5 units.

For Odoo, no line means quantity 0. That's understandable in many cases for example when restricting to an specific location (how to decide if an element should go here with 0 quantity), but this is a debate for doing with Odoo itself.

PR #77, opened 2015-08-25
Initial inventory values to be allowed and inserted in below grid

dmsmkd:
This fix allows to filter and then create initial inventory values for products which did not had any inventory move before this is used.

### DIFF
--- a/stock_inventory_preparation_filter/models/stock_inventory.py
+++ b/stock_inventory_preparation_filter/models/stock_inventory.py
@@ -40,6 +40,21 @@ class StockInventoryFake(object):
         self.partner_id = inventory.partner_id
         self.package_id = inventory.package_id
 
+    def default_value(self):
+        value = {}
+        value.update({
+            'theoretical_qty': 0.0, 
+            'product_id': self.product_id.id, 
+            'location_id': self.location_id.id if self.location_id else False,  
+            'prod_lot_id': self.lot_id.id if self.lot_id else False, 
+            'inventory_id': self.id, 
+            'package_id': self.package_id.id if self.package_id else False, 
+            'product_qty': 0.0, 
+            'product_uom_id': self.product_id.uom_id.id, 
+            'partner_id': False
+        })
+        return [value]
+
 
 class StockInventory(models.Model):
     _inherit = 'stock.inventory'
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As explained in https://github.com/OCA/stock-logistics-warehouse/issues/76, I'm not sure if this is a good idea.

dmsmkd:
Thanks Pedro for the quick answer. What i intend to do here is to make a initial stock values to the filtered products. This is allowed originally from Odoo and is even presented in a official video. Only thing to make sure is to set new value and not leave it to zero, since it won't make sense.

pedrobaeza:
I would suggest then to add an option to select if you want to add zero quantity products or not before starting inventory (with a selection field or a boolean).

ybofr:
Hi Pedro,
The logic of Odoo is wrong from a functional point of view. When preparing the list of products to inventory, I should be able to select all products from my warehouse. If some exclusion should be done, then not based on the on hand quantity, but based on the fact that the product has never been moved in the system.

So, best would be to have a checkbox asking if products with stock zero should be selected in order to keep the (wrong) odoo behaviour and to allow the standard warehouse behaviour.

Thanks and regards
Eva P.

pedrobaeza:
OK, that what I was asking for: please add a flag or a selection to decide if you want products with zero quantity or not. These are the tasks to do in order to achieve this:
- Add the flag.
- Document this new use case on README.
- Include all possible selection cases (standard ones plus the added by this module).
- Make tests for each new possible scenario.

jjscarafia:
Any news on this PR? Can I take it and add the flag?

gurneyalex:
@jjscarafia If you want to for the PR, please do.

jjscarafia:
We have make a new PR (https://github.com/OCA/stock-logistics-warehouse/pull/153) to replace this one as suggested here.

Please check functionality and then we improove README and test cases

jjscarafia:
@dmsmkd we have made this new pr #153 of yours as suggested by @pedrobaeza and @gurneyalex.
 @ybofr For now it options are all products (new one) and default odoo one (products with stock). I agree with you that it makes more sence to use "products with moves in the location", but I think that this should be fixed on odoo with another PR. 
What do you think? We keep going with PR #153?

rousseldenis:
Closing this as it is quite old and superseeded by #153. But closed too.

PR #81, opened 2015-09-10
Migrate stock_available_sale to v8

clonedagain:
- [x] Please review/merge this before: https://github.com/OCA/stock-logistics-warehouse/pull/100

Correctly segregates templates and variants.
Rewritten in new API.
Uses the ORM instead of SQL calls.
Conforms better to OCA standards

You may be interested in this other PR too: https://github.com/OCA/stock-logistics-warehouse/pull/93


clonedagain:
@lbellier for your information

clonedagain:
WIP, please bear with me as I iron out the details.

clonedagain:
The remaining failures seem unrelated to these changes, so I'd say this is ready for review.

stock_available_sale/views/product_template_view.xml
```diff
@@ -3,8 +3,8 @@
     <data>
         <!-- Add the quantity available to promise in the product form -->
         <record id="view_product_form_quoted_qty" model="ir.ui.view">
-            <field name="name">product.form.quoted_qty</field>
-            <field name="model">product.product</field>
+            <field name="name">Quoted quantity on product form</field>
+            <field name="model">product.template</field>
```
moylop260, line 7: Why not `product.product`?

stock_available_sale/models/product_template.py
```diff
@@ -0,0 +1,47 @@
...
+##############################################################################
+
+from openerp import models, fields, api
+from openerp.addons import decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    quoted_qty = fields.Float(
```
moylop260, line 12: Why `product.template`?

moylop260:
@clonedagain 
travis red. Is a know error?

stock_available_sale/models/product_template.py
clonedagain, line 12: We need both: product.template shows the total of the variants. That's what the standard does with quantities too if I read it correctly.
The one for product.product is at stock_available_sale/models/product_product.py, line 27.

stock_available_sale/views/product_template_view.xml
clonedagain, line 7: If I'm not mistaken, this view is reused for both product templates and variants so doing it on product.template serves both purposes at once.
The tree view on the other hand seems to have been specifically inherited for product.product in the standard, but I'm not adding the new field there.

clonedagain:
@moylop260 The error seems unrelated but please double-check if you can.
The main 8.0 is red too, but I don't know if someone is working to make it green.

clonedagain:
@moylop260 if you care about this PR you may want to test #80 too, which makes the product/variant situation more correct.

stock_available_sale/models/product_template.py
moylop260, line 12: Sorry, I don't understand why we need the same field in both models.
If `product_template.quoted_qty` show total of the variants then IMHO should be called different to  `product_prodcut.quoted_qty` because first one will be overwritten.

gurneyalex:
@clonedagain build should be fixed by #82

stock_available_sale/models/product_template.py
clonedagain, line 12: Yes, the method of product.product is overriding product.template on purpose, similar to what the standard does: check out https://github.com/odoo/odoo/blob/8.0/addons/stock/product.py#L345 vs https://github.com/odoo/odoo/blob/8.0/addons/stock/product.py#L139

gurneyalex:
:+1: 

if you have the opportunity to add an automated test using the location / warehouse context it would be a very nice addition.

gurneyalex:
@clonedagain can you rebase this PR and resolve the merge conflict?

clonedagain:
@gurneyalex I merged the latest 8.0 branch, I hope I didn't break anything.

clonedagain:
@gurneyalex I've improved the tests, hopefully now it's going to get the coverall quota.

clonedagain:
I've also removed the `@api.one` decorators like was done recently in `stock_available_immediately`.

clonedagain:
I had to rebase this on another PR: https://github.com/OCA/stock-logistics-warehouse/pull/100
This is not needed for this PR by itself, but I fear that the changes demanded by #100 will be forgotten if I don't include them right now.

Anyway, if runbot & travis are still green, this still needs review.

clonedagain:
Travis is red on a Transifex error. Anything I can do about it?

pedrobaeza:
Retry as you have done.

stock_available_sale/models/product_product.py
```diff
@@ -0,0 +1,118 @@
...
+        for group in self.env['sale.order.line'].read_group(
+                domain, ['product_uom_qty', 'product_id', 'product_uom'],
+                ['product_id', 'product_uom'],
+                lazy=False):
+            product_id = group['product_id'][0]
+            uom_id = group['product_uom'][0]
+            # Compute the quoted quantity in the product's UoM
+            # Rounding is OK since small values have not been squashed before
+            results += Counter({
+                product_id: self.env['product.uom'].sudo()._compute_qty(
```
clonedagain: @lmignon I'm doing a `sudo` also in `stock_available_sale`, do you think it could break multi-company setups like it did in `stock_available_mrp`?

lmignon: @clonedagain Why do you need to use `sudo()` here? `product.uom` can be read by `base.group_user`. Moreover, each time you use `sudo()`, you build a new empty environment and the _compute_qty will need to query the database to retrieve the product, the uom, etc... IMO, you introduce here a performance issue. What do you think?

stock_available_sale/models/product_product.py
```diff
@@ -0,0 +1,118 @@
...
+        for product in self:
+            product.immediately_usable_qty -= product.quoted_qty
+
+    @api.multi
+    def _get_quoted_qty(self):
+        """Compute the quantities in Quotations."""
+
+        domain = [
+            ('state', '=', 'draft'),
+            ('product_id', 'in', [p.id for p in self])]
```
lmignon: self._ids

clonedagain:
@lmignon I changed as you suggested.

lmignon:
:+1:  (Code review) Tahnk you @clonedagain

moylop260:
:+1: (w/o test)

PR #92, opened 2015-10-26
product_available improvements

gurneyalex:
- fix the dependencies for the computed field
- use api.multi instead of api.one to avoid calling `super()._immediately_usable_qty` in a loop (this improves perfs on a tree view  display)

### DIFF
--- a/stock_available/models/product_product.py
+++ b/stock_available/models/product_product.py
@@ -29,7 +29,7 @@ class ProductProduct(models.Model):
     """
     _inherit = 'product.product'
 
-    @api.one
+    @api.multi
     @api.depends('virtual_available')
     def _immediately_usable_qty(self):
         """No-op implementation of the stock available to promise.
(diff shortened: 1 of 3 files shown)


gurneyalex:
@clonedagain your impression on this is welcome.

lepistone:
:+1: thanks

yvaucher:
I think we could change `@api.one` to `@api.multi` only in  `stock_available_immediately/models/product_product.py`.

Nevertheless it does what we want :+1:

gurneyalex:
@yvaucher using api.one means that the return value of the method is accumulated in a list. In this case, the methods is implicitely returning `None`, so we are generating lists of `None` which are being garbage collected immediately, and this is not very efficient. So I argue that api.one should be removed in this place (the gain is way lower than avoiding the `super()` call but still :smile_cat: )

clonedagain:
I understand the logic, :+1: for me

PR #93, opened 2015-11-04
stock_available_mrp: migrate to v8

clonedagain:
- [x] Please review/merge this before: https://github.com/OCA/stock-logistics-warehouse/pull/100

Now computes potential quantities for both product templates and variants. To keep the code simple, only the biggest potential of any single variant is accounted for in the template's potential.
Now takes all levels of phantom BoM into account, respects validity dates etc. thanks to the use of the standard method _bom_explode, as suggested by @gdgellatly in OCA#5 (comment)
Tests rewritten in python and improved.
Code now adheres to new file/manifest conventions.

Still to do: 
- [x] add tests using several levels of (phantom) BoM
- [x] add tests with BoMs lines having an UoM for the component product different than the main UoM of that product
- [x] add tests with a BoMs making finished products in an UoM different than the main UoM of that product
- [x] add tests with warehouse & location contexts (we don't touch that but we never know)
- [x] add tests for the stock available for sale
- [x] respect multi-company rules

You may be interested in this other PR too: https://github.com/OCA/stock-logistics-warehouse/pull/81


clonedagain:
@gdgellatly @pedrobaeza @guewen @gurneyalex you have shown interest for this module in v7, you may want to review this v8 port.

I wanted to make an early PR to expose my progress - and I will appreciate early feedback - but please consider this as still work in progress.
The tests should pass but they are badly lacking in certain areas and the module is not field-tested either.

clonedagain:
I'm rather happy with the state of this PR: if Travis and runbot run green, I'd say this is "NEED REVIEW"

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,58 @@
...
+        This is the same implementation as for templates."""
+        super(ProductProduct, self)._immediately_usable_qty()
+        for product in self:
+            product.immediately_usable_qty += product.potential_qty
+
+    @api.multi
+    def _get_potential_qty(self):
+        """Compute the potential qty based on the available components."""
+        # Browse the BOMs as superuser to bypass access rights
+        bom_obj = self.env['mrp.bom'].sudo()
```
lmignon: Is `sudo()`multi-company aware?

lmignon:
@clonedagain Thank you for migrating this module. IMO you shouldn't find the bom defined for a product as a sudoer. In a multi company env, if two boms are defined for two different companies, the method will always return the first bom found.  Am'I right or do I miss something?

clonedagain:
I have no idea whether `sudo` would break multi-company situations, but it would make sense and be a problem.

On the other hand, if I don't `sudo()` then I need to grant read access on BoMs to everyone, to be able to read the stock available to promise.
Would you agree with this change?

stock_available_mrp/models/product_product.py
clonedagain: Good question but no idea!

lmignon:
@clonedagain A bom is defined for a company. For a same product set if you define different bom for different companies, by using sudo to call `_bom_find` the search to the bom definition will be done as superuser (https://github.com/odoo/odoo/blob/9.0/addons/mrp/mrp.py#L225) and the record rules will not be applied to restrict the result to the company of the current_user.

IMO, you can solve this problem in 2 differents ways.
- use base_suspend_security from https://github.com/OCA/server-tools/tree/9.0/base_suspend_security in place of sudo
- put the user's company into the context since if the field is present it will be used as criteria into the search domain (https://github.com/odoo/odoo/blob/9.0/addons/mrp/mrp.py#L216)

What do you think?

lmignon:
@clonedagain IMO the first proposed solution is the only one valid even if it add a new dependency since it's the only one that let's the system apply the record rules defined for `mrp.bom`. If we put a company_id into the context, the domain will be defined with `domain + [('company_id', '=', context['company_id'])]` . With the record rule, the domain will be `['|',('company_id','child_of',[user.company_id.id]),('company_id','=',False)]`

I'm even not sure that using the method `_bom_find` is a good idea since it doesn't break if more than one mrp.bom are found for a product_id.

clonedagain:
@lmignon I used `sudo()` indeed to bypass all restrictions (even record rules), but I agree this is probably not the best to do.
I agree to use  `base_suspend_security` instead. I'm short on time at the moment so if you can send me a patch I'll appreciate - otherwise it's OK, I'll make the change some time later.

Regarding the fact that _bom_find selects only one BoM, this is on purpose:
- `_bom_find` selects the "best" BoM, and is the logic used when creating MOs from procurements. Then again, I admit that using `sudo()` makes it less smart
- I can't do anything smart with multiple BoMs without a quite complex manufacturing policy (like, "I can promise X using this BoM + Y using that BoM if there are components left"). I have no intention to handle such complexity but I can add it as a limitation/roadmap item in the Readme.

lmignon:
@clonedagain I'll adapt the code and make a PR on yours tomorrow morning. You already have done an incredible job on this PR.

clonedagain:
Thanks that's very nice of you.

clonedagain:
I merged @lmignon's proposal and squashed the commits.

lmignon:
@clonedagain  why do you have squashed  my commit?  Sorry but I feel offensed.

lmignon:
@clonedagain I insist you can't squash the commit from different authors. We lost here the authorship of some changes in the code. I'm sure you understand the potential impact.

clonedagain:
Sure, I'll undo ASAP and fast-forward your commit instead. No offense 
intended.

lmignon:
@clonedagain Thank you, you have done the hard work on this PR and thank you to you for that. I was a little bit surprised. I've spent a lot of time trying to write a test failing if the call to `sudo` was removed and ensuring that the multi-company is supported to finally conclude that nor `sudo` nor `suspend_security` was required. IMO, this addons is usefull and very well designed.

lmignon:
:+1:  (Code review + functional tests)

moylop260:
Could you rebase it, please?

moylop260:
Please,
Check RST syntax error
https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/90260641

clonedagain:
Rebased, RST fixed.

moylop260:
:+1:

clonedagain:
Updated README using latest template.

gurneyalex:
@clonedagain any idea why the tests are red on odoo and green on OCB?

clonedagain:
@gurneyalex I never noticed that, maybe something changed upstream. I'll check next week if no-one finds the problem in between.

clonedagain:
@gurneyalex I rebased the PR and the "good" news is that now it fails on both the standard and OCB. So I guess we'll find out what changed rather easily.

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,229 @@
...
+                                               self.product_wo_bom]}
+
+        # Get the warehouses
+        self.wh_main = self.browse_ref('stock.warehouse0')
+        self.wh_ch = self.browse_ref('stock.stock_warehouse_shop0')
+
+    def assertPotentialQty(self, record, qty, msg):
+        record.refresh()
+        #  Check the potential
+        self.assertEqual(record.potential_qty, qty, msg)
```
clonedagain: Travis seems to fail here. @lmignon could it be related to the change in multi-company rules at https://github.com/odoo/odoo/commit/2fd14db57433d08ab368a7e18d18ae8d49dafab1 by any chance?

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,229 @@
...
+                         'company_ids': [(4, chicago_id)]})
+        test_user = test_user.sudo(test_user)
+        bom.company_id = self.ref('base.main_company')
+        self.assertPotentialQty(
+            test_user_tmpl, 0,
+            "The bom should not be visible to non members of the bom's "
+            "company or company child of the bom's company")
+        bom.company_id = chicago_id
+        self.assertPotentialQty(
+            test_user_tmpl, 1000.0, '')
```
clonedagain, line 140: Travis seems to fail here. @lmignon could it be related to the change in multi-company rules at odoo/odoo@2fd14db by any chance?

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # set the bom on the main company (visible to members of the main comp
+        # and all products without company (visible to all)
+        # and the demo user on Chicago (child of main company)
+        self.env['product.product'].search([
+            TRUE_LEAF]).write({'company_id': False})
+        chicago_id = self.ref('stock.res_company_1')
+        test_user.write({'company_id': chicago_id,
+                         'company_ids': [(4, chicago_id)]})
+        test_user = test_user.sudo(test_user)
```
clonedagain: @lmignon I think this sudo is useless, you're not reusing the variable `test_user` after this line do you?

stock_available_mrp/tests/test_potential_qty.py
clonedagain, line 140: @lmignon I think the problem is that Odoo now checks the record rules on quants, and our test does not include stock belonging to Chicago.
I'll add an inventory for the Chicago company to fix that and let you know.

clonedagain:
@lmignon @gurneyalex @lbellier @moylop260 I think it was only the test which was wrong, I fixed it at b8c2014 and didn't have to change the main code.

All is green, is it OK to merge please ?

clonedagain:
@gdgellatly @pedrobaeza @guewen your opinions are still welcome of course.

lbellier:
:+1: seems ok.

gurneyalex:
:+1:

PR #94, opened 2015-11-05
8.0 stock quant merge

clonedagain:
Proposal to adopt OdooMRP's module to merge quants when they were split and it's not needed anymore (ie. reservation is undone).

To be done:
- [x] Add OCA as co-author
- [x] Use OCA's README template
- [x] Add automatic tests

### DIFF
--- a/stock_quant_merge/models/stock.py
+++ b/stock_quant_merge/models/stock.py
@@ -1,0 +1,43 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+# For copyright and license notices, see __openerp__.py file in root directory
+##############################################################################
+
+from openerp import models, api
+
+
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.multi
+    def merge_stock_quants(self):
+        pending_quants = self.filtered(lambda x: True)
+        for quant2merge in self:
+            if (quant2merge in pending_quants and
+                    not quant2merge.reservation_id):
+                quants = self.search(
+                    [('id', '!=', quant2merge.id),
+                     ('product_id', '=', quant2merge.product_id.id),
+                     ('lot_id', '=', quant2merge.lot_id.id),
+                     ('package_id', '=', quant2merge.package_id.id),
+                     ('location_id', '=', quant2merge.location_id.id),
+                     ('reservation_id', '=', False),
+                     ('propagated_from_id', '=',
+                      quant2merge.propagated_from_id.id)])
+                cont = 1
+                cost = quant2merge.cost
+                for quant in quants:
+                    if (self._get_latest_move(quant2merge) ==
+                            self._get_latest_move(quant)):
+                        quant2merge.sudo().qty += quant.qty
+                        cost += quant.cost
+                        cont += 1
+                        pending_quants -= quant
+                        quant.sudo().unlink()
+                quant2merge.sudo().cost = cost / cont
+
+    @api.model
+    def quants_unreserve(self, move):
+        quants = move.reserved_quant_ids
+        super(StockQuant, self).quants_unreserve(move)
+        quants.merge_stock_quants()
--- a/stock_quant_merge/models/stock.py
+++ b/stock_quant_merge/models/stock.py
@@ -11,20 +11,24 @@ class StockQuant(models.Model):
     _inherit = 'stock.quant'
 
     @api.multi
+    def _mergeable_domain(self):
+        """Return the quants which may be merged with the current record"""
+        self.ensure_one()
+        return [('id', '!=', self.id),
+                ('product_id', '=', self.product_id.id),
+                ('lot_id', '=', self.lot_id.id),
+                ('package_id', '=', self.package_id.id),
+                ('location_id', '=', self.location_id.id),
+                ('reservation_id', '=', False),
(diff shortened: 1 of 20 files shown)


clonedagain:
For you information @pedrobaeza @oihane @anajuaristi @lbellier

anajuaristi:
Stock_quant_merge is working in production for several customers so :+1: RTM from my side

rvalyi:
Looks good. It's something I always wanted.
But @clonedagain remarks should be implemented so we can merge it in the OCA.

clonedagain:
The code now adheres to the conventions and has automatic tests.
Reviewers wanted.

elicoidal:
Formally LGTM (no test, no technical check)

stock_quant_merge/models/stock.py
```diff
@@ -0,0 +1,44 @@
...
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.multi
+    def merge_stock_quants(self):
+        pending_quants = self.filtered(lambda x: True)
+        for quant2merge in self:
+            if (quant2merge in pending_quants and
+                    not quant2merge.reservation_id):
+                quants = self.search(
```
rvalyi: could you please isolate this domain in a method that would take quant2merge as an argument and return the domain?
Some may need to override this domain. Example: in Brazil we will probably separate quants of the same product depending on if it comes from national origin or if it's imported. It's the same product in the MRP for instance. But for fiscal reports we cannot mix quants and we don't want to freak with lots for product origin. In this case we would need to override this domain, but may be there are other use cases.

stock_quant_merge/models/stock.py
```diff
@@ -0,0 +1,44 @@
...
+
+from openerp import models, api
+
+
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.multi
+    def merge_stock_quants(self):
+        pending_quants = self.filtered(lambda x: True)
```
oihane: Could you please explain why do you use pending_quants, wouldn't it been enough by processing those with `reservation_id` not set?

clonedagain: Wait, _you_ wrote that didn't you? :-)
I think it avoids trying to merge a quant we just removed in a previous iteration.

oihane: Yes, I did, but I don't remember why :worried: 
Maybe you should also start by doing the for only with those quants that do not have reservation...
I mean `for quant2merge in self.filtered(lambda x: not x.reservation_id)`

clonedagain:
@rvalyi i've changed it as you suggested.

rvalyi:
@clonedagain thanks! and kudos for the work!

:+1:

stock_quant_merge/models/stock.py
clonedagain: I suppose that's a way to get a copy of the recordset?

stock_quant_merge/models/stock.py
```diff
@@ -0,0 +1,49 @@
...
+                ('lot_id', '=', self.lot_id.id),
+                ('package_id', '=', self.package_id.id),
+                ('location_id', '=', self.location_id.id),
+                ('reservation_id', '=', False),
+                ('propagated_from_id', '=', self.propagated_from_id.id)]
+
+    @api.multi
+    def merge_stock_quants(self):
+        # Get a copy of the recorset
+        pending_quants = self.browse(self.ids)
```
clonedagain, line 28: @oihane is it clearer this way?

oihane, line 28: why don't you change the _for_? If you get only the ones with reservation_id not set it will do less work, don't you think?

clonedagain, line 28: The `for` is on `self` not `pending_quants`.

clonedagain, line 28: For the main part I think it's OK because the module itself only calls this method on unreserved quants, but there may be corner cases if called through RPC or custom modules.

I'm sorry but it's a bit too subtle for me to research given the time I can spare on this.
Maybe you could propose a test to make sure it's OK?

oihane, line 28: This is what I mean...

```
    @api.multi
    def merge_stock_quants(self):
        # Get a copy of the recorset
        pending_quants = self.browse(self.ids)
        for quant2merge in self.filtered(lambda x: not x.reservation_id):
            if quant2merge in pending_quants:
                quants = self.search(self._mergeable_domain())
                cont = 1
                cost = quant2merge.cost
                for quant in quants:
                    if (self._get_latest_move(quant2merge) ==
                            self._get_latest_move(quant)):
                        quant2merge.sudo().qty += quant.qty
                        cost += quant.cost
                        cont += 1
                        pending_quants -= quant
                        quant.sudo().unlink()
                quant2merge.sudo().cost = cost / cont
```

So this way you will only **for** those quants that have not been reserved yet.

@pedrobaeza, am I wrong?

clonedagain, line 28: Ok changed as you suggested.

clonedagain:
@lbellier is it OK for you?

lbellier:
@clonedagain Yes, it's good for me.

Cedric-Pigeon:
:+1: (code review) Thanks !

oihane:
I see it ok :+1:

PR #99, opened 2015-11-06
[IMP] stock_quant_manual_assign: Add lot and package to selection view

pedrobaeza:


pedrobaeza:
@oihane @Daniel-CA, please review

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -11,15 +11,11 @@
                             <tree editable="top" delete="0" create="0">
                                 <field name="quant" />
                                 <field name="location_id" />
+                                <field name="lot_id" />
```
oihane, line 14: This should have group, so that you make it visible only for those in "group_stock_production_lot"

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -11,15 +11,11 @@
...
+                                <field name="package_id" />
```
oihane, line 15: Same as lot, this one has its own group

oihane:
@pedrobaeza if you include those changes I will consider :+1: :wink:

Daniel-CA:
I agree with @oihane, include those groups and :+1:

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
pedrobaeza, line 14: I have added the groups to the fields definition instead.

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
pedrobaeza, line 15: I have added the groups to the fields definition instead.

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
oihane, line 14: Ok, it does not mind because it works the same way, thanks

ddiazpinto:
:+1:

PR #100, opened 2015-11-06
[IMP] Decouple the quantity for templates and variants

clonedagain:
There are cases where we dot NOT want to simply sum the quantities of all the variants. For example when dealing with manufacturing capacities, we may have to chose between variants because we can't make ALL of them with the same components.
So instead of a simple non-modular implementation, we'll let each module define his own implementation of how to compute the product template's quantity available for sale.

### DIFF
--- a/stock_available/__openerp__.py
+++ b/stock_available/__openerp__.py
@@ -1,23 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    This module is copyright (C) 2014 Numérigraphe SARL. All Rights Reserved.
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
+# © 2014 Numérigraphe SARL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Stock available to promise',
     'version': '8.0.3.0.0',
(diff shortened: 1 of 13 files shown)


stock_available/__openerp__.py
```diff
@@ -20,7 +20,7 @@
 
 {
     'name': 'Stock available to promise',
-    'version': '8.0.2.0.0',
+    'version': '8.0.3.0.0',
     "author": u"Numérigraphe,Odoo Community Association (OCA)",
```
elicoidal, line 7: Remove u

stock_available/models/product_product.py
```diff
@@ -36,8 +36,10 @@ def _immediately_usable_qty(self):
 
         By default, available to promise = forecasted quantity.
 
-        Must be overridden by another module that actually implement
-        computations."""
+        EACH sub-module MUST override this method in BOTH product.product
```
elicoidal: Remove the capitals.

stock_available/models/product_template.py
```diff
@@ -28,12 +28,16 @@ class ProductTemplate(models.Model):
     @api.multi
     @api.depends('product_variant_ids.immediately_usable_qty')
     def _immediately_usable_qty(self):
-        """Compute the quantity using all the variants"""
+        """No-op implementation of the stock available to promise.
+
+        By default, available to promise = forecasted quantity.
+
+        EACH sub-module MUST override this method in BOTH product.product
```
elicoidal: remove capitals

stock_available_immediately/models/product_template.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: Simplify header following: https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

stock_available/__openerp__.py
pedrobaeza, line 7: You can't remove it because of the `é`

pedrobaeza, line 7: Anyway, it's harmless

elicoidal, line 7: Understood

stock_available/models/product_product.py
clonedagain: I will. I always forget RST is richer than plain text, sorry. To much 
reading RFCs in my young time probably.

clonedagain:
@elicoidal Ok I've simplified all the headers in the modules I'm changing, and I've removed the CAPS in the docstrings.

elicoidal:
@clonedagain Thanks for your contribution!

lmignon:
:+1:  (Code review + functional tests)

moylop260:
:+1: (w/o test)

PR #101, opened 2015-11-17
[ADD] stock_traceability_operation

clonedagain:
Module to show the detail of pack operations in the traceability.
Useful in multi-location setups: the location indicated on stock moves
is less precise that the on on the operations because it does not
account for putaway strategies, manual changes, and reservations made by
the scheduler.

Work in progress:
- [x] should display total quantities in grouped view
- [x] must add access rules
- [x] must add translation template + French
- [x] could be faster
- [x] Coverall wants automatic tests
- [x] needs user testing


clonedagain:
@lbellier for your information.

@lmignon, @pedrobaeza, @jgrandguillaume I think this may be interesting to you. If you know others who would be interested too, @-mention them here please.

stock_traceability_operation/README.rst
```diff
@@ -0,0 +1,65 @@
...
+button "Traceability" on the serial number/lot form, or the
+"Quant history" button on the quant form.
+
+Unfortunately the standard only presents the associated stock moves, which is
+a problem because the actual locations can be different: when executing the
+transfer, stock workers may select any of the sub-locations though various
+means (automatic reservations, automatic putaway strategies, manual entry).
+
+In order to improve the traceability, this module lets you "unfold" the
+stock moves to see the actual locations which were used in the transfer.
```
elicoidal: remove 'which were'?

clonedagain: Done.

clonedagain:
Coverall was stuck but I rebased and now it's all green. Users are Ok with the feature.
Can we merge?

stock_traceability_operation/README.rst
```diff
@@ -0,0 +1,69 @@
...
+* Warehouse (and most other apps) > Product > form view > button "Moves"
+* Warehouse (and most other apps) > Configuration > Product Variants > form view > button "Moves"
+* Warehouse > Traceability > Serial Number/Lot > form view > button "Traceability"
+* Warehouse > Traceability > Quants > form view > button "Quant history"
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/8.0
+
+Kown issues
```
JosDeGraeve: small typo, replace by 'Known issues'

clonedagain:
@JosDeGraeve the typo is corrected.

clonedagain:
Fixed a bug in lot traceability

stock_traceability_operation/models/stock_quant.py
```diff
@@ -0,0 +1,66 @@
...
+from openerp import api, models
+from openerp.tools.translate import _
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    @api.multi
+    def action_view_quant_history(self):
+        '''Replace the action on stock moves into an action on the report
```
pedrobaeza: Use double-quote for docstring

stock_traceability_operation/models/stock_quant.py
```diff
@@ -0,0 +1,66 @@
...
+                # Notice: product_qty is in the product'sUoM
+                report_vals.update(
+                    {'product_uom_qty': move.product_qty,
+                     'location_id': move.location_id.id,
+                     'location_dest_id': move.location_dest_id.id})
+                report_ids.append(report_obj.create(report_vals).id)
+        if not report_ids:
+            return False
+        return {
+            'domain': "[('id', 'in', "
```
pedrobaeza: You can use here the Python list and tuple elements, so no weird `str` + `map` operation is needed.

pedrobaeza:
Very great module. I think that you have to add another button instead of replacing old ones, called "Improved history" or similar, and the known issue is easily solved. You can hide the old buttons if you want in the customer custom module, or include a security group that hides them if not marked.

clonedagain:
@lbellier do you agree with @pedrobaeza that we should keep the standard button and add ours alongside?
We're not calling super so we may be breaking other modules.

clonedagain:
@pedrobaeza the domain is `safe_eval`ed, so it has to be a string.

pedrobaeza:
@clonedagain, no, it's first checked if is an instance of a string, and if so, safe_evaled, but if not, is passed as is. Try and tell me.

clonedagain:
@pedrobaeza ok it seems the error is in the tests then. I'll try to improve that.

gurneyalex:
:+1:

clonedagain:
@pedrobaeza I've changed as you suggested (fixed the test) and rebased.

lbellier:
@clonedagain i agree with @pedrobaeza.

clonedagain:
@pedrobaeza @lbellier I've changed the code as you suggested: create new buttons and actions instead of replacing the standard ones.

stock_traceability_operation/views/stock_production_lot_view.xml
```diff
@@ -0,0 +1,24 @@
...
+
+<openerp>
+<data>
+
+<record model="ir.ui.view" id="stock_production_lot_traceability_operation">
+    <field name="name">Add operation traceability on lots</field>
+    <field name="model">stock.production.lot</field>
+    <field name="inherit_id" ref="stock.view_production_lot_form"/>
+    <field name="arch" type="xml">
+        <xpath expr="//button[@name='action_traceability']" position="attributes">
```
pedrobaeza: Better not to hide the button, to allow other improvements made to standard view accessible to the user when installing this module. You can hide the button in a your customer customization module if you want to clear the screen.

stock_traceability_operation/views/stock_production_lot_view.xml
```diff
@@ -0,0 +1,24 @@
...
+            <attribute name="attrs">{'invisible': True}</attribute>
+        </xpath>
+        <xpath expr="//button[@name='action_traceability']" position="after">
+            <button name="action_traceability_operation" icon="fa-arrows-v" class="oe_stat_button"
+                string="Traceability" type="object" attrs="{'invisible': [('quant_ids','=',[])]}"/>
```
pedrobaeza: Call it "Traceability (by operation)" or similar to differentiate it from the other button.

clonedagain:
@pedrobaeza @lbellier I've pushed a new version which keeps the standard buttons and calls the new ones "Detailed traceability". Do you agree with this one?

pedrobaeza:
Thanks :+1:

clonedagain:
Can you merge this please ?

clonedagain:
@nhomar can you please comment on this?

From the community list I understand you have a similar module but much more advanced, do you intend to hand it over to OCA ?

moylop260:
👍

PR #107, opened 2015-11-30
[IMP] make stock_reserve compatible with multi-company

JordiBForgeFlow:
The stock_reserve module will propose a default reserve location from ir_model_data. But what we found is that in multi-company environments a stock reservation will produce a cross-company stock move, because only one global location is available, and this location is internal to one single company.

This PR will add a flag in the stock location to indicate that it is meant to be used as a stock reservation location. Then you can create one reserved location for each company.

### DIFF
--- a/stock_available/i18n/de.po
+++ b/stock_available/i18n/de.po
@@ -3,54 +3,55 @@
 # * stock_available
 # 
 # Translators:
+# Rudolf Schnapka <rs@techno-flex.de>, 2016
 msgid ""
 msgstr ""
 "Project-Id-Version: stock-logistics-warehouse (8.0)\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2016-01-14 01:38+0000\n"
-"PO-Revision-Date: 2016-01-13 16:35+0000\n"
-"Last-Translator: OCA Transbot <transbot@odoo-community.org>\n"
-"Language-Team: English (http://www.transifex.com/oca/OCA-stock-logistics-warehouse-8-0/language/en/)\n"
+"PO-Revision-Date: 2016-01-14 10:16+0000\n"
+"Last-Translator: Rudolf Schnapka <rs@techno-flex.de>\n"
+"Language-Team: German (http://www.transifex.com/oca/OCA-stock-logistics-warehouse-8-0/language/de/)\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: \n"
-"Language: en\n"
+"Language: de\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 #. module: stock_available
 #: view:product.template:stock_available.view_stock_available_kanban
 msgid "Available to promise:"
-msgstr "Available to promise:"
+msgstr "Frei für Zusage:"
 
 #. module: stock_available
 #: field:stock.config.settings,module_stock_available_sale:0
 msgid "Exclude goods already in sale quotations"
-msgstr "Exclude goods already in sale quotations"
+msgstr "Produkte aus Angeboten ausschließen"
 
 #. module: stock_available
 #: field:stock.config.settings,module_stock_available_immediately:0
 msgid "Exclude incoming goods"
-msgstr "Exclude incoming goods"
+msgstr "Eingehende Ware ausschließen"
 
 #. module: stock_available
 #: field:stock.config.settings,module_stock_available_mrp:0
 msgid "Include the production potential"
-msgstr "Include the production potential"
+msgstr "Zugänge aus Fertigung berücksichtigen"
 
 #. module: stock_available
 #: model:ir.model,name:stock_available.model_product_product
 msgid "Product"
-msgstr "Product"
+msgstr "Produkt"
 
 #. module: stock_available
 #: model:ir.model,name:stock_available.model_product_template
 msgid "Product Template"
-msgstr "Product Template"
+msgstr "Produktvorlage"
 
 #. module: stock_available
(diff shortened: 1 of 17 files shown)


pedrobaeza:
Why don't simply put the location with the field company_id empty?

JordiBForgeFlow:
Because this would result on a move between a location from your own company to a location outside of your company, which from an inventory accounting perspective would trigger journal entries:
Debit: Stock Output
Credit: Stock Account

see in stock_account.py, line 144:
if company_from and (move.location_id.usage == 'internal' and move.location_dest_id.usage not in ('internal', 'transit') or company_from != company_to):

It is correct to tag some of your warehouse locations as 'reserved', but not correct to move them outside of your company.

pedrobaeza:
OK, thanks for the explanation. I think you should create one location per existing company then on a post-init-hook, instead of harcoding one in XML.

stock_reserve/__openerp__.py
```diff
@@ -32,6 +32,7 @@
              ],
  'demo': [],
  'data': ['view/stock_reserve.xml',
+          'view/stock_location.xml',
```
elicoidal: remove empty key
version should change as well: not sure 8.0.0.2.1 or 8.0.0.3.0

stock_reserve/data/stock_data.xml
```diff
@@ -3,6 +3,7 @@
     <data noupdate="1">
         <record id="stock_location_reservation" model="stock.location">
             <field name="name">Reservation Stock</field>
```
elicoidal, line 5: s/Stock/Location

stock_reserve/data/stock_data.xml
```diff
@@ -3,6 +3,7 @@
...
+            <field name="reserved">True</field>
             <field name="location_id" ref="stock.stock_location_locations"/>
```
elicoidal, line 6: remove 's' at the end of the id?

stock_reserve/__openerp__.py
pedrobaeza: 8.0.2.0.0, for 2 reasons:
- Having the full version nomenclature.
- This new version implies a migration script.

stock_reserve/model/__init__.py
```diff
@@ -19,5 +19,7 @@
 #
 ##############################################################################
 
+from . import stock_location
```
elicoidal: Please sort in alphabetical order following: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

stock_reserve/model/stock_location.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
elicoidal: Please use simple headers following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

stock_reserve/__openerp__.py
JordiBForgeFlow: @pedrobaeza what does a post-init-hook mean? Can you provide an example?

stock_reserve/data/stock_data.xml
JordiBForgeFlow, line 5: @elicoidal can you clarify the comment?

stock_reserve/data/stock_data.xml
JordiBForgeFlow, line 6: But 'stock_location_locations' is the default name given to that location by Odoo.

stock_reserve/__openerp__.py
pedrobaeza: Here you have one example:

https://github.com/OCA/manufacture/blob/8.0/quality_control_stock/__init__.py#L9

It's launched after installing the module. You can also follows last proposed convention about this: https://github.com/OCA/maintainer-tools/pull/166

stock_reserve/data/stock_data.xml
elicoidal, line 6: Is it correct stock.stock_location_location**s**?

elicoidal, line 6: If it is standard, forget about it
It is strange though because I thought ids wouldnot have plurals

JordiBForgeFlow:
Thanks @pedrobaeza and @elicoidal . Commited the changes as per your review comments.

stock_reserve/migrations/8.0.2.0.0/post-migration.py
```diff
@@ -16,8 +14,19 @@
 #
 #    You should have received a copy of the GNU Affero General Public License
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
 
-from . import stock_reserve
-from . import product
+
+def migrate(cr, installed_version):
```
pedrobaeza: I think this migration script is now useless

JordiBForgeFlow: The migration script intends to update the reserved locations already created in previous versions. The post-init hook is now going to create extra reserved locations for additional companies.

pedrobaeza: OK, understood.

pedrobaeza:
Please check Travis status

JordiBForgeFlow:
Travis and Runbot are now green

pedrobaeza:
:+1:

mileo:
:+1: LGTM

mileo:
If you can please put a configuration example:
- Physical locations / WH / STOCK
- Physical locations / WH / Reserved

NEVER LIKE THIS:
- Physical locations / WH / STOCK
- Physical locations / WH / STOCK / Reserved

[]s

mileo:
@jbeficent Why you close the pr?

PR #110, opened 2015-12-15
8.0 stock lot supplier importer

ismaelcj:
I propose a new module. It allows to import Serial Number (lots) by the supplier product_code (SKU - Stock-Keeping Unit) from a xls file.

The xls needs, at least, the columns 'sku_no' and 'serial_no' (no matter whether it is case sensitive). The program will search the file sku columns on 'Supplier Product Code' field from product supplier info.

### DIFF
--- a/stock_quant_manual_assign/i18n/nl.po
+++ b/stock_quant_manual_assign/i18n/nl.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: stock-logistics-warehouse (8.0)\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-11-18 03:44+0000\n"
-"PO-Revision-Date: 2016-11-25 14:57+0000\n"
+"POT-Creation-Date: 2017-02-01 01:51+0000\n"
+"PO-Revision-Date: 2017-02-01 11:20+0000\n"
 "Last-Translator: OCA Transbot <transbot@odoo-community.org>\n"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
What is the purpose of importing lots, but not quantities?

Thanks for the tests! :wink:

ismaelcj:
The purpose is to have the lots created before processing the picking in. The supplier give us a list of serial number associated with a sku number before send the products. So we have not to read the serial numbers of each product when they come.

stock_lot_supplier_importer/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# © 2015 FactorLibre - Ismael Calvo <ismael.calvo@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Lot Importer by Supplier",
+    "summary": "Import Lots from xls by product's SKU or name, from supplier.",
+    "version": "8.0.1.0.0",
+    "category": "arehouse Management",
```
pedrobaeza: s/arehouse/Warehouse

pedrobaeza:
Write the purpose then on the README to make it clear (and please respect 80 cols also on README.rst)

pedrobaeza:
You have to add external_dependencies key with xlrd python dependency on the manifest, and add also on .travis.yml the corresponding package to be installed in order to have Travis happy.

gurneyalex:
@ismaelcj could you rebase the PR in order to trigger a CI rebuild?

ismaelcj:
@gurneyalex, Is it now correct? I don't know why there is three commits...

PR #113, opened 2016-01-05
[IMP][stock_mts_mto_rule] set pull rule of customer location like spl…

JulioSerna:
…it_procurement to allow when warehouse is two/three steps delivery propagate move by mts-mto in pick/pack/out

### DIFF
--- a/stock_mts_mto_rule/model/warehouse.py
+++ b/stock_mts_mto_rule/model/warehouse.py
@@ -70,6 +70,26 @@ class Warehouse(models.Model):
             'picking_type_id': warehouse.mto_pull_id.picking_type_id.id,
         }
 
+    @api.model
+    def _get_push_pull_rules(self, warehouse, active, values, new_route_id):
+        pull_obj = self.env['procurement.rule']
+        res = super(Warehouse, self)._get_push_pull_rules(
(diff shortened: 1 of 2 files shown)


nhomar:
:+1:

nhomar:
Hello....

Do I need another +1 here @jbeficent @pedrobaeza @rvalyi  @ somebody :-)

moylop260:
:+1: wo tests

pedrobaeza:
:+1: (code review)

asoral:
I have an issue on the latest version of MTO/MTS used in odoo8 (recent)
1. create a chained pull rules like -> when a sales order is created procurement chain is created like - Customer location <- sales Staging <- MTO/MTS <- MO/stock move
2. Cancel the procurement MTO/MTS, it does not have a STOCK MOVE associated since it was a dummy procurement without a stock move to check the inventory level
3. System will not propagate the cancellation because it looks for the stock move in the procurement to cancel but this procurement will have none, so it is not canceled.

moylop260:
@asoral 
Please, Could you create a [new issue](https://github.com/OCA/stock-logistics-warehouse/issues/new)?
Explain Steps, Saw, Expected, sha's of your environment, add images or video of your steps

PR #115, opened 2016-01-28
8.0 add column "Available to sale" to saleorder

lbellier:
Hello.
I propose this patch to add column "Available for sale" on the sale order view.
This is very helpful when a person take a command on phone for example. He can directly tell to his customer he don't have product on stock.


clonedagain:
your proposal is targeting v9 but I think only v8 is tested

lbellier:
Bad branch ! sorry...

PR #116, opened 2016-01-28
[IMP]: Add column "Available for sale" on sale order view.

lbellier:
Hello,
I propose this patch to add a new column "Available for sale" on the sale order view.
This can be helpful to directly inform customer than the product he want is out of stock.


clonedagain:
:+1: looks fine and useful.

lbellier:
All green !!!

lmignon:
@lbellier that looks a useful functionality but a shortened title for the new column will improve the display.
image

lbellier:
Why not... 
"Available" ?
"Avail. 2 promise" ?
Any suggestions ?

gurneyalex:
globally favorable on this, but I'd like a modification: I've used a similar config on a customer deployement, and we had a significant perf hit when displaying SO with a large number of lines. What I would suggest is:
- add a group and a Sales setting to enable the group
- make the view specific to this group (don't set the visiblility of the column based on the group, as this will not prevent the computation from happening)

@clonedagain, @lmignon @lbellier your opinion on the above suggestion is of course very welcome.

lmignon:
@gurneyalex Thank you for your advice. I've not taken into account the potential performance issue on first read. But the way you propose to modify the code to be able to limit this potential issue to only members of a specific group looks good to me.

clonedagain:
@gurneyalex how do I "make the view specific to this group" ? can you point me to an example please?

stock_available_sale/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# © 2016 Numérigraphe SARL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
+
+
+class SaleOrder(models.Model):
```
alexis-via, line 8: SaleOrderLine

gurneyalex:
@clonedagain sorry for taking so long to answer... See for instance addons/sale/sale_view.xml the view with xml_id `view_order_form_editable_list` https://github.com/odoo/odoo/blob/8.0/addons/sale/sale_view.xml#L230 : this inheriting `ir.ui.view` has a groups_id field, that way the changes made in it are only visible to users from these groups.

clonedagain:
Thanks @gurneyalex . We've moved on to other projects so I'm not sure we can amend this proposal soon.
If someone feels like making a better PR we'll help review it.

Cedric-Pigeon:
@clonedagain Hello, do you plan to achieve this PR ? Thanks

rousseldenis:
The original branch of this has been deleted, so closing

PR #118, opened 2016-02-10
[ADD] stock_inventory_revaluation

JordiBForgeFlow:
# 
# Stock Account Inventory Revaluation

If your company runs a perpetual inventory system, you may need to perform
inventory revaluation.

You can re-valuate inventory values by:
- Changing the price for a specific product. The inventory price is changed
  and inventory value is recalculated according to the new price. In case of
  real price, you can select which quants you want to change the price on.
- Changing the value of the inventory. The quantity of inventory remains
  unchanged, resulting in a change in the price.
# Configuration
- Go to _Inventory / Configuration / Products / Product Categories_ and
  define, for each category, a Valuation Increase Account and a Valuation
  Decrease Account. These accounts will be used as contra-accounts to the
  Stock Valuation Account during the inventory re-valuation.
# Usage
- Go to _Inventory / Inventory Control / Inventory Revaluation / Products_
  to create a new Inventory Revaluation.


stock_inventory_revaluation/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+Configuration
+=============
+
+
+* Go to *Inventory / Configuration / Products / Product Categories* and
+  define, for each category, a Valuation Increase Account and a Valuation
+  Decrease Account. These accounts will be used as contra-accounts to the
```
pedrobaeza, line 29: s/contra/counterpart

stock_inventory_revaluation/README.rst
```diff
@@ -0,0 +1,87 @@
...
+  Stock Valuation Account during the inventory re-valuation.
```
pedrobaeza, line 30: s/re-valuation/revaluation

stock_inventory_revaluation/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+* Eficent Business and IT Consulting Services S.L. <contact@eficent.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
```
elicoidal, line 77: https (x2)

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,478 @@
...
+                       states={'draft': [('readonly', False)]},
+                       default='/')
+
+    revaluation_type = fields.Selection(
+        [('price_change', 'Price Change'),
+         ('inventory_value', 'Inventory Debit/Credit')],
+        string="Revaluation Type",
+        readonly=True, required=True,
+        default='price_change',
+        help="'Price Change': You can re-valuate inventory values by Changing "
```
elicoidal, line 40: s/Changing/changing

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,478 @@
...
+             "the price for a specific product. The inventory price is "
+             "changed and inventory value is recalculated according to the "
+             "new price.\n "
+             "'Inventory Debit/Credit': Changing the value of the inventory. "
+             "The quantity of inventory remains unchanged, resulting in a "
+             "change in the price",
+        states={'draft': [('readonly', False)]})
+
+    remarks = fields.Text('Remarks',
+                          help="Displays by default Inventory Revaluation. "
```
elicoidal, line 50: By default Inventory Revaluation.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,478 @@
...
+    def _create_accounting_entry(self, amount_diff):
+        timenow = time.strftime('%Y-%m-%d')
+        move_data = self._prepare_move_data(timenow)
+        datas = self.env['product.template'].get_product_accounts(
+            self.product_template_id.id)
+        self.move_id = self.env['account.move'].create(move_data).id
+        move_line_obj = self.env['account.move.line']
+
+        if not self.decrease_account_id or not self.increase_account_id:
+            raise UserError(_("Please add a  Increase Account and "
```
elicoidal, line 327: Valuation Increase Account and Valuatin Decrease Account

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,478 @@
...
+                if not diff:
+                    raise UserError(_("No difference between current and new "
+                                      "valuation for product %s."
+                                      % self.product_template_id.name))
+                if self.new_value < 0:
+                    raise UserError(_("The new value for product %s cannot be "
+                                      "negative"
+                                      % self.product_template_id.name))
+                if self.qty_available <= 0.0:
+                    raise UserError(_("Cannot do an inventory value change if "
```
elicoidal: Inventory value change not possible if

stock_inventory_revaluation/security/stock_inventory_revaluation_security.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data noupdate="1">
+
+    <record model="ir.rule" id="stock_inventory_revaluation_comp_rule">
```
elicoidal, line 5: Did I miss something or there is no ACL group for changing revaluation?

nhomar:
@jbeficent 

I am a little confused, this feature is already working on odoo (at least conceptually what you mentioned), can you share with us which is the difference between your approach and the one in odoo?

Thanks.

stock_inventory_revaluation/wizards/stock_inventory_revaluation_get_quants.py
```diff
@@ -0,0 +1,60 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models, _
+
+
+class StockInventoryRevaluationGetQuants(models.TransientModel):
+
+    _name = 'stock.inventory.revaluation.line.get.quant'
+    _description = 'Inventory revaluation line get Quants'
+
+    date_from = fields.Date('Date From')
```
elicoidal, line 13: From?
Until?

stock_inventory_revaluation/wizards/stock_inventory_revaluation_get_quants_view.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+
+        <record model="ir.ui.view"
+                id="stock_inventory_revaluation_line_get_quant_form">
+          <field name="name">stock.inventory.revaluation.line.get.quant.form</field>
+          <field name="model">stock.inventory.revaluation.line.get.quant</field>
+          <field name="arch" type="xml">
+            <form string="Get Quants">
```
elicoidal, line 10: Select or Pick?

stock_inventory_revaluation/wizards/stock_inventory_revaluation_get_quants_view.xml
```diff
@@ -0,0 +1,38 @@
...
+                or
+                <button string="Cancel" class="oe_link" special="cancel"/>
+              </footer>
+            </form>
+          </field>
+        </record>
+
+        <record id="action_stock_inventory_revaluation_line_get_quant"
+                model="ir.actions.act_window">
+          <field name="name">Get Quants for Inventory Revaluation</field>
```
elicoidal, line 27: Pick

elicoidal:
Overall LGTM!

JordiBForgeFlow:
@nhomar, I can't really find now in Odoo the feature tha allows one to change the cost of a quant, with the subsequent accounting entries.

Support for change of standard and average is possible in odoo, but only through a wizard. If you  want to apply a revaluation for many products it's not user friendly.

That's why I thought of integrating thw whole inventory revaluation in one place. This is really a mimick of SAP B1 inventory revaluation tool. Seehttps://m.youtube.com/watch?v=CWUizZ8Twpk

JordiBForgeFlow:
@nhom also the contra account of inventory revaluation in standard is wrong. It's not correct to use stock input and stock output! The user should be able to define, and the proposal should be different from stock in/out

nhomar:
@jbeficent 

Well we just finished a huge reevaluation inventory process in a 10k SKU company with 4 subsidiaries.

I need to say "yes" to the fact that some stuff was done in order to achieve such process.

For example, in th SAP video you just show I have a quick question (which is the main concern of my question).

Where

The example number you are putting there is the one you set actually on the standard_cost field, and the wizard is necesary in order to achieve the account move topics. IF we are talking of 1 product IMHO If I reevaluate the inventory the stock in/out account is the one affected and teh ones of dofference account (all already in odoo).

You can see some little fixes we did [here](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/product_extended_variants/model/template.py) and we ran a [cronjob](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/product_extended_variants/wizard/wizard_price.py#L68) in order to set the new standard for those which depends of a BoM price, but all using the original objects, no new fields, no new behaviors just "fixes".

That's my point here is that may be I am not seeing correctly the value of the new feature, and I want to be enlghted whether you are right or we forgot something huge on our side ;-)

As a side Note. With those approached I am showing you we achieve have consistent:

Accounting - Quants - Sotck Moves. All say the same in the moment th cronjob runs, and all account moves are generatd accordingly with correct NIIF approach (nothing special just to say a pretty name to what odoo is doing right now).

It should be GREAT if you share a little video of your approach, because I think that the account you are adding to the category is not necesary (but may be I am worng) and I think the actual Odoo'd algorithms only need little fixes (and not the creation of a new one).

But YES this is a hard topic our analysis, and deployed results took to us something like 4 months in order to follow all the auditory and so on (well we manage financially segmentation costs but that's another history).

If you want we can have a little meeeting in spanish to understand better and maybe we can help on this PR a little bit more.

Regard.s

JordiBForgeFlow:
@nhomar In your module [product_extended_variants](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/product_extended_variants):
- You are overriding the standard method `do_change_standard_price` to insert the price difference account.
- The user is not given the option to change the GL contra-accounts accounts. An inventory revaluation should use separate GL accounts. Price Differences account is to be used when there are differences between invoice price and PO price under a standard price cost method. An accountant would be confused when seeing a "Price Differences" account mixing inventory revaluations with Invoice-PO adjustments. That's why it's important to PROPOSE to the user what account should be used. In some scenarios the GL account to use may be different (e.g. having a specific 'Inventory Obsolescence account' that is used in specific revaluations. I reported this issue to Odoo [here](https://github.com/odoo/odoo/issues/9856), and they will be implementing the proposed changed in master [here](https://github.com/odoo/odoo/commit/40bd4a2f051996d4ab5ca7956562c8f552c4c969).

Odoo's module [product_extended](https://github.com/odoo/odoo/tree/8.0/addons/product_extended):
- Provides support to change a product that is manufactured, and has BOM, only under the condition that the product is managed under standard price. Does not support revaluation under average cost method.

No tool in Odoo now supports inventory revaluation under the real cost method. That is, change the cost of the quants. 

The module proposed here covers all possible scenarios of price revaluation, and ideally, once Odoo has proper support for price change with ability to define what GL account should be use, we could call the same wizard from this module, and thus this would really represent another way to achieve the same thing.

Another option is to fix / alter the standard methods offered by odoo for standard & average (possibly using a 'hooks' module that overrides the standard), and we build a separate tool for inventory revaluation specifically suited for the 'real' cost method. Perhaps it's better from a consistency perspective. Your feedback is welcome here.

I'm not sure that I see how your approach can achieve Accounting = Quants = Stock Moves. When you change a standard price of a product (standard or average price) the cost of quants is really not changing. And that's fine, because you're not operating under real cost method. Note that the field 'Inventory Value' of a quant is calculated field that gives you the product's standar price under standard / average and the true cost of the quant under real cost method.

From your statements I can see that your use case of inventory revaluation dealt with standard prices in raw materials and finished products. I have the same issue, where a company has screwed inventory valuation due to lack of proper maintenance of standard prices. That's why I'm going for a real costing approach, where you ensure that COGS will be recognized under FIFO costing method, which is the most accurate method of COGS recongition and inventory valuation as well.

JordiBForgeFlow:
@nhomar I'd be more than happy to discuss this with you in a meeting. You can reach me on skype or Hangouts.

jgrandguillaume:
Hi,

That's a very good work, again @jbeficent thank you for all your work. The only remark, such a module would deserve a little test IMO as it really have important impact.

Otherwise LGTM :+1: 

Regards,

Joël

JordiBForgeFlow:
For some reason Runbot is killed with the following error message:
2016-02-23 12:24:49,999 1505 ERROR ? openerp.sql_db: Connection to the database failed
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3142379-118-a0b00e/openerp/sql_db.py", line 560, in borrow
    result = psycopg2.connect(dsn=dsn, connection_factory=PsycoConnection)
  File "/srv/openerp/instances/openerp-oca-runbot/sandbox/local/lib/python2.7/site-packages/psycopg2/**init**.py", line 179, in connect
    connection_factory=connection_factory, async=async)
OperationalError: could not connect to server: No such file or directory
    Is the server running locally and accepting
    connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?

JordiBForgeFlow:
Runbot is killed because I created a branch from 9.0, not from 8.0. I'm closing this PR and creating a separate one, https://github.com/OCA/stock-logistics-warehouse/pull/121, that uses the correct 8.0 branch.

PR #119, opened 2016-02-15
[IMP] stock_quant_merge: Merge quant at install + hook for inherit merge

pedrobaeza:
This PR improves 2 things:
- It provides an installation hook to clean up existing split quants.
- It provides a hook for inheriting the values that are written in the merged quant.

### DIFF
--- a/stock_quant_merge/README.rst
+++ b/stock_quant_merge/README.rst
@@ -19,7 +19,8 @@ Installation
 ============
 
 On installation time, a first scan is done over all your quants to see which
-ones are split, and join them.
+ones are split, and join them, which can take a while, so this may fail to
+install on big databases if the CPU limit is too low.
 
 Usage
 =====
--- a/stock_quant_merge/README.rst
+++ b/stock_quant_merge/README.rst
@@ -7,12 +7,19 @@ Quant merge
 ===========
 
 Odoo splits quants each time a reservation is done: this module makes Odoo
-merge them back if they still meet the following requirements:
+merge them back when you unserve them if they still meet the following
+requirements:
 
 * same product
 * same serial number/lot
 * same location
 * same package
+
+Installation
+============
+
+On installation time, a first scan is done over all your quants to see which
+ones are split, and join them.
 
 Usage
 =====
--- a/stock_quant_merge/models/stock.py
+++ b/stock_quant_merge/models/stock.py
@@ -19,14 +19,20 @@ class StockQuant(models.Model):
                 ('lot_id', '=', self.lot_id.id),
                 ('package_id', '=', self.package_id.id),
                 ('location_id', '=', self.location_id.id),
-                ('reservation_id', '=', False),
+                ('reservation_id', '=', self.reservation_id.id),
                 ('propagated_from_id', '=', self.propagated_from_id.id)]
+
+    @api.model
+    def _get_merged_vals(self, quant1, quant2):
+        return {
+            'qty': quant1.qty + quant2.qty,
+        }
 
     @api.multi
     def merge_stock_quants(self):
         # Get a copy of the recorset
         pending_quants = self.browse(self.ids)
-        for quant2merge in self.filtered(lambda x: not x.reservation_id):
+        for quant2merge in self:
             if quant2merge in pending_quants:
                 quants = self.search(quant2merge._mergeable_domain())
                 cont = 1
@@ -34,7 +40,8 @@ class StockQuant(models.Model):
                 for quant in quants:
                     if (self._get_latest_move(quant2merge) ==
                             self._get_latest_move(quant)):
-                        quant2merge.sudo().qty += quant.qty
+                        quant2merge.sudo().write(
+                            self._get_merged_vals(quant2merge, quant))
                         cost += quant.cost
                         cont += 1
                         pending_quants -= quant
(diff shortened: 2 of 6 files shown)


pedrobaeza:
cc @oihane @clonedagain

stock_quant_merge/models/stock.py
```diff
@@ -19,14 +19,20 @@ def _mergeable_domain(self):
                 ('lot_id', '=', self.lot_id.id),
                 ('package_id', '=', self.package_id.id),
                 ('location_id', '=', self.location_id.id),
-                ('reservation_id', '=', False),
+                ('reservation_id', '=', self.reservation_id.id),
```
clonedagain, line 22: is this change related to the new features or is it a bug fix?

clonedagain:
That looks sane, but the tests are not passing anymore. Maybe you need to update them since you use another product?

stock_quant_merge/models/stock.py
pedrobaeza, line 22: This is to allow a merging from the init hook of any of possible quants. Imagine that you have reserved 2 quants for the same move. This one allows to merge then in only one.

pedrobaeza:
Sorry about the test. I try only with the second test (the one that I have added), and forget the first one. I'll change it.

clonedagain:
@pedrobaeza hehe so it does happen even to the best of us :)
No problem, just ping me when it's fixed.

pedrobaeza:
Done

clonedagain:
Looks fine from reading the code. I'll try to test it on our big DB (~3M quants) but I fear the install may be too long for default CPU limits. If such is the case, maybe a warning should be added to the README.

pedrobaeza:
I have already added https://github.com/OCA/stock-logistics-warehouse/pull/119/files#diff-674e9db8ba7d8f6cd1fd1c408bc5d61bR22. Isn't enough?

clonedagain:
I was specifically thinking of a warning in the line of "this may fail to install on big databases if the CPU limit is too low"

pedrobaeza:
Warning added on the README

pedrobaeza:
Can you please review this again?

stock_quant_merge/README.rst
```diff
@@ -7,13 +7,21 @@ Quant merge
 ===========
 
 Odoo splits quants each time a reservation is done: this module makes Odoo
-merge them back if they still meet the following requirements:
+merge them back when you unserve them if they still meet the following
```
gurneyalex, line 10: s/unserve/unreserve/

gurneyalex:
this looks quite nice. :+1: once small typo is fixed. 

I'm wondering: why should I install this? Is there a perf gain on a db with a large number of quants? Something else? If so, I think it deserves mentioning in the README. 

And if there is a huge perf gain, maybe we could make a PR on odoo master?

pedrobaeza:
@gurneyalex, this is more or less like the Windows Disk Defragmenter, but with quants instead of sectors.

clonedagain:
@pedrobaeza I was wondering if doing this would not introduce wrong results to the dated valuation?
In the view `stock_history` the standard joins the moves and the quants, wouldn't merging the quants make it count wrong quantities?

pedrobaeza:
Well, I don't know, because I haven't digged in that view. It depends on what quantity field takes. Can you check it?

stock_quant_merge/models/stock.py
clonedagain, line 22: I think if we merge quants with different "in date" values, we'll be introducing wrong results in the FIFO/LIFO removal strategy.

pedrobaeza, line 22: OK, we can add that condition on the mergeable domain.

clonedagain:
My understanding is that the `stock_history` counts the quantity of the moved quants, so I think that if we merge all the possible quants we'll be counting some quantities double.

So I think we should only merge quants which are used in the same moves.

stock_quant_merge/models/stock.py
```diff
@@ -34,7 +40,8 @@ def merge_stock_quants(self):
                 for quant in quants:
                     if (self._get_latest_move(quant2merge) ==
                             self._get_latest_move(quant)):
-                        quant2merge.sudo().qty += quant.qty
+                        quant2merge.sudo().write(
+                            self._get_merged_vals(quant2merge, quant))
                         cost += quant.cost
                         cont += 1
```
clonedagain, line 46: If you're going to merge all the possible quants, some will have different costs, so you should be doing a weighted average like the standard does in the dated valuation (it's changed recently: https://github.com/odoo/odoo/commit/1d7c12022dcffce4792b9eec85706e19338cfa9f)

gurneyalex:
setting needs fixing following @clonedagain's remark

pedrobaeza:
cc @Tecnativa

rafaelbn:
@pedrobaeza could you please review comments of this PR? Thanks

gurneyalex:
needs rebasing

yajo:
Code seems alright, glitches have already been commented. Waiting for fixes and rebase to do functional review (note to self: uninstall module, unreserve quant, install module, only 1 quant should be there).

pedrobaeza:
We have lost interest in this after handling this with the customer in another way. If someone wants to take over the work, please make another PR. I close meanwhile.

PR #124, opened 2016-03-01
[8.0] stock_available_mrp : configure compute potential

laetitia-gangloff:
On configuration add new field : stock_available_mrp_based_on to configure which field to use when compute potential.

For example, in my case, I want potential quantity based on available to promise and not on quantity on hand.

### DIFF
--- a/stock_available/models/res_config.py
+++ b/stock_available/models/res_config.py
@@ -2,12 +2,21 @@
 # © 2014 Numérigraphe SARL
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields
+from openerp import api, models, fields
+from openerp.tools.safe_eval import safe_eval
 
 
 class StockConfig(models.TransientModel):
     """Add options to easily install the submodules"""
     _inherit = 'stock.config.settings'
+
+    @api.model
+    def _get_stock_available_mrp_based_on(self):
+        """Gets the available languages for the selection."""
+        fields = self.env['ir.model.fields'].search(
+            [('model', '=', 'product.product'),
+             ('ttype', '=', 'float')])
+        return [(field.name, field.field_description) for field in fields]
 
     module_stock_available_immediately = fields.Boolean(
         string='Exclude incoming goods',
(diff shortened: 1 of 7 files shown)


gurneyalex:
@laetitia-gangloff I like this. Can you update the README.rst file to explain the new feature?

laetitia-gangloff:
@gurneyalex thank you !
I updated README.rst.

cyrilgdn:
@laetitia-gangloff @gurneyalex Discussion on this feature also in PR #136 (similar PR)

stock_available/models/res_config.py
```diff
@@ -31,3 +40,27 @@ class StockConfig(models.TransientModel):
...
+        if not res['stock_available_mrp_based_on']:
+            res['stock_available_mrp_based_on'] = 'qty_available'
+        return res
+
+    @api.multi
+    def set_stock_available_mrp_based_on(self):
+        if self.stock_available_mrp_based_on:
+            icp = self.env['ir.config_parameter']
+            icp.set_param('stock_available_mrp_based_on',
+                          repr(self.stock_available_mrp_based_on))
```
cyrilgdn, line 66: Why save the repr() of field's name and safe_eval() after? Why not directly saved the field's name ?
(and call `icp.get_param('stock_available_mrp_based_on', 'qty_available')` )

stock_available/views/res_config_view.xml
```diff
@@ -22,6 +22,8 @@
                                 <div>
                                     <field name="module_stock_available_mrp" class="oe_inline" />
                                     <label for="module_stock_available_mrp" />
+                                    <label for="stock_available_mrp_based_on" />
+                                    <field name="stock_available_mrp_based_on" class="oe_inline" attrs="{'required':[('module_stock_available_mrp','=',True)]}"/>
```
cyrilgdn, line 26: Is there a reason not to put this new setting directly in an inherit model in stock_available_mrp (so this PR changes only stock_available_mrp)?

(Sorry if this question is strange, I'm a new Odoo developer :-))

laetitia-gangloff, line 26: Yes, there is a reason :)

If you set the option in stock_available_mrp, you should first install stock_available_mrp (check the box and valid) and in a second step you can configure the stock field. If the option is in stock_available, you can directly configure all you need in one step.

stock_available/models/res_config.py
laetitia-gangloff, line 66: Is it not because stock_available_mrp_based_on is not mandatory ? (not sure, I think it is copy/paste before all :()

stock_available/views/res_config_view.xml
cyrilgdn, line 26: Ok !  Thank you!

cyrilgdn:
PR #136 updated with your commit merged 

@laetitia-gangloff  I remove safe_eval and repr, seems working and a little nicer ;-)
If you still have a database with key stock_available_mrp_based_on already configure you will need to execute a query like:
`update ir_config_parameter set value = regexp_replace(value, '(^u''|'')', '', 'g') where key = 'stock_available_mrp_based_on';`

laetitia-gangloff:
In PR #136

PR #125, opened 2016-03-01
fixes procurement name

levkar:
Stock move names are taken from rule instead of procurement. Normally it shows product name on other rules but it becomes "WH:Stock -> Customer"  in mto_mts_rule. (It is not intentional, is it?)

This commit turns is to product name.


levkar:
sorry this should go to 8.0

Issue #111, opened 2015-12-30
stock_mts_mto_rule :: extra moves created when recreate delivery order


ghost:
product A = 5 units ( in stock)
make product A ( MTO+MTS )
make new sale order and order 10 units of product A
confirm and recheck availability in delivery order
confirm and receive products from the PO made
cancel the delivery order and recreate it from sale order
!!! now delivery order has extra lines

levkar:
I confirm this bug but more. You don't even have to confirm and receive PO. Just cancel delivery order and recreate it from the sale order to see the magic.

Things really get serious once a customer realises that he gave some duplicate orders! 

FYI
@florian-dacosta @pedrobaeza

StefanRijnhart:
Closing as the issue of duplicate lines was fixed in the attached PR

PR #132, opened 2016-03-10
Do no run mts mto procurement if it already has been splited into at least another procurement

florian-dacosta:
Fix the issue https://github.com/OCA/stock-logistics-warehouse/issues/111

But the fix is not perfect, that is why I have updated the readme file to put a known issue.

Explanation : 
With mts+mto rule, a procurement is created and then splited into 1 or 2 procurements depending of the stock level:  one MTS and the other MTO.
If we cancel the delivery order, the 3 procurement are also canceled.
If we recreate the delivery order, from the sale order form, then, the 3 procurement are back to confirm state (because they are all linked to sale order line) and are run again.
So the mto procurement create one move, the mts procurement create one move, and the mts+mto procurement split again and create 2 more procurements which create also 2 moves.
Consequently, the quantity in the delivery order is doubled.

I made a quick fix so if the mts+mto procurement has already been splited, it can't do it again.
So if we run it again, it won't do anything.
The problem is that when we recreate the delivery order, we won't take the actual stock to split into mts and mto procurement. So the old mto and mts are taken from the old procurements.
Still, it is way better than have too much product in the delivery order.

If you have any idea to improve it, I would be glad to hear it.
An idea would be not to propagate sale_line_id on the splited procurements, but it may have other bad consequences. And I think it is better if they are linked to a sale order line anyway...

### DIFF
--- a/stock_mts_mto_rule/README.rst
+++ b/stock_mts_mto_rule/README.rst
@@ -49,6 +49,15 @@ Configuration
 
 You have to select 'Use MTO+MTS rules' on the company's warehouse form.
 
+Known issues
+============
+
+If you cancel a delivery order and then recreate it from Recreate
+Delivery Order button in sale order form, then the stock level at the time of
+the Re-Creation won't be taken into account. So if a purchase order was created
+when the sale order was first validated, a similar purchase order will be created
+during the Re-creation of the delivery order, even if not needed regarding the actual stock.
+
 Bug Tracker
 ===========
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can't you do this with a proper configuration of the cancel propagation?

florian-dacosta:
What do you mean? 
The problem is not during the cancellation.
It comes from the method `_ship_create method`, which put all 3 procurements back to confirmed state, even if we juste want the mts+mto procurement to go back to confirmed state.

pedrobaeza:
OK, I understand

:+1:

gurneyalex:
:+1:

PR #135, opened 2016-03-16
stock_available_mrp: fix BOM qty (multi-units, efficiency)

cyrilgdn:
Fix BOM quantity, manage: 
- Multi units for same component in multi-levels BOM
- BOM produced qty (with unit)

### DIFF
--- a/stock_available_mrp/tests/test_potential_qty.py
+++ b/stock_available_mrp/tests/test_potential_qty.py
@@ -12,6 +12,14 @@ class TestPotentialQty(TransactionCase):
     def setUp(self):
         super(TestPotentialQty, self).setUp()
 
+        self.product_model = self.env["product.product"]
+        self.bom_model = self.env["mrp.bom"]
+        self.bom_line_model = self.env["mrp.bom.line"]
+        self.stock_quant_model = self.env["stock.quant"]
+
+        self.setup_demo_data()
+
+    def setup_demo_data(self):
         #  An interesting product (multi-line BoM, variants)
         self.tmpl = self.browse_ref(
             'product.product_product_4_product_template')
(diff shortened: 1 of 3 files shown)


clonedagain:
@lbellier we should review this, it seems there is a bug in stock-available-mrp.
@cyrilgdn we'll need a bit of time to understand your patch.

laetitia-gangloff:
:+1: (functional)

gurneyalex:
:+1:

@clonedagain @lbellier The bug fixed by this PR will not bite you unless you have complicated BOMs e.g. using non default UoM for the produced qty, or using an efficiency != 100%, or unless a product appears in different BOMs with different units.

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -12,6 +12,14 @@ class TestPotentialQty(TransactionCase):
...
         super(TestPotentialQty, self).setUp()
 
+        self.product_model = self.env["product.product"]
+        self.bom_model = self.env["mrp.bom"]
+        self.bom_line_model = self.env["mrp.bom.line"]
+        self.stock_quant_model = self.env["stock.quant"]
+
+        self.setup_demo_data()
+
+    def setup_demo_data(self):
```
clonedagain, line 22: Why split this into another method?

cyrilgdn, line 22: @clonedagain Just to separate loading of models and datas.
(Some tests don't use demo datas. Although this data are still loaded, I can more easily see the part that interests me.)

clonedagain, line 22: Would `SavepointCase` help you with this? You'd set up the demo data once for all at class setup and you can get the models in instance setup.
(not sure if it's more readable to put models here, but I won't object).

cyrilgdn, line 22: It was just for readability, if you're not agree I can revert if you want. (and rebase other PRs)

PR #136, opened 2016-03-16
BOM stock: add the ability to choose the type of stock for components.

cyrilgdn:
Depends on PR #135.

Merge  PR #124 which add the ability to configure the type of stock for components to use to compute BOM stock (between product's float fields) .

Add small improvements and a test for field choice.

### DIFF
--- a/stock_available/models/res_config.py
+++ b/stock_available/models/res_config.py
@@ -2,12 +2,21 @@
 # © 2014 Numérigraphe SARL
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields
+from openerp import api, models, fields
+from openerp.tools.safe_eval import safe_eval
 
 
 class StockConfig(models.TransientModel):
     """Add options to easily install the submodules"""
     _inherit = 'stock.config.settings'
+
+    @api.model
+    def _get_stock_available_mrp_based_on(self):
+        """Gets the available languages for the selection."""
+        fields = self.env['ir.model.fields'].search(
+            [('model', '=', 'product.product'),
+             ('ttype', '=', 'float')])
+        return [(field.name, field.field_description) for field in fields]
 
     module_stock_available_immediately = fields.Boolean(
         string='Exclude incoming goods',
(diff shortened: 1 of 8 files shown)


laetitia-gangloff:
It seems similar to this PR : https://github.com/OCA/stock-logistics-warehouse/pull/124

cyrilgdn:
Hello @laetitia-gangloff ,

Thanks for your comment. Indeed, this PR adds the same feature than yours. 
I reviewed your commit and put the field setting in StockConfig is much better than in company's setting! 

But I think that basing this setting on all float product's fields is a bit strange because it's possible to choose price, weight, ... for computing the bom stock.

In this PR, I wrote a little test for this feature and this PR depends on PR #135 which fix some bugs.

So, if you want, I can modify this PR by:
- Take the StockConfig setting in your commit
- Fix the possible field's list (real, virtual, immediately available)
- Adapt the model and my test to use StockConfig

Regards.

laetitia-gangloff:
I agree to set all list of float fields is a bit strange, but it is to stay coherent with this module : 
https://github.com/OCA/connector-magento/blob/8.0/magentoerpconnect/magento_model.py#L153

And the other advantage is, that it is possible to easily configure another custom quantity field (for example in a multi-warehouse environment, if you want only the quantity in one warehouse).

cyrilgdn:
Ok, so if you agree I can merge your modifications in my PR?

I just add two comments in your PR, so i'll wait for your answers on both PRs for merge.

laetitia-gangloff:
It is OK for me.

cyrilgdn:
Rebase on @laetitia-gangloff PR #124

laetitia-gangloff:
Thank You !
I will try to take time Monday to review this one and close the #124

clonedagain:
The feature is interesting. @lbellier do you think we can spare some time to look into this?

gurneyalex:
:+1:

laetitia-gangloff:
:+1:

lmignon:
:+1:

stock_available/models/res_config.py
```diff
@@ -31,3 +42,27 @@ class StockConfig(models.TransientModel):
...
+    stock_available_mrp_based_on = fields.Selection(
+        _get_stock_available_mrp_based_on,
+        string='based on',
+        help="Choose the field of the product which will be used to compute "
+             "potential.\nIf empty, Quantity On Hand is used.\n"
+             "Only the quantity fields have meaning for computing stock",
+    )
+
+    @api.model
+    def get_default_stock_available_mrp_based_on(self, fields):
```
clonedagain, line 55: coveralls says this methods are never called and I don't see it used anywhere either.
Should you remove it ?

stock_available/models/res_config.py
```diff
@@ -31,3 +42,27 @@ class StockConfig(models.TransientModel):
...
+        res = {}
+        icp = self.env['ir.config_parameter']
+        res['stock_available_mrp_based_on'] = icp.get_param(
+            'stock_available_mrp_based_on', 'qty_available'
+        )
+        return res
+
+    @api.multi
+    def set_stock_available_mrp_based_on(self):
```
clonedagain, line 64: Same here: not called

clonedagain:
~~Please do not fast-merge until we find out why coveralls is red.~~

stock_available/models/res_config.py
cyrilgdn, line 55: This method is used by res_config, returns the value to display in settings if none exists in database.

re_config search all methods starting with 'get_default_', look at openerp.addons.base.res.res_config.default_get.

stock_available/models/res_config.py
cyrilgdn, line 64: Same, loot at 'execute' method in same file.

clonedagain:
Coveralls complains about methods auto-discovered by the standard. Not sure how we can have this tested in unittests. Maybe we should merge as-is?

stock_available/models/res_config.py
clonedagain, line 55: OK fine!

atchuthan:
code reviewed:+1:

PR #142, opened 2016-03-21
[9.0] update stock_available / stock_available_immediately

gurneyalex:
The PRs porting stock_available and stock_available_immediately to v9 did not use the latest available version from 8.0. 

cherry-picked commits:
- 6e5525bd0de1b9f00ab46ddbb211f92a32d4e01e
- 0ba72645aa6c9fb4f94b63c1ea58b24736bc08bc

### DIFF
--- a/stock_available/__openerp__.py
+++ b/stock_available/__openerp__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Stock available to promise',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     "author": "Numérigraphe, Sodexis, Odoo Community Association (OCA)",
     'category': 'Warehouse',
     'depends': ['stock'],
--- a/stock_available_immediately/__openerp__.py
+++ b/stock_available_immediately/__openerp__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Ignore planned receptions in quantity available to promise",
-    "version": "9.0.1.0.0",
+    "version": "9.0.1.1.0",
     "depends": ["stock_available"],
     "author": "Camptocamp,Sodexis,Odoo Community Association (OCA)",
     "license": "AGPL-3",
--- a/stock_available/models/product_product.py
+++ b/stock_available/models/product_product.py
@@ -21,8 +21,11 @@ class ProductProduct(models.Model):
 
         By default, available to promise = forecasted quantity.
 
-        Must be overridden by another module that actually implement
-        computations."""
+        **Each** sub-module **must** override this method in **both**
+            `product.product` **and** `product.template`, because we can't
+            decide in advance how to compute the template's quantity from the
+            variants.
+        """
         for prod in self:
             prod.immediately_usable_qty = prod.virtual_available
 
--- a/stock_available/models/product_product.py
+++ b/stock_available/models/product_product.py
@@ -14,7 +14,7 @@ class ProductProduct(models.Model):
     """
     _inherit = 'product.product'
 
-    @api.one
+    @api.multi
     @api.depends('virtual_available')
     def _immediately_usable_qty(self):
         """No-op implementation of the stock available to promise.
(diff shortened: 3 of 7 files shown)


gurneyalex:
@atchuthan @clonedagain your opinion on this PR is welcome.

stock_available/__openerp__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Stock available to promise',
-    'version': '9.0.1.0.0',
+    'version': '9.0.3.0.0',
```
pedrobaeza: Change this according new version scheme

stock_available_immediately/__openerp__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Ignore planned receptions in quantity available to promise",
-    "version": "9.0.1.0.0",
+    "version": "9.0.2.0.1",
```
pedrobaeza: The same

stock_available/__openerp__.py
clonedagain: Where is it documented please?

pedrobaeza: I'm talking about the normal version increment numbering: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers. As the version numbering has been reset in 9.0, you cannot jump from 9.0.1.0.0 to 9.0.3.0.0 in one PR. This happens because the cherry-pick.

clonedagain:
how about merging the whole v8 branch instead?

pedrobaeza:
It has been merged. The v8 version in that moment was (for example, 8.0.2.1.0), and the diff for the following change, 8.0.3.0.0 it's correct. But when migrating from v8 to v9, version 8.0.2.1.0 became version 9.0.1.0.0, and thus, now for the same change, it doesn't make sense to go to 9.0.3.0.0.

stock_available/__openerp__.py
clonedagain: OK I didn't understand we had to reset version numbers. Maybe it could be more explicit.

clonedagain: https://github.com/OCA/maintainer-tools/compare/master...clonedagain:patch-3 makes it clearer

pedrobaeza: I don't think so. It's not the same. We have reset the version number to 9.0.1.0.0, so the documentation is correct, and you have to increment x, y or z according this new version, not according 8.0 version.

clonedagain:
~~@pedrobaeza I think we should merge regularly, cherry-picking each change is the wrong thing to do.
At the moment if I checkout 9.0 and merge 8.0 into it, I see a **very large** number of files that have not been merged yet.~~
`stock_available_mrp` for example seems obsolete in v9, you'll want to keep sale_available_\* all at the same level.

clonedagain:
:-1: for me at the moment

pedrobaeza:
We cannot merge regularly as Odoo makes because there are lots of changes (new/old API, version specific documentation...)

pedrobaeza:
@clonedagain, please, that was already discussed largely on the contributor list. Don't start again the same discussion or make it there, that is the correct place.

clonedagain:
@pedrobaeza OK I didn't know, I find the conclusion regrettable but I retract my comment.
@gurneyalex I still suggest you should import all stock_available_\* changes at the same time.

pedrobaeza:
@clonedagain, well, if you want, take again the topic to be discussed on the list, but please first search for the previous discussion, that leads to this conclusion. I would like also to make regularly merges, but I'm afraid is not possible.

About this PR, I suppose @gurneyalex thought that both commits are the only difference.

gurneyalex:
@pedrobaeza @clonedagain these are the only differences related to modules which have been ported to v9. We are preparing a global merge for the unported modules in order to help porters.

gurneyalex:
@clonedagain a separate PR is being prepared for stock_available_mrp

gurneyalex:
@pedrobaeza versions fixed.

pedrobaeza:
Thanks! :+1:

yvaucher:
:+1:

atchuthan:
:+1:
I agree with gurneyalex's cherry-pick
(Sorry for late reply)

PR #143, opened 2016-03-21
[9.0] move new V8 module: stock_quant_manual_assign

cyrilgdn:
stock_quant_manual_assign is V8 module which was not moved in V9.

### DIFF
--- a/stock_quant_manual_assign/__openerp__.py
+++ b/stock_quant_manual_assign/__openerp__.py
@@ -1,27 +1,12 @@
-# -*- encoding: utf-8 -*-
-##############################################################################
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as published
-#    by the Free Software Foundation, either version 3 of the License, or
-#    (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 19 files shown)


stock_quant_manual_assign/__openerp__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Mikel Arregi - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Stock - Manual assignment of quants",
+    "version": "8.0.1.0.0",
```
pedrobaeza, line 7: Change version number

pedrobaeza:
Is this finished, because the module remains uninstallable?

stock_quant_manual_assign/__openerp__.py
cyrilgdn, line 7: @pedrobaeza Sorry, my PR message is not very clear. I'm just adding this V8 module as unported  in V9 (i will create similar PR for stock_quant_merge and partner_location_auto_create).

We are merging, with @gurneyalex, 8.0 into 9.0 (splitted by module).

pedrobaeza, line 7: OK, understood. Then it's correct, except you can omit the line in the README.md, that is going to be changed in the next automated cron that scans modules.

cyrilgdn, line 7: Ok :)  I did not know it was automatic, thanks!

pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #147, opened 2016-03-21
[9.0] merge v8 stock_available

cyrilgdn:
Depends on PR #142 

Based on V8 PR #140 

Need this PR for stock_available_mrp V9 migration.

PR #140, opened 2016-03-17
Stock available mrp fix list view recursive qty

cyrilgdn:
Depends on PRs #135  and #136 

Bug: 
For example  I have a product P1 with BOM like that: P1 needs 1 P2 which needs 1 P3.

With 5 P3 in stock, when available to promise includes the production potential based on component's available to promise, we should have 5 P1, 5 P2, 5P3  

In form view no problem, in list view we have 0 available to promise P1.

PR #148, opened 2016-03-21
[9.0] stock_available_mrp migration

cyrilgdn:
Depends on PR #147 

Based on V8 PR #140


gurneyalex:
:+1:

atchuthan:
:+1:

teeling:
I have a problem, not sure if it's a bug or if it's as expected. 
The "potential" field on the product page seem to take BoM products with type "consumable" in consideration when figuring out how many that can be built. In my opinion, only products of type "stockable" should be calculated here.
Would you consider this a bug, or is there a clean workaround?

PR #153, opened 2016-04-04
Inventory initial values

nicomacr:
This PR replace https://github.com/OCA/stock-logistics-warehouse/pull/77 adding a please add a selection to decide if you want products with zero quantity or not no fill inventory.

Please check functionality and then we improove README and test cases


nicomacr:
https://github.com/OCA/stock-logistics-warehouse/pull/77

Issue #152, opened 2016-04-04
stock_inventory_hierarchical_location cause import error when starting server


andhit-r:
Hi,

I notice that **stock_inventory_hierarchical_location** module cause import error when starting server. The problems are in **inventory_hierarchical_location.py** line **24** and **27**. Bellow are the messages:

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 177, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 165, in execute
    application_iter = app(environ, start_response)
  File "/opt/odoo/odoo/openerp/service/wsgi_server.py", line 216, in application
    return application_unproxied(environ, start_response)
  File "/opt/odoo/odoo/openerp/service/wsgi_server.py", line 202, in application_unproxied
    result = handler(environ, start_response)
  File "/opt/odoo/odoo/openerp/http.py", line 1289, in **call**
    self.load_addons()
  File "/opt/odoo/odoo/openerp/http.py", line 1308, in load_addons
    m = **import**('openerp.addons.' + module)
  File "/opt/odoo/odoo/openerp/modules/module.py", line 80, in load_module
    mod = imp.load_module('openerp.addons.' + module_part, f, path, descr)
  File "/opt/odoo/oca_stock_logistics_warehouse/stock_inventory_hierarchical_location/**init**.py", line 21, in <module>
    from . import inventory_hierarchical_location
  File "/opt/odoo/oca_stock_logistics_warehouse/stock_inventory_hierarchical_location/inventory_hierarchical_location.py", line 27, in <module>
    from openerp.addons.stock_inventory_hierarchical.hierarchial_inventory import PARENT_VALUES
ImportError: No module named hierarchial_inventory

Could someone confirm this?

andhit-r:
PR for this issue https://github.com/OCA/stock-logistics-warehouse/pull/154

PR #154, opened 2016-04-04
Fix import error on stock_inventory_hierarchical_location

andhit-r:
Fix issue https://github.com/OCA/stock-logistics-warehouse/issues/152

### DIFF
--- a/stock_inventory_hierarchical_location/inventory_hierarchical_location.py
+++ b/stock_inventory_hierarchical_location/inventory_hierarchical_location.py
@@ -21,10 +21,10 @@
 from openerp.osv import orm
 from openerp.tools.translate import _
 
-from stock_inventory_hierarchical import HierarchicalInventoryException
+from openerp.addons.stock_inventory_hierarchical.exceptions import HierarchicalInventoryException
 
 # Add the date to the list of fields we must propagate to children inventories
-from stock_inventory_hierarchical import PARENT_VALUES
+from openerp.addons.stock_inventory_hierarchical.hierarchical_inventory import PARENT_VALUES
 PARENT_VALUES.append('exhaustive')
 
 


stock_inventory_hierarchical_location/inventory_hierarchical_location.py
```diff
@@ -21,10 +21,10 @@
 from openerp.osv import orm
 from openerp.tools.translate import _
 
-from stock_inventory_hierarchical import HierarchicalInventoryException
+from openerp.addons.stock_inventory_hierarchical.exceptions import HierarchicalInventoryException
```
pedrobaeza: The more suitable form is `from .stock_inventory_hierarchical import ...`

andhit-r:
@pedrobaeza done

pedrobaeza:
Sorry, I though the file was in the same module, but it isn't. The way you put it was correct. Sorry for making you change it.

andhit-r:
@pedrobaeza no worries. I didn't check it either. Still in panic mode because all my servers went down after pulled latest code :)

pedrobaeza:
Let's fast-track this one then.

:+1:

Daniel-CA:
I am having another error in this module now.
  File "/opt/odoo/demo_odoomrp/oca/stock-logistics-warehouse/stock_inventory_hierarchical_location/inventory_hierarchical_location.py", line 24, in <module>
    from openerp.addons.stock_inventory_hierarchical.exceptions import HierarchicalInventoryException
ImportError: No module named exceptions

Daniel-CA:
Forget about this, looks like that there is some kind of problem in the dependencies of the linux server, I am having another problem in another module so probably fixing dependency problem will fix this too.

PR #159, opened 2016-04-15
8.0 stock_quant_merge: Better code and better perf

alexis-via:
This is a small code improvement:
- don't update the cost price on quant2merge if there are no merge (small perf improvement)
- write cost and qty at the same time (small perf improvement)
- the cost should not be a simple average, but an average weighted by qty (in most cases, the cost if the same because the 2 quants originate from the same stock move... but it's bad for my eyes to read code where we do a simple average and not a weighted average).

@pedrobaeza That's your code... tell us what you think of these changes. The code that manipulate quants is always "dangerous code" !

### DIFF
--- a/stock_quant_merge/models/stock.py
+++ b/stock_quant_merge/models/stock.py
@@ -19,6 +19,7 @@ class StockQuant(models.Model):
                 ('lot_id', '=', self.lot_id.id),
                 ('package_id', '=', self.package_id.id),
                 ('location_id', '=', self.location_id.id),
+                ('company_id', '=', self.company_id.id),
                 ('reservation_id', '=', False),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It seems good :+1: 

Please check my PR https://github.com/OCA/stock-logistics-warehouse/pull/119, where the only concern is addressed here in this PR.

alexis-via:
@pedrobaeza I guess you should take some of the good idea of this PR and move the interesting pieces of code to your PR #119 and then I'll close this PR and approve your PR.

gurneyalex:
:+1:

rousseldenis:
:+1:

PR #165, opened 2016-05-11
FIX stock_inventory_preparation_filter search EAN13 for scanned products

JosDeGraeve:
When using the feature 'allows to make an inventory based on scanned
products', users might expect they can use the EAN13 barcode for
scanning their products.

Steps to reproduce:
- install module stock_inventory_preparation_filter
- create new stock.inventory record, select 'Empty list'
- Use your barcode scanner to enter EAN13 barcodes instead of typing the product reference them on the keyboard
- Click "start inventory"

Observed:
- No products are added to the inventory when their EAN13 barcode is used

Expected:
- The products are added to the inventory.

### DIFF
--- a/stock_inventory_preparation_filter/models/stock_inventory.py
+++ b/stock_inventory_preparation_filter/models/stock_inventory.py
@@ -109,8 +109,10 @@ class StockInventory(models.Model):
                     tmp_lines[line.product_code] = line.product_qty
             inventory.empty_line_ids.unlink()
             for product_code in tmp_lines.keys():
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

elicoidal:
:+1:

JosDeGraeve:
Is there something more needed to have this merged ?

pedrobaeza:
Not really, because there are more than 5 days since the opening. For less than 5 days 3 approvals are needed.

PR #168, opened 2016-05-27
[FIX] stock_quant_manual_assign: Don't let the browser call name_get …

mlaitinen:
…for each Many2one field in the wizard. Prefetch the record names instead.

**Steps to reproduce:**
1. Create a product `x` and generate `db_maxconn` + 10 quants for it
2. Create a picking and add a move with `x` as the product
3. Set the picking as "to do"
4. Open the stock move
5. Click "Manual quants"

**Expected result:**
The wizard opens and user is able to manually select quants

**Actual result:**
You get Connection Pool Error because Odoo is out of database connections.

At least `stock.quant` name is `store=False` and computed on the fly. When you have a large number of quants in the wizard list view, the browser will call `name_get` as many times, and attempt to open as many database connections. This will result in Connection Pool Error when `db_maxconn` is smaller than the number of rows.

The Many2one records (quant, package, lot, location) aren't needed for anything else than to display their names, so instead of letting browser call `name_get` a few hundred times when the wizard opens, prefetch the Many2one record names in `assign.manual.quants.default_get()`.

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # (c) 2015 Mikel Arregi - AvanzOSC
 # (c) 2015 Oihane Crucelaegui - AvanzOSC
+# (c) 2016 Miku Laitinen - Avoin.Systems
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from openerp import api, exceptions, fields, models, _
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think you have a problem in your installation, as this uses the standard behavior, so with your reasoning, a simple list view that call name_get will fail.

mlaitinen:
I just have rather low `db_maxconn` (2 workers, 12 `db_maxconn`, 5 users).

The `name_get` of `stock.quant` will be called simultaneously many many times and each of them will use a separate database cursor. My solution calls `name_get` sequentially instead of in parallel. [`stock.quant._get_quant_name`](https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L269) generates the name of `stock.quant` on the fly. 

I don't think raising `db_maxconn` is a long-lasting solution. What if I have 500 quants to choose from?

pedrobaeza:
What I'm stating is that the operation on name_get is not launched that way AFAIK. @hbrunn has worked with this one in OCB and maybe can tell something.

Cedric-Pigeon:
@mlaitinen Hello, do you plan to achieve this Pr ? Is it still useful ? Thanks

mlaitinen:
No, not really. Maybe I'll just throw some money at the problem and get more database connections.

Just monitor the database while Odoo opens up connections for a single tree view. Could they have designed Odoo in a way that lets a single user grab ~30 simultaneous connections to the database while opening a simple view?

PR #170, opened 2016-06-06
stock_orderpoint_generator: Migration V9

cyrilgdn:
Was stock_orderpoint_creator, migrated from V7.


stock_orderpoint_generator/wizard/orderpoint_generator_view.xml
```diff
@@ -0,0 +1,40 @@
...
+        <field name="arch" type="xml">
+            <form string="Product warehouse config" version="7.0">
+                <label string="This wizard will apply the following orderpoint to selected product(s)"/>
+                <group string="Templates" colspan="4">
+                    <field name="orderpoint_template_id" colspan="4"/>
+                </group>
+                <footer>
+                    <button name="action_configure" string="Apply" type="object" class="oe_highlight"
+                            icon="gtk-execute"/>
+                    or
```
yvaucher: The "or" was removed in 9.0 Please remove it

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,67 @@
...
+from openerp import api, fields, models
+
+
+class OrderpointTemplate(models.Model):
+    """ Template for orderpoints
+
+    Here we use same model as stock.warehouse.orderpoint but set product_id
+    as non mandatory as we cannot remove it. This field will be ignored.
+
+    This has the advantage the advantage of ensuring that the order point
```
guewen: 'the advantage' is repeated

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    _name = 'stock.warehouse.orderpoint.template'
+
+    _inherit = 'stock.warehouse.orderpoint'
+    _table = 'stock_warehouse_orderpoint_template'
+
+    name = fields.Char(copy=True)
+
+    procurement_ids = fields.One2many(
+        'procurement.order', 'orderpoint_id', copy=False
```
guewen: `fields.One2many` are `copy=False` by default (but it does not hurt to explicit it)

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    group_id = fields.Many2one('procurement.group', copy=True)
```
guewen: `fields.Many2one` are `copy=True` by default

stock_orderpoint_generator/wizard/orderpoint_generator.py
```diff
@@ -0,0 +1,45 @@
...
+        product_ids = self.env.context.get('active_ids')
+        assert product_ids and isinstance(product_ids, list)
+
+        if self.env.context.get('active_model') == 'product.template':
+            templates = self.env['product.template'].browse(product_ids)
+            product_ids = templates.mapped('product_variant_ids.id')
+            if len(product_ids) != len(templates):
+                raise UserError(_(
+                    'Cannot apply because some of selected '
+                    'products has multiple variants.'
```
guewen, line 42: Not sure to understand what this limitation is for. Having multiple variants on a template is common, and the "orderpoint templates" are based on variants, so?

stock_orderpoint_generator/models/orderpoint_template.py
cyrilgdn: Yes but this field has copy=False in stock.warehouse.orderpoint (https://github.com/OCA/OCB/blob/9.0/addons/stock/stock.py#L4706) so it shows that copy is overridden.

guewen: All right! you shouldn't need to specify again `'procurement.group'` then.

stock_orderpoint_generator/wizard/orderpoint_generator.py
cyrilgdn, line 42: Not sure  to understand too :) !

But in V7, this wizard was only available in product.product tree view. In V9, most of products list are product.template so I added the same button on these views but I wasn't sure that users want selected templates to be applied on all variants.... 

Multiple variants may mean different orderpoints? I don't really know....

guewen, line 42: Yes each variant will have its own orderpoint.

I think we should just generate the orderpoints for all the variants of a template.

If one day someone really want to generate the orderpoints for only a subselection of the variants of a template, he'll make the wizard to be usable on both `product.template` and `product.product` views.

stock_orderpoint_generator/models/orderpoint_template.py
cyrilgdn: Like group_id, I added this line to override copy value so if copy is already False I think I can remove this line.

stock_orderpoint_generator/models/orderpoint_template.py
cyrilgdn: Oh nice, I did not know that.

stock_orderpoint_generator/wizard/orderpoint_generator.py
guewen, line 42: > If one day someone really want to generate the orderpoints for only a subselection of the variants of a template, he'll make the wizard to be usable on both product.template and product.product views.

Already there btw :)

guewen, line 42: Those who use variants can use the 'variants' menu, so it's ok for me to keep the error.

guewen:
:+1: Thanks

stock_orderpoint_generator/README.rst
```diff
@@ -0,0 +1,42 @@
...
+=====================
+
+Add a wizard to configure order points for multiple products in one go.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
yvaucher: `{project_repo}` place holder to replace

stock_orderpoint_generator/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Order point generator',
+    'summary': 'Mass configuration of stock order points',
+    'version': '9.0.1.0.0',
+    'author': "Camptocamp, Odoo Community Association (OCA)",
+    'category': 'Warehouse',
+    'license': 'AGPL-3',
+    'complexity': 'easy',
```
yvaucher: I think we can drop this line. I don't think it is even still present in modules.

stock_orderpoint_generator/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Cyril Gaudin (Camptocamp)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
yvaucher: No need for a license here

stock_orderpoint_generator/models/orderpoint_template.py
yvaucher: Thus here too you can define only the `copy` attribute.

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,67 @@
...
+
+    product_id = fields.Many2one(
+        'product.product',
+        'Product',
+        required=False,
```
yvaucher: Here we might want as well to define only the `required` attribute

stock_orderpoint_generator/tests/test_orderpoint_generator.py
```diff
@@ -0,0 +1,92 @@
...
+            'qty_multiple': 1,
+            'warehouse_id': self.ref('stock.warehouse0')
+        })
+
+    def check_orderpoint(self):
+        orderpoints = self.orderpoint_model.search([
+            ('name', '=', 'OP/000445')
+        ])
+
+        orderpoints = sorted(orderpoints, key=lambda o: o.product_id.id)
```
yvaucher: You can sort while doing the search using attribute `order`

```
model.search(domain, order="product_id")
```

cyrilgdn:
Updated with all remarks taken in account.

stock_orderpoint_generator/tests/test_orderpoint_generator.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(2, len(orderpoints))
+
+        self.assertEqual(self.p1, orderpoints[0].product_id)
+        self.assertEqual(self.p2, orderpoints[1].product_id)
```
yvaucher, line 47: You could also check if `self.p1 | self.p2 == orderpoints.mapped('product_id')`

stock_orderpoint_generator/wizard/orderpoint_generator_view.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="orderpoint_generator_view" model="ir.ui.view">
+        <field name="name">stock.warehouse.orderpoint.generator</field>
+        <field name="model">stock.warehouse.orderpoint.generator</field>
+        <field name="arch" type="xml">
+            <form string="Product warehouse config" version="7.0">
```
yvaucher: `version="7.0"` is deprecated simply remove it

stock_orderpoint_generator/tests/test_orderpoint_generator.py
cyrilgdn, line 47: For tests I preferred to check separately as I think this is really more readable.

cyrilgdn:
@yvaucher Updated with your remarks!

yvaucher:
Thanks :+1:

lasley (approved):
Thanks for the contribution @cyrilgdn  👍

PR #172, opened 2016-07-01
[9.0][MIG] stock_manual_assign: port to 9.0

lbelorgey:
Port to v9 and fix issue if move and quants haven't the same uom.

But there is still an issue if you uncheck a selected quant, the remaining quantity doesn't update. I don't understand why, if someone has an idea.


pedrobaeza:
You have to follow migration guide to rescue latest commits from v8: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

PR #173, opened 2016-07-02
[9.0][MIG] stock_manual_assign: port to 9.0

lbelorgey:
Port to v9 and fix issue if move and quants haven't the same uom.

In README.rst, I don't update the runbot link because I don't find the right one for v9.

There's still an issue if you uncheck a selected quant, the remaining quantity doesn't update. I don't understand why, if someone has an idea.

### DIFF
--- a/stock_quant_manual_assign/README.rst
+++ b/stock_quant_manual_assign/README.rst
@@ -30,7 +30,7 @@ help us smashing it by providing a detailed and welcomed `feedback
 <https://github.com/OCA/
 stock-logistics-warehouse/issues/new?body=module:%20
 stock_quant_manual_assign%0Aversion:%20
-8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
 
 Credits
--- a/stock_quant_manual_assign/README.rst
+++ b/stock_quant_manual_assign/README.rst
@@ -6,7 +6,7 @@
 Stock - Manual Quant Assignment
 ===============================
 
-This module allow you to manually change the automatic quant selection.
+This module allows you to manually change the automatic quant selection.
 
 Usage
 =====
--- a/stock_quant_manual_assign/README.rst
+++ b/stock_quant_manual_assign/README.rst
@@ -2,23 +2,24 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-===========================
-Manual assignment of quants
-===========================
+===============================
+Stock - Manual Quant Assignment
+===============================
 
-With this module, user can manually change the automatic selection of quants.
+This module allow you to manually change the automatic quant selection.
 
 Usage
 =====
 
 To use this module, you need to:
 
-* You will able to select the quants from each stock.move. There will be a
-  button that opens a wizard directly from the move or from the picking.
+#. Select a stock move or a stock picking.
+#. Open the wizard with the button "Manual Quants".
+#. Select the quants to assign (reserve).
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/153/8.0
+   :target: https://runbot.odoo-community.org/runbot/153/9.0
 
 Bug Tracker
 ===========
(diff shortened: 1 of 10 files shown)


stock_quant_manual_assign/README.rst
```diff
@@ -30,7 +30,7 @@ help us smashing it by providing a detailed and welcomed `feedback
 <https://github.com/OCA/
 stock-logistics-warehouse/issues/new?body=module:%20
 stock_quant_manual_assign%0Aversion:%20
-8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
elicoidal: Actually this line is to be removed as per the newest template

stock_quant_manual_assign/__openerp__.py
```diff
@@ -25,5 +25,5 @@
         "wizard/assign_manual_quants_view.xml",
         "views/stock_move_view.xml",
     ],
-    "installable": False,
```
elicoidal: in this file, you should remove the contributor (they should be in the README.rst

elicoidal: Name would be better: Stock - Manual Quant Assignment

stock_quant_manual_assign/README.rst
```diff
@@ -30,7 +30,7 @@ help us smashing it by providing a detailed and welcomed `feedback
 <https://github.com/OCA/
```
elicoidal: up: 
"With this module, user can manually change the automatic quant selection."
and
"How to use this module:

#. Select the quants from each stock.move. 
#. A button opens a wizard directly from the move or from the picking."

stock_quant_manual_assign/README.rst
```diff
@@ -2,36 +2,32 @@
...
 
-===========================
-Manual assignment of quants
-===========================
+===============================
+Stock - Manual Quant Assignment
+===============================
 
-With this module, user can manually change the automatic selection of quants.
+This module allow you to manually change the automatic quant selection.
```
pedrobaeza: s/allow/allows

stock_quant_manual_assign/i18n/stock_quant_manual_assign.pot
```diff
@@ -4,10 +4,10 @@
 #
```
pedrobaeza, line 4: Remove this file as it's not needed anymore

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -17,7 +17,7 @@ def check_qty(self):
             if record.quants_lines:
                 move = self.env['stock.move'].browse(
                     self.env.context['active_id'])
-                if record.lines_qty > move.product_uom_qty:
+                if record.lines_qty > move.product_qty:
```
pedrobaeza: Why do you change this? The correct value is put in product_uom_qty, not product_qty. Does this changes in 9.0?

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -63,15 +63,18 @@ def default_get(self, var_fields):
...
-        quants_lines = [{
-            'quant': x.id,
-            'lot_id': x.lot_id.id,
-            'package_id': x.package_id.id,
-            'selected': x in move.reserved_quant_ids,
-            'qty': x.qty if x in move.reserved_quant_ids else 0,
-            'location_id': x.location_id.id,
-        } for x in available_quants]
-        return {'quants_lines': quants_lines}
+        quants_lines = []
```
pedrobaeza: Why do you change this if previous version is even more compact?

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -1,38 +1,36 @@
...
+                            <field name="qty" attrs="{'readonly':[('selected', '=' ,False)]}" sum="qty"/>
+                        </tree>
+                    </field>
+                    <field name="move_qty" colspan="2" readonly="1"/>
+                    <field name="lines_qty" invisible="1" />
+                </group>
+                <footer>
+                    <button name="assign_quants" type="object"
+                        string="Confirm" class="oe_highlight" />
+                    or
```
pedrobaeza: In version 9.0 the `or` is now omitted

stock_quant_manual_assign/wizard/assign_manual_quants.py
lbelorgey: I use this module to assign quants in "Manufacturing Orders/Products to Consume" (MRP module).
UoM in my BoM (I use "g" because quantities are very small) is different from the default UoM of the product ("Kg" because it is the UoM I use for the inventory).
When I select a "stock.move" in "Products to consume", the UoM is the BoM one ("g") and when I open the wizard "Manual Quants", the quant UoM is the default UoM of the product ("Kg"). So there is an issue. That's the reason why I use move.product_qty which seems to be in the same UoM of the quant. In my tests, move.product_qty and move.product_uom_qty have the same value when the UoM ratio is applied.

But Anoher solution, perhaps a better one, is to convert move.product_uom_qty to the default product UoM by using move.product_uom.ratio. I saw a conversion method in UoM. Do you known if quant.qty is always in the default UoM of the product because there isn't UoM field in Quant ?

JordiBForgeFlow:
The PR https://github.com/OCA/stock-logistics-warehouse/pull/196 supersedes this one, and fixes the issue of computing the remaining qty. 

Can you close this one?

PR #157, opened 2016-04-12
9.0 stock mts mto rule

JordiBForgeFlow:
Replaces #156

PR #177, opened 2016-07-14
port stock_mts_mto_rule to v9

nedaszilinskas:


Garamotte:
A PR for this module's migration is already pending here : #157 

I checked the diff between your 2 PRs, and the diff is really small.
Functionnally, there is no change, except a fix (8a8e6f1) made in 8.0, which is absent in this PR but applied in #157 

@pedrobaeza : I think you can close this PR as it's an outdated duplicate

pedrobaeza:
OK, closing.

PR #181, opened 2016-08-08
FIX picking edition by warehouse user due to related none readonly field

nicomacr:
With "stock_quant_manual_assign", if a user with "warehouse / user" group try to add a line in a picking he raise an error due to no rights for writing on "stock.picking.type". 
This error is because the related field "picking_type_code" must be readonly

### DIFF
--- a/stock_quant_manual_assign/__openerp__.py
+++ b/stock_quant_manual_assign/__openerp__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Stock - Manual assignment of quants",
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change the module version to 8.0.1.1.0

jjscarafia:
:+1:

pedrobaeza:
This is a known side effect of some related+computed chain that should be blocked this dirty way, so :+1: and fast-tracking it.

PR #183, opened 2016-08-26
[ADD] new module stock_inventory_line_price

esthermm:


stock_inventory_line_price/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+# (c) 2015 Esther Martín - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Standard price at inventory level",
+    "version": "8.0.1.0.0",
+    "category": "Generic Modules",
+    "author": "OdooMRP team, "
+              "AvanzOSC, "
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza",
```
pedrobaeza: Put `Tecnativa` here please

stock_inventory_line_price/models/stock_inventory.py
```diff
@@ -0,0 +1,36 @@
...
+
+class StockInventoryLine(models.Model):
+    _inherit = 'stock.inventory.line'
+
+    standard_price = fields.Float('Cost Price',
+                                  digits=dp.get_precision('Product Price'))
+
+    @api.model
+    def _resolve_inventory_line(self, inventory_line):
+        inventory_line.product_id.standard_price = \
```
pedrobaeza: This line is not valid when the cost price is average. Please query @agaldona, that knows about this process

stock_inventory_line_price/models/stock_inventory.py
```diff
@@ -0,0 +1,36 @@
...
+
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    @api.model
+    def _get_inventory_lines(self, inventory):
+        vals = super(StockInventory, self)._get_inventory_lines(inventory)
+        for line in vals:
+            product = self.env['product.product'].browse(line['product_id'])
+            line.update({'standard_price': product.standard_price})
```
pedrobaeza: It's better to let the field empty, and only act when having a explicit defined price, because you can create an inventory in one moment, but don't validate it after several weeks, and then, the price in case of average price will be outdated. @anajuaristi will understand this concept.

stock_inventory_line_price/README.rst
```diff
@@ -0,0 +1,18 @@
...
+* This module allows you to handle the standard price at inventory level.
+
+Credits
+=======
+
+Contributors
+------------
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Ana Juaristi <ajuaristio@gmail.com>
+* Esther Martín <esthermartin@avanzosc.es>
```
lasley, line 35: Please use OCA readme template - https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

stock_inventory_line_price/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Esther Martín - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
```
lasley: Please use OCA header - https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

lasley:
Thanks for the submission @esthermm - comments inline + I agree with Pedro's.

stock_inventory_line_price/models/stock_inventory.py
agaldona: I'm not completely sure, but if the concept is "there are 10 units of ProductA with a cost of 0'43", in that case it must not compute the average cost, it must rewrite it. @anajuaristi can you give as a point?

anajuaristi:
@pedrobaeza @lasley I agree to let the field cost empty  
:+1:

anajuaristi:
About the average cost...  To change the standar price on product form should be exactly the same behavior when you change the price manually.
This is... if you make inventory:
- the existing quants will be no affected/changed at all but only the product form field will be changed. 
- The value to set on product form is just the one set on inventory line. 
- The adjusment quant value will be also the one set on inventory line.
  There is no way to make "average cost" because inventory is setting the stock quantity. There is no "average" with existing stock.

esthermm:
@pedrobaeza @lasley  Requested changes done
There have been added a field to show the standard price of the product in order to have it as reference and other cost price field that user can change, the same way as quantity field.

lasley (approved):
LGTM thanks @esthermm 👍

oihane (approved):
LGTM :+1:

anajuaristi:
LGTM :+1:

anajuaristi:
So, it is ready to merge. I haven't got permissions so I would ask someone to do it asap, please. 

Thank you:
Ana

stock_inventory_line_price/README.rst
```diff
@@ -0,0 +1,50 @@
...
+
+=================================
+Standard price at inventory level
+=================================
+
+* This module allows you to handle the standard price at inventory level.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
oihane: There is no repo_id nor branch defined

stock_inventory_line_price/README.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
oihane: You have to add a project_repo name

oihane (approved):
LGTM

PR #184, opened 2016-08-29
[8.0][ADD] stock_orderpoint_automatic_creation

Daniel-CA:
This module generates stock orderpoints automatically for a product when is created taking create user company as reference.

### DIFF
--- a/stock_orderpoint_automatic_creation/README.rst
+++ b/stock_orderpoint_automatic_creation/README.rst
@@ -7,20 +7,21 @@ Stock Orderpoint Automatic Creation
 ===================================
 
 This module generates stock orderpoints automatically for a product when is
-created taking created user company as reference.
+created taking the company of the user who created the product as reference.
+Orderpoints creation can be automate from product, product category or company.
 
 Configuration
 =============
 
 To configure this module, you need to:
 
-Go to company form, in the "Settings" tab, you can customize the default
-orderpoints data.
+Go to company form, in the "Settings" tab, you can customize the default orderpoints data.
+Also you can configure orderpoint creation in product and product_category checking 'Yes' or 'No' in the proper field.
 
 Bug Tracker
 ===========
 
-Bugs are tracked on `GitHub Issues <https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of trouble, please
 check there if your issue has already been reported. If you spotted it first,
 help us smashing it by providing a detailed and welcomed `feedback <https://github.com/OCA/stock-logistics-warehouse/issues/new?body=module:%20stock_orderpoint_automatic_creation%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
--- a/stock_orderpoint_automatic_creation/README.rst
+++ b/stock_orderpoint_automatic_creation/README.rst
@@ -1,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================================
+Stock Orderpoint Automatic Creation
+===================================
+
+This module generates stock orderpoints automatically for a product when is
+created taking created user company as reference.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+Go to company form, in the "Settings" tab, you can customize the default
+orderpoints data.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback <https://github.com/OCA/stock-logistics-warehouse/issues/new?body=module:%20stock_orderpoint_automatic_creation%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+* Daniel Campos <danielcampos@avanzosc.es>
+* Ana Juaristi <ajuaristio@gmail.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/stock_orderpoint_automatic_creation/__openerp__.py
+++ b/stock_orderpoint_automatic_creation/__openerp__.py
@@ -11,7 +11,8 @@
     'website': 'http://www.odoomrp.com',
     "depends": ['procurement'],
     "category": 'Inventory, Logistic, Storage',
-    "data": ['views/res_company_view.xml'],
+    "data": ['views/res_company_view.xml',
+             'views/product_view.xml'],
     "installable": True,
     "application": True
 }
--- a/stock_orderpoint_automatic_creation/__openerp__.py
+++ b/stock_orderpoint_automatic_creation/__openerp__.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Daniel Campos <danielcampos@avanzosc.es> - Avanzosc S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Stock Orderpoint Automatic Creation',
+    'version': "8.0.1.0.0",
+    'author': "AvanzOSC, Odoo Community Association (OCA)",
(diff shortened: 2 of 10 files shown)


stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,43 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Stock Orderpoint Automatic Creation
```
pedrobaeza, line 6: Update to latest README template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,43 @@
...
+====================================
+
+This module generates stock orderpoints automatically for a product when is
+created taking create user company as reference.
+* In the company form, in the "Settings" tab, you can customize the default
+orderpoints data.
```
pedrobaeza: Put 2 spaces for correct indentation and move this to Configuration section

stock_orderpoint_automatic_creation/__openerp__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Put short headers

stock_orderpoint_automatic_creation/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+
+{
+    'name': 'Stock Orderpoint Automatic Creation',
+    'version': "8.0.1.0.0",
+    'author': "AvanzOSC",
+    'contributors': ["Daniel Campos <danielcampos@avanzosc.es>",
+                     "Ana Juaristi <ajuaristio@gmail.com>"],
+    'website': 'http://www.odoomrp.com',
+    "depends": ['procurement'],
+    "category": "Generic Modules",
```
pedrobaeza: Put the same category as procurement

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,24 @@
...
+        result = super(ProductProduct, self).create(vals)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
+        if result.type != 'service':
+            orderpoint_obj.create(
+                {'name': result.name,
+                 'product_id': result.id,
+                 'product_max_qty':
+                 self.env.user.company_id.orderpoint_product_max_qty,
+                 'product_min_qty':
+                 self.env.user.company_id.orderpoint_product_min_qty
```
pedrobaeza: How do you define the location_id?
And what happens when you have more than 1 warehouse?

Daniel-CA: Yes, I was thinking that maybe is better to add this information on warehouse but we have the problem that we don't know for which warehouse is the product being created...
I have to evaluate what is the best option.

Daniel-CA: Finally added default warehouse to company configuration so only one warehouse by company could be configured for automatic orderpoint creation, not sure if I should explain it in Known issues / Roadmap.

agaldona:
LGTM :+1:

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Stock Orderpoint Automatic Creation
```
pedrobaeza, line 6: Line above

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,49 @@
...
+===================================
+
+This module generates stock orderpoints automatically for a product when is
+created taking create user company as reference.
```
pedrobaeza: s/create/created

stock_orderpoint_automatic_creation/__openerp__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Daniel Campos <danielcampos@avanzosc.es> - Avanzosc S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Stock Orderpoint Automatic Creation',
+    'version': "8.0.1.0.0",
+    'author': "AvanzOSC",
```
pedrobaeza: Add OCA

pedrobaeza:
The problem is still there with multiple warehouses.

Daniel-CA:
Changes done, I have added a boolean field to check if the automatic orderpoints applies for the current company and if yes do for all warehouses on it.

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+    @api.model
+    def create(self, vals):
+        result = super(ProductProduct, self).create(vals)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
+        wh_obj = self.env['stock.warehouse']
+        if result.type != 'service':
+            if self.env.user.company_id.create_orderpoints:
+                company = self.env.user.company_id
+                warehouses = wh_obj.search([('company_id', '=', company.id)])
+                for warehouse in warehouses:
```
carlosdauden: What do you think the next idea?
Add a selection field product category and company with Yes, No or None values to determine whether or not created the rule.
If this field is set to None in the product, the value is obtained from the search. If the category also has None is obtained from the company.
This selection field should be company_dependent.

Daniel-CA: Hi @carlosdauden, if I have understood your idea is to define the rules creation from the product category and otherwise from the company, right?

carlosdauden: Yes. It's a little more complex but far more flexible.

Daniel-CA: I see it a good approach, I will add the code or you prefer to do it yourself?

carlosdauden: Implement it you, but if you need anything you tell me and I help you delighted

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================================
+Stock Orderpoint Automatic Creation
+===================================
+
+This module generates stock orderpoints automatically for a product when is
+created taking created user company as reference.
```
gurneyalex: s/taking created user company/taking the company of the user who created the product/

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, api
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.model
+    def create(self, vals):
```
gurneyalex, line 21: you need to add the following decorator when overriding `create()`:

```
@api.returns('self', lambda value: value.id)
```

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+        result = super(ProductProduct, self).create(vals)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
+        wh_obj = self.env['stock.warehouse']
+        if result.type != 'service':
```
gurneyalex: It's weird to also create an orderpoint for `consumable` products. I think this should be:

```
if result.type == 'product':
```

gurneyalex (changes requested):
Thanks for your contribution. 

The create() method needs changing. The product type issue needs to be checked. 

I also suggested a clarification in the README.

Daniel-CA:
Hi @gurneyalex, thanks for the review, I will do the changes and the improvement proposed by @carlosdauden.

pedrobaeza:
What this still lacks is the support for multi-warehouses.

Daniel-CA:
Hi @pedrobaeza, you mean that only create ordepoint rules for the defined warehouses of the company?

pedrobaeza:
Configuration at product/product template/product category should be a one2many where you specify for each warehouse (or even better, for each location you want), what to do: create or not create orderpoints.

Daniel-CA:
Don't know if I have understood you approach, but I see it as complicating the configuration too much.

pedrobaeza:
Yeah, but that's needed for multi-warehouse. Maybe you want automatic procurement on one warehouse (location), but not in another.

pedrobaeza:
I have just discovered https://github.com/OCA/stock-logistics-warehouse/pull/170 as an alternative solution.

Daniel-CA:
Hi @pedrobaeza thanks for the info, but as I can see the module is only for v9 and is good for making mass creation of orderpoints in already  generated products, but if you create new products you should launch the wizard again, in IMHO are different purpose modules.

pedrobaeza:
Yeah, they are not equivalent, but other way to do things.

Cedric-Pigeon:
@Daniel-CA Hello, do you plan to achieve this PR ? Is it still usefull ? Thanks.

Daniel-CA:
Hi @Cedric-Pigeon,
This module is finished to me, but if someone wants to improve it with more functionality will be welcomed. 
I have also a PR for the proposed module in v8 (although the module lacks of functionality I mentioned in a previous message) at #214 
Anyway if the PR does not interest anyone I can cancel it without any problem. I will move the module to one of my own branches.

Regards

PR #186, opened 2016-09-07
Port stock_reserve and stock_reserve_sale to 9.0

katyukha:
Following addons ported to Odoo 9.0 branch
- stock_reserve
- stock_reserve_sale


pedrobaeza:
Please, only one module per PR.

katyukha:
Is there an easy way to split this commit on two?

pedrobaeza:
No, the best bet is that you duplicate your branch, and let this one with only one module, and propose the other branch with the other module (or both if one depends on the other as is this case I think). Indicate in the other PR that depends on this one, and you can rebase that one when merging this. For achieving this, you need to separate changes for each module in at least one commit for each.

katyukha:
Would it be good enough to include in that new PRs also changes from #187 and #188, because they both depends on this PR?

pedrobaeza:
Yeah, I think so

katyukha:
Code is splited: #189, #190

Is it Ok?

pedrobaeza:
Yes! I close this one then.

PR #187, opened 2016-09-08
stock_reserve_sale: security fixes

katyukha:
This PR is based on #186 

Allow salesman to create reservations without need of extra rights

PR #188, opened 2016-09-09
Stock reserve sale refactor

katyukha:
Based on PR #186 and #187 

stock_reserve_sale: Refactor to simplify wizard
1. `stock_reserve` refactored to correctly compute default value for
   `location_id` if `default_picking_type_id` is passed via context.
2. `stock_reserve_sale` refactored to simplify wizard with following
   ideas in mind:
   - sales person is responsible to sale, not move products across
     warehouse, so he should not have ability to decide what and where to
     move in reservation process, thus all stock related fields was
     remvoed from _Reservation wizard_ and now is computed
     automaticaly (got from SO -> warehouse -> Pick Type)
   - Stock reservation logic is not so specific to place it all inside
     wizard, so reservation logic moved to `sale.order.line` and
     thus is accessible from other code and RPC.


katyukha:
Closed: these modifications are already included in #189 and #190

PR #191, opened 2016-09-14
[MIG] stock_quant_partner_info migrated to OCA

oihane:


stock_quant_partner_info/migrations/8.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Remove this file

oihane: Can you explain me why?

stock_quant_partner_info/models/stock_quant.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dest_partner_id = fields.Many2one(
+        comodel_name='res.partner', compute='_compute_partner_id',
+        string='Partner', store=True)
+
+    @api.depends('location_id', 'history_ids')
+    def _compute_partner_id(self):
+        for quant in self.filtered(
+                lambda q: q.location_id.usage in ['customer', 'supplier']):
+            move = quant.history_ids.filtered(
```
pedrobaeza: Don't filter again the history_ids by location. You already have it on `[:-1]`, as you are checking previously the location_id usage.

pedrobaeza:
Description should be clear to say that this module put the partner of the picking that moves the quant to a supplier/customer location, because you have the concept of location owner that conflicts with this one.

stock_quant_partner_info/migrations/8.0.1.1.0/pre-migration.py
pedrobaeza: Because this is the first module that gets to OCA, so no migration is needed. If you internally need it, update all your installations and then the script is useless

oihane: Ok, it is a public module I've thought that it was required, just in case. Thanks for the explanation

stock_quant_partner_info/README.rst
```diff
@@ -0,0 +1,43 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================
+Stock - Quant partner info
+==========================
+
+With this module you can see in which is the related partner if it is in a supplier or
```
elicoidal: With this module, when you are in a customer or supplier location, the name of the partner of the last picking to move the quant is displayed.

oihane: Thanks

stock_quant_partner_info/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    "name": "Stock - Quant partner info",
+    "version": "8.0.1.1.0",
+    "category": "Warehouse Management",
+    "license": "AGPL-3",
+    "author": "AvanzOSC, "
+              "Odoo Community Association (OCA)",
+    "website": "http://www.avanzosc.es",
+    "contributors": [
```
andhit-r: Remove this as this information already exist on README

andhit-r (approved):
:+1:  just minor comment

Cedric-Pigeon (approved):

PR #192, opened 2016-09-15
[9.0][FIX] add button to product form

SodexisTeam:
Added the potential quantity button on product form too.


pedrobaeza:
Can't this be put in the original module (stock_available) in the view that is available both for template and product?

atchuthan:
no, the field "potential_qty" is added only in stock_available_mrp module.

pedrobaeza:
Anyway, my point is that there's a view that is valid for both template and product. I'm asking about adding there the field for not having to duplicate the work each time, in whatever module you need to do it.

atchuthan:
@pedrobaeza Moved the field to stock_available module as per our previous discussion.

change label to "needs review"

stock_available/views/product_template_view.xml
```diff
@@ -20,6 +19,15 @@
                             <span class="o_stat_text">Available</span>
                         </div>
                     </button>
+                    <button type="action" name="%(stock.product_open_quants)d"
+                        attrs="{'invisible':[('type', '!=', 'product')]}"
```
pedrobaeza: What about consumable products?

atchuthan: I tried to use the same visibility as "immediately_usable_qty" field
https://github.com/OCA/stock-logistics-warehouse/pull/192/files/70c8ae8602260ef375d4c765a5d7a02e02eb3e34#diff-33210f15e1b07dbc688105994395937aR14

pedrobaeza: Yeah, but both can make sense also for consumable products I think

atchuthan: Yes, I agree.

pedrobaeza (approved):
Squashing on merge

stock_available/models/product_product.py
```diff
@@ -29,11 +29,27 @@ def _immediately_usable_qty(self):
...
-        compute='_immediately_usable_qty',
+        compute='_compute_immediately_usable_qty',
         string='Available to promise',
         help="Stock for this Product that can be safely proposed "
              "for sale to Customers.\n"
              "The definition of this value can be configured to suit "
              "your needs")
+    potential_qty = fields.Float(
+        compute='_compute_potential_qty',
+        type='float',
```
Garamotte: You don't need to define the type attribute on new API.

stock_available/models/product_template.py
```diff
@@ -24,11 +24,36 @@ def _immediately_usable_qty(self):
...
-        compute='_immediately_usable_qty',
+        compute='_compute_immediately_usable_qty',
         string='Available to promise',
         help="Stock for this Product that can be safely proposed "
              "for sale to Customers.\n"
              "The definition of this value can be configured to suit "
              "your needs")
+    potential_qty = fields.Float(
+        compute='_compute_potential_qty',
+        type='float',
```
Garamotte: You don't need to define the type attribute on new API.

stock_available_mrp/__openerp__.py
```diff
@@ -5,15 +5,14 @@
...
               u"Odoo Community Association (OCA)",
     'category': 'Hidden',
     'depends': [
         'stock_available',
         'mrp'
     ],
-    'data': [
-        'views/product_template_view.xml',
-    ],
+    'data': [],
```
Garamotte: Please remove empty key.

Garamotte (commented):
Some comments, but not blocking.

atchuthan:
@pedrobaeza squashed all commits
@sylvain-garancher changes done

gurneyalex (approved):

PR #197, opened 2016-09-29
9.0 stock_putaway_product: Migration to v9.0

carlosdauden:
This module extends the functionality of the odoo putaway strategy.
It defines a new type of putaway strategy where users can set a specific
stock location per product.

@Tecnativa


pedrobaeza:
Please squash all the commits from Jos together.

stock_putaway_product/README.rst
```diff
@@ -59,7 +61,7 @@ Bugs are tracked on `GitHub Issues
 case of trouble, please
 check there if your issue has already been reported. If you spotted it first,
 help us smashing it by providing a detailed and welcomed `feedback
-<https://github.com/OCA/stock-logistics-warehouse/issues/new?body=module:%20stock_putaway_product%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+<https://github.com/OCA/stock-logistics-warehouse/issues/new?body=module:%20stock_putaway_product%0Aversion:%209.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Use new template version without this part.

stock_putaway_product/wizard/stock_change_product_qty.py
```diff
@@ -21,9 +22,8 @@ def default_get(self, fields):
             putaway_ids = self.env['stock.product.putaway.strategy'].search([
```
pedrobaeza: Rename this to `putaway` (without `s` nor `ids`), as it reflects better the content of the variable.

pedrobaeza (changes requested):

stock_putaway_product/models/product_putaway.py
```diff
@@ -0,0 +1,79 @@
...
+            ]
+            for strategy in putaway_strategy.product_location_ids.search(
+                    strategy_domain, limit=1):
+                return strategy.fixed_location_id.id
+        else:
+            return super(ProductPutawayStrategy, self).putaway_apply(
+                putaway_strategy, product)
+
+
+class FixedPutawayStrat(models.Model):
```
pedrobaeza: Don't short the class name

stock_putaway_product/views/product.xml
```diff
@@ -0,0 +1,30 @@
...
+     © 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3). -->
+<odoo>
+    <record id="view_template_property_form" model="ir.ui.view">
+        <field name="name">product.template.product.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="stock.view_template_property_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='stock_property']" position="before">
+                <group colspan="4" name="putaway">
```
pedrobaeza: Put a string in the group like "Put Away Locations"

stock_putaway_product/models/product_putaway.py
```diff
@@ -0,0 +1,79 @@
...
+    )]
+
+    @api.model
+    def get_default_inventory_id(self):
+        return self.env['stock.inventory'].search(
+            [('id', '=', self.env.context.get('active_id', False))])
+
+    putaway_id = fields.Many2one(
+        comodel_name='product.putaway',
+        sting='Put Away Method',
```
pedrobaeza: This should be "Put Away Strategy"

stock_putaway_product/models/product_putaway.py
```diff
@@ -0,0 +1,79 @@
...
+    def _get_putaway_options(self):
+        ret = super(ProductPutawayStrategy, self)._get_putaway_options()
+        return ret + [('per_product', 'Fixed per product location')]
+
+    product_location_ids = fields.One2many(
+        comodel_name='stock.product.putaway.strategy',
+        inverse_name='putaway_id',
+        string='Fixed per product location',
+        copy=True)
+    method = fields.Selection(
```
pedrobaeza: Don't overwrite the field. Just use `selection_add`

pedrobaeza:
I don't see the product table in the put away strategy view.

pedrobaeza:
Please check also that in the product view, the label for putaway_id field is "Putaway Id" instead of "Put Away Strategy".

stock_putaway_product/models/product_putaway.py
carlosdauden: Original field declared in old API with defined function _get_putaway_options to load selection options: https://github.com/odoo/odoo/blob/6d43f280c9bfa589a88dec9433948861a93b39d7/addons/stock/product.py#L542

carlosdauden:
@pedrobaeza changes done.

gurneyalex:
@pedrobaeza I'm not sure of the review flow with the new tool. Carlos says the requested changes were made but your NOK review is still there. Is there a manual action required on your side?

pedrobaeza:
Yes, I need to validate the changes manually and press on "Approve changes" then

stock_putaway_product/models/product_putaway.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @api.model
+    def putaway_apply(self, putaway_strategy, product):
+        if putaway_strategy.method == 'per_product':
+            offset = self.env.context.get('starting_strategy', 0)
+            strategy = putaway_strategy.product_location_ids.filtered(
+                lambda x: (x.product_product_id == product or
+                           (not x.product_product_id and
+                            x.product_tmpl_id == product.product_tmpl_id))
+            )[offset:offset+1]
```
gurneyalex: using `filtered` here seems very convoluted to me, especially since for some weird reason `offset` comes from the context. 

Using `[offset:offset+1]` over `[offset]` only makes sense if the offset may be greater than the number of filtered results, in which case `strategy` will be an empty resultset and the method returns `None`. 

Could you add some documentation on the behavior of this method?

pedrobaeza:
@carlosdauden, what have you finally do? Have you removed the context argument?

rousseldenis (approved):
:+1: Tested on runbot. Great job!

pedrobaeza (approved):
Reason explained directly by Carlos, so :+1:

PR #198, opened 2016-09-30
[8.0] stock_account_quant_merge

JordiBForgeFlow:
This module is an extension of "stock_quant_merge", and adds the cost as a
criteria to merge quants.

Odoo splits quants each time a reservation is done: this module makes Odoo
merge them back if they still meet the following requirements:
- same product
- same serial number/lot
- same location
- same package
- same unit cost


pedrobaeza:
You can change the base branch in the same PR without closing it, so please don't open new ones.

PR #203, opened 2016-10-09
[10.0] Migrate stock_mts_mto_rule

Garamotte:
Changes in this PR :
- Updated from v8.0 branch + included #157
- The `_compute_qty` method of product.uom has been replaced by `_compute_quantity`
- The `_get_mts_mto_procurement` and `_run` methods of procurement.order have been converted from @api.model to @api.multi
- Added the `_get_route_name` method override, to add the `mts_mto` type (+ adapted the corresponding `_format_routename` calls)
- The `change_route` method of stock.warehouse has been replaced by `_update_routes`, and no longer has any argument

PR #204, opened 2016-10-15
[MIG][9.0] Migrate stock_mts_mto_rule

Garamotte:
Superseds #157 :
- Add missing commits from changes in 8.0 branch
- Includes changes made in #157
- Updated README.rst
#203 was based on #157, and has already been merged, so I think this one can be quickly merged.

### DIFF
--- a/stock_mts_mto_rule/README.rst
+++ b/stock_mts_mto_rule/README.rst
@@ -49,6 +49,15 @@ Configuration
 
 You have to select 'Use MTO+MTS rules' on the company's warehouse form.
 
+Known issues
+============
+
+If you cancel a delivery order and then recreate it from Recreate
+Delivery Order button in sale order form, then the stock level at the time of
+the Re-Creation won't be taken into account. So if a purchase order was created
+when the sale order was first validated, a similar purchase order will be created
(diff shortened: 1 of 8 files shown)


pedrobaeza (approved):

pedrobaeza:
Some tests that don't reduce coverage would be good.

Garamotte:
Updated unit tests.

I didn't achieve to reach this line : https://coveralls.io/builds/8367627/source?filename=stock_mts_mto_rule/model/procurement.py#L95

pedrobaeza:
Well, it's a lot. If you don't want to be faulted by that line and it's an strange condition that is rarely given, you can bypass coverage using `# pragma: no cover` comment in the `if` sentence.

Garamotte:
I think this is better to know that this line isn't tested than adding a no cover, isn't it ?

pedrobaeza:
OK, good to me. Let's wait for a second reviewer to merge.

Doscaal:
:+1:

pedrobaeza:
We need a third one before 5 working days for merging. Please ping me if there's no one but 5 working days has past.

andhit-r (approved):
:+1:

PR #206, opened 2016-10-23
[9.0] [NEW] stock_available_unreserved

JordiBForgeFlow:
# 
# Stock Available Unreserved

This module allows users to check the quantity of a stocked product that is
available on-hand, and that has not yet been reserved for use anywhere else.

This key figure is very important during the monitoring of the warehouse
execution, because it assists users to ensure that the flow of products will
not be stuck due to a sudden unavailability of stock.

If the warehouse personnel ensures that the unreserved quantity on hand > 0,
then nobody will be stuck in pickings or manufacturing orders waiting for
the availability of unreserved stock.


JordiBForgeFlow:
cc @sbidoul

stock_available_unreserved/models/product.py
```diff
@@ -0,0 +1,141 @@
...
+        domain_quant += domain_products
+
+        domain_quant.append(('reservation_id', '=', False))
+
+        domain_quant += domain_quant_loc
+
+        return domain_quant
+
+    @api.multi
+    def _product_available_not_res_hook(self, quants):
```
AaronHForgeFlow, line 107: I don't understand why this is necessary. The hook is only at the end of the '_product_available' method so we just need to call super to inherit, don't we?

AaronHForgeFlow (commented):
Runbot is not available but tested on local :+1:

Cedric-Pigeon (approved):

JordiBForgeFlow:
Thanks for your feedback @Cedric-Pigeon

pedrobaeza:
Can you provide tests?

stock_available_unreserved/models/product.py
JordiBForgeFlow: Not really because you do not have quants when you inherit

rousseldenis:
Even if this one was pushed some months ago, maybe more reviews should be preferable. Moreover, one function is not covered by tests.

JordiBForgeFlow:
Thanks @rousseldenis. We will make a separate PR to cover with tests the missing function.
cc @mreficent

rousseldenis:
@jbeficent Otherwise :+1:  ;-)

MiquelRForgeFlow:
@rousseldenis 
The method <code>action_open_quants_unreserved()</code> is an action that returns a simple dictionary from <code>product.template</code>. Thus, I think that the method is not necessary to be covered in the test but just tested manually.

PR #209, opened 2016-10-24
[8.0] [NEW] stock_orderpoint_procure_uom

JordiBForgeFlow:
This module allows users users to define what unit of measure should be used
in procurements created from minimum stock rules.

A typical use case would be a product that is stocked in centimeters, and
needs to be restocked in meters from another warehouse. When the picking is
created, the quantity to be transferred will be expressed in meters, making
it easier for the people responsible for the transfers to understand the
requirement.


pedrobaeza:
Name should be _stock_warehouse_orderpoint_uom_, don't you think?

JordiBForgeFlow:
Hmm, yes, you're right. Now changing it.

JordiBForgeFlow:
When using the run_scheduler() I get this error:
https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/170173383#L971

Not sure why. the standard odoo module also run the scheduler in tests:
https://github.com/odoo/odoo/blob/8.0/addons/stock/tests/test_resupply.py#L37

pedrobaeza:
There shouldn't be a problem with that if you use it correctly. Maybe you should mask your behavior while executing the rest of the tests. See https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

JordiBForgeFlow:
Ok, fixed. Thanks again @pedrobaeza !!

JordiBForgeFlow:
I was incorrectly calling super() on a test method :S

stock_orderpoint_uom/__openerp__.py
```diff
@@ -0,0 +1,19 @@
...
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Orderpoint UoM",
+    "summary": "Allows to create procurement orders in the UoM indicated in "
+               "the orderpoint",
+    "version": "8.0.1.0.0",
+    "author": "Eficent Business and IT Consulting Services S.L,"
+              "Odoo Community Association (OCA)",
+    "website": "https://www.odoo-community.org",
```
andhit-r, line 12: Not your company website?

stock_orderpoint_uom/__openerp__.py
```diff
@@ -0,0 +1,19 @@
...
+    "category": "Warehouse Management",
+    "depends": ["stock"],
+    "data": ["views/stock_warehouse_orderpoint_view.xml"],
+    "license": "AGPL-3",
+    'installable': True,
+    'application': True,
```
andhit-r: Is it really need to be an aplication?

andhit-r (changes requested):
Functionality: :+1: tested on runbot
Code: waiting for green travis

stock_orderpoint_uom/__openerp__.py
JordiBForgeFlow: Removed that.

stock_orderpoint_uom/__openerp__.py
JordiBForgeFlow: Hmm, not in this case.

stock_orderpoint_uom/models/procurement_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+from openerp import api, models
+
+
+class ProcurementOrder(models.Model):
+    _inherit = "procurement.order"
+
+    @api.multi
+    @api.onchange('product_id')
+    def onchange_product_id(self):
```
AaronHForgeFlow: I'm getting an error in runbot when creating a new procurement order and selecting a product, in the on_change method.

AaronHForgeFlow (commented):
The rest looks fine, Both the runbot and the code.

JordiBForgeFlow:
fixed. @aheficent @pedrobaeza Can you approve?

AaronHForgeFlow (approved):
The error I commented is fixed. So :+1:

andhit-r (approved):
:+1:

PR #210, opened 2016-10-24
[8.0] [NEW] stock_orderpoint_manual_procurement

JordiBForgeFlow:
# 
# Stock Orderpoint Manual Procurement

This module allows users to manually start procurements from the list of
reordering rules, based on the quantity that is recommended to be procured.
# Usage

Go to 'Configuration / Reordering Rules' and review the quantity recommended
to be procured. You can now start the procurement for a single or a list of
reordering rules.

The recommended quantity to procure is adjusted to the procurement unit of
measure indicated in the reordering rule.


JordiBForgeFlow:
Depends on merging https://github.com/OCA/stock-logistics-warehouse/pull/209

andhit-r:
@jbeficent thanks for your contributions. Few questions:
1. Why do you need to depend on stock_orderpoint_uom?
2. Why user can adjust qty procured? Why not just read min max qty? Isn't that defeat the purpose of order point?

JordiBForgeFlow:
@andhit-r 
I want to depend on stock_orderpoint_uom because I want to show the recommended qty to procure in the procurement_uom, if the user maintained it.

The orderpoints min and max quantity become recommendations. The user can stick to them, or propose his own procurement. That's the beauty of it. You DON'T have to stick to what the reordering rule recommends in order to properly manage your warehouse.

andhit-r:
@jbeficent having a feature to change procure qty surely has it own benefit. I am 100% agree with that. But procurement rule are agreement between department (purchase, warehouse, finance). You can not change it easily.

My propose alternatives is to restrict qty changes to specific group

JordiBForgeFlow:
@andhit-r This module introduces the possibility to control WHEN and HOW MUCH to procure, and converts the reordering rules in recommendations to reorder. But the user has the last word on both parameters.

If a company does not want to let users have this level of control, they should not install this module.

Perhaps this does not make sense to you know from a business perspective. It is part of a bigger scheme that I'm putting in place to empower users to better plan and execute the logistics in order to reach the optimum stock levels. Give me a few days and you'll see other modules that go in this same direction.

This module is, IMHO, formally correct, and has a valid business case.

andhit-r:
@jbeficent i also have a business case that require manual procurement, but the manual procurement has to respect procurement rule. That's why i suggest the additional group to change qty. If user belong to that group then the select qty table showed and if not than the select qty table is invisible.

JordiBForgeFlow:
@andhit-r I have incorporated your feedback comments regarding the group allowed to change the quantity. And I also added tests.

andhit-r:
@jbeficent thanks for the changes. :+1:

JordiBForgeFlow:
cc @eficent

stock_orderpoint_uom/models/procurement_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class ProcurementOrder(models.Model):
+    _inherit = "procurement.order"
+
+    @api.multi
+    @api.onchange('product_id')
+    def onchange_product_id(self):
+        for rec in self:
+            rec.procure_uom_id = rec.product_id.uom_id
```
AaronHForgeFlow, line 16: I'm getting an error when selecting the product in the procurement form: "TypeError: 'int' object is not iterable". May the cause of this error be here?

stock_orderpoint_manual_procurement/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,65 @@
...
+            op.procure_recommended_date = \
+                procurement_model._get_orderpoint_date_planned(
+                    op, datetime.today())
+            procure_recommended_qty = 0.0
+            prods = procurement_model._product_virtual_get(op)
+            if prods is None:
+                continue
+            if float_compare(prods, op.product_min_qty,
+                             precision_rounding=op.product_uom.rounding) < 0:
+                qty = max(op.product_min_qty, op.product_max_qty) - prods
```
AaronHForgeFlow, line 32: I've created a Reordering rule for the product [DVD] Blank DVD-RW with the following values:
Minimum Quantity 5.000 
Maximum Quantity 25.000
Quantity Multiple 1.000
Unit of measure: Dozens
The quantity on hand is 0.00 and the forecasted quantity is 0.00 dozens. 

When I create a procurement  and I select the product and then the unit of measure, the quantity proposed is the maximum quantity in the reordering rule and the unit of measure is Dozens.

It works as expected, however why do we choose the maximum instead of the minimum?

andhit-r: @aheficent minimum qty is a trigger to let procurement engine start procuring. So if the stock level drop below minimum qty it will trigger procurement. Max qty will tell procurement engine how many will it procure. Here are few links regarding stock level:

https://www.lokad.com/min-max-inventory-planning-definition
http://accountlearning.blogspot.co.id/2010/05/minimum-level-or-safety-stock-and-its.html

andhit-r:
@mikevhe18 @azmimr67 @m-sani would you kindly help review this PR. Many thanks.

m-sani:
👍

mikevhe18:
:+1:

JordiBForgeFlow:
Thanks all for your reviews! This is now merged. 

cc @lreficent can you port to v9?

LoisRForgeFlow:
@jbeficent This module depends on _stock_orderpoint_uom_ module, which is not available in v9. Might we migrate that one first?

PR #214, opened 2016-10-28
[8.0][ADD] stock_orderpoint_generator: backport from 9.0

Daniel-CA:
Backport from 9.0 version

### DIFF
--- a/stock_orderpoint_generator/models/orderpoint_template.py
+++ b/stock_orderpoint_generator/models/orderpoint_template.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+# © 2012-2016 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from openerp import api, fields, models
+
+
+class OrderpointTemplate(models.Model):
+    """ Template for orderpoints
+
+    Here we use same model as stock.warehouse.orderpoint but set product_id
+    as non mandatory as we cannot remove it. This field will be ignored.
+
+    This has the advantage of ensuring that the order point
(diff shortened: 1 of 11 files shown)


bealdav:
Really thanks :+1: 

except test stock_traceability_operation.tests.test_traceability: FAIL

Maybe solved in origin ? could you rebase to test again ?

Daniel-CA:
Rebase done, let's see if now is green.

Cedric-Pigeon (approved):

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,55 @@
...
+
+    This has the advantage of ensuring that the order point
+    and the order point template have the same fields.
+
+    _table is redefined to separate templates from orderpoints
+    """
+    _name = 'stock.warehouse.orderpoint.template'
+
+    _inherit = 'stock.warehouse.orderpoint'
+    _table = 'stock_warehouse_orderpoint_template'
```
LoisRForgeFlow: This is the default name it would take for this model, isn't it? thus not needed I think.

Daniel-CA: Hi @lreficent,
Looks like that is not necessary as it is the same as the new class name, I just maintain the code as it was in the backport module.

LoisRForgeFlow: Yes, then better to take the occasion to remove it!

Daniel-CA:
Change done.

agaldona (approved):

bealdav (approved):

PR #217, opened 2016-11-07
[8.0] [stock_warehouse_orderpoint_stock_info]

JordiBForgeFlow:
This is a port from OdooMRP. https://github.com/odoomrp/odoomrp-utils/tree/8.0/stock_warehouse_orderpoint_stock_info


JordiBForgeFlow:
cc @mikelarre, @anajuaristi @pedrobaeza

JordiBForgeFlow:
@pedrobaeza @anajuaristi You cannot store the field `available` (https://github.com/OCA/stock-logistics-warehouse/pull/217/files#diff-954246850314d5d706bb6495fd5f4400R45) because the computation depends on field 'product_location_qty', which is not an stored fields, and depends on the quantity on hand.

JordiBForgeFlow:
@pedrobaeza @anajuaristi I removed the field 'is enough product available?' (https://github.com/odoomrp/odoomrp-utils/blob/8.0/stock_warehouse_orderpoint_stock_info/models/stock_warehouse_orderpoint.py#L36) because it was incorrectly defined.

It was meant to be used to search for products that do not meet the minimum, and was defined as a stored field. However after running tests we found that this stored field was not being recomputed correctly, because it's computation depends on field 'product_location_qty', which is not a stored field.

If you want to be able to apply such field you have to provide the correct dependencies in the method. It's a complex topic, and we're also addressing it, but will need to be done in a separate module, in any case.

pedrobaeza:
I don't know this module, Jordi. I let in Ana's side.

oihane:
I don't remember exactly what was it meant to do. As @jbeficent said is only used in tree view, I think is a way to access easily to those that have available qty less than minimum stock.

It was asked to include by @pedrobaeza https://github.com/odoomrp/odoomrp-utils/pull/27#issuecomment-74638920

JordiBForgeFlow:
@oihane It is used in the search. And that's where it's critical, because searching for this field will for sure drive wrong results.

pedrobaeza:
Wow, I didn't even remember that I reviewed this module... bad memory. Anyways, being in odoomrp-utils, you know that I deny of modules from that second repo, hehe.

JordiBForgeFlow:
In any case, I removed that field. You can check the PR and provide your feedback.

stock_warehouse_orderpoint_stock_info/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,46 @@
...
+                location=rec.location_id.id
+                )._product_available()[rec.product_id.id]
+            rec.product_location_qty = product_available['qty_available']
+            rec.incoming_location_qty = product_available['incoming_qty']
+            rec.outgoing_location_qty = product_available['outgoing_qty']
+            rec.virtual_location_qty = product_available['virtual_available']
+
+    @api.multi
+    @api.depends('product_location_qty', 'product_min_qty')
+    def _compute_product_available(self):
```
oihane: You can remove this part as it does not have `available` field.

oihane (changes requested):

oihane (approved):
:+1:  LGTM

mikelarre:
:+1:  when travis goes green

JordiBForgeFlow:
This one is ready to merge.

PR #226, opened 2016-12-05
[ADD]stock_location_type_accounting_info

mikevhe18:
In current state, accounting information can only be defined when type of locations is inventory or production. With this module, user allows to define it not only type locations is inventory or production but customer, supplier, or transit too.


rousseldenis:
@mikevhe18 This will be closed due to inactivity

PR #228, opened 2016-12-14
[8.0] stock_valuation_account_manual_adjustment

JordiBForgeFlow:
Stock Inventory Account Manual Adjustment
=========================================

This module shows in the product inventory stock value and the accounting
value, and allows to create accounting adjustment entries to align the
accounting value to match with the inventory stock value.

Configuration
=============

Assign the group 'Stock Valuation Account Manual Adjustments' to the users
that should be allowed to list the valuation discrepancies and to reconcile
them.

Usage
=====

* Go to 'Warehouse / Inventory Control / Real Time Inventory Valuation'

* Filter on 'Valuation Discrepancy != 0.0' using the advanced filter.

* Select the products that you wish to reconcile and press 'More /
  Adjust Stock Valuation Account Discrepancies'.

Known issues / Roadmap
======================

* In order to properly manage the inventory valuation from an accounting
  perspective all journal items created for inventory accounts should
  include the product.

* The price change function of Odoo creates journal entries to the inventory
  account without specifying the product. This is bad because you loose the
  ability to control the inventory valuation from accounting at the level of
  the product. In order to fix this, use the module
  'stock_inventory_revaluation', to be found in the same OCA repository.


JordiBForgeFlow:
cc @eficent

andhit-r:
@jbeficent i have a case study that you might want to consider

* YourCompany has an **Finished Good Account** to hold inventory accounting value
* A manual journal entry created to open balance for **Finished Good Account**. Let say:
&nbsp;&nbsp;&nbsp;&nbsp;Finished Good Account Dr. 700
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Opening Balance Cr. 700
This journal entry obviously still create discrepancies because manually created journal entry (and move) could not fill product_id field on account.move.line
* Manual Valuation then should be performed to adjust. Here is the catch, (1) YourCompany does not want to create additional account as contra-valuation-account, (2) **Finished Good Account** balance after manual valuation perform should not change.
* So manual valuation performed with wizard value: (1)  Valuation Increase Contra-Account == **Finished Good Account**, and (2) Valuation Decrease Contra-Account == **Finished Good Account**. This to make sure that final balance will not change. Journal entry will be:
&nbsp;&nbsp;&nbsp;&nbsp;Finished Good Account Dr. 700
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Finished Good Account Cr. 700
* Now both journal items above will have product_id filled
* The discrepancies will still != 0

Screenshot for manual valuation:
manual_valuation_1

Screenshot for created journal entry:
manual_valuation_2

Screenshot discrepency after valuation:
manual_valuation_3

IMO you should only assign product_id to one journal item only.

stock_valuation_account_manual_adjustment/wizards/mass_create_view.xml
```diff
@@ -0,0 +1,68 @@
...
+                 <form string="Reconcile Inventory Account">
+                     <group>
+                         <field name="increase_account_id"/>
+                         <field name="decrease_account_id"/>
+                         <field name="journal_id"/>
+                         <field name="remarks"/>
+                     </group>
+                     <newline/>
+                     <group colspan="2">
+                         <button name="process"
```
andhit-r, line 20: It's better to put the buttons on <footer>

JordiBForgeFlow: What do you mean?

JordiBForgeFlow: @andhit-r Regarding your long comment, if a company runs perpetual inventory the inventory account should be detailed at a product level. I recently performed an initial balance for a company, and I asked the quantity on stock and the valuation of each product, in order to create the opening entry, detailed at product level.

In order to load what I did was:
1-. Enter the correct standard cost in the product
2.- Set the product valuation to real_time
3.- Change the virtual inventory location to use as debit and credit account "Opening Balance Equity"
4.- Perform an inventory adjustment, to add the on hand quantities as they reported

As a result journal entries were created for each product as:
Dr. Finished Goods
Cr. Opening Balance Equity

That's the correct way to do IMHO. No need to use the manual adjustments tool at all.

andhit-r: True, but not all implementation project start that way. What is a company use accounting only module first?

andhit-r: You should put all buttons inside footer tag. 
Without footer tag:
manual_valuation_wizard1

With footer tag:
manual_valuation_wizard2

JordiBForgeFlow: If a company starts using accounting only, when they switch to stock accounting they should open a new GL account to manage inventory valuation, of course! Then they will need to credit the balance from the old and debit into the new. Same procedure as above, but replace "opening balance equity" with "old inventory account". :)

Thanks for the feedback!

stock_valuation_account_manual_adjustment/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Assign the group 'Stock Valuation Account Manual Adjustments' to the users
+that should be allowed to list the valuation discrepancies and to reconcile
+them.
+
+Usage
+=====
+
+* Go to the list of products or product variants. You can see in the tree
+  view the inventory value and the accounting value.
```
andhit-r: Inventory Value, Accounting Value, and Discrepancy can only be seen on *Real Time Inventory Valuation*

stock_valuation_account_manual_adjustment/models/stock_valuation_account_manual_adjustment.py
```diff
@@ -0,0 +1,177 @@
...
+            'date': date_move,
+            'product_id': prod.id,
+            'account_id': account_id,
+            'move_id': move.id,
+            'credit': amount
+        }
+
+    @api.multi
+    def post(self):
+        timenow = time.strftime('%Y-%m-%d')
```
andhit-r, line 141: Why not use fields.Datetime.now()? You use it on https://github.com/OCA/stock-logistics-warehouse/pull/228/files#diff-ecb3b276206e9b56ac80d5cccbfe6518R167.

andhit-r:
@jbeficent i'm still sure that you only need to assign product_in on one move line. Strange things always happen on implementation hahaha. But for now it's fine. I will create PR if my use case happen in the future. Waiting to pass all test before tumbs. Thank you for this PR. Can't wait to use it on production.

JordiBForgeFlow:
@eficent can you please review?

stock_valuation_account_manual_adjustment/init_hook.py
```diff
@@ -0,0 +1,55 @@
...
+def set_stock_valuation_account_manual_adjustment_in_account_move_line(cr):
+    cr.execute("""SELECT column_name
+    FROM information_schema.columns
+    WHERE table_name='account_move_line' AND
+    column_name='stock_valuation_account_manual_adjustment_id'""")
+    if not cr.fetchone():
+        cr.execute(
+            """
+            ALTER TABLE account_move_line
+            ADD COLUMN stock_valuation_account_manual_adjustment_id
```
JordiBForgeFlow, line 50: @pedrobaeza Can you help me with this? I am adding field 'stock_valuation_account_manual_adjustment_id' in account.move.line as a related with move_id, and in order to speed-up the install I am pre-creating the field. However at the time of install the field is being still computed. Why?

stock_valuation_account_manual_adjustment/README.rst
JordiBForgeFlow: Thanks. Done!

LoisRForgeFlow (approved):
LGTM :+1:

JordiBForgeFlow:
@andhit-r Runbot is green. However there's still one issue related to the pre-init hook. I'm not sure if it's being effective.

AaronHForgeFlow (approved):
LGTM :+1: 

I think also that it would be a good idea to have the filter 'Valuation Discrepancy != 0.0'  in the "Real-Time Inventory valuation" tree view.

stock_valuation_account_manual_adjustment/init_hook.py
AaronHForgeFlow: I think that computed with store=True are always computed when the module is installed.

andhit-r:
:+1: . It's better if all status checks pass though.

stock_valuation_account_manual_adjustment/init_hook.py
JordiBForgeFlow: In fact the pre-init code is running correctly.

JordiBForgeFlow:
Thanks @andhit-r. Now merged

PR #230, opened 2016-12-19
[8.0] stock_account_change_product_valuation

JordiBForgeFlow:
Stock Account Change Product Valuation
======================================

If your company runs a perpetual inventory system, changing the
product's costing method or the type of product requires proper readjustments:

Changing the type of product.
* From consumable to stockable: the resulting inventory value will be reset
  to 0, since everything that was received as a  stockable was expensed out at
  the time of entering the supplier invoice.

Changing the costing method:

* From standard or average to real: the cost of all the existing quants will
  be reset to the product's standard price.
* From real to standard or average: the standard price of the product
  template will be set to the average unit cost of all the internal quants.

.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
   :alt: Try me on Runbot
   :target: https://runbot.odoo-community.org/runbot/154/8.0

Known issues / Roadmap
======================

Pending to do:

* From stockable to consumable: The inventory value should be
  completely removed. Most importantly when you run a perpetual inventory
  (connected with accounting)


andhit-r:
LGTM :+1:

LoisRForgeFlow (approved):
Tested on runbot :+1:

AaronHForgeFlow (commented):
When I select 'stockable' as product type and 'standard price' as stock method and I buy, the inventory value of all quants are readjusted to product standard price. This behaviour is not due to this module but maybe this module should prevent this to happen, what do you think?

AaronHForgeFlow (approved):
Forget about my previous comment. If the product is standard price the value of the quants should be the standard price * quantity. It does not matter what was the previous value.

:+1:

JordiBForgeFlow:
@lreficent Can you migrate to 9.0?

PR #231, opened 2016-12-20
[8.0] account_move_line_product

JordiBForgeFlow:
Account Move Line Product
=========================

This module enhances the views associated to account moves and account move
lines to display the product that is associated to the move line.

This will be relevant in perpetual inventory, for inventory-related account
moves.


JordiBForgeFlow:
cc @eficent

LoisRForgeFlow (approved):
Tested on runbot + code review :+1:

AaronHForgeFlow (approved):
:+1: 

Very useful when adjusting stock valuation discrepancies

rousseldenis:
@jbeficent I think there is a problem of context here ...
Maybe you could have waited a little bit more reviews before merging ?
See my remark in https://github.com/OCA/stock-logistics-warehouse/pull/238

Functional need (inventory) can not just justify the location of this module here IMO
@lmignon @adrienpeiffer

JordiBForgeFlow:
@rousseldenis Oh, I see... This feature is, however, a grey zone... It is sufficiently justified in the README that this feature is associated to the perpetual inventory management. I propose that we stick to this branch now, in order to avoid futher confusion between versions.

rousseldenis:
@jbeficent Ok. That was just for consistency reasons (as if I want accounting modules, I go naturally in accounting repos :-) )

PR #233, opened 2016-12-22
[9.0]-[mig]-stock_orderpoint_uom

LoisRForgeFlow:
Stock Orderpoint UoM
================

This module allows users users to define what unit of measure should be used
in procurements created from minimum stock rules.

A typical use case would be a product that is stocked in centimeters, and
needs to be restocked in meters from another warehouse. When the picking is
created, the quantity to be transferred will be expressed in meters, making
it easier for the people responsible for the transfers to understand the
requirement.

Usage
=====

Go to 'Configuration / Reordering Rules' and indicate a Procurement UoM.

### DIFF
--- a/stock_orderpoint_uom/models/stock_warehouse_orderpoint.py
+++ b/stock_orderpoint_uom/models/stock_warehouse_orderpoint.py
@@ -13,13 +13,6 @@ class StockWarehouseOrderpoint(models.Model):
     procure_uom_id = fields.Many2one(comodel_name='product.uom',
                                      string="Procurement UoM")
 
-    @api.multi
-    @api.onchange('product_id')
-    def onchange_product_id(self):
-        for rec in self:
-            if rec.procure_uom_id:
-                rec.procure_uom_id = False
-
     @api.constrains('product_uom', 'procure_uom_id')
     def _check_procure_uom(self):
         if any(orderpoint.product_uom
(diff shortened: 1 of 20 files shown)


andhit-r:
:+1:

JordiBForgeFlow (approved):
:+1: Technical and functional review

JordiBForgeFlow:
@aheficent can you review?

AaronHForgeFlow (approved):
Functional test and code review :+1:  Good job!

stock_orderpoint_uom/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models, _
+from openerp.exceptions import Warning as UserError
```
bealdav, line 7: No more Warning in v9 ?, import UserError directly, I think

LoisRForgeFlow: I don't know about that, have you got some reference/documentation for that? Anyway the PR is already merged.

bealdav: https://github.com/odoo/odoo/blob/9.0/openerp/exceptions.py#L31-L38

LoisRForgeFlow: Yes, I've seen now that this is also mentioned in CONTRIBUTING.md file.

cc @jbeficent should we add this change?

JordiBForgeFlow:
Yes, as a separate PR.

PR #234, opened 2016-12-23
[9.0]-[mig]-stock_orderpoint_manual_procurement

LoisRForgeFlow:
Stock Orderpoint Manual Procurement
===================================

This module allows users to manually start procurements from the list of
reordering rules, based on the quantity that is recommended to be procured.

Usage
=====

Go to 'Inventory > Inventory Control > Reordering Rules' and review the
quantity recommended to be procured. You can now start the procurement for a
single or a list of reordering rules.

The recommended quantity to procure is adjusted to the procurement unit of
measure indicated in the reordering rule.

If you want users to be able to change the recommended quantity to procure,
you should assign them to the security group 'Change quantity in manual
procurements from reordering rules', under 'Settings / Users / Users'.


JordiBForgeFlow:
Depends on https://github.com/OCA/stock-logistics-warehouse/pull/233

JordiBForgeFlow (approved):
Technical and functional review :+1:

stock_orderpoint_manual_procurement/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    "depends": ["stock",
+                "stock_orderpoint_uom"],
+    "data": ["security/stock_orderpoint_manual_procurement_security.xml",
+             "wizards/make_procurement_orderpoint_view.xml",
+             "views/procurement_order_view.xml",
+             "views/stock_warehouse_orderpoint_view.xml"
+             ],
+    "license": "AGPL-3",
+    'installable': True,
+    'application': True,
```
AaronHForgeFlow: Is it not a single module?
Modules extend functionality. An app is a collection of modules, isn't it?

AaronHForgeFlow (changes requested):
I'm getting this qty when I procure less than the recommended qty:
image

LoisRForgeFlow:
@aheficent I cannot reproduce the error. But I think it can be something related with rounding precision of the uom.
image
In that case you probably have to set up a rounding of a whole unit.

stock_orderpoint_manual_procurement/__openerp__.py
LoisRForgeFlow: Yes, you likely right. I'll fix it

AaronHForgeFlow:
Yes you are right. By default the rounding precision for units of measure is 0.001, which is wrong.

AaronHForgeFlow (approved):
LGTM now :+1:

andhit-r (approved):
:+1:

JordiBForgeFlow:
@lreficent thanks. Can you squash your own commits? https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

PR #236, opened 2016-12-27
[8.0] [stock_inventory_revaluation] fix quant unlink

JordiBForgeFlow:
The stock inventory revaluation tool allows to revaluate negative quants. But these quants can ultimately be deleted automatically by the application, once they are merged with positive quants.

This PR allows to remove the records of 'stock.inventory.revaluation.quant' when the quant associated to them is deleted. Otherwise the users are stuck with the message:

`Integrity Error

The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set`

### DIFF
--- a/stock_inventory_revaluation/README.rst
+++ b/stock_inventory_revaluation/README.rst
@@ -51,6 +51,15 @@ Usage
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/154/8.0
 
+Known issues / Roadmap
+======================
+
+* A negative quant can be revaluated, but can also ultimately be merged with
+  a positive quant. In that case, the associated revaluation quant record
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
cc @eficent

andhit-r:
:+1:

It would be nice if there is a log somewhere to inform that a stock.inventory.revaluation.quant has been deleted. Or at least inform it on known issue section.

JordiBForgeFlow:
Thanks @andhit-r. I have extended the readme.

AaronHForgeFlow (approved):
:+1: 

stock.inventory.revaluation.quant records are deleted when the quant is deleted, avoiding the integrity error

LoisRForgeFlow (approved):
Tested :+1:

PR #239, opened 2017-01-05
[10.0][MIG] stock_reserve

andrius-preimantas:


rousseldenis (changes requested):
@andrius-preimantas Please check https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
to keep commit history.

Oops did it too fast...

rousseldenis (approved):
Code review

andrius-preimantas:
Ping :-)

rousseldenis:
Can we move on this ?
Need review

andrius-preimantas:
@rousseldenis can we ping the ones who should do the review in order to finally have PR merged?

rousseldenis:
@katyukha @gfcapalbo @jbeficent @lreficent @aheficent

LoisRForgeFlow:
I've forced rebuilt of runbot twice (yesterday and today) and it is not working yet :confused:

rousseldenis:
@lreficent It's working now

stock_reserve/view/product.xml
```diff
@@ -1,33 +1,31 @@
...
-        </record>
-    </data>
-</openerp>
+<record model="ir.ui.view" id="product_product_form_view_reservation_button">
+    <field name="name">product.template.reservation.button</field>
+    <field name="model">product.product</field>
+    <field name="inherit_id" ref="product.product_template_form_view"/>
+    <field name="arch" type="xml">
+        <xpath expr="//div[@name='button_box']" position="inside">
+            <button class="oe_inline oe_stat_button" name="action_view_reservations"
```
LoisRForgeFlow: This button is not being showed for the product.product.
image
I accessed it directly from the reservation.

stock_reserve/test/stock_reserve.yml
```diff
@@ -6,7 +6,7 @@
     # Travis, the hook performing this operation for the new stock reservation
     # location is run after the test execution. This causes the stock level
     # computation to be wrong at the time the tests are run.
-    self._parent_store_compute(cr)
+    self._parent_store_compute()
```
LoisRForgeFlow, line 9: I think it is a good occasion to move this to python test.

stock_reserve/model/stock_reserve.py
```diff
@@ -56,28 +56,65 @@ class StockReservation(models.Model):
...
+        #     3. look up property fields
+        #     4. look up field.default
+        #     5. delegate to parent model
+        #
+        # If picking_type_id is present and location_id is not, try to find
+        # default value for location_id
+        if not res.get('picking_type_id', None):
+            res['picking_type_id'] = self._default_picking_type_id()
+
+        picking_type_id = res.get('picking_type_id', None)
```
LoisRForgeFlow: Some reason to prefer <code>None</code> instead of <code>False</code>? Just wondering, for me makes no difference.

LoisRForgeFlow (commented):
Tested on runbot + code review

stock_reserve/model/stock_reserve.py
lasley: `.get` implicitly returns `None` when a key is not existing. `False` would only be returned in the event that the key exists, and is simply not set in the data. This would be the same as if directly accessing via `res.picking_type_id`

That said, the `None` here is superfluous & can be removed.

stock_reserve/test/stock_reserve.yml
andrius-preimantas: I agree. Although this PR focuses on migration and not improving/refactoring the module. My suggestion would be to push it through and improve it later on.

stock_reserve/view/product.xml
andrius-preimantas: Fixed it

andrius-preimantas:
@lreficent @lasley thanks for the feedback. Made the improvements based on your comments. It'd be nice to merge it with next occasion.

Cheers

LoisRForgeFlow:
Functionally works fine, but now the commit history of previous versions is lost and looking @rousseldenis first review I would say that it was fine at the beginning. Also, when you recover the history you can take the occasion to squash your commits.

lmignon:
@andrius-preimantas Can you take into account the last comment from @lreficent  so we can move forward with this PR and #274

stock_reserve/test/stock_reserve.yml
BetaUliansyah: I use it in my v10 odoo, and it works. Is it only me who notice that stock_reserve in apps.odoo.com refers to different module than this one?

BetaUliansyah:
I use it in my v10 odoo, and it works. Is it only me who notice that stock_reserve in apps.odoo.com refers to different module than this one?

rousseldenis:
@BetaUliansyah It seems that the github user stellaf had massively copied modules and pushed them on apps.odoo.com!
Moreover modules are full of craps (.pyc files, missing __manifest__.py, ...).
Maybe should we warn Odoo of this

BetaUliansyah:
@rousseldenis yeah, that's partly because OCA is much late to push their great repositories to apps.odoo.com.

lasley:
Hey all - OCA board has looked into this and is working with stellaf, who is being more than cooperative. This was just a simple mistake, and it should be noted that it was completely legal and within licensing permissions.

We're looking into some ways to avoid this from happening again. I do feel that @BetaUliansyah is correct, and that this is due to our slow merge process.

That said, I highly recommend that anyone interested in a faster merge process take a look at [our WIP release process](https://docs.google.com/document/d/1wwnu7oe5cDTyjZH3hdKkGLnF6fH4jPP3RLMPAhm3gjs), which may help with the speed of merges.

rousseldenis:
@lasley OK. The process is obviously weird and is not reflecting the work we do on OCA to people browsing odoo apps. The release process was done so in order to let the review process being done. Maybe the rules can evolve...

lasley:
The core of the issue is that Odoo SA apps store:
* Does not have an authenticated user - only the authors key is observed
* Accepts the first module of a name submitted for an Odoo version, regardless of previous name ownership
* Does not have a user prefix for module namespacing, such as we see in more evolved solutions like Dockerhub or Github


Given the above, stellaf was able to take precedence of the module name due to our slow release cycle. Hopefully the new cycle is able to speed things up and thus reserve the namespaces quicker.

Rules can most certainly evolve, and we are open to evaluating all suggestions.

rousseldenis:
@lasley Ok, got it. Thanks for explanation

gurneyalex (approved):

PR #246, opened 2017-01-31
[8.0][stock_putaway_strategy][FIX] Allows a user that is not Warehouse user or manager to access to the product form

JordiBForgeFlow:
When module "stock_putaway_strategy" is installed only Warehouse/User and Warehouse/Manager can access to the product form. 

After this PR, all users can access to the product form, but only users that belong to the group "Warehouse/User" can access to the section for Putaway strategies in the Inventory tab.

PR #247, opened 2017-01-31
[9.0][stock_putaway_strategy][FIX] Allows a user that is not Warehouse user or manager to access to the product form

JordiBForgeFlow:
Warehouse/User and Warehouse/Manager can access to the product form.
After this PR, all users can access to the product form,
but only users that belong to the group "Warehouse/User"
can access to the section for Putaway strategies in the Inventory tab.

Same as #246 but for 9.0


JordiBForgeFlow:
cc @mreficent @JosDeGraeve @lasley

lasley (approved):

LoisRForgeFlow (approved):

PR #251, opened 2017-02-17
[9.0][ADD] stock_cycle_count

LoisRForgeFlow:
This module provides the capability to execute a cycle count strategy in a
warehouse through different rules defined by the user. Cycle count is an
alternative to full wall-to-wall physical inventories in which little
portions (stock locations) of the stock are selected to count on a regular
basis.

The system propose locations in which to perform a inventory adjustment every
day based on a set of rules defined for the warehouse. In addition the system
can propose Zero-Confirmations which are simply and opportunistic counts to
check whether a locations has actually became empty or not.

With this strategy it is possible to:
* Remove the need to perform full physical inventories and to stop the
  production in the warehouse.
* Measure the accuracy of the inventory records and improve it.
* Correct inventory errors earlier and prevent them to become bigger.

Configuration
=============

You can configure the rules to compute the cycle count, acting as follow:

#. Go to "Inventory > Configuration > Cycle Count Rules"
#. Create as much cycle count rules as you want.
#. Assign the rules to the Warehouse where you want to apply the rules in.
#. Set a "Cycle Count Planning Horizon" for each warehouse.

Usage
=====

Once you have some rules configured for your warehouses, you can proceed as
is described below.

#. Go to "Inventory > Configuration > Warehouse Management > Warehouses".
#. Select all the warehouses you want to compute the rules in.
#. Click on "Action" and then in "Compute Cycle Count Rules".
#. Go to "Inventory Control > Cycle Counts".
#. Select a Cycle Count planned an confirm it, this will create a draft
    Inventory Adjustment.
#. In the right top corner of the form view you can access the generated
    Inventory Adjustment.
#. Proceed with the Inventory Adjustment as usual.


stock_cycle_count/README.rst
```diff
@@ -0,0 +1,93 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
+Stock Cycle Count
+=================
+
+This module adds the capability to execute a cycle count strategy in a
```
rousseldenis: Maybe, you should use the 'Inventory' term in the introduction as it is not obvious at first sight (you talk about it further).
That's the term used in Odoo. Just to be clearer.

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -0,0 +1,88 @@
...
+                                 default=_company_get)
+
+    @api.one
+    def do_cancel(self):
+        self.state = 'cancelled'
+
+    @api.model
+    def _prepare_inventory_adjustment(self):
+        return {
+            'name': 'INV/{}'.format(self.name),
```
rousseldenis, line 61: This one is not generated by sequence at create ?

stock_cycle_count/models/stock_cycle_count_rule.py
```diff
@@ -0,0 +1,198 @@
...
+        elif self.rule_type == 'turnover':
+            proposed_cycle_counts = self._compute_rule_turnover(locs)
+        elif self.rule_type == 'accuracy':
+            proposed_cycle_counts = self._compute_rule_accuracy(locs)
+        return proposed_cycle_counts
+
+    @api.model
+    def _propose_cycle_count(self, date, location):
+        cycle_count = {
+            'date': date.strftime('%Y-%m-%d %H:%M:%S'),
```
rousseldenis: DEFAULT_SERVER_DATETIME_FORMAT ?

stock_cycle_count/models/stock_cycle_count_rule.py
```diff
@@ -0,0 +1,198 @@
...
+            last_inventories = self.env['stock.inventory'].search([
+                ('location_id', '=', loc.id),
+                ('state', 'in', ['confirm', 'done', 'draft'])]).mapped('date')
+            if last_inventories:
+                latest_inventory = sorted(last_inventories, reverse=True)[0]
+                try:
+                    period = self.periodic_count_period / \
+                        self.periodic_qty_per_period
+                    next_date = datetime.strptime(
+                        latest_inventory, '%Y-%m-%d %H:%M:%S') + timedelta(
```
rousseldenis: Same as above

stock_cycle_count/models/stock_cycle_count_rule.py
```diff
@@ -0,0 +1,198 @@
...
+        moves = self.env['stock.move'].search([
+            '|', ('location_id', '=', location.id),
+            ('location_dest_id', '=', location.id),
+            ('date', '>', date),
+            ('state', '=', 'done')])
+        return moves
+
+    @api.model
+    def _compute_turnover(self, move):
+        turnover = move.product_uom_qty * move.product_id.standard_price
```
rousseldenis: Shouldn't you use the method :
`move.product_id.price_compute('standard_price', uom=move.product_uom.id)[move.product_id.id]`
to get product standard unit price and then multiply by quantity ?

stock_cycle_count/models/stock_inventory.py
```diff
@@ -0,0 +1,34 @@
...
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    @api.one
+    def _compute_inventory_accuracy(self):
+        total_qty = sum(self.line_ids.mapped('theoretical_qty'))
+        abs_discrepancy = sum(self.line_ids.mapped(
+            lambda x: abs(x.discrepancy_qty)))
+        if total_qty:
+            self.inventory_accuracy = 100 * (total_qty - abs_discrepancy) / \
```
rousseldenis: Shouldn't be 100.0 ?
And this one shouldn't be as static variable (I don't like hardcoded values... subjective) ?

rousseldenis (changes requested):
First code review
Seems great !

stock_cycle_count/README.rst
LoisRForgeFlow: Yes, you are probably right.

stock_cycle_count/models/stock_cycle_count.py
LoisRForgeFlow: No, the sequence is defined for the stock.cycle.count object. The stock.inventories created from the cycle counts just use the name of the parent cycle count for its own name without any sequence.

stock_cycle_count/models/stock_cycle_count_rule.py
LoisRForgeFlow: Good point! I have to check that, but seem that you are right.

stock_cycle_count/models/stock_cycle_count_rule.py
LoisRForgeFlow: will that make any difference?

stock_cycle_count/models/stock_inventory.py
LoisRForgeFlow: You mean just create a variable in the class, for instance <code>percent = 100.0</code>, and use it in the line you pointed?

Thanks for the review btw!  :grinning:

stock_cycle_count/models/stock_cycle_count.py
rousseldenis: Ok, sorry for that

stock_cycle_count/models/stock_cycle_count_rule.py
rousseldenis: Check the method in product. currency, uom, user context

stock_cycle_count/models/stock_inventory.py
rousseldenis: Yes!

stock_cycle_count/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. repo_id is available in https://github.com/OCA/stock-logistics-warehouse
+.. branch is "9.0" for example
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
MiquelRForgeFlow: project_repo → stock-logistics-warehouse

stock_cycle_count/models/stock_cycle_count_rule.py
LoisRForgeFlow: I'm attending this now and I found that the method you refer to (price_compute) is an odoo v10 method. What do you think about keeping it like this now and to add that change in a future migration to v10?

rousseldenis: In v9, I think this is  _compute_price(self, cr, uid, from_uom_id, price, to_uom_id=False), but maybe it's unnecessary

https://github.com/odoo/odoo/blob/9.0/addons/product/product.py

LoisRForgeFlow: With that method you get the 'sale price' while I'm trying to use the 'cost' which I think is what is used for inventory valuations and similar, right? So finally I found that the proper method is get_unit_price from stock.move.

I will use it, thanks!

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -0,0 +1,88 @@
...
+        vals['name'] = self.env['ir.sequence'].next_by_code(
+            'stock.cycle.count') or ''
+        return super(StockCycleCount, self).create(vals)
+
+    @api.model
+    def _company_get(self):
+        company_id = self.env['res.company']._company_default_get(self._name)
+        return company_id
+
+    name = fields.Char(string='Name', readonly=True)
```
alan196, line 28: The string is not necesary.

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -0,0 +1,88 @@
...
+    cycle_count_rule_id = fields.Many2one(
+        comodel_name='stock.cycle.count.rule',
+        string='Cycle count rule',
+        required=True)
+    state = fields.Selection(selection=[
+        ('draft', 'Planned'),
+        ('open', 'Execution'),
+        ('cancelled', 'Cancelled'),
+        ('done', 'Done')
+    ], string='State', default='draft')
```
alan196, line 44: The string is not necesary.

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -0,0 +1,88 @@
...
+    stock_adjustment_ids = fields.One2many(comodel_name='stock.inventory',
+                                           inverse_name='cycle_count_id',
+                                           string='Inventory Adjustment')
+    inventory_adj_count = fields.Integer(compute=_count_inventory_adj)
```
alan196, line 48: This field don't have string

stock_cycle_count/models/stock_cycle_count_rule.py
```diff
@@ -0,0 +1,202 @@
...
+            raise UserError(
+                _('You cannot define a negative or null number of counts per '
+                  'period.')
+            )
+        if self.periodic_count_period < 0:
+            raise UserError(
+                _('You cannot define a negative period.')
+            )
+
+    name = fields.Char('Name', required=True)
```
alan196, line 80: string is not required

stock_cycle_count/models/stock_cycle_count.py
LoisRForgeFlow: What do you mean? the label? I just like to use almost always labels to make the code more readable.

stock_cycle_count/models/stock_cycle_count.py
LoisRForgeFlow: Again, not sure what do you mean. In the tree view we need an string to be displayed.

stock_cycle_count/models/stock_cycle_count.py
LoisRForgeFlow: I didn't put string here because this is only used to display the number in a button, any string defined would be actually not used.

stock_cycle_count/models/stock_cycle_count_rule.py
LoisRForgeFlow: It's required for the tree view. What I'm missing here it's the 'string=' label in the definition to be consistent with all the code.

rousseldenis: @lreficent I think he wants to say that if string argument is not declared, a default string is built from variable name with Camelcase, so 'name' becomes 'Name' but I don't know v9, in v10 that's true. Even if I prefer explicit declaration...

LoisRForgeFlow: I was not aware of that, thanks you for point it out. Always good to learn!

alan196: That is what I was trying to say. The default label is working since v9.
Also there is a pylint-odoo to check this and I have configured my repos with travis to fail if you put the same name of the field.

JordiBForgeFlow:
@gmeficent can you review?

JordiBForgeFlow:
@alan196 @rousseldenis @mreficent Do you approve the PR after your feedback comments?

MiquelRForgeFlow (approved):
:+1:

alan196 (approved):
👍

gmeficent (approved):
Code and functional review. Works smoothly :+1:

PR #258, opened 2017-03-08
[9.0] Add Automatic Reordering Rules

guewen:
When a product variant is created with a (or several) rule template(s), it
automatically generates (on the fly) the corresponding rule(s). When a rule
template is modified, it automatically updates the existing rules of the
linked products (it uses the mechanism in place, disable the old and create
a fresh one). The latter update is done by a cron because it might take a
long time to update all the products reordering rules.

### DIFF
--- a/stock_orderpoint_generator/README.rst
+++ b/stock_orderpoint_generator/README.rst
@@ -6,7 +6,40 @@
 Order point generator
 =====================
 
-Add a wizard to configure order points for multiple products in one go.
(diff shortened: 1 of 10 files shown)


simahawk (approved):
code review: LGTM

lmignon (approved):
@guewen Thank you for this work. LGTM (Code review only!)

pedrobaeza (approved):

PR #259, opened 2017-03-14
[10.0] [ADD] migrate packaging management

lmignon:
#61 forward ported in 10.0

PR #261, opened 2017-03-16
[10.0][mig] stock_orderpoint_uom

LoisRForgeFlow:
Changes done:
* Basic migration: odoo tags, odoo imports, bump version...
* Update README.
* Fix moved inherited methods.
* Adapt method calls to new version.
* Squash commit history.


JordiBForgeFlow (approved):
:+1: Technical and functional review.

AaronHForgeFlow (approved):
:+1: LGTM

MiquelRForgeFlow (approved):
fine :+1:

lmignon:
@lreficent The same kind of functionality is provided by the purchase_packaging into #259 
The main difference between the two approaches is that the uom to use is taken on the supplierinfo into purchase_packaging.... It would be nice to have the 2 PRs converging. What do you think?

MiquelRForgeFlow:
@lmignon I think that they are similar features, but some companies do not use packaging at all. And also, as documented in the README, sometimes you have a need in 'mm' but you request in 'm'. Not related to packages.

MiquelRForgeFlow:
@lmignon If you agree we can merge this module.

LoisRForgeFlow:
I will proceed to merge this, after all, #259 is already merged so the possibility to converge both modules is not possible (at least for this version).

PR #266, opened 2017-03-21
Fix: avoid remove assigned lots on manual assignement

cubells:
When a user are assigned lots on a picking and recheck availability, lots are removed.

With this, lots are not removed, and button is shown in order to allow user to change or not lots assigned.

cc @Tecnativa


pedrobaeza (approved):
@jbeficent please review

rousseldenis (approved):
Code review

PR #267, opened 2017-03-23
[10.0][ADD] stock_available_product_expiry

rousseldenis:


rousseldenis:
As it breaks stock_available modules, I will add a config parameter to enable this feature or not.

rousseldenis:
Fix commits squashed. Please don't squash these.

lmignon (approved):
@rousseldenis Thank you for this useful addon (IMO)
:+1: (Code review /functional)

pedrobaeza (commented):
Shouldn't this be `stock_available_product_expiry` to follow the same terminology?

lmignon:
@pedrobaeza module renamed and commit squashed/fixed

pedrobaeza:
Setup files haven't been renamed

pedrobaeza:
Can you add also some tests for not dropping coverage?

lmignon:
I'm improving this PR. -> WIP

lmignon:
Works done. This PR is ready for reviews

lmignon:
PLZ don't squash commits on this PR

stock_available_product_expiry/models/stock_quant.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import fields, models
+
+
+class StockQuant(models.Model):
+
+    _inherit = 'stock.quant'
+
+    lot_id = fields.Many2one(
+        'stock.production.lot',
+        auto_join=True
```
lmignon: @rousseldenis IMO the auto_join works only with required fields. Can you check that quants without lot and reservation_id are still visible with `auto_join=True`?

rousseldenis: Ok, I'll check

rousseldenis: Indeed, not visible... :-(

I try to find another way to compute

stock_available_product_expiry/models/product_product.py
```diff
@@ -107,16 +106,35 @@ def _compute_quantities(self):
             product.qty_expired -= product.outgoing_expired_qty
         return res
 
+    def _get_quants(self, quant_domain, quant_out_domain):
```
lmignon, line 145: @rousseldenis You must be sure that the original value for `auto_join` on the field is restored even in case of exception....

```python
@api.model
@contextmanager
def _force_auto_join(self, model_field_list):
    """ Force  auto_join for each item (model_name, field_name) in model_field_list
    The original value are restored on exit
    """
    vals = {}
    for model_name, field_name in model_field_list:
        field = self.env[model_name]._fields[field_name]
        vals[field] = field.auto_join
        field.auto_join = True
    try:
        yield
    finally:
        for field, value in vals.iteritems():
            field.auto_join = value

def  _get_quants(self, quand_domain, quant_out_domain):
    stock_quant_obj = self.env['stock.quant']
    with self._force_auto_join([('stock.quant', 'lot_id')('stock.quant', 'reservation_id')]:
        quants = stock_quant_obj.with_context(lang='').read_group(
            quant_domain, ['product_id', 'qty'], ['product_id'])
        quants_out = stock_quant_obj.with_context(lang='').read_group(
            quant_out_domain, ['product_id', 'qty'], ['product_id'])
        return quants, quants_out
```

rousseldenis: Done

lmignon:
@rousseldenis Can you squash the two commits 'Imrove computation performance' PLZ

rousseldenis:
@lmignon Seems ready to be merged (or final review) ?

@pedrobaeza cc@eficent

rousseldenis:
@lreficent I think this one can be merged but missing review

LoisRForgeFlow:
I just rebuilt the runbot, it seems ok, but I'm not able to access :confused:

gurneyalex (approved):

stock_available_product_expiry/models/product_product.py
```diff
@@ -0,0 +1,179 @@
...
+                quant_res.get(product.id, 0.0),
+                precision_rounding=product.uom_id.rounding
+            )
+        return res
+
+    @api.multi
+    def action_open_expired_quants(self):
+        action = self.env.ref('stock.product_open_quants').read()[0]
+        domain = [
+            ('product_id', '=', self.id),
```
gurneyalex: maybe change to ('product_id', 'in', self.ids) and this can get called from a list view

gurneyalex (approved):

rousseldenis:
@gurneyalex Done

rousseldenis:
@Cedric-Pigeon Can you approve this as you tested latest changes ?

Cedric-Pigeon (approved):

PR #268, opened 2017-03-27
[FIX] stock_quant_manual_assign: computed field is not the same

agaldona:
I've got a stock move with 0.015 qty. When I launch the quant manual assign wizard, I get one quant with qty 16 to select. When I select this quant, the qty reserved changes to 0.015. There is a computed field which sums all reserved qty in quants in the wizard, its value is 0.015. When comparing the computed field with the move qty 0.015 > 0.015 it returns True, this is wrong, but If I compare the sum instead of the computed field it returns False.

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -5,6 +5,7 @@
 
 from openerp import api, exceptions, fields, models, _
 import openerp.addons.decimal_precision as dp
+from openerp.tools.float_utils import float_compare
 
 
 class AssignManualQuants(models.TransientModel):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not the solution. You should use float_compare instead with a reasonable delta (for example, 8).

elicoidal:
Actually it should be comparing with the decimal precision of the weight or UoM.

agaldona:
Do you mean something like this? @pedrobaeza @elicoidal

elicoidal (approved):

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -13,13 +14,14 @@ class AssignManualQuants(models.TransientModel):
...
-            if record.quants_lines:
-                move = self.env['stock.move'].browse(
-                    self.env.context['active_id'])
-                if record.lines_qty > move.product_uom_qty:
-                    raise exceptions.Warning(
-                        _('Quantity is higher than the needed one'))
+        precision_digits = dp.get_precision('Product Unit of Measure'
+                                            )(self.env.cr)[1]
+        for record in self.filtered(lambda x: x.quants_lines):
+            move = self.env['stock.move'].browse(self.env.context['active_id'])
```
pedrobaeza: Make the browse only once

agaldona:
Changes done.

pedrobaeza:
Merged and forward-ported to v9. Thanks.

PR #270, opened 2017-03-29
[9.0][stock_inventory_revaluation] Allow to delete quants

JordiBForgeFlow:
Now when an inventory revaluation has been posted for a product that has real price costing methods, if you revaluate some quants and then these quants need to be deleted afterwards, the user gets an error.

Deleting quants is done automatically by the system in some cases. For example, when you unreserve a delivery order and serveral quants are merged together.

### DIFF
--- a/stock_inventory_revaluation/tests/test_stock_inventory_revaluation.py
+++ b/stock_inventory_revaluation/tests/test_stock_inventory_revaluation.py
@@ -234,6 +234,12 @@ class TestStockInventoryRevaluation(TransactionCase):
                                  'Incorrect inventory revaluation for '
                                  'type Price Change.')
 
+        quants = self.env['stock.quant'].search([('product_id', '=',
+                                                 self.product_real_1.id)])
+
+        # We should be able to delete the quants afterwards
+        quants.with_context(force_unlink=True).unlink()
+
     def create_inventory_revaluation_price_change_average(self):
         revaluation_type = 'price_change'
         # Create an Inventory Revaluation for average cost product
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@lreficent @gmeficent can you review?

LoisRForgeFlow:
Travis says that the tests are failing: <code>stock_inventory_revaluation.tests.test_stock_inventory_revaluation: ERROR</code>

stock_inventory_revaluation/tests/test_stock_inventory_revaluation.py
```diff
@@ -234,6 +234,13 @@ def test_inventory_revaluation_price_change_real(self):
                                  'Incorrect inventory revaluation for '
                                  'type Price Change.')
 
+        quants = self.env['stock.quant'].search([('product_id', '=',
+                                                 self.product_real_1.id)])
+
+        # We should be able to delete the quants afterwards
+        for quant in quants:
+            quant.unlink()
```
gmeficent: Maybe using .sudo() would solve the issue.

image

gmeficent (changes requested):
I couldn't replicate the User Error on runbot and the added test fails due to constraints so I can't quite tell if it works.

stock_inventory_revaluation/tests/test_stock_inventory_revaluation.py
lmignon: @jbeficent @gmeficent It's not allowed by odoo to call unlink on a quant... https://github.com/odoo/odoo/blob/9.0/addons/stock/stock.py#L746
UNLESS you call unlink with `force_unlink=True` into the context 
```python
quants..with_context(force_unlink=True).unlink
```
You don't need to loop on the quants to call unlink...

lmignon (changes requested):

JordiBForgeFlow:
Thanks for your feedback @lmignon. Travis is green now.

gmeficent (approved):
Seems like that solves the issue, LGTM :+1:

LoisRForgeFlow (approved):
Code review LGTM :+1:

PR #255, opened 2017-02-28
[MIG][10.0] stock_putaway_product

rousseldenis:
Migration of stock_putaway_product

Issue #201, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [ ] base_product_merge
- [ ] business_product_location - By @rousseldenis - #446 
- [ ] configurable_stock_level
- [ ] partner_location_auto_create
- [x] stock_available - By @SodexisTeam - #219
- [x] stock_available_immediately - By @SodexisTeam - #219
- [x] stock_available_mrp - @Cedric-Pigeon #280
- [x] stock_available_sale - By @chienandalu - #334 
- [x] stock_available_unreserved - By @StefanRijnhart - #283
- [ ] stock_inventory_line_price - @dlasley #276
- [ ] stock_inventory_preparation_filter
- [ ] stock_location_area_data
- [ ] stock_location_area_management
- [ ] stock_location_lockdown - By @alexis-via  - #516 
- [ ] stock_location_ownership
- [ ] stock_lot_valuation
- [ ] stock_move_location
- [x] stock_mts_mto_rule - By @sylvain-garancher - #203 
- [ ] stock_optional_valuation
- [ ] stock_orderpoint_creator
- [ ] stock_partner_lot
- [x] stock_product_location_sorted_by_qty - By @chafique-delli - #444 
- [x] stock_putaway_product - By @rousseldenis - #255
- [x] stock_quant_manual_assign - By @fanha99 - #326
- [x] stock_quant_merge - By @lreficent - #265 
- [ ] stock_reord_rule
- [ ] stock_reserve - By @andrius-preimantas - #239 
- [ ] stock_reserve_sale

Garamotte:
Migration of stock_mts_mto_rule here : https://github.com/OCA/stock-logistics-warehouse/pull/203
(Including #157 which migrates from v8.0 to v9.0, but is not yet merged)

SodexisTeam:
We are planning to migrate stock_available and stock_available_immediately.

pedrobaeza:
OK, written in the main comment

atchuthan:
Created PR https://github.com/OCA/stock-logistics-warehouse/pull/219 for stock_available & stock_available_immediately.

kkarolis:
Planning to migrate stock_reserve & stock_reserve_sale (blocking #190)

Edit:
Added a longstanding PR

mrparadies:
[...]

PR #273, opened 2017-04-14
[MIG] Migrate stock_putaway_product to v10.0

nim-odoo:
This is a copy-paste of the 9.0 version. There was only a call to `putaway_apply` to correct to make it work.


nim-odoo:
Whooops, just saw it already exists in #255. Nevermind :)

pedrobaeza:
Hi, Nicolas, good to see you in the OCA side!

It's a pity that the module is already being ported. Next time, you can check also the issue we open on each repository for tracking the migration of the modules. In this case, #201, but as they are manually maintained, the best bet is to look at the pending PRs.

PR #275, opened 2017-04-22
[9.0][MIG] stock_inventory_line_price: v9 Migration

lasley:
* Split models by file
* Simplify logic in `stock.inventory.line._resolve_inventory_line`
* Remove out of spec manifest keys
* Bump version

### DIFF
--- a/stock_inventory_line_price/models/stock_inventory.py
+++ b/stock_inventory_line_price/models/stock_inventory.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Esther Martín - AvanzOSC
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, fields, models
+import openerp.addons.decimal_precision as dp
+
+
+class StockInventoryLine(models.Model):
+    _inherit = 'stock.inventory.line'
+
+    theoretical_std_price = fields.Float(
+        'Theorical Cost Price', digits=dp.get_precision('Product Price'),
+        readonly=True)
+    standard_price = fields.Float('Cost Price',
+                                  digits=dp.get_precision('Product Price'))
+
+    @api.model
+    def _resolve_inventory_line(self, inventory_line):
+        if inventory_line.theoretical_std_price !=\
+                inventory_line.standard_price:
+            inventory_line.product_id.standard_price = \
+                inventory_line.standard_price
+            move_id = super(StockInventoryLine,
+                            self)._resolve_inventory_line(inventory_line)
+            if move_id:
+                move = self.env['stock.move'].browse(move_id)
+                move.price_unit = inventory_line.standard_price
+                return move_id
+        return super(StockInventoryLine, self)._resolve_inventory_line(
+            inventory_line)
+
+
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    @api.model
+    def _get_inventory_lines(self, inventory):
+        vals = super(StockInventory, self)._get_inventory_lines(inventory)
+        for line in vals:
+            product = self.env['product.product'].browse(line['product_id'])
+            line.update({'theoretical_std_price': product.standard_price,
+                         'standard_price': product.standard_price})
+        return vals
--- a/stock_inventory_line_price/models/stock_inventory.py
+++ b/stock_inventory_line_price/models/stock_inventory.py
@@ -2,33 +2,7 @@
 # (c) 2016 Esther Martín - AvanzOSC
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import api, fields, models
-import openerp.addons.decimal_precision as dp
-
-
-class StockInventoryLine(models.Model):
-    _inherit = 'stock.inventory.line'
-
-    theoretical_std_price = fields.Float(
-        'Theorical Cost Price', digits=dp.get_precision('Product Price'),
-        readonly=True)
-    standard_price = fields.Float('Cost Price',
-                                  digits=dp.get_precision('Product Price'))
-
-    @api.model
-    def _resolve_inventory_line(self, inventory_line):
-        if inventory_line.theoretical_std_price !=\
-                inventory_line.standard_price:
-            inventory_line.product_id.standard_price = \
-                inventory_line.standard_price
-            move_id = super(StockInventoryLine,
-                            self)._resolve_inventory_line(inventory_line)
-            if move_id:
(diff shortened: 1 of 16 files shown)


LoisRForgeFlow:
I couldn't use the runbot so I rebuilt it, but I cannot log in yet (after the rebuild)

lasley:
@lreficent - looks like login works now 🚀

LoisRForgeFlow:
@lasley now I tried and again I can't. Why so many issues with the runbot latetly?

lasley:
@lreficent - Got me, looks like Runbot was restarted last night. I'm rebuilding this one again.

stock_inventory_line_price/views/inventory_view.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="name">view.inventory.form.inherit</field>
+        <field name="model">stock.inventory</field>
+        <field name="inherit_id" ref="stock.view_inventory_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='line_ids']//field[@name='product_id']" position="after">
+                <field name="theoretical_std_price"/>
+                <field name="standard_price"/>
+            </xpath>
+            <xpath expr="//tree" position="attributes">
+                <attribute name="colors">blue: product_qty != theoretical_qty or theoretical_std_price != standard_price; red: theoretical_qty &lt; 0</attribute>
```
LoisRForgeFlow, line 14: I have also modified the colors of the lines [here](https://github.com/OCA/stock-logistics-warehouse/blob/9.0/stock_inventory_discrepancy/views/stock_inventory_view.xml#L43). What would be the proper way to make all this "visual notifications" to coexist? Use icons in one of the cases?

LoisRForgeFlow (commented):
Finally tested! functionally works as expected. Just a comment inline

stock_inventory_line_price/views/inventory_view.xml
lasley: Wow I never considered the modularity concerns of changing line colors - great catch.

What if we conformed to some sort of interface mechanism with a field and a method? Something like:

```
class StockInventory(Model):
    _inherit = 'stock.inventory'

    attribute_colors = fields.Text(compute='_compute_attribute_colors')

    def _compute_attribute_colors(self):
        colors = '; '.join(["%s: %s" % (k, v) for k, v in colors.items()])
        for record in self:
            record.attribute_colors = colors

    @api.model
    def _get_attribute_colors(self):
        return {
            'blue': 'product_qty != theoretical_qty or theoretical_std_price != standard_price',
            'red': 'theoretical_qty &lt; 0',
        }
```

We'd need a base module though.

LoisRForgeFlow: I'm not sure, that way the only thing you do is move the problem to another place I think. Because then you and me would overwrite the `_get_attribute_colors()` separately instead of the attributes in the view, leading to the same problem: whoever overrides the last would prevail, right?

lasley: Damn yeah you're right - for some reason I was only considering different colors being used, but yeah we're actually conflicting at the color level.

What if we took the same proposal, but instead combined the colors from all parents. Something like:

``` python
class StockInventory(Model):
    _inherit = 'stock.inventory'

    attribute_colors = fields.Text(compute='_compute_attribute_colors')

    def _compute_attribute_colors(self):
        """ This would be in the base module """
        colors = defaultdict([])
        for color, logic in self._get_attribute_colors():
            colors[color].append('(%s)' % logic)
        colors = '; '.join(["%s: %s" % (k, ' or '.join(v)) for k, v in colors.items()])
        for record in self:
            record.attribute_colors = colors

    @api.model
    def _get_attribute_colors(self):
        """ This would be in our modules. The base would simply return `[]` """
        res = super(StockInventory, self)._get_attribute_colors()
        res += [
            ('blue', 'product_qty != theoretical_qty or theoretical_std_price != standard_price'),
            ('red', 'theoretical_qty &lt; 0'),
        ]
        return res
```

LoisRForgeFlow: Yes, that makes more sense to me. If no one see some problem with it, we can go like this.

Another point, that is more like a question is: I use the decoration classes of bootstrap because that was what I saw in v9 code to color lines, do you think that is good to use standard colors or try to keep with bootstrap classes?

lasley: I prefer bootstrap classes, because then they can be themed.

LoisRForgeFlow: So, if you want you can add this little module here in this PR and I'll open another one adapting to your changes

stock_inventory_line_price/models/stock_inventory.py
```diff
@@ -0,0 +1,18 @@
...
+from openerp import api, models
+
+
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    @api.model
+    def _get_inventory_lines(self, inventory):
+        vals = super(StockInventory, self)._get_inventory_lines(inventory)
+        for line in vals:
```
lmignon: @lasley IMO we should browse all the product at once to avoid performance issue...

stock_inventory_line_price/views/inventory_view.xml
lmignon: @lasley Do you plan to work on this PR to take into account the comment from @lreficent?

rousseldenis: @lasley ping

stock_inventory_line_price/models/stock_inventory.py
rousseldenis: @lasley Can you look into @lmignon remarks to move forward on this ?

lasley:
Sorry guys - I suck and missed this a bunch. My workload's pretty intense at the moment, so I probably won't have time to make the base module in https://github.com/OCA/stock-logistics-warehouse/pull/275#discussion_r115679274 for a while unfortunately. 

I did attend to @lmignon's performance comment though, and added the colorization conflict note to the readme

rousseldenis (approved):
Code review

LoisRForgeFlow (changes requested):
Still one comment unattended, oventhough Lasley is likely not going to do it, in order to merge it should be solved.

rousseldenis:
@lreficent So, closing ?

LoisRForgeFlow:
@rousseldenis I guess so, if someone is interested on continuing the work can open a new PR.

PR #278, opened 2017-05-08
[9.0][stock_cycle_count] Assign zones to cycle count rules

JordiBForgeFlow:
This PR improves the cycle count by allowing users to define zones (locations) to which the rules should be applied. If the user does not indicate a zone, the application will use all the locations of the warehouse that was selected.

cc @lreficent

### DIFF
--- a/stock_cycle_count/README.rst
+++ b/stock_cycle_count/README.rst
@@ -32,6 +32,15 @@ To install this module, you need to:
 * Download this module to your addons path.
 * Install the module in your database.
 
+Recommendations
+---------------
+
+It is highly recommended to use this module in conjunction with:
+
+* `stock_inventory_verification_request`: Adds the capability to request Slot
+  Verifications.
+* `stock_inventory_lockdown`: Lock down locations during inventories.
+
 Configuration
 =============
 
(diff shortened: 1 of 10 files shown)


lreficent:
Coverage is red because of a onchange method, I'm wondering whether it is really needed to test onchange method in tests?  :confused:

LoisRForgeFlow:
Coverage is red because of an onchange method, I'm wondering whether it is really needed to test onchange method in tests?  :confused:

lasley:
> I'm wondering whether it is really needed to test onchange method in tests? 

If it's logic, it should be tested.

LoisRForgeFlow:
Yes, I suppose I have no scape  :sweat_smile:

It's done now!

LoisRForgeFlow:
This PR also fix the following issues:

* Do not propose cycle counts in the past.
* ~~Able to handle several pre-existing and planned cycle counts for the same location.~~
* Cycle counts only editable in state draft.
* Track some fields in the mail thread.
* Enhance the cycle count tree view with filters and colors.
* The configuration of an inventory adjustment related to a cycle count cannot be modified.
* Not cancel but update pre-existing cycle counts.
* Add server action to confirm several CC at a time.

lasley:
@lreficent - seeing as there's still commits coming into this PR, I'm going to set as WIP. Let me know when ready for review and I'll switch the tag.

LoisRForgeFlow:
@lasley I totally forgot, this is ready to review. You can change the tag.

Thanks!

MiquelRForgeFlow (approved):
LGTM! :+1:

PR #249, opened 2017-02-14
[8.0]stock_available: improve computation performance

Cedric-Pigeon:

PR #279, opened 2017-05-09
[10.0]stock_available: improve computation performance

Cedric-Pigeon:
Forward port of https://github.com/OCA/stock-logistics-warehouse/pull/249

### DIFF
--- a/stock_available/models/product_template.py
+++ b/stock_available/models/product_template.py
@@ -24,15 +24,15 @@ class ProductTemplate(models.Model):
                 continue
             immediately_usable_qty = 0.0
 
+            potential_qty = 0
             for p in product.product_variant_ids:
                 qty = variant_available[p.id]["immediately_usable_qty"]
                 immediately_usable_qty += qty
-            if product.product_variant_ids:
-                potential_qty = max(
-                    [v.potential_qty for v in product.product_variant_ids])
+                if p.potential_qty > potential_qty:
+                    potential_qty = p.potential_qty
+
             res[product.id].update({
-                "immediately_usable_qty": immediately_usable_qty})
-            res[product.id].update({
+                "immediately_usable_qty": immediately_usable_qty,
                 "potential_qty": potential_qty})
 
         return res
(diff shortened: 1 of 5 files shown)


stock_available/models/product_template.py
```diff
@@ -10,6 +10,33 @@
...
+        for product in self:
+
+            if isinstance(product.id, models.NewId):
+                continue
+            immediately_usable_qty = 0.0
+
+            for p in product.product_variant_ids:
+                qty = variant_available[p.id]["immediately_usable_qty"]
+                immediately_usable_qty += qty
+            if product.product_variant_ids:
```
rousseldenis: @Cedric-Pigeon Can you maybe put this computation in the above loop?

stock_available/models/product_template.py
```diff
@@ -10,6 +10,33 @@
...
+                potential_qty = max(
+                    [v.potential_qty for v in product.product_variant_ids])
+            res[product.id].update({
```
rousseldenis: @Cedric-Pigeon One update ?

rousseldenis (changes requested):

Cedric-Pigeon:
@rousseldenis Remarks integrated, thanks.

rousseldenis (approved):
:+1:

lmignon (approved):
:+1: Thank you @Cedric-Pigeon

Cedric-Pigeon:
@pedrobaeza What about this one ?

lmignon:
@lasley can you take a look at this one. It's an important improvement. The pr is falsely red due to a decrease in the coverage since the fix implies more lines of code.

pedrobaeza:
Well, the patch seems legit, but unfortunately, there's not automated test to confirm that it works correctly at least one time. That's why we shouldn't look down on that check. But if the PSC decides to merge it, it doesn't go against OCA rules, as tests are not (yet) mandatory.

Cedric-Pigeon:
@pedrobaeza You are right. We just provided a test.

PR #281, opened 2017-05-10
[MIG] stock_move_partner_info v9

serpentcs-dev1:
## Stock Move Partner Info

- This module displays the partner of the picking in its stock movements and also allows grouping by a partner in stock movements.


darshan-serpent:
pending : readme and manifest file

serpentcs-dev1:
@aheficent Could you please check?

stock_move_partner_info/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+
+from openerp import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    picking_partner = fields.Many2one(
+        'res.partner', string='Picking Partner', store=True,
+        related='picking_id.partner_id', help='Partner of the picking')
```
rousseldenis: @darshan-serpent It is preferable to set related as readonly=True

rousseldenis (changes requested):

stock_move_partner_info/models/stock_move.py
serpentcs-dev1: Right, of course, I missed that, thanks!

stock_move_partner_info/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Serv. Tec. Avanzados - Pedro M. Baeza (http://www.serviciosbaeza.com)
```
serpentcs-dev1: @pedrobaeza How should I put this to fit it within 79 characters? Could you please suggest?

pedrobaeza: Please switch this to my new company "Tecnativa". You can put:
```
Copyright 2015 Tecnativa - Pedro M. Baeza
```

LoisRForgeFlow (changes requested):
You have a Invalid view definition, check runbot and travis logs please

LoisRForgeFlow (changes requested):
The groupby is not working too...
image

AaronHForgeFlow (changes requested):
I think the problem in the group_by is because the field is related. I think if we add store=True to the field it'd work.

LoisRForgeFlow (approved):
This seems fine now! :+1:

AaronHForgeFlow (approved):
LGTM :+1:

StefanRijnhart:
Because of the way this module was migrated from Odoomrp, the attribution is lost from the commits. @pedrobaeza do you have an opinion on this?

pedrobaeza:
Indeed commit attribution should be respected. Please follow the general migration guide (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0), but replacing the branch by the odoomrp-wip one.

serpentcs-dev1:
#285 This new PR preserves the commit attribution.

PR #282, opened 2017-05-12
[9.0] [MIG] stock_inventory_lockdown

MiquelRForgeFlow:
Standard migration.


StefanRijnhart (approved):
Thanks! Code reviewed and tested on runbot.

LoisRForgeFlow (approved):
Functional and technical review :+1: 

Just to minor (not blocking) things:
* It would be good to update the images of the description, they are v8.
* Squash transifex commits.

Cedric-Pigeon (approved):

stock_inventory_lockdown/models/stock_inventory.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, api
+
+
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    @api.model
+    def _get_locations_open_inventories(self):
```
lmignon: @mreficent IMO if would be more efficient to allow to pass a location_ids parameter to this method and t use this paramaterd  (if provided) into the domain for the search on inventories...
ex:
```python
 @api.model
 def _get_locations_open_inventories(self, locations_ids = None):
    inventory_domain = [('state', '=', 'confirm')]
    if locations_ids:
        inventory_domain.append(('location_id', 'child_of', location_ids))
    inventories = self.search(inventory_domain)
    location_ids = inventories.mapped('location_id')
    location_domain = [
        ('location_id', 'child_of', location_ids.ids),
        ('usage', 'in', ['internal', 'transit'])]
     if locations_ids:
        location_domain.append(('location_id', 'child_of', location_ids))
    return self.env['stock.location'].search(location_domain)
```
In this way we will filter the result to the locations for which we want to check if an inventory exist.

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,46 @@
...
+
+class StockLocation(models.Model):
+    """Refuse changes during exhaustive Inventories"""
+    _inherit = 'stock.location'
+    _order = 'name'
+
+    @api.multi
+    def _check_inventory(self):
+        """Error if an inventory is being conducted here"""
+        location_inventory_open_ids = self.env['stock.inventory'].sudo(
```
lmignon: @mreficent With the change proposed for `_get_locations_open_inventories` the method will be simplified
```python
location_inventory_open_ids = self.env['stock.inventory'].sudo()._get_locations_open_inventories(self.ids)
if location_inventory_open_ids:
    raise ValidationError(...)
```

lmignon (changes requested):
@mreficent IMO we can improve the logic to check for locked locations.... and improve the performances. see my comments. What do you think about?

MiquelRForgeFlow:
@lmignon suggestion attended to 😊

stock_inventory_lockdown/models/stock_quant.py
```diff
@@ -0,0 +1,58 @@
...
+        """Check that the location is not locked by an open inventory.
+        Check both the location as it was (source) and the location as
+        it will be (destination).
+        We verify the locations even if they are unchanged, because changing
+        ie. the quantity is not acceptable either.
+        @raise ValidationError if they are.
+        """
+        if not self.env.context.get('bypass_lockdown', False):
+            # Find the locked locations
+            locked_location_ids = self.env[
```
lmignon: @mreficent This code should be also modified to take advantage of the new paramater on `_get_locations_open_inventories`

```python
if 'location_id' in vals.keys():
    locked_location_ids = self.env['stock.inventory']._get_locations_open_inventories(
        self.env['stock.location'].browse(vals['location_id']) + self.mapped('location_id')
    )
   if locked_location_ids:
        raise ...
```

stock_inventory_lockdown/models/stock_quant.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @api.model
+    def create(self, vals):
+        """Check that the locations are not locked by an open inventory.
+        @raise ValidationError if they are.
+        """
+        quant = super(StockQuant, self).create(vals)
+        if not self.env.context.get('bypass_lockdown', False):
+            locked_location_ids = self.env[
+                'stock.inventory']._get_locations_open_inventories()
```
lmignon: @mreficent same comment as above

lmignon (changes requested):

stock_inventory_lockdown/models/stock_quant.py
MiquelRForgeFlow: But then the raise message will not have explicit locations 😕

lmignon: @mreficent 
```python
location_names = locked_location_ids.mapped('name')
raise ValidationError(
    _('An inventory is being conducted at the following '
      'location(s):\n%s') % "\n - ".join(location_names))
```

lmignon (approved):

lmignon:
Lgtm  code review only

JordiBForgeFlow (approved):
:+1: Functional review

PR #284, opened 2017-05-15
[10.0][ADD] Module stock_operation_package_mandatory

rousseldenis:
This adds a module to make destination package mandatory for every
product operation in a picking


stock_operation_package_mandatory/models/stock_picking.py
```diff
@@ -0,0 +1,35 @@
...
+        for picking in self:
+            if picking.picking_type_id.destination_package_mandatory:
+                operations = picking.pack_operation_product_ids.filtered(
+                    lambda o: float_compare(
+                        o.qty_done,
+                        0,
+                        o.product_id.uom_id.rounding) > 0)
+                if not all([operation.result_package_id
+                            for operation in operations]):
+                    raise UserError(
```
lmignon, line 27: @rousseldenis How can we provide to the user the informations required to solve the problem? It could be hard to find the operation where the package is missing...

rousseldenis: @lmignon As this function is called in the do_new_transfer one (which is applied on the picking), it's quite difficult to extend error message with product names list for example (as we can have several products in that case).
If you have an idea...

lmignon: no idea...

lmignon (approved):
LGT%

stock_operation_package_mandatory/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'name': 'Stock Operation Package Mandatory',
+    'summary': """
+        Makes destination package mandatory on stock pack operations""",
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu',
+    'depends': ['product',
+                'stock'
+                ],
```
Cedric-Pigeon: Fix wrong indentation

Cedric-Pigeon (approved):
Small remark, otherwise LGTM

rousseldenis:
@Cedric-Pigeon done!

Issue #286, opened 2017-05-16
[RFC] Prioritize locations during removal strategy


JordiBForgeFlow:
We found an issue during the warehouse removal operations where quants could be fetched from a location that is less desired than others, for quants that were received in the same date. We want to establish a general priority on locations to better resolve this kind of issues.

Example
------------
We purchase 20 widgets, and after receiving them, we place 10 in shelf A, and 10 in shelf B (because shelf A has limited space). Shelf B is way less reachable than shelf A. Self A and B are under Stock location.

When a delivery order is created, and 5 units need to be fetched from  Stock, Odoo will apply the FIFO rule, and may fetch the 5 items from Shelf A, or Shelf B with the same probability.

In those cases we want to make sure that Shelf B has less priority than A, so that Odoo fetches first all inventory from Shelf A, before reaching out to B. At the same time respecting the FIFO rule.

Proposed approach
---------------------------
* Add field 'Removal priority' in stock.location
* Add the above field as related, stored field in stock.quant
* sort by date_in,removal_priority in a method that inherits from https://github.com/odoo/odoo/blob/9.0/addons/stock/stock.py#L714

JordiBForgeFlow:
cc @mreficent can you look into this one?

JordiBForgeFlow:
Implemented in https://github.com/OCA/stock-logistics-warehouse/tree/9.0/stock_removal_location_by_priority

PR #287, opened 2017-05-19
[9.0][ADD] stock_removal_location_by_priority

MiquelRForgeFlow:
Stock Removal Location by Priority
==================================

This module adds a removal priority field on stock locations.
This priority applies when removing a product from different stock locations
and the incoming dates are equal in both locations.

Solves https://github.com/OCA/stock-logistics-warehouse/issues/286.


stock_removal_location_by_priority/README.rst
```diff
@@ -0,0 +1,60 @@
...
+==================================
+
+This module adds a removal priority field on stock locations.
+This priority applies when removing a product from different stock locations
+and the incoming dates are equal in both locations.
+
+Configuration
+=============
+
+You can configure the rules to compute the cycle count, acting as follow:
```
LoisRForgeFlow: I think this line does not belong here :wink:

stock_removal_location_by_priority/models/stock_quant.py
```diff
@@ -0,0 +1,34 @@
...
+        if any(move.mapped('location_id.removal_priority')):
+            if removal_strategy == 'fifo':
+                order = 'in_date, removal_priority, id'
+                return self._quants_get_order(
+                    quantity, move, ops=ops, domain=domain, orderby=order)
+            elif removal_strategy == 'lifo':
+                order = 'in_date desc, removal_priority desc, id desc'
+                return self._quants_get_order(
+                    quantity, move, ops=ops, domain=domain, orderby=order)
+            raise UserError(_('Removal strategy %s not implemented.') % (
```
LoisRForgeFlow: Are you sure that to raise an error is the good call here? Maybe just logging a warning message in the server logs could do the job and do not block new possible removal strategies.

stock_removal_location_by_priority/models/stock_quant.py
```diff
@@ -0,0 +1,34 @@
...
+    @api.model
+    def apply_removal_strategy(self, quantity, move, ops=False,
+                               domain=None, removal_strategy='fifo'):
+        if any(move.mapped('location_id.removal_priority')):
+            if removal_strategy == 'fifo':
+                order = 'in_date, removal_priority, id'
+                return self._quants_get_order(
+                    quantity, move, ops=ops, domain=domain, orderby=order)
+            elif removal_strategy == 'lifo':
+                order = 'in_date desc, removal_priority desc, id desc'
```
LoisRForgeFlow: I think that removal_priority should be sorted crescent here too. No matter what removal strategy you use when the removal_priority is the one that would decide we want it to always pick the same way.

LoisRForgeFlow (commented):
Some comments in line

stock_removal_location_by_priority/models/stock_quant.py
MiquelRForgeFlow: The base apply_removal_strategy method has this UserError. Perhaps, other possible strategies that may exist they also are included via Super like in this case.

LoisRForgeFlow: Yes, you are right, better to go along with standard

LoisRForgeFlow (approved):

lmignon (approved):
@mreficent Thank you for this new module: LGTM (functional + code review)

Cedric-Pigeon (approved):

PR #292, opened 2017-05-30
[9.0][ddmrp] Demand Driven Material Requirements Planning

JordiBForgeFlow:
DDMRP
=====

The Demand Driven Material Requirements Planning (DDMRP) is a multi-echelon
formal planning and execution technique with five distinct components:

* *Strategic inventory positioning*. Answers to the question: 'Given our system
  and environment, where should we place inventory to have the best protection?'
* *Buffer profiles and levels*. Allows to define the strategically positioned
  buffers (reordering rules in Odoo) to achieve the optimum levels of inventory.
* *Dynamic adjustments*. Buffers (reordering rules in Odoo) adapt dynamically
 to a changing environment conditions, such as demand, lead time or
 variability in supply.
* *Demand-driven planning*. Represents an intuitive method of supply order
  generation, based on the information provided by the buffers.
* *Highly visible and collaborative execution*. Simply launching purchase
  orders (POs), manufacturing orders (MOs) and transfer orders (TOs) from any
  planning system does not end the materials and order management challenge.
  These POs, MOs and TOs have to be effectively managed to synchronize with
  the changes that often occur within the "execution horizon."

These five components work together to greatly dampen, if not eliminate,
the nervousness of traditional MRP systems and the bullwhip effect in complex
and challenging environments.


JordiBForgeFlow:
cc @lreficent @BenjaHe @christdum @espo-tony Can you review?

ddmrp/models/mrp_production.py
```diff
@@ -0,0 +1,93 @@
...
+
+class MrpProduction(models.Model):
+    _inherit = 'mrp.production'
+
+    @api.model
+    def _search_procurements(self):
+        return [('production_id', '=', self.id)]
+
+    @api.model
+    def _search_orderpoints(self):
```
espo-tony, line 19: This method looks like never been used. Is it here for allowing further implementation or could be removed?

ddmrp/models/mrp_production.py
```diff
@@ -0,0 +1,93 @@
...
+        return [('name', '=', self.move_prod_id.origin)]
+
+    @api.model
+    def _find_orderpoint_from_procurement(self, procurement):
+        orderpoint = procurement.move_dest_id.procurement_id.orderpoint_id
```
espo-tony, line 24: As for the ORM definition, the fields _move_dest_id_ on _procurement.order_ and _procurement_id_ on _stock.move_ are not required, so they could be blank here. If that happens (is there maybe some constraint on a parent method, or anywhere else that I didn't remark?), the method will raise an exception here.

ddmrp/models/mrp_production.py
```diff
@@ -0,0 +1,93 @@
...
+    def _compute_orderpoint_id(self):
+        for rec in self:
+            domain = rec._search_procurements()
+            procurements = rec.env['procurement.order'].search(domain)
+            orderpoints = [procurement.orderpoint_id for procurement in
+                           procurements if procurement.orderpoint_id]
+            if orderpoints:
+                rec.orderpoint_id = orderpoints[0]
+            else:
+                for procurement in procurements:
```
espo-tony, line 39: Is the structure of these nested loops correct? Should this method change rec.orderpoint_id as many times as length of procurements?

ddmrp/models/mrp_production.py
```diff
@@ -0,0 +1,93 @@
...
+                        if not originating_procurement:
+                            break
+
+    @api.multi
+    @api.depends("orderpoint_id")
+    def _compute_execution_priority(self):
+        for rec in self:
+            if rec.state not in ['done', 'cancel']:
+                rec.execution_priority_level = \
+                    rec.orderpoint_id.execution_priority_level
```
espo-tony, line 57: As for the ORM definition, _rec.orderpoint_id_ is not required, so it could not exist here. Is it possible it's changed to a blank value? In that case this method would raise an exception here.

ddmrp/models/procurement_order.py
```diff
@@ -0,0 +1,37 @@
...
+    add_to_net_flow_equation = fields.Boolean(
+        string='Pending add net flow equation', default=True,
+        readonly=True,
+        help="If this flag is set, the procurement is pending to be included "
+             "into the net flow equation.")
+
+    @api.multi
+    def write(self, vals):
+        res = super(ProcurementOrder, self).write(vals)
+        if 'state' in vals.keys() \
```
espo-tony, line 22: This condition could maybe be written as 

`if vals.get('state', False) not in ['draft', 'cancel'] and not vals.get('add_to_net_flow_equation', False):`

It covers the case (I don't know how likely) that 'add_to_net_flow_equation' is explicitly passed as False in the method arguments) and, IMHO is also more readable.

ddmrp/models/product_adu_calculation_method.py
```diff
@@ -0,0 +1,43 @@
...
+
+    company_id = fields.Many2one(
+        'res.company', string='Company', required=True,
+        default=lambda self:
+        self.env['res.company']._company_default_get(
+            'product.adu.calculation.method'))
+
+    @api.multi
+    @api.constrains('method', 'horizon')
+    def _check_horizon(self):
```
espo-tony, line 39: Wouldn't be better to replace this constraint (or at least to enforce it) by making  horizon required on the view whenever method is 'past' or 'future'?

Something like (in the XML definition, of course):
`<field name="horizon" class="oe_inline" attrs="{'required':[('method','in',['past','future'])]}"/>`

ddmrp/views/product_adu_calculation_method_view.xml
```diff
@@ -0,0 +1,67 @@
...
+                <form string="ADU Calculation Method">
+                    <group name="name">
+                        <field name="name"/>
+                    </group>
+                    <group name="method">
+                        <field name="method"/>
+                        <field name="use_estimates"/>
+                        <label for="horizon"/>
+                        <div name="horizon">
+                            <field name="horizon" class="oe_inline"/> days
```
espo-tony, line 29: As remarked before, I suggest to make this field required when _horizon_ is _past_ or _future_:
`
<field name="horizon" class="oe_inline" attrs="{'required':[('method','in',['past','future'])]}"/> days`

espo-tony (commented):
Some questions and suggestions.

ovnicraft:
@jbeficent do you plan continue this work ? 
I see all about DDMRP and your video and its a great job ! 

if you consider need help for this or migrate to v10.0 please count with me as a resource at half time.

LoisRForgeFlow:
@ovnicraft The project has its own repo now and it is on v11, please review https://github.com/OCA/ddmrp/pull/16 :smiley:

MiquelRForgeFlow:
@lreficent then, shouldn't this PR be closed?

LoisRForgeFlow:
@mreficent Yes, you are right.

Closing since the project is now at [OCA/ddmrp](https://github.com/OCA/ddmrp)

PR #299, opened 2017-06-20
[10.0][MIG] stock_inventory_lockdown

MiquelRForgeFlow:
Standard migration.

- [x] Added https://github.com/OCA/stock-logistics-warehouse/pull/304.


LoisRForgeFlow (commented):
I just tested on runbot and realized of one issue, which is that the lockdown is not taking into account the children of the location where an inventory adjustment is being conducted. For instance:

1. Start an inventory Adjustment (IA) in Stock.
2. Try to deliver some apple headphones (which are stored in Stock/Shelf1).
3. No warning raises and the picking is completed.

At the moment this module works perfect if you are using `stock_inventory_exclude_sublocation` and excluding sublocations in IA. But not with standard odoo.

LoisRForgeFlow:
Pending to solve this: https://github.com/OCA/stock-logistics-warehouse/pull/304

MiquelRForgeFlow:
https://github.com/OCA/stock-logistics-warehouse/pull/304 was merged, so this PR is ready to be reviewed and merged.

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,45 @@
...
+        # If changing the parent, no inventory must conducted there either
+        if vals.get('location_id'):
+            locations_to_check |= self.browse(vals['location_id'])
+        locations_to_check._check_inventory()
+        return super(StockLocation, self).write(vals)
+
+    @api.model
+    def create(self, vals):
+        """Refuse create if an inventory is being conducted at the parent"""
+        if 'location_id' in vals:
```
serpentcs-dev1: This is giving error when you try to create a new stock location and when there is an 'In Progress' Inventory, because here the if the condition gets satisfied even when `location_id` is empty.

This should be like `if 'location_id' in vals and vals['location_id']:` or simply `if vals.get('location_id'):` like you have done in write method.

stock_inventory_lockdown/models/stock_inventory.py
```diff
@@ -0,0 +1,31 @@
...
+            # Early exit if no match found
+            return []
+        location_ids = inventories.mapped('location_id')
+
+        # Extend to the children Locations
+        location_domain = [
+            ('location_id', 'child_of', location_ids.ids),
+            ('usage', 'in', ['internal', 'transit'])]
+        if locations_ids:
+            location_domain.append(('location_id', 'child_of', locations_ids))
```
serpentcs-dev1, line 30: The domain  `('location_id', 'child_of', locations_ids)` is already added above in the list!!

serpentcs-dev1 (changes requested):

stock_inventory_lockdown/README.rst
```diff
@@ -0,0 +1,60 @@
...
+======================================
+Lock down locations during inventories
+======================================
+This module lets you lock down the locations during an inventory.
+
+Usage
+=====
+
+.. image:: stock_inventory_lockdown/static/images/move_error.png
+   :alt: Error message
```
nikul-serpentcs, line 14: @mreficent This link does not work, Can you please take a look?

nikul-serpentcs (changes requested):

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,45 @@
...
+        """Error if an inventory is being conducted here"""
+        location_inventory_open_ids = self.env['stock.inventory'].sudo().\
+            _get_locations_open_inventories(self.ids)
+        if location_inventory_open_ids:
+            raise ValidationError(
+                _('An inventory is being conducted at this '
+                  'location'))
+
+    @api.multi
+    def write(self, vals):
```
rousseldenis: @mreficent I know this is migration, but shouldn't be checked through constrains ? Moreover that avoid write and create overrides.

stock_inventory_lockdown/models/stock_inventory.py
MiquelRForgeFlow: <code>location_ids</code> != <code>locations_ids</code>

stock_inventory_lockdown/README.rst
MiquelRForgeFlow: There is no way to solve this. We can put it in a way that it works in github or in a way it works in Odoo, but not in both.

rousseldenis (changes requested):

stock_inventory_lockdown/models/stock_location.py
MiquelRForgeFlow: @rousseldenis  Will it work also for the unlink method?

rousseldenis: Of course not, but you can call the _check_inventory in a constrains method and call the _check in unlink specifically

MiquelRForgeFlow:
Please, could everyone update their reviews?

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,32 @@
...
+        location_inventory_open_ids = self.env['stock.inventory'].sudo().\
+            _get_locations_open_inventories(self.mapped('location_id').ids)
+        if location_inventory_open_ids:
+            raise ValidationError(
+                _('An inventory is being conducted at this location'))
+
+    @api.multi
+    def unlink(self):
+        """Refuse unlink if an inventory is being conducted"""
+        location_inventory_open_ids = self.env['stock.inventory'].sudo(). \
```
LoisRForgeFlow: This is not working as expected, maybe due to the space after `sudo()`, please remove it.

LoisRForgeFlow (changes requested):

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo.exceptions import ValidationError
+
+
+class StockLocation(models.Model):
+    """Refuse changes during exhaustive Inventories"""
+    _inherit = 'stock.location'
+    _order = 'name'
+
+    @api.multi
+    @api.constrains('location_id')
```
LoisRForgeFlow, line 15: This is also not working as it worked before, raises on creation but not on write...

LoisRForgeFlow (changes requested):

stock_inventory_lockdown/models/stock_location.py
rousseldenis: Weird as the trigger is location_id.

The code is different from the initial migration commit.

Here you don't take into account self in recordset to check.

MiquelRForgeFlow: But if we change the `_check_inventory_location_id` to check the self, then the create won't work correctly :S

rousseldenis (approved):

rousseldenis:
@mreficent Can you add test on the unlink to get good coverage ?

LoisRForgeFlow (approved):
This is working fine now! :+1:

LoisRForgeFlow:
@nikul-serpentcs @darshan-serpent could you update your reviews please?

stock_inventory_lockdown/models/stock_inventory.py
serpentcs-dev1: @mreficent I am talking about the [line 27](https://github.com/OCA/stock-logistics-warehouse/pull/299/files/4f391ba257942d07fb874f32b16d2483483bc339#diff-3817be29f13c1c2ff56a0a821a025338R27) the child locations are already appended to the domain list.

serpentcs-dev1 (approved):
Code Review LGTM :+1:

nikul-serpentcs (approved):
Functional Testing, Works Fine!!! :+1:

stock_inventory_lockdown/models/stock_inventory.py
MiquelRForgeFlow: @darshan-serpent

in line 18: you have `locations_ids`, added in inventory_domain
in line 27: you have `location_ids.ids`
in line 30: you have `locations_ids`, added in location_domain

The three things are different.

serpentcs-dev1: Oh Yes right, now I noticed, thanks! :smile:

PR #302, opened 2017-06-21
[10.0][MIG] stock_removal_location_by_priority

MiquelRForgeFlow:
In v10, the removal method is different, so in this migration I went with the approach of creating in settings an option to activate this module or not.


AaronHForgeFlow (commented):
Hi @mreficent Can you please mention this new approach in the readme?

stock_removal_location_by_priority/README.rst
```diff
@@ -0,0 +1,69 @@
...
+This priority applies when removing a product from different stock locations
+and the incoming dates are equal in both locations.
+
+Installation
+============
+
+To install this module, you need to do as follows:
+
+#. Go to "Inventory > Configuration > Settings"
+#. In 'Location & Warehouse' section, mark "Use 'Removal Priority' in Locations *"
```
LoisRForgeFlow: This explanation belong also to the configuration section, this option appears once the module actually installed.

stock_removal_location_by_priority/models/stock_config_settings.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class StockConfigSettings(models.TransientModel):
+    _inherit = 'stock.config.settings'
+
+    removal_priority_active = fields.Boolean(
+        related='company_id.removal_priority_active',
+        string="Use 'Removal Priority' in Locations *")
```
LoisRForgeFlow: remove last `*`

LoisRForgeFlow (commented):
Some comments inline

stock_removal_location_by_priority/README.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Configuration
+=============
+
+You can configure the removal priority as follows:
+
+#. Go to "Inventory > Configuration > Warehouse Management > Locations"
+#. In each Location form, put a Removal Priority.
+
+
```
LoisRForgeFlow: I would add also the 'usage' section how this behaves and how do you have to configure this priority. 
**tip:** You can try to adapt a little the original issue https://github.com/OCA/stock-logistics-warehouse/issues/286

stock_removal_location_by_priority/models/res_company.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    removal_priority_active = fields.Boolean(
+        default=False,
```
rousseldenis: That's default

rousseldenis (changes requested):
Code review. Minor comment.

@mreficent Can you add test for 'lifo' removal strategy ?

stock_removal_location_by_priority/models/res_company.py
MiquelRForgeFlow: What happens with the default?

LoisRForgeFlow: For booleans default is false, you don't need to specify it.

gustavovalverde:
At the actual state, Is this PR safe enough for production?

MiquelRForgeFlow:
I think so.

May @rousseldenis  update his review :smirk:

stock_removal_location_by_priority/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,17 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record id="view_stock_config_settings" model="ir.ui.view">
+        <field name="name">stock.config.view - removal_priority</field>
+        <field name="model">stock.config.settings</field>
+        <field name="inherit_id" ref="stock.view_stock_config_settings"/>
+        <field name="arch" type="xml">
+            <field name="group_warning_stock" position="after">
+                <field name="removal_priority_active"/>
```
LoisRForgeFlow: Every time I check this box and apply, it goes back to False. Please have a look.

LoisRForgeFlow (changes requested):
Tested on runbot.

LoisRForgeFlow:
We have realized that the company setting approach is not the best, especially if working with multi-company so we plan to move to implied-group setting.

stock_removal_location_by_priority/views/stock_location_view.xml
```diff
@@ -10,7 +10,7 @@
         <field name="inherit_id" ref="stock.view_location_form"/>
         <field name="arch" type="xml">
             <field name="active" position="after">
-                <field name="removal_priority"/>
+                <field name="removal_priority" group="removal_priority_active"/>
```
LoisRForgeFlow: @gustavovalverde if you are in a hurry and you don't use multi-company, by removing in this view `group="removal_priority_active"` and checking the box *Use 'Removal Priority' in Locations* in the setting it will work ok for you.

gustavovalverde: I'm indeed moving to multicompany , what could I expect in a multicompany environment?

LoisRForgeFlow (approved):
Code review + tested on runbot :+1:

Thanks!

LoisRForgeFlow:
@rousseldenis @aheficent @gustavovalverde Could you update your review?

gurneyalex:
This feels like a wrong approach to me. Why are you not creating a new product.removal strategy to implement this?

LoisRForgeFlow:
@gurneyalex We want to respect each product removal strategy (FIFO, LIFO...) while just setting some priority for locations (which is BTW the second priority after date, so FIFO and LIFO will always apply first). [code](https://github.com/Eficent/stock-logistics-warehouse/blob/9302972ccacb75b9a5b852576f926fbc22f0e8eb/stock_removal_location_by_priority/models/stock_quant.py#L22)

See related issue for more clarification: https://github.com/OCA/stock-logistics-warehouse/issues/286

rousseldenis:
I agree with @lreficent . @gurneyalex FIFO, LIFO, FEFO are strategy concepts based on dates. Here, it is a different concept.

Creating new strategies should have implied creation of three more strategies (each existing one + removal priority).

rousseldenis (approved):
Code review

PR #305, opened 2017-06-27
[10.0] stock_available modules: optimize stock computation by avoiding to call useless compute

Cedric-Pigeon:
In V8.0 we used _product_available to feed a dict with data.
The problem of this method is that you compute a lot of thing several times because for each stock computed values Odoo computes the other ones.
It doesn't have big impact on small database but on biggest one it could really slow.

This PR intends to avoid these useless computations and try to benefit of the prefetches values as much as possible.

On a big database, I reduced computation time by 6.


congdpt (approved):

stock_available/models/product_product.py
```diff
@@ -16,54 +16,35 @@ class ProductProduct(models.Model):
...
-            field_names=field_names, arg=arg)
-        for prod_id in res:
-            res[prod_id]['immediately_usable_qty'] = res[prod_id][
-                'virtual_available']
-            res[prod_id]['potential_qty'] = 0.0
-
+    def _compute_available_quantities_dict(self):
+        res = {}
+        for product in self:
+            res[product.id] = {}
```
lasley: Why the individual assignments? Just create the dict with the vals:

```
res[product.id] = {
    'immediately_usable_qty': ...
```

stock_available/models/product_product.py
```diff
@@ -16,54 +16,35 @@ class ProductProduct(models.Model):
...
-        other modules to inherit it
-        """
-        res = self._product_available()
-        for prod in self:
-            prod.immediately_usable_qty = res[prod.id][
-                'potential_qty']
+    def _compute_available_quantities(self):
+        res = self._compute_available_quantities_dict()
+        for product in self:
+            data = res[product.id]
```
lasley: Why the intermediary assignment instead of just looping `res[product.id]` directly?

stock_available/models/product_product.py
```diff
@@ -16,54 +16,35 @@ class ProductProduct(models.Model):
...
+            for key, value in data.iteritems():
+                if hasattr(product, key):
+                    product[key] = value
 
     immediately_usable_qty = fields.Float(
```
lasley: Please move field declarations to top - https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

stock_available/models/product_template.py
```diff
@@ -10,75 +10,45 @@
...
-            variants += product.product_variant_ids
-        variant_available = variants._product_available()
-
+    @api.multi
+    @api.depends('product_variant_ids.immediately_usable_qty',
+                 'product_variant_ids.potential_qty')
+    def _compute_available_quantities(self):
+        res = self._compute_available_quantities_dict()
         for product in self:
+            data = res[product.id]
```
lasley: Same question regarding intermediary assignment instead of direct loop

stock_available/models/product_product.py
```diff
@@ -16,54 +16,35 @@ class ProductProduct(models.Model):
...
-    @api.depends()
-    def _compute_potential_qty(self):
-        """Set potential qty to 0.0 to define the field defintion used by
-        other modules to inherit it
-        """
-        res = self._product_available()
-        for prod in self:
-            prod.immediately_usable_qty = res[prod.id][
-                'potential_qty']
+    def _compute_available_quantities(self):
```
lasley, line 30: It looks like this is duplicated in `product.template` why not just escalate to the template method and just pass in the `product.product` instead? Something like:

```
self.env['product.template']. _compute_available_quantities(self)
```

Admittedly I haven't tried it in this context, but I'm almost positive I've done something similar elsewhere without triggering an apples to oranges error.

stock_available/models/product_template.py
```diff
@@ -10,75 +10,45 @@
...
-        We can't add the potential of variants: if they share components we
-        may not be able to make all the variants.
-        So we set the arbitrary rule that we can promise up to the biggest
-        variant's potential.
-        """
-        res = self._product_available()
-        for tmpl in self:
-            tmpl.potential_qty = res[tmpl.id]['potential_qty']
-
     immediately_usable_qty = fields.Float(
```
lasley, line 41: Move to top

stock_available_immediately/models/product_product.py
```diff
@@ -3,18 +3,20 @@
...
+    @api.multi
+    def _compute_available_quantities_dict(self):
+        res = super(ProductProduct, self)._compute_available_quantities_dict()
+        for product in self:
+            res[product.id]['immediately_usable_qty'] -= \
+                product.incoming_qty
         return res
+
+    @api.depends('virtual_available', 'incoming_qty')
+    def _compute_available_quantities(self):
```
lasley, line 21: What is the purpose of this method declaration? Are the depends changing or something?

Maybe add a docblock indicating the reasoning? Otherwise it's just a method calling the super.

lasley (changes requested):
Thanks @Cedric-Pigeon - comments inline

rousseldenis:
@Cedric-Pigeon Can you take into account remarks from @lasley ?

rousseldenis:
@Cedric-Pigeon Any news on this? Relaunched runbot and all green

rousseldenis:
@lreficent @Cedric-Pigeon @lmignon I rebased this as there is new modules that use the stock_available process.

Keep in mind that this is a change in the manner we compute available quantities. We construct a dictionary with all values and then use that dictionary to build computed ones.

That avoid to have too much confidence in how orm works for many chained computed values (but this work is based on how Odoo did the job in stock module).

MiquelRForgeFlow (approved):
LGTM :+1:

lmignon (approved):
LGTM

JordiBForgeFlow (approved):

PR #308, opened 2017-06-29
[ADD] stock_orderpoint_automatic_creation: migration to 10.0

cubells:
This module generates stock orderpoints automatically for a product when is created taking the company of the user who created the product as reference.
Orderpoints creation can be automate from product, product category or company.

cc @ Tecnativa


stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,61 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================================
+Stock Orderpoint Automatic Creation
+===================================
+
+This module generates stock orderpoints automatically for a product when is
+created taking the company of the user who created the product as reference.
+Orderpoints creation can be automate from product, product category or company.
```
pedrobaeza: s/automate/automated

stock_orderpoint_automatic_creation/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "author": "AvanzOSC, "
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "http://www.odoomrp.com",
+    "depends": [
+        "procurement",
+        "stock",
+    ],
+    "category": "Inventory, Logistic, Storage",
```
pedrobaeza, line 18: Is this category correct?

stock_orderpoint_automatic_creation/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "data": [
+        "views/res_company_views.xml",
+        "views/product_views.xml",
+    ],
+    "installable": True,
+    "application": True
```
pedrobaeza: This is not an application

stock_orderpoint_automatic_creation/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Daniel Campos <danielcampos@avanzosc.es> - Avanzosc S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import product
```
pedrobaeza: product_product

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
+    def create(self, vals):
+        result = super(ProductProduct, self).create(vals)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
+        wh_obj = self.env['stock.warehouse']
+        create = False
+        if not result.type == 'product' or result.create_orderpoint == 'no':
+            return result
```
pedrobaeza: If you make here return, you don't need this if chain

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
+    def create(self, vals):
+        result = super(ProductProduct, self).create(vals)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
```
pedrobaeza: Don't declare this until it's cleared that you need to create the orderpoint

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+        if not result.type == 'product' or result.create_orderpoint == 'no':
+            return result
+        elif result.create_orderpoint == 'yes':
+            create = True
+        elif result.categ_id.create_orderpoints == 'no':
+            return result
+        elif (result.categ_id.create_orderpoints == 'yes' or
+                self.env.user.company_id.create_orderpoints):
+            create = True
+        if create:
```
pedrobaeza: Make that if you reach this point, you don't need the `if`

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+            company = self.env.user.company_id
+            warehouses = wh_obj.search([('company_id', '=', company.id)])
+            for warehouse in warehouses:
+                orderpoint_obj.create(
+                    {'name': result.name,
```
pedrobaeza: Put this in a hook method `_prepare_orderpoint_vals` for allowing extensions

stock_orderpoint_automatic_creation/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+                     'product_id': result.id,
+                     'product_max_qty': company.orderpoint_product_max_qty,
+                     'product_min_qty': company.orderpoint_product_min_qty,
+                     'warehouse_id': warehouse.id,
+                     'location_id': warehouse.lot_stock_id.id
+                     })
+        return result
+
+
+class ProductCategory(models.Model):
```
pedrobaeza: This should be in a separated file

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import models, fields, api, exceptions, _
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    orderpoint_product_max_qty = fields.Integer(
+        'Max. product quantity',
+        help='Default orderpoint Max. product quantity', default=0)
```
pedrobaeza: Default is already 0, so no need to add it.

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,25 @@
...
+    orderpoint_product_min_qty = fields.Integer(
+        'Min. product quantity',
+        help='Default orderpoint Max. product quantity', default=0)
+    create_orderpoints = fields.Boolean(
+        'Create Orderpoints', help='Check this for automatic orderpoints')
+
+    @api.constrains('orderpoint_product_max_qty', 'orderpoint_product_min_qty')
```
pedrobaeza: This is not @api.multi compliant

pedrobaeza (changes requested):
Add your copyright in files

stock_orderpoint_automatic_creation/__manifest__.py
cubells: Yes, It is.

stock_orderpoint_automatic_creation/models/__init__.py
cubells: Are you sure?

In oca module, It's defined product.py and product_template.py:

https://github.com/OCA/OCB/tree/10.0/addons/product/models

pedrobaeza: The rule is the model name, and the model name is product.product, not only product

pedrobaeza (approved):

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, exceptions, fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    orderpoint_product_max_qty = fields.Integer(
```
rousseldenis: @cubells Why not using Product Decimal Precision here (as defined in stock.warehouse.orderpoint)?

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+        string='Max. product quantity',
+        help='Default orderpoint Max. product quantity',
+    )
+    orderpoint_product_min_qty = fields.Integer(
```
rousseldenis: Same as above ?

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+    create_orderpoints = fields.Boolean(
+        string='Create Orderpoints',
+        help='Check this for automatic orderpoints',
+    )
+
+    @api.multi
+    @api.constrains('orderpoint_product_max_qty', 'orderpoint_product_min_qty')
+    def _check_orderpoint_product_qty(self):
+        for orderpoint in self:
+            if orderpoint.orderpoint_product_max_qty < 0 \
```
rousseldenis: @cubells If you take into account decimal precision, I'd prefer using float_compare

rousseldenis (changes requested):

cubells:
@rousseldenis done!

stock_orderpoint_automatic_creation/models/res_company.py
```diff
@@ -0,0 +1,38 @@
...
+    create_orderpoints = fields.Boolean(
+        string='Create Orderpoints',
+        help='Check this for automatic orderpoints',
+    )
+
+    @api.multi
+    @api.constrains('orderpoint_product_max_qty', 'orderpoint_product_min_qty')
+    def _check_orderpoint_product_qty(self):
+        for company in self:
+            if float_compare(company.orderpoint_product_max_qty, 0.0,
```
rousseldenis: @cubells You cannot set an arbitrary rounding (2) here but you have to take `product_id.uom_id.rounding` !

rousseldenis (changes requested):

stock_orderpoint_automatic_creation/models/res_company.py
cubells: And from where I take the product rounding if I'm in a company @rousseldenis ?

rousseldenis: Ok, sorry for that, you can take the Product decimal precision instead I think

cubells:
@rousseldenis done!

rousseldenis (approved):
:+1: Code review

stock_orderpoint_automatic_creation/models/product_category.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    create_orderpoints = fields.Selection(
```
elicoidal, line 12: A strange choice: why not a simple checkbox? Usability is better: one click instead of 2
Any reason for this choice?

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================================
+Stock Orderpoint Automatic Creation
+===================================
+
+This module generates stock orderpoints automatically for a product when is
```
elicoidal: This module  automatically generates stock orderpoints for a product when it is
created, based on the information from the user's company.

stock_orderpoint_automatic_creation/README.rst
```diff
@@ -0,0 +1,61 @@
...
+created taking the company of the user who created the product as reference.
+Orderpoints creation can be automated from product, product category or company.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to company form and, in the "Settings" tab, you can customize the default
```
elicoidal: Go to company form, tab "settings" and customize the default orderpoints data

elicoidal (commented):
some details and a question on usability but overall LGTM

stock_orderpoint_automatic_creation/models/product_category.py
cubells: Original author decide that when user select "No" option in product or categories, then orderpoint is not created even if company settings are configured to create them.

if user select not to create orderpoint in product:
https://github.com/OCA/stock-logistics-warehouse/pull/308/files#diff-0743ce7eb081903f7439f511d86f3439R28
or in category:
https://github.com/OCA/stock-logistics-warehouse/pull/308/files#diff-0743ce7eb081903f7439f511d86f3439R30
then orderpoint is not created.

This allows user to avoid create orderpoint always for all products and categories if company are configured to do it.

pedrobaeza:
Thanks for the merge, Cédric. Please take into account for other merges to see if some commits can be squashed as this case, because they are result of the review process.

Issue #288, opened 2017-05-22
[RFC] Add Chatter to stock.inventory


JordiBForgeFlow:
We have identified that it is usual for people to log some insights around inventory adjustments. We also identify that changes key attributes of the inventory adjustment should be logged. E.g. when the status of the adjustment changes.

We plan to introduce a module 'stock_inventory_chatter', that just adds this capability.

JordiBForgeFlow:
cc @mreficent @lreficent

JordiBForgeFlow:
@mreficent can you work on this one?

rousseldenis:
@jbeficent I think this is solved.

PR #309, opened 2017-06-30
[9.0][ADD] stock_inventory_chatter

MiquelRForgeFlow:
Request of this module: https://github.com/OCA/stock-logistics-warehouse/issues/288

Stock Inventory Chatter
=======================

This module adds the capability to log the changes being done in Inventory Adjustments.


andhit-r (approved):

elicoidal (approved):
LGTM

stock_inventory_chatter/models/stock.py
```diff
@@ -0,0 +1,23 @@
...
+
+from openerp import api, fields, models
+
+
+class StockInventory(models.Model):
+    _name = 'stock.inventory'
+    _inherit = ['stock.inventory', 'mail.thread', 'ir.needaction_mixin']
+
+    partner_id = fields.Many2one(track_visibility='always')
+    state = fields.Selection(track_visibility='onchange')
```
LoisRForgeFlow, line 14: I would also track `location_id` and `filter`

stock_inventory_chatter/models/stock.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @api.multi
+    def _track_subtype(self, init_values):
+        self.ensure_one()
+        if 'state' in init_values and self.state == 'confirm':
+            return 'stock_inventory_chatter.mt_inventory_confirmed'
+        elif 'state' in init_values and self.state == 'done':
+            return 'stock_inventory_chatter.mt_inventory_done'
```
LoisRForgeFlow: This msgs seems that are not being logged.

LoisRForgeFlow (commented):
Some little comments

AaronHForgeFlow (changes requested):
messages mt_inventory_confirmed and mt_inventory_done not being logged in runbot

stock_inventory_chatter/models/stock.py
MiquelRForgeFlow: I think this msgs are sent through mail to the followers of the thread.

stock_inventory_chatter/models/stock.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class StockInventory(models.Model):
+    _name = 'stock.inventory'
+    _inherit = ['stock.inventory', 'mail.thread', 'ir.needaction_mixin']
```
AaronHForgeFlow, line 11: is it really needed to inherit form ir.needaction_mixin?

stock_inventory_chatter/models/stock.py
AaronHForgeFlow: In that case I'd also track the cancelled state @mreficent

stock_inventory_chatter/models/stock.py
MiquelRForgeFlow: It seems so.

MiquelRForgeFlow:
Comments attended.

LoisRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: Thanks

PR #310, opened 2017-07-04
[9.0][IMP] stock_inventory_exclude_sublocaton: track visibility and only editable in draft state

LoisRForgeFlow:
Adds a dependency with `stock_inventory_chatter` to track the `exclude_sublocation` field in the chatter. Also taking the occasion to fix the readonly states.


stock_inventory_exclude_sublocation/__openerp__.py
```diff
@@ -6,12 +6,12 @@
...
     "summary": "Allow to perform inventories of a location without including "
                "its child locations.",
-    "version": "9.0.1.0.0",
+    "version": "9.0.1.0.1",
     "author": "Eficent,"
               "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-warehouse",
     "category": "Warehouse Management",
-    "depends": ["stock"],
+    "depends": ["stock", "stock_inventory_chatter"],
```
aheficent: Why do you remove this?

AaronHForgeFlow (approved):
:+1:

MiquelRForgeFlow (approved):
LGTM :+1:

mreficent:
Added minor fixes from other two modules just to avoid to make two other PRs.

AaronHForgeFlow (approved):
:+1: cool

LoisRForgeFlow:
:+1:

MiquelRForgeFlow:
Added some minor fixes for others modules to solve some warnings. They are minor fixes, so I preferred to group here them than just make several PRs.

MiquelRForgeFlow:
@lreficent , you should edit the title as your initial commit has been changed. You can't change the dependencies of an already merged module. So, we should add this dependency when migrating to v10

LoisRForgeFlow:
@mreficent let's wait then to v10 to do the fix. The main purpose of this PR is lost then, do you want me to close this one so you can open a new PR for the other changes you have proposed?

PR #307, opened 2017-06-29
[9.0] Add stock_reorder_forecast module with tests

gfcapalbo:
Allows to predict date stock levels will reach minimum by
analizying sales volume in a period and therefore to trigger RFQ's ahead
of time",

reopening of #222

PR #311, opened 2017-07-04
Backport of stock_reorder_forecast from version 9.0 to version 8.0

daramousk:
See https://github.com/OCA/stock-logistics-warehouse/pull/307


NL66278:
@daramousk Please look at Travis errors. Both Pylint and test give errors. And wondering about the message from the CLAbot, als obviously @gfcapalbo has signed the cla.

daramousk:
@NL66278 The tests work when the `purchase` module is not installed and fail when it is installed. The purchase module is installed by the forecast module as a dependency.

So as I understand there is a bug in the tests in the `stock_change_qty_reason` module and they do not work properly when purchase is installed so this module has nothing to do with them failing.

PR #313, opened 2017-07-10
[8.0] stock reorder limit

NL66278:
Do not create procurements processing minimum stock rules, when this makes no sense. For instance for obsolete products, or products that have set the purchase_ok flag to False.


stock_reorder_limit/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sensible limits on minimum stock rule processing",
```
bealdav: HI @NL66278 thanks for your good contribution.
Just an idea. I don't want block the merge but I imagine that technical module name and name key content are so different than it could be confusing. Don't you think ? 
Note your name key content here fit perfectly for summary content which is also searchable in apps.odoo.com

NL66278: @bealdav I changed the name and summary according to your suggestion.

gurneyalex:
looks nice but the tests are red...

rousseldenis:
@NL66278 Do you plan to fix tests to allow this to be merged ?

rousseldenis:
@NL66278

rousseldenis:
Closing this as no activity. Feel free to reopen it if needed.

PR #314, opened 2017-07-10
[9.0][MIG] stock_account_change_product_valuation

MiquelRForgeFlow:
Standard migration.


stock_account_change_product_valuation/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+        if values.get('type', False):
+            new_type = values.get('type')
+            if new_type == 'product':
+                for rec in self:
+                    if rec.type == 'consu':
+                        for variant in rec.product_variant_ids:
+                            quants = self.env['stock.quant'].search(
+                                [('product_id', '=', variant.id)])
+                            for quant in quants:
+                                quant.cost = 0.0
```
LoisRForgeFlow: We can improve this: 
`quants = self.env['stock.quant'].search([('product_id', '=', variant.id)]).write('cost': 0.0)`

stock_account_change_product_valuation/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+                        rec.standard_price = 0.0
+        if values.get('cost_method', False):
+            new_method = values.get('cost_method')
+            if new_method == 'real':
+                for rec in self:
+                    type = values.get('type', False) or rec.type
+                    if type == 'product':
```
LoisRForgeFlow: 'type' is a python keyword, it would be good to avoid to use it as variable.

stock_account_change_product_valuation/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+                        for variant in rec.product_variant_ids:
+                            quants = self.env['stock.quant'].search(
+                                [('product_id', '=', variant.id),
+                                 ('location_id.usage', '=', 'internal')])
+                            quants.write(
+                                {'cost': rec.standard_price})
```
lreficent: same as above

LoisRForgeFlow (approved):
Some minor comments - nothing blocking

Thanks for the work!

espo-tony (approved):
LGTM

stock_account_change_product_valuation/models/product.py
```diff
@@ -0,0 +1,57 @@
...
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    @api.multi
+    def write(self, values):
+        if values.get('type', False):
+            new_type = values.get('type')
+            if new_type == 'product':
+                for rec in self:
```
rousseldenis: @mreficent I would have done this : `for variant in self.filtered(lambda r: r.type == 'consu').mapped('product_variant_ids'):`

That avoid a supplementary loop. Idem for next ones.

Then, I would have modified all quants of variant_ids at once, to have just one write().

So, 
`
variants = self.filtered(lambda r: r.type == 'consu').mapped('product_variant_ids')
quants = self.search([('product_id', 'in', variants.ids)])
quants.write({'cost': 0.0'})
variants.mapped('product_tmpl_id').standard_price = 0.0
`
I'm not sure about the last line. The point is making quite loops in a write is bulky. But that's my point of view ;-)

rousseldenis (changes requested):

rousseldenis (approved):
Code review :+1:

PR #325, opened 2017-08-03
[FIX] stock_mts_mto_rule: _update_routes now handle multiple records

zaoral:
This method is called from the write() method the one who support multiple records since is a api.multi method.

I was making some test when I tried to activate the mto+mts option for multiple warehouse and the next error appears: ValueError: Expected singleton

In order to fix this error I only added a loop to manage the multiple registers.


moylop260:
cc @sylvain-garancher @florian-dacosta

florian-dacosta (approved):
It seems good to me, thanks for the fix!

Garamotte (approved):
LGTM :)

LoisRForgeFlow (approved):
:+1:

PR #329, opened 2017-08-21
[9.0][FIX] stock_inventory_revaluation: able to cancel

LoisRForgeFlow:
Currently no one able to cancel revaluations and the following error raises:

  Sorry, you are not allowed to modify this document. Please  contact your system administrator if you think 
  this is an error.
  (Document model: stock.quant)

### DIFF
--- a/stock_inventory_revaluation/models/stock_inventory_revaluation.py
+++ b/stock_inventory_revaluation/models/stock_inventory_revaluation.py
@@ -404,7 +404,8 @@ class StockInventoryRevaluation(models.Model):
     def button_cancel(self):
         for revaluation in self:
             for reval_quant in revaluation.reval_quant_ids:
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
Also, this is blocking when using perpetual inventory valuation: [account_move.py#L24](https://github.com/OCA/stock-logistics-warehouse/blob/9.0/stock_inventory_revaluation/models/account_move.py#L24)

I'm not sure whether it is good to work around this when canceling from the inventory revaluation itself.

cc @jbeficent

lreficent:
superseded by https://github.com/OCA/stock-logistics-warehouse/pull/329

LoisRForgeFlow:
Superseded by https://github.com/OCA/stock-logistics-warehouse/pull/333

PR #330, opened 2017-08-23
[10.0][MIG] stock_inventory_discrepancy

LoisRForgeFlow:
Migration from v9.0 to v10.0:
* Adapt methods.
* Bump version
* Rename manifest and replace openerp by odoo.
* Remove copyrights on init files.
* Squash transifex commits into just one.
* Replace api.one with api.multi.

PR #331, opened 2017-08-23
[10.0][MIG] stock_inventory_verification_request

LoisRForgeFlow:
Migration from v9 to v10:
* Replace openerp with odoo.
* Fix renamed references.
* Use fa icon
* Replace api.one with api.multi.
* Bump version and rename manifest


LoisRForgeFlow:
* [x] : Depends on https://github.com/OCA/stock-logistics-warehouse/pull/330

stock_inventory_verification_request/models/stock_inventory.py
```diff
@@ -0,0 +1,57 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+
+
+class StockInventory(models.Model):
+    _inherit = 'stock.inventory'
+
+    requested_verification = fields.Boolean(
+        string='Requested Verification?', default=False, copy=False)
```
Garamotte: `default=False` is [explicitely ignored on Boolean fields](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L2289-L2290).

stock_inventory_verification_request/models/stock_inventory.py
```diff
@@ -0,0 +1,57 @@
...
+    slot_verification_ids = fields.One2many(
+        comodel_name='stock.slot.verification.request',
+        string='Slot Verification Requests', inverse_name='inventory_id')
+
+    @api.multi
+    def action_request_verification(self):
+        self.requested_verification = True
```
Garamotte: I think you should add `self.ensure_one()` as first line on this method.

stock_inventory_verification_request/models/stock_inventory.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @api.multi
+    def action_open_svr(self):
+        '''
+        Open the corresponding Slot Verification Request directly from the
+        Inventory Lines.
+        '''
+        request_svr_ids = []
+        for line in self:
+            request_svr_ids += line.slot_verification_ids.ids
```
Garamotte: `request_svr_ids = self.mapped('slot_verification_ids').ids`

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+from odoo import api, fields, models
+
+
+class SlotVerificationRequest(models.Model):
+    _name = 'stock.slot.verification.request'
+    _inherit = 'mail.thread'
+
+    @api.model
+    def create(self, vals):
+        vals['name'] = self.env['ir.sequence'].next_by_code(
```
Garamotte: Maybe we should add a check to avoid forcing the name if someone calls the `create` method with a name already defined ?

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+            'stock.slot.verification.request') or ''
+        return super(SlotVerificationRequest, self).create(vals)
+
+    @api.multi
+    def _count_involved_moves(self):
+        for rec in self:
+            rec.involved_move_count = len(self.involved_move_ids)
```
Garamotte: You forgot to replace `self` by `rec` in the `len` method.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @api.multi
+    def _count_involved_inv_lines(self):
+        for rec in self:
+            rec.involved_inv_line_count = len(self.involved_inv_line_ids)
```
Garamotte: You forgot to replace `self` by `rec` in the `len` method.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+        ('wait', 'Waiting Actions'),
+        ('open', 'In Progress'),
+        ('cancelled', 'Cancelled'),
+        ('done', 'Solved')
+    ], string='Status', default='wait')
+    responsible_id = fields.Many2one(comodel_name='res.users',
+                                     string='Assigned to')
+    product_id = fields.Many2one(comodel_name='product.product',
+                                 string='Product', required=True)
+    notes = fields.Text('Notes')
```
Garamotte: Explicitely name the `string` attribute.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+    involved_move_count = fields.Integer(compute=_count_involved_moves)
+    involved_inv_line_ids = fields.Many2many(
+        comodel_name='stock.inventory.line',
+        relation='slot_verification_inv_line_involved_rel',
+        column1='slot_verification_request_id',
+        column2='inventory_line_id',
+        string='Involved Inventory Lines')
+    involved_inv_line_count = fields.Integer(compute=_count_involved_inv_lines)
+
+    @api.model
```
Garamotte: Should this method be `@api.multi`, as many `self.product_id` are used in it ?

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+    def _get_involved_moves_domain(self):
+        domain = [('product_id', '=', self.product_id.id), '|',
+                  ('location_id', '=', self.location_id.id),
+                  ('location_dest_id', '=', self.location_id.id)]
+        return domain
+
+    @api.model
```
Garamotte: `@api.multi`, like the previous method.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+    product_id = fields.Many2one(comodel_name='product.product',
+                                 string='Product', required=True)
+    notes = fields.Text('Notes')
+    involved_move_ids = fields.Many2many(
+        comodel_name='stock.move',
+        relation='slot_verification_move_involved_rel',
+        column1='slot_verification_request_id',
+        column2='move_id',
+        string='Involved Stock Moves')
+    involved_move_count = fields.Integer(compute=_count_involved_moves)
```
Garamotte: The computed field method name should be a string, not the method itself.
This allows overriding the method without the need to override the field too.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+    involved_inv_line_ids = fields.Many2many(
+        comodel_name='stock.inventory.line',
+        relation='slot_verification_inv_line_involved_rel',
+        column1='slot_verification_request_id',
+        column2='inventory_line_id',
+        string='Involved Inventory Lines')
+    involved_inv_line_count = fields.Integer(compute=_count_involved_inv_lines)
```
Garamotte: The computed field method name should be a string, not the method itself.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+        return domain
+
+    @api.model
+    def _get_involved_lines_domain(self):
+        domain = [('product_id', '=', self.product_id.id),
+                  ('location_id', '=', self.location_id.id)]
+        return domain
+
+    @api.model
+    def _get_involved_lines_and_locations(self):
```
Garamotte: If you change the two previous methods to `@api.multi`, then you should change this one too.

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+    def action_solved(self):
+        self.write({'state': 'done'})
+        return True
+
+    @api.multi
+    def action_view_moves(self):
+        action = self.env.ref('stock.stock_move_action')
+        result = action.read()[0]
+        result['context'] = {}
+        moves_ids = sum([svr.involved_move_ids.ids for svr in self], [])
```
Garamotte: `moves_ids = self.mapped('involved_move_ids').ids`

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+        if len(moves_ids) > 1:
+            result['domain'] = \
+                "[('id','in',[" + ','.join(map(str, moves_ids)) + "])]"
```
Garamotte: `result['domain'] = [('id', 'in', move_ids)]`

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+            result['res_id'] = moves_ids and moves_ids[0] or False
+        return result
+
+    @api.multi
+    def action_view_inv_lines(self):
+        action = self.env.ref(
+            'stock_inventory_verification_request.action_inv_adj_line_tree')
+        result = action.read()[0]
+        result['context'] = {}
+        line_ids = sum([svr.involved_inv_line_ids.ids for svr in self], [])
```
Garamotte: `line_ids = self.mapped('involved_inv_line_ids').ids`

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -0,0 +1,136 @@
...
+        if len(line_ids) > 1:
+            result['domain'] = \
+                "[('id','in',[" + ','.join(map(str, line_ids)) + "])]"
```
Garamotte: `result['domain'] = [('id', 'in', line_ids)]`

Garamotte (changes requested):

stock_inventory_verification_request/models/stock_slot_verification_request.py
LoisRForgeFlow: Oh, didn't know this. Just to make it clearer: if I define it with `compute=_count_involved_moves`and I inherit the method `_count_involved_moves` in other module I would have to override the field as well, but if I do the definition like this `compute='_count_involved_moves'` not so?

stock_inventory_verification_request/models/stock_slot_verification_request.py
LoisRForgeFlow: It also works as a recordset with the api.model but you are right in the sense that api.multi makes it clearer.

LoisRForgeFlow:
@sylvain-garancher Thanks for the review! All your comments are attended, just asked a little clarification in one of them.

stock_inventory_verification_request/models/stock_slot_verification_request.py
Garamotte: This is it.
When using `compute=_method_name`, Python links the method itself (the python object), and you'll need to override the field to set a new method if you want to override the method.
When using `compute='_method_name'`, Python sets the method name, and Odoo calls that method on the model (including automatic call of the overrides), instead of directly the method of your file.

LoisRForgeFlow:
@sylvain-garancher dependencies are met. Could you please update your review?

Garamotte (approved):

MiquelRForgeFlow (approved):

lreficent:
Rebased

LoisRForgeFlow:
Rebased and squashed commits

PR #346, opened 2017-10-10
[9.0][IMP] stock_cycle_count: Add views and reports

LoisRForgeFlow:
Improve tools to analyse how accuracy evolve in locations.


stock_cycle_count/models/stock_inventory.py
```diff
@@ -12,23 +12,27 @@
...
+        for inv in self:
+            theoretical = sum(inv.line_ids.mapped(
+                lambda x: abs(x.theoretical_qty)))
+            abs_discrepancy = sum(inv.line_ids.mapped(
+                lambda x: abs(x.discrepancy_qty)))
+            if theoretical:
+                inv.inventory_accuracy = max(
+                    PERCENT * (theoretical - abs_discrepancy) / theoretical,
+                    0.0)
+            if not inv.line_ids and inv.state == 'done':
```
rousseldenis, line 27: You can get these records out of the loop (as you based your computation on inv_lines : substract self recordset with records with no inv_lines)

stock_cycle_count/reports/report_stock_location_accuracy.py
```diff
@@ -0,0 +1,38 @@
...
+from openerp import api, fields, models
+
+
+class LocationAccuracyReport(models.AbstractModel):
+    _name = "report.stock_location_accuracy"
+
+    @api.model
+    def _get_inventory_domain(self, loc_id):
+        return [('location_id', '=', loc_id),
+                ('exclude_sublocation', '=', True),
```
rousseldenis: Is this shouldn't be optional through parameter ?

rousseldenis (changes requested):
Code review. Minor comment.

Good improvement!

Same remarks about api on v10 (to make travis green)

stock_cycle_count/models/stock_inventory.py
LoisRForgeFlow: if I filter the recordset as you propose the inventories with no lines and in state done will end up with an accuracy of 0.0 (the default), while here I'm setting them to 100.0. Am I missing something?

stock_cycle_count/reports/report_stock_location_accuracy.py
LoisRForgeFlow: yes, why not? adding now

stock_cycle_count/models/stock_inventory.py
rousseldenis: No, I want to say you should compute this in batch before the loop and substract these recordset to enter the loop

LoisRForgeFlow: Ok, I got you. Fixed in https://github.com/OCA/stock-logistics-warehouse/pull/346/commits/d1832362eddedebd13813202aa4709f768991611

rousseldenis:
@lreficent Can you check travis ?

stock_cycle_count/models/stock_inventory.py
```diff
@@ -15,7 +15,10 @@ class StockInventory(models.Model):
     @api.multi
     @api.depends("state", "line_ids")
     def _compute_inventory_accuracy(self):
-        for inv in self:
+        zero_inventories = self.filtered(
+            lambda i: not i.line_ids and i.state == 'done')
+        zero_inventories.inventory_accuracy = PERCENT
```
LoisRForgeFlow: @rousseldenis This line of course give singleton error, but the obivous solution (`zero_inventories.write({'inventory_accuracy': PERCENT})`) is creating an infinite loop, I really don't understand why. Any idea?

stock_cycle_count/models/stock_inventory.py
```diff
@@ -17,7 +17,8 @@ class StockInventory(models.Model):
     def _compute_inventory_accuracy(self):
         zero_inventories = self.filtered(
             lambda i: not i.line_ids and i.state == 'done')
-        zero_inventories.inventory_accuracy = PERCENT
+        for inv in zero_inventories:
+            inv.inventory_accuracy = PERCENT
```
LoisRForgeFlow: And the only way to fix it is this commit (https://github.com/OCA/stock-logistics-warehouse/pull/346/commits/2b4103365fa4c31e041fda3f8ee387c1aa967e97) which I think leave us in a position very similar to the starting one and I hardly see any performance improvement. What do you think?

LoisRForgeFlow:
@rousseldenis See my two comments above.

stock_cycle_count/models/stock_inventory.py
rousseldenis: No, in fact. I didn't realized.

Indeed you cannot write on compute. Wrong! You have to change recordset fields one by one(the save is done at the end).

The improvement is only valid outside computes. Sorry.

LoisRForgeFlow: I didn't get a notification of this answer. No worries, it is always good to learn :wink:

LoisRForgeFlow: @rousseldenis reverted to initial state, could you please review again?

rousseldenis (approved):

aheficent:
:heart_eyes:

MiquelRForgeFlow (approved):

PR #347, opened 2017-10-13
[10.0][IMP] stock_putaway_product: Boost get_product_putaway_strategies

tfossoulw:
This fix boost the execution to the def get_product_putaway_strategies.

You can see result when system try to generate a backorder for an incoming shipment:

Create an incoming shipment with 1000 products with all differents fixed_location_id.
If you process only one line and you validate the picking you'll see the system will need about 2 min and half to get all the product fixed_location_ids.

### DIFF
--- a/stock_putaway_product/README.rst
+++ b/stock_putaway_product/README.rst
@@ -72,6 +72,7 @@ Contributors
 * Jos De Graeve - Apertoso N.V. <Jos.DeGraeve@apertoso.be>
 * Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
 * Denis Roussel - ACSONE SA/NV <denis.roussel@acsone.eu>
+* Thomas Fossoul - WINK SA/NV <tfossoul@wink.be>
 
 
 Maintainer
--- a/stock_putaway_product/__manifest__.py
+++ b/stock_putaway_product/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Putaway strategy per product',
     'summary': 'Set a product location and put-away strategy per product',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Inventory',
     'website': 'http://www.apertoso.be',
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Please change module version number

tfossoulw:
Done :)

pedrobaeza:
@sergio-teruel please check

rousseldenis (approved):
Code review. Great!

JordiBForgeFlow (approved):
:+1: Technical review

LoisRForgeFlow (approved):
LGTM, what is the speed boost with this change? edit: I mean in your usecase.

tfossoulw:
@lreficent its 2min30 faster for the case: generate a backorder of a 1000 products incoming shipment.

Calculate each time this takes times:
```
return self.product_location_ids.filtered(lambda x: (
            x.product_product_id == product or
            (not x.product_product_id and
             x.product_tmpl_id == product.product_tmpl_id)))
```
But this simply read data in memory for each product loaded in the environment which is much more fast:
```
        strategy = product.product_putaway_ids.filtered(
            lambda strategy: (strategy.putaway_id == self))
        if not strategy:
            strategy = product.product_tmpl_id.product_putaway_ids.filtered(
                lambda strategy: (strategy.putaway_id == self))
        return strategy
```

rousseldenis:
@tfossoulw You have to say there is a product putaway strategy for each product so for each move line we had to filter on product_location_ids which can contain thousands of records.

pedrobaeza:
You can put a comment in the code as reminder for future refactorings

rousseldenis:
@pedrobaeza It can be difficult to explain some parts of code that do not exist anymore. It's just a better way to do it.

pedrobaeza:
@rousseldenis yeah, sometimes it's difficult to say: "hey, don't do this again", but with latest commit I think that now is very correct.

tfossoulw:
@pedrobaeza Done

pedrobaeza (approved):
Merging when CIs finish

PR #349, opened 2017-10-30
[10.0][FIX] purchase_packaging: fix price_subtotal recomputation

tfossoulw:
purchase_line.price_subtotal is not recalculated when you adapt quantity.

**To reproduce:**
- Create a request for quotation
- Add some line without setting quantity.
- Save
- Edit
- Set quantity. You'll see the price subtotal will be recomputed.
- Save
The price subtotal will back to the previous amount.

### DIFF
--- a/purchase_packaging/models/purchase_order_line.py
+++ b/purchase_packaging/models/purchase_order_line.py
@@ -47,6 +47,13 @@ class PurchaseOrderLine(models.Model):
             fields.Date.from_string(self.order_id.date_order),
             uom_id=self.product_uom)
 
+    @api.depends('product_purchase_qty',
+                 'product_qty',
+                 'price_unit',
+                 'taxes_id')
+    def _compute_amount(self):
+        return super(PurchaseOrderLine, self)._compute_amount()
+
     @api.multi
     @api.depends('product_purchase_uom_id', 'product_purchase_qty')
     def _compute_product_qty(self):
--- a/purchase_packaging/models/purchase_order_line.py
+++ b/purchase_packaging/models/purchase_order_line.py
@@ -14,7 +14,8 @@ class PurchaseOrderLine(models.Model):
 
     product_tmpl_id = fields.Many2one(
         related='product_id.product_tmpl_id',
-        comodel_name='product.template'
+        comodel_name='product.template',
+        readonly=True
     )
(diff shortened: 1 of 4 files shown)


purchase_packaging/tests/test_purchase_order_line.py
```diff
@@ -70,6 +70,15 @@ def test_po_line(self):
                          self.product_uom_8.id)
         self.assertAlmostEqual(po_line.product_purchase_qty, 2)
         self.assertAlmostEqual(po_line.product_qty, 16)
+
+        # test subtotal computation
```
lmignon, line 79: @tfo Can you add a line into the test description to describe what's expected.

lmignon (changes requested):
@tfossoulw Thank you for the proposed fix. Nevertheless the problem should be resolved in an other way. We should only put into the  'depends'  list, the fields used by the method even if we have into this list a computed field depending of others fields. It's the responsibility of the orm to compute and resolve the dependency hierarchy of a compute method...

tfossoulw:
New version:

Add a store=True on the product_qty field will allow to trigger the price_subtotal field when edditing the product_purchase_qty field via its transitivity field dependency.

Sometimes we must recalculate the product_qty and it must be done at a precise moment.
We don' t update product_qty on the inverse def to avoid infinitive loop.

When the system create/write purchase order line, the system will trigger the "inverse def" that recalculate the product_purchase_qty.
Then we will recalculate the product_qty based on the new product_purchase_qty.

purchase_packaging/models/purchase_order_line.py
```diff
@@ -34,7 +35,8 @@ def _default_product_purchase_uom_id(self):
     product_qty = fields.Float(
         compute="_compute_product_qty",
         string='Quantity',
-        inverse='_inverse_product_qty'
+        inverse='_inverse_product_qty',
+        store=True
```
gurneyalex, line 39: you can't do that: this fields value depends on the context (you can compute the quantity in the context of a warehouse, or of a stock location...). Changing product_qty to a stored field will break odoo.

gurneyalex (changes requested):

rousseldenis:
Closing this as it is quite old. Feel free to reopen it.

PR #351, opened 2017-11-08
[10.0] purchase_packaging fix _get_product_seller with no date_order

tfossoulw:
on the def _select_seller() the system try to compare "date" with seller.date_start and seller.date_end.
seller.date_start and seller.date_end are str type and here we send date() type.
It generate an error.

### DIFF
--- a/purchase_packaging/models/procurement_order.py
+++ b/purchase_packaging/models/procurement_order.py
@@ -18,7 +18,8 @@ class ProcurementOrder(models.Model):
         seller = self.product_id._select_seller(
             partner_id=supplier.name,
             quantity=res['product_qty'],
-            date=po.date_order and fields.Date.from_string(po.date_order),
+            date=po.date_order and fields.Date.to_string(
+                fields.Date.from_string(po.date_order)) or None,
             uom_id=self.product_id.uom_po_id)
         if seller.packaging_id:
                 res['packaging_id'] = seller.packaging_id.id
--- a/purchase_packaging/__manifest__.py
+++ b/purchase_packaging/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Purchase Packaging",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "author": 'ACSONE SA/NV, '
               'Odoo Community Association (OCA)',
     "category": "Warehouse",
--- a/purchase_packaging/models/purchase_order_line.py
+++ b/purchase_packaging/models/purchase_order_line.py
@@ -43,8 +43,8 @@ class PurchaseOrderLine(models.Model):
         return self.product_id._select_seller(
             partner_id=self.order_id.partner_id,
             quantity=self.product_qty,
-            date=self.order_id.date_order and
-            fields.Date.from_string(self.order_id.date_order),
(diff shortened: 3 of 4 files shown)


rousseldenis (approved):
Indeed it was a bug.

Fast merge

purchase_packaging/models/procurement_order.py
```diff
@@ -18,7 +18,7 @@ def _prepare_purchase_order_line(self, po, supplier):
         seller = self.product_id._select_seller(
             partner_id=supplier.name,
             quantity=res['product_qty'],
-            date=po.date_order and fields.Date.from_string(po.date_order),
+            date=po.date_order or None,
```
lmignon: @tfossoulw Into the server the date_order is always a string! Therefore the fix must be:
```python
date=po.date_order and fields.Date.from_string(po.date_order) or None,
```

lmignon (changes requested):

rousseldenis (changes requested):
No.

Was not good

purchase_packaging/models/procurement_order.py
tfossoulw: On the _select_seller, it compare "seller.date_start" with "date".
Ex: if seller.date_start and seller.date_start > date:

In this way  "seller.date_start" is a String and is not converted in to a date. So if we send a date format it will not work.


I tested it and I got this error : 
  File "/home/odoo/wink/colombo_odoo/odoo-colombo/src/odoo/odoo/addons/product/models/product.py", line 508, in _select_seller
    if seller.date_start and seller.date_start > date:
TypeError: can't compare datetime.date to str

Edit: I also put a date on the test to validate it.

rousseldenis: Ok, I re-checked the _select_seller function and you are right. The expected date argument is string.

rousseldenis (approved):

purchase_packaging/models/procurement_order.py
rousseldenis: @tfossoulw The problem is that in PO, date_order is Datetime, so the solution would be 

```python
date=po.date_order and fields.Date.to_string(fields.Date.from_string(po.date_order)) or None,
```

rousseldenis (changes requested):

rousseldenis (approved):
Ok now

purchase_packaging/models/procurement_order.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2015-2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import api, fields, models
+from odoo import api, models, fields
```
pedrobaeza: Why do you reorder this? It was sorted alphabetically, so it was even better

pedrobaeza (changes requested):
Change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

rousseldenis:
@tfossoulw Can you attend @pedrobaeza suggested changes ?

tfossoulw:
@pedrobaeza @rousseldenis version changed + revert import order.

rousseldenis:
Good!

pedrobaeza (approved):

PR #356, opened 2017-11-16
[9.0][FIX] stock_inventory_lockdown: locking unneded locations.

LoisRForgeFlow:
**issue:**
Locations used to get the implied inventories are being added after all as locked locations, which doesn't make sense.

**example scenario:**
1. Inventory adjustment running at Stock/shelf 1
2. Stock move from Vendors to Stock.

**Current behavior:**
Locked locations obtained: Vendors, Stock and all its children.
Stock move is blocked.

**Expected behavior:**
Locked locations obtained: Stock/shelf 1 and all its children.
Stock move ~is blocked because an inventory adjustment is being done in a child of Stock.~ is not blocked.

### DIFF
--- a/stock_inventory_lockdown/__openerp__.py
+++ b/stock_inventory_lockdown/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Inventory Lock Down",
     "summary": "Lock down stock locations during inventories.",
-    "version": "9.0.1.0.1",
+    "version": "9.0.1.0.2",
     "depends": ["stock"],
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow (approved):

JordiBForgeFlow:
@lreficent I found an issue. As a user I can create and fully process a Receipt to a location that is being inventoried (e.g. WH/Stock/Shelf 1). That should not be acceptable.

JordiBForgeFlow (approved):

PR #361, opened 2017-12-28
[11.0] [MIG] stock_quant_manual_assign

fanha99:
[11.0] [MIG] stock quant manual assign
please help to check, thx.

NOTE: in v11, the stock_quant is take different role in compare to v10.
This module must be stock lot/serial manual assign


pedrobaeza:
You can review other PRs and ask in exchange that they review yours

fanha99:
Hi,
Am I need to fix all the test_flake8 in the Travis-CI, most of them is "line too long (96 > 79 characters)"?

2 other Travis-CI report is relative to the "test_stock_quant_manual_assign.py" which is from v9.0 -> v10.0. I just don't know how to use them. is there any document about the it? Is that good to remove them?

pedrobaeza:
You should fix everything, and test shouldn't be removed, but fixed. Tests assures that everything goes correct. Tests can fail because:

* The migration is not correct and it's not covering the expected results.
* Something in the code has changed (field names, model names, method names...) and you need to adapt the test to this new situation.

fanha99:
hi,
I think the Tests fail because of the Odoo itself.
I try to delete all the lines in the test file, except create product, it fails.
I try other tests from other module, which is contain create product command, it fails also.

fanha99:
I tried the Tests of stock module. it fails.

rousseldenis:
@fanha99 The quantity field in stock.quant model has changed.

It's called 'quantity' now (old name : qty).

fanha99:
@rousseldenis, thanks. I can run the test now. but there are still bugs
is there a document online about how to debug the test?

fanha99:
hi all,
I removed some line in the Test file which is related to move.reserved_quant_ids.
move.reserved_quant_ids was remove in v11.0
not sure how to rewrite these lines.

fanha99:
@DbTechGit you can try this module to "lock" assigned serial

rousseldenis:
@fanha99 Is this ok ?

fanha99:
@rousseldenis : i'm using this module in our production Odoo. I think it is ok for us. :)
- it is work well on PC, but not good on the mobile interface (as some problem with the kanban refresh)
- it will not work in case the stock move don't have the init demand quantity. because the module check and don't allow user to transfer the quantity is bigger then init demand quantity.

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -43,17 +43,25 @@ def _compute_qties(self):
...
-        if move.picking_id:  # pragma: no cover
-            move.picking_id.recompute_pack_op = True
-        for quant_id in move.reserved_quant_ids.ids:
-            move.write({'reserved_quant_ids': [[3, quant_id]]})
-        for line in self.quants_lines.filtered('selected'):
-            quants.append([line.quant, line.qty])
-        self.env['stock.quant'].quants_reserve(quants, move)
+        move._do_unreserve()
+        for line in self.quants_lines:
+            if line.qty > 0:
```
rousseldenis: @fanha99 Please use float_compare

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -43,17 +43,25 @@ def _compute_qties(self):
...
+                available_quantity = quant._get_available_quantity(
+                    move.product_id,
+                    move.location_id,
+                    lot_id=line.lot_id)
+                if available_quantity <= 0:
```
rousseldenis: Maybe a float_compare on this should be used too

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -63,20 +71,28 @@ def default_get(self, fields):
...
+            line = {}
+            line['quant'] = x.id
+            line['lot_id'] = x.lot_id.id
+            line['on_hand'] = x.quantity
+            line['package_id'] = x.package_id.id
+            line['selected'] = False
+            line['qty'] = 0
+            move_lines = move.move_line_ids.filtered(
+                lambda ml:
+                ml.location_id == x.location_id and ml.lot_id == x.lot_id)
```
rousseldenis, line 93: I think this is too restrictive (or I don't understand the flow) ? IMHO you have to add condition as you maybe want to match quants that don't have lot if move line has no lot.

`... and (ml.lot_id == False and x.lot_id = False or ml.lot_id == x.lot_id)`

stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
```diff
@@ -99,33 +96,32 @@ def test_quant_manual_assign(self):
...
-        self.assertEqual(len(wizard.quants_lines.filtered('selected')),
-                         len(self.move.reserved_quant_ids))
-        selected_quants = wizard.quants_lines.filtered(
-            'selected').mapped('quant')
-        self.assertEqual(
-            sum(self.move.reserved_quant_ids.mapped('qty')),
-            wizard.lines_qty)
-        for quant in self.move.reserved_quant_ids:
-            self.assertTrue(quant in selected_quants)
+        # self.assertEqual(len(wizard.quants_lines.filtered('selected')),
```
rousseldenis: Please remove comments if they are not needed

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -8,11 +8,11 @@
                 <group col='4' colspan="4">
                     <field name='quants_lines' colspan="4">
                         <tree editable="top" delete="0" create="0">
-                            <field name="quant" />
-                            <field name="location_id" />
-                            <field name="in_date" />
-                            <field name="lot_id" />
-                            <field name="package_id" />
+                            <field name="lot_id" readonly="1"/>
```
rousseldenis, line 11: Why changing to readonly from v10 ? Is the flow changed ?

rousseldenis (changes requested):
Code review

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
fanha99: these fields were set to readonly in the python code. Just not remember why need to set readonly the the xml view too. maybe there are bugs or somethings else.
will double check.
there are 10 months since the last commit (Dec 28, 2017).

stock_quant_manual_assign/wizard/assign_manual_quants.py
fanha99: multi locations?
when the delivery order is form location A, you could not choice the product from location B to send out.

stock_quant_manual_assign/wizard/assign_manual_quants.py
fanha99: should I use math.isclose(line.qty, 0.0, rel_tol=1e-6)
the math.isclose is with the python 3.5

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
rousseldenis: Ok, no problem. Try to make this merged.

stock_quant_manual_assign/wizard/assign_manual_quants.py
rousseldenis: No, Odoo orm uses a layer on top of some float fields to interpret what is stored in database. These fields are defined with decimal precision. So, to interpret values, you have to use orm-integrated functions as float_compare. Other libraries are useless.

stock_quant_manual_assign/wizard/assign_manual_quants.py
rousseldenis: On move lines, you can have multiple locations if products come from children of move location.

Try to store product A in STOCK > Shelf 1 and product B in STOCK > Shelf 2. Create a picking from stock to out with product A and product B.

Moves are from stock to out and move lines are from children locations.

fanha99: these lines of code is getting the correct reserved quant/lot info (form previous action) and display to the wizard form.

cubells:
@fanha99 can I push to this repository and finish this pull request?

fanha99:
@cubells : yes, please :)

LoisRForgeFlow:
@fanha99 Thanks for the work here!

I just tested it on runbot and it does not work as I expected if you select quants in sublocations. Example:
image

In the example above I select the quant in the small refrigerator and confirm. After that the quantity reserved is still 0.
image

pedrobaeza:
This is because there's no stock.move.line created.

stock_quant_manual_assign/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+# For copyright and license notices, see __openerp__.py file in root directory
```
rousseldenis, line 3: This is not relevant

rousseldenis (approved):
Code review. LGTM. Could you attend @lreficent remarks?

pedrobaeza:
@rousseldenis I don't think this is working properly. See our previous comments.

cubells:
I'm working on a pr to approve this pr.

cubells:
Superseded by https://github.com/OCA/stock-logistics-warehouse/pull/508

PR #363, opened 2018-01-04
[MIG] stock_inventory_preparation_filter: Migration to 10.0

chienandalu:
More preparation filters for inventories
========================================

Includes more options for making an inventory out of:

* Multiple products.
* Products of a category.
* Multiple lots

It also allows to make an inventory based on scanned products, adding a line
with product code and quantity.

cc @Tecnativa


stock_inventory_preparation_filter/__manifest__.py
```diff
@@ -1,41 +1,25 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     "name": "Extended Inventory Preparation Filters",
-    "version": "8.0.1.0.0",
+    "version": "10.0.1.0.0",
     "depends": [
         "stock",
     ],
     "author": "OdooMRP team,"
```
pedrobaeza: Remove OdooMRP

stock_inventory_preparation_filter/models/stock_inventory.py
```diff
@@ -1,34 +1,30 @@
...
+from odoo import _, api, fields, models
 
 
 class StockInventoryEmptyLines(models.Model):
     _name = 'stock.inventory.line.empty'
 
     product_code = fields.Char(
-        string='Product Code', size=64, required=True)
+        string='Product Code',
+        size=64,
```
pedrobaeza: Check if the product code is still limited to 64

stock_inventory_preparation_filter/models/stock_inventory.py
```diff
@@ -61,7 +57,7 @@ def _get_available_filters(self):
         return res_filters
 
     filter = fields.Selection(
-        selection=_get_available_filters, string='Selection Filter',
+        selection=_selection_filter, string='Selection Filter',
```
pedrobaeza: Remove the redefinition of the field. You are even breaking inheritance, because you are redefining the selection method without quotes, which makes this non inheritable.

pedrobaeza (changes requested):
Check Travis also

pedrobaeza (approved):

PR #375, opened 2018-02-13
[11.0][mig] stock_available_unreserved

mpanarin:
Migration of stock_available_unreserved to Odoo 11.
A big one as functionality of stock.quant was changed a lot in Odoo 11


stock_available_unreserved/models/product.py
```diff
@@ -0,0 +1,118 @@
...
+        quants = self.env['stock.quant'].search(
+            [
+                ('product_id', 'in', products),
+            ]
+        )
+        quant_ids = quants.filtered(
+            lambda x: x.product_id.qty_available_not_res > 0
+        ).ids
+        result = self.env.ref('stock.product_open_quants').read()[0]
+        result['domain'] = "[('id', 'in', {})]".format(quant_ids)
```
simahawk: not a requirement, just FTR: this 2 keys can be python lists and dict, no need to convert to strings ;)

stock_available_unreserved/models/product.py
```diff
@@ -0,0 +1,118 @@
...
+        compute='_compute_qty_available_not_reserved',
+    )
+
+    qty_available_stock_text = fields.Char(
+        compute='_compute_qty_available_not_reserved',
+        string='Available per stock',
+    )
+
+    @api.multi
+    def _product_available_not_res_hook(self, quants):
```
simahawk: seems not used anywhere, which means no hook at all: still needed?

stock_available_unreserved/models/product.py
```diff
@@ -0,0 +1,118 @@
...
+    @api.multi
+    def _compute_product_available_not_res_dict(self):
+
+        res = {}
+
+        domain_quant = self._prepare_domain_available_not_reserved()
+        quants = self.env['stock.quant'].with_context(lang=False).search(
+            domain_quant,
+        )
+        for prod in self:
```
simahawk: just a note for future perf-wise refactoring: wouldn't be better to pick up the quants already grouped by product.
This multiple loop + filtered sounds sub-optimal when you have tons of products and quants.
Would you mind adding a comment like `'TODO: refactor this for better performance (maybe using a raw query)`?

simahawk (changes requested):

simahawk:
@jbeficent ping :wink:  Can you have a look at my comments?

stock_available_unreserved/models/product.py
mpanarin: got lost during refactoring :face_with_head_bandage: 
will reimplement it )

stock_available_unreserved/models/product.py
mpanarin: @simahawk this actually requires further testing and optimisation.
as:
1. This is used only in compute field, so i don't think it will hold a ton of product every time
2. Seems that there can no longer be several quants for one product on same location. And they are basically per product per location. 
I mimicked the behaviour here, but a review from more knowledgable person in stock BP is required :)

max3903 (approved):

JordiBForgeFlow:
@simahawk we are going to look into it now.

stock_available_unreserved/models/product.py
yvaucher: (fixed as reimplemented here https://github.com/OCA/stock-logistics-warehouse/pull/375/files#diff-65190c8c38ef3b3e1094f032c1298735R107)

stock_available_unreserved/models/product.py
```diff
@@ -0,0 +1,119 @@
...
+
+    qty_available_stock_text = fields.Char(
+        compute='_compute_product_available_not_res',
+        string='Unreserved stock quantity',
+    )
+
+    @api.multi
+    @api.depends('product_variant_ids.qty_available_not_res')
+    def _compute_product_available_not_res(self):
+        no_new = self.filtered(lambda x: not isinstance(x.id, models.NewId))
```
yvaucher: Wouldn't be more efficient to filter those directly in the `for tmpl in self:` loop and skip them with a `continue` ?

stock_available_unreserved/models/product.py
yvaucher: I don't see a constraint on the quant model. Thus it is not safe to assume uniqueness of quants on product + location

You can have a look at the stock module how to perfectly mimic the count using a read_group.

https://github.com/odoo/odoo/blob/d28f8f704febe030f9c068f895adf0b1ce3e8dc1/addons/stock/models/product.py#L125

yvaucher (changes requested):
I'm also concerned about the loop + filter on quants

stock_available_unreserved/models/product.py
guewen: Doesn't need to be part of this pull request, this is a migration, not a refactoring :) But as @simahawk says a note would be worth to add

stock_available_unreserved/models/product.py
guewen: Yes, prevent to loop twice on the same list

stock_available_unreserved/models/product.py
guewen: I missed the fact that in 10.0 it was a `read_group`

stock_available_unreserved/models/product.py
```diff
@@ -15,102 +13,105 @@ class ProductTemplate(models.Model):
...
+            tmpl.qty_available_stock_text = "/".join(
+                tmpl.mapped('product_variant_ids.qty_available_stock_text')
+            )
 
     @api.multi
     def action_open_quants_unreserved(self):
         products = self.mapped('product_variant_ids').ids
+        quants = self.env['stock.quant'].search(
+            [
+                ('product_id', 'in', products),
```
grindtildeath: Would be nice to rename `products` to `product_ids` to distinct it's not a recordset. 
Or do `.ids` here instead of above.

stock_available_unreserved/models/product.py
```diff
@@ -15,102 +13,105 @@ class ProductTemplate(models.Model):
...
 
     qty_available_not_res = fields.Float(
-        string='Quantity On Hand Unreserved', digits=UNIT,
-        compute='_compute_product_available_not_res')
+        string='Quantity On Hand Unreserved',
+        digits=UNIT,
+        compute='_compute_product_available_not_res',
+    )
 
     qty_available_stock_text = fields.Char(
```
grindtildeath: Can't this field be removed ?
If I understood correctly it just adds ' On Hand' after the value qty_available_not_res, and it could be done in the view instead of having a field for that.

grindtildeath (approved):
Two small non blocking comments. Looks good :+1:

stock_available_unreserved/models/product.py
rgarnau: So, as we understand, would be:

```
domain_quant = self._prepare_domain_available_not_reserved()
quants_res = self.env['stock.quant'].with_context(lang=False).read_group(domain_quant, ['product_id', 'quantity', 'reserved_quantity'], ['product_id'], orderby='id')
for prod in self:
    quantity = sum(datum['quantity'] - datum['reserved_quantity'] for 
    datum in quants_res)
    vals['qty_available_not_res'] = quantity
    res[prod.id] = vals
```

guewen:
@yvaucher can you update your review?

simahawk (approved):

guewen (approved):

LoisRForgeFlow:
@mpanarin seems that tests are failing, could you have a look?

gurneyalex (approved):

PR #379, opened 2018-02-16
[11.0] Putaway strategy: add module to restore the method field

asaunier:
The method field was removed in module 'stock' of Odoo 11. See
https://github.com/odoo/odoo/commit/f2926823e848f4aee89542ced9fb846c74a524d2

Since the field is used by OCA modules such as stock_putaway_product
(eg. to add a 'per_product' method), the new stock_putaway_method module
is added to restore this feature if needed.

Current putaway strategy editing panel in the inventory configuration:

capture du 2018-02-16 09-59-10

With the new stock_putaway_method module:

capture du 2018-02-15 17-40-35

### DIFF
--- a/stock_putaway_method/README.rst
+++ b/stock_putaway_method/README.rst
@@ -1,0 +1,50 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+=======================
+Putaway strategy method
+=======================
+
+This module adds the putaway strategy method back, removed from the stock module in Odoo 11.
+
+Installation
+============
+
+To install this module, just click the install button. This module is automatically installed when installing relying modules such as *stock_putaway_product*.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* Alexandre Saunier - Camptocamp SA <alexandre.saunier@camptocamp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/stock_putaway_method/models/product_strategy.py
+++ b/stock_putaway_method/models/product_strategy.py
@@ -1,0 +1,18 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class PutAwayStrategy(models.Model):
+    _inherit = 'product.putaway'
+
+    method = fields.Selection(
+        selection='_get_putaway_options',
+        string='Method',
+        default='fixed',
+        required=True,
+    )
+
+    @api.model
+    def _get_putaway_options(self):
+        return [('fixed', 'Fixed Location')]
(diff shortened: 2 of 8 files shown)


pedrobaeza:
But the commit says that is not used anymore...

Garamotte:
@pedrobaeza Maybe it's not used anymore because Odoo removed some features, this already happened in the past. Or maybe it's not used anymore because another features allows to manage this in a different way...

@asaunier Did you try to get the same behaviour in a plain standard Odoo without adding this field back ?

asaunier:
@pedrobaeza @sylvain-garancher Thanks for reviewing this PR :)

In Odoo 10 the ``method`` selector was available but with only one option, ``fixed location``. This is the only method supported by the ``stock`` module.

According to https://github.com/odoo/odoo/blob/11.0/addons/stock/doc/stock.rst#preparing-pack-operations
> In the stock module, there is one putaway strategy: fixed location. For each such strategy you can also specify the related location. Of course, based on this, custom developments make it possible to implement the putaway strategy you want (as it is applied on all of the stock being moved at once).

In Odoo 11 the field has been removed (perhaps because a one-option selector is not that useful?) by odoo/odoo@f292682 making ``fixed location`` the enforced method.

But the ``stock_putaway_product`` module makes it possible to choose another putaway strategy, thus we need to add the selector back.

> Did you try to get the same behaviour in a plain standard Odoo without adding this field back ?

I have found out no other way to choose an alternate putaway strategy. With the standard stock module in Odoo11 , the putaway strategy seems "hard-coded" to "fixed location, according to https://github.com/odoo/odoo/blob/11.0/addons/stock/views/product_strategy_views.xml#L7-L19

yvaucher:
@pedrobaeza @sylvain-garancher this fied is required to have different method implemented. `stock_putaway_product` module was relying on this field to choose whether it is the standard behavior or the `pre_product`.

https://github.com/OCA/stock-logistics-warehouse/blob/8c5039ca1052eeae730f16ee09c20a25f96aff48/stock_putaway_product/models/product_putaway.py#L38-L44

EDIT: sneak peek of stock_putaway_product

yvaucher:
@asaunier please find here the template to add a `README.rst` file:

https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

bealdav (approved):

Garamotte:
Ok, thanks for the explanation @yvaucher :)

stock_putaway_method/README.rst
```diff
@@ -0,0 +1,50 @@
...
+Installation
+============
+
+To install this module, just click the install button. This module is automatically installed when installing relying modules such as *stock_putaway_product*.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
LoisRForgeFlow: fill project_repo with `stock-logistics-warehouse`

LoisRForgeFlow (approved):
LGTM

asaunier:
This PR is now ready and required by PR #380 

After @bealdav and @lreficent's approvals, the ``putaway_apply`` function has been removed from https://github.com/asaunier/stock-logistics-warehouse/blob/1ec85097b9dafcd9a61893d69d92be9912b06de2/stock_putaway_method/models/product_strategy.py#L6 since it was somehow a duplicate of the function already available in the parent ``stock`` module at https://github.com/odoo/odoo/blob/11.0/addons/stock/models/product_strategy.py#L24 => "fixed location" is indeed the standard strategy.

This function is overridden in the child ``stock_putaway_product`` module at https://github.com/asaunier/stock-logistics-warehouse/blob/68110806569c07f7f8dcc45932bdf783be2fa3bd/stock_putaway_product/models/product_putaway.py#L37 to support the "per_product" strategy.

stock_putaway_method/tests/test_putaway_method.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+
+
+class TestPutawayMethod(SavepointCase):
```
mpanarin: No real reason to use SavepointCase when no actual data is created or updated in setUpClass. Can be TransactionCase

stock_putaway_method/models/product_strategy.py
```diff
@@ -0,0 +1,13 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PutAwayStrategy(models.Model):
+    _inherit = 'product.putaway'
+
+    method = fields.Selection('_get_putaway_options', "Method",
```
mpanarin: YAPF makes weird decisions sometimes ;)
try to keep formatting with parameters on separate lines with trailing commas )

mpanarin (approved):
LGTM, comments are non-blocking

asaunier:
Thanks @mpanarin PR updated according to your remarks.

guewen (approved):

PR #385, opened 2018-02-27
[11.0] [MIG] stock orderpoint manual procurement

bodedra:


stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint_view.xml
```diff
@@ -0,0 +1,50 @@
...
+                             <field name="qty" groups="stock_orderpoint_manual_procurement.group_change_orderpoint_procure_qty"/>
+                            <field name="uom_id" groups="product.group_uom"/>
+                        </tree>
+                    </field>
+                </group>
+                <footer>
+                    <button name="make_procurement" string="Execute"
+                            type="object"  class="oe_highlight"  />
+                    or
+                    <button string="Cancel" class="oe_link" special="cancel" />
```
guewen: Should be updated to the latest changes, example from the account module:

```xml
                    <footer>
                        <button string="Reconcile" name="trans_rec_reconcile" type="object" default_focus="1" class="btn-primary"/>
                        <button string="Cancel" class="btn-default" special="cancel"/>
                    </footer>
```

The `or` must be moved, the class attribute change on both buttons and there is a default_focus on the primary button.

guewen:
When the user does not have the `Change quantity in manual procurements from reordering rules` right, the behavior is strange. The button is displayed:

selection_229

But the wizard only displays the name of the product:

selection_228

I was very confused by this, like, what is going to happen if I use the `Execute` button?
In my opinion, the recommended value should be displayed as readonly if you don't have the right to change the value.

That being said, it was the same in 10.0 so it should probably not be in the scope of the migration.

guewen (approved):
A small nice-to-have change in the wizard but overall looks good.

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint_view.xml
bodedra: Corresponded. Thanks.

bodedra:
@guewen Agree with comment. "The recommended value should be displayed as readonly if you don't have the right to change the value"

Corresponded. Thanks.

max3903 (approved):

guewen:
Thanks @bodedra!

PR #388, opened 2018-03-01
[10.0][IMP] purchase_packaging: reduce sql query calls for search uom reference category

tfossoulw:
We move the search launch on 2 def before the loop to avoid to call it for each lines.

So we will only launch the search one time whatever the number of line we will have on the PO.

### DIFF
--- a/purchase_packaging/models/purchase_order_line.py
+++ b/purchase_packaging/models/purchase_order_line.py
@@ -64,8 +64,6 @@ class PurchaseOrderLine(models.Model):
              ('uom_type', '=', 'reference')])
         uom_by_category = {to_uom.category_id: to_uom for to_uom in to_uoms}
         for line in self:
-            if not line.product_purchase_uom_id:
-                continue
             line.product_qty = line.product_purchase_uom_id._compute_quantity(
                 line.product_purchase_qty,
                 uom_by_category.get(line.product_purchase_uom_id.category_id))
--- a/purchase_packaging/models/purchase_order_line.py
+++ b/purchase_packaging/models/purchase_order_line.py
@@ -48,38 +48,48 @@ class PurchaseOrderLine(models.Model):
             uom_id=self.product_uom)
 
     @api.multi
-    @api.depends('product_purchase_uom_id', 'product_purchase_qty')
+    @api.depends('product_purchase_uom_id',
+                 'product_purchase_qty',
+                 'product_purchase_uom_id.category_id')
     def _compute_product_qty(self):
         """
         Compute the total quantity
         """
+        uom_categories = self.mapped("product_purchase_uom_id.category_id")
+        uom_obj = self.env['product.uom']
+        to_uoms = uom_obj.search(
+            [('category_id',
+              'in',
+              uom_categories.ids),
+             ('uom_type', '=', 'reference')])
+        uom_by_category = {to_uom.category_id: to_uom for to_uom in to_uoms}
         for line in self:
-            uom_obj = self.env['product.uom']
-            to_uom = uom_obj.search(
-                [('category_id',
-                  '=',
-                  line.product_purchase_uom_id.category_id.id),
-                 ('uom_type', '=', 'reference')], limit=1)
             if not line.product_purchase_uom_id:
-                return
+                continue
             line.product_qty = line.product_purchase_uom_id._compute_quantity(
                 line.product_purchase_qty,
-                to_uom)
+                uom_by_category.get(line.product_purchase_uom_id.category_id))
 
     @api.multi
     def _inverse_product_qty(self):
         """ If product_quantity is set compute the purchase_qty
         """
+        uom_categories = self.mapped("product_purchase_uom_id.category_id")
+        uom_obj = self.env['product.uom']
+        from_uoms = uom_obj.search(
+            [('category_id',
+              'in',
+              uom_categories.ids),
+             ('uom_type', '=', 'reference')])
+        uom_by_category = {
+            from_uom.category_id: from_uom for from_uom in from_uoms}
         for line in self:
             if line.product_id:
                 supplier = line._get_product_seller()
                 if supplier:
                     product_purchase_uom = supplier.min_qty_uom_id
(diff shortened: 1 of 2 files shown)


purchase_packaging/models/purchase_order_line.py
```diff
@@ -53,33 +53,44 @@ def _compute_product_qty(self):
         """
         Compute the total quantity
         """
+        uom_categories = self.mapped("product_purchase_uom_id.category_id")
```
rousseldenis, line 58: @tfossoulw I think this is good but you should put product_purchase_uom_id.category_id in the depends as recordset would be already filled in with these records before entering function.

rousseldenis: (As your 'mapped' could induce a query if records are not fetched)

rousseldenis (changes requested):
Minor comments

purchase_packaging/models/purchase_order_line.py
tfossoulw: @rousseldenis done :)

rousseldenis:
@Cedric-Pigeon @lmignon

purchase_packaging/models/purchase_order_line.py
```diff
@@ -48,38 +48,51 @@ def _get_product_seller(self):
...
+        uom_obj = self.env['product.uom']
+        to_uoms = uom_obj.search(
+            [('category_id',
+              'in',
+              uom_categories.ids),
+             ('uom_type', '=', 'reference')])
+        uom_by_category = {}
```
lmignon:
```python
uom_by_category = {to_uom.category_id: to_uom for to_uom in to_uoms}
```

purchase_packaging/models/purchase_order_line.py
```diff
@@ -48,38 +48,51 @@ def _get_product_seller(self):
...
         """
+        uom_categories = self.mapped("product_purchase_uom_id.category_id")
+        uom_obj = self.env['product.uom']
+        from_uoms = uom_obj.search(
+            [('category_id',
+              'in',
+              uom_categories.ids),
+             ('uom_type', '=', 'reference')])
+        uom_by_category = {}
+        for from_uom in from_uoms:
```
lmignon: `uom_by_category = {to_uom.category_id: to_uom for to_uom in to_uoms}`

lmignon (changes requested):
@tfossoulw Thank you for the improvement. PLZ Can you reword the commit message to be more meaningful?

tfossoulw:
@lmignon @rousseldenis i rename the commit and adapt the code

rousseldenis:
@tfossoul Please fix flake8 (line 84).

Please improve also PR title

rousseldenis (approved):

rousseldenis:
@tfossoulw Sorry but can you just update module version (patch) ?

tfossoulw:
@rousseldenis so 10.0.1.0.1

lmignon (approved):
(Code review Only)

acsonefho (approved):

rousseldenis:
@Cedric-Pigeon ok ?

purchase_packaging/models/purchase_order_line.py
```diff
@@ -48,38 +48,48 @@ def _get_product_seller(self):
...
-                [('category_id',
-                  '=',
-                  line.product_purchase_uom_id.category_id.id),
-                 ('uom_type', '=', 'reference')], limit=1)
             if not line.product_purchase_uom_id:
                 return
             line.product_qty = line.product_purchase_uom_id._compute_quantity(
                 line.product_purchase_qty,
-                to_uom)
+                uom_by_category.get(line.product_purchase_uom_id.category_id))
```
gurneyalex: disclaimer : I don't use this module and I don't know how it is supposed to work. I agree that your change is not modifying the original behavior, but I have doubts about that behavior. Why are we using the reference unit of the category of the uom as target for a conversion? In the standard stock addon, the quantities are converted to the uom of the product which makes much more sense in my opinion. Since this is the unit which will be used when processing in the move, should we not use it in here?

gurneyalex: cc @lmignon who is the original author of the module

lmignon: @gurneyalex Indeed, I'm the original author in v7.... Your comment makes sense and I've any motivation in mind to oppose to you. @rousseldenis can you check this?

rousseldenis: @gurneyalex Indeed, this code is correct as in this module we change the behaviour of standards Odoo fields product_qty and product_uom_id. They are computed now.

The fields that appear in forms are new ones to make conversion easier. We use here two dimensional units.

So, we compute the original fields considering reference purchase units to make procurements and stock moves values conversion transparent for user.

@Cedric-Pigeon I know you worked on this too. An advice ?

purchase_packaging/models/purchase_order_line.py
```diff
@@ -48,38 +48,48 @@ def _get_product_seller(self):
...
+             ('uom_type', '=', 'reference')])
+        uom_by_category = {to_uom.category_id: to_uom for to_uom in to_uoms}
         for line in self:
-            uom_obj = self.env['product.uom']
-            to_uom = uom_obj.search(
-                [('category_id',
-                  '=',
-                  line.product_purchase_uom_id.category_id.id),
-                 ('uom_type', '=', 'reference')], limit=1)
             if not line.product_purchase_uom_id:
```
rousseldenis: @Cedric-Pigeon @lmignon This is not correct as we must not stop computation. Maybe a self.filtered('product_purchase_uom_id') should be more convenient at the loop beginning.

lmignon: @rousseldenis good catch... should be 'continue"

tfossoulw:
Last commit: product_purchase_uom_id is mandatory so there is no need to test the filled value on _compute_product_qty.

I also update the version.

PR #389, opened 2018-03-09
[ENH][11.0] stock_orderpoint_uom

bodedra:
Need to review:

- check context to avoid twice quantity calculation.


max3903 (approved):

LoisRForgeFlow:
@bodedra could you explain a little why are you doing this?

bodedra:
@lreficent sorry for late response on these. Would you please review new updated change on it and also https://github.com/OCA/stock-logistics-warehouse/pull/396

rousseldenis (approved):

LoisRForgeFlow (changes requested):
This is not the way to fix the issue. see https://github.com/OCA/stock-logistics-warehouse/pull/396#pullrequestreview-105699909

LoisRForgeFlow:
This is not needed, proper fix is being done at https://github.com/OCA/stock-logistics-warehouse/pull/396

PR #392, opened 2018-03-15
[FIX] stock_request routes on childs

etobella:
When you are working with complex warehouse structures, you could have multiple locations with their parents. Routes of the the parents also applies to child location, so, it should be necessary to add the option to select all the possible rules a location on stock request.
Right now, it only shows the specific rules of the location.
This PR solves this issue.
@jbeficent

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -177,6 +177,14 @@ class StockRequest(models.Model):
                 request.product_id.uom_id._compute_quantity(
                     open_qty, request.product_uom_id)
 
+    def get_parents(self):
+        location = self.location_id.sudo()
+        result = location
+        while location.location_id:
+            location = location.location_id
+            result |= location
+        return result
+
     @api.constrains('product_id')
     def _check_product_uom(self):
         ''' Check if the UoM has the same category as the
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):
Good catch!

rgarnau (approved):
LGTM :+1:

stock_request/models/stock_request.py
```diff
@@ -177,6 +177,14 @@ def _compute_qty(self):
                 request.product_id.uom_id._compute_quantity(
                     open_qty, request.product_uom_id)
 
+    def get_parents(self):
+        location = self.location_id.sudo()
```
rousseldenis, line 181: Why this?

Use of sudo is to be confidential. Remember you switch to new environment.

In which case sudo is mandatory ?

rousseldenis (changes requested):

stock_request/models/stock_request.py
etobella: We search with sudo because the user may not have permissions to view the locations, but we must access them in order to obtain routes as it is used on https://github.com/OCA/stock-logistics-warehouse/pull/392/commits/262672cb4f223c54d853a93e4a9868989e5eeb5e#diff-cb7ac1537b7c039cfc237796ef052a90L212

rousseldenis (approved):

PR #393, opened 2018-03-16
[10.0] Add module stock_inventory_pos_category

alexis-via:
See README for more info.

### DIFF
--- a/stock_inventory_pos_category/README.rst
+++ b/stock_inventory_pos_category/README.rst
@@ -1,0 +1,51 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Stock Inventory POS category
+============================
+
+When you create an inventory, you can choose to select a particular product category. With this module, you can also choose to select one or more point of sale categories.
+
+Usage
+=====
+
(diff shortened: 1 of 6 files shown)


pedrobaeza:
This should be in OCA/pos, isn't it?

alexis-via:
haha, I was expecting this one :) I'm sure that if I had proposed it in OCA/pos, someone would have said "Shouldn't it be in OCA/stock-xxx" :) I really don't care.

gurneyalex (approved):

rousseldenis:
@alexis-via Shouldn't this be included in stock_inventory_preparation_filter ?

gurneyalex:
@rousseldenis @alexis-via is the last question blocking from merging this PR?

rousseldenis:
@alexis-via

alexis-via:
@rousseldenis the module stock_inventory_preparation_filter only depends on "stock", it doesn't depend on "point_of_sale", so it can't use pos categories. So my answer would be "no"... but maybe I didn't understand your question...

rousseldenis:
@alexis-via What I wanted to say is that you could extend stock_inventory_preparation_filter with stock_inventory_preparation_filter_pos as it covers the same functionnal perimeter.

Moreover, that makes repository more readable for people looking for some modules

rousseldenis:
@alexis-via Could you rebase ?

alexis-via:
@rousseldenis  I took into account your remark while porting the feature to v14: new module stock_inventory_preparation_filter_pos for v14 here: https://github.com/OCA/stock-logistics-warehouse/pull/1224

PR #395, opened 2018-03-19
[11.0][IMP] stock_request_purchase: Remove hook

MiquelRForgeFlow:
The module `purchase_procurement_run_buy_hook` is not needed anymore.

### DIFF
--- a/stock_request_purchase/models/procurement_rule.py
+++ b/stock_request_purchase/models/procurement_rule.py
@@ -15,13 +15,10 @@ class ProcurementRule(models.Model):
             vals['stock_request_ids'] = [(4, values['stock_request_id'])]
         return vals
 
-    def _prepare_purchase_order_line_update(self, line,
-                                            procurement_uom_po_qty,
-                                            price_unit, values):
-        vals = super(ProcurementRule,
-                     self)._prepare_purchase_order_line_update(
-            line, procurement_uom_po_qty, price_unit, values
-        )
+    def _update_purchase_order_line(self, product_id, product_qty, product_uom,
+                                    values, line, partner):
+        vals = super(ProcurementRule, self)._update_purchase_order_line(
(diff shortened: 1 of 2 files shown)


stock_request_purchase/models/procurement_rule.py
```diff
@@ -15,13 +15,10 @@ def _prepare_purchase_order_line(self, product_id, product_qty,
...
-    def _prepare_purchase_order_line_update(self, line,
-                                            procurement_uom_po_qty,
-                                            price_unit, values):
-        vals = super(ProcurementRule,
-                     self)._prepare_purchase_order_line_update(
-            line, procurement_uom_po_qty, price_unit, values
-        )
+    def _update_purchase_order_line(self, product_id, product_qty, product_uom,
+                                    values, line, partner):
+        vals = super(ProcurementRule, self).__update_purchase_order_line(
```
LoisRForgeFlow: there is double underscore here

MiquelRForgeFlow: good catch!

MiquelRForgeFlow:
:green_apple:

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM now :+1: 

FTR, https://github.com/odoo/odoo/commit/71d5425c83edd212b4acf00f1d55bd5ee5f96761

JordiBForgeFlow:
My mistake. This PR is bad, bad. The fact that the module purchase_procurement_run_buy_hook is not needed does not mean that the method should not be called! As you can see in odoo/odoo@71d5425, Odoo has now added this same method in core, that we need to inherit.

So, the only thing needed was to remove the module purchase_procurement_run_buy_hook...

I am reverting here https://github.com/etobella/stock-logistics-warehouse/pull/1

PR #396, opened 2018-03-20
[ADD] updated parameter values pass to run() method, to avoid twice qty calculation

bodedra:
Need to review:

- Remove `_prepare_procurement_values()` method
- Pass context to `run()` method to avoid twice quantity calculation.

### DIFF
--- a/stock_orderpoint_manual_procurement/models/stock_warehouse_orderpoint.py
+++ b/stock_orderpoint_manual_procurement/models/stock_warehouse_orderpoint.py
@@ -24,14 +24,6 @@ class StockWarehouseOrderpoint(models.Model):
         compute="_compute_procure_recommended",
     )
 
-    def _prepare_procurement_values(self, product_qty, date=False,
-                                    group=False):
-        res = super(StockWarehouseOrderpoint,
-                    self)._prepare_procurement_values(product_qty, date=date,
-                                                      group=group)
-        res.update({'qty': product_qty})
-        return res
-
     @api.multi
     def _get_procure_recommended_qty(self, virtual_qty, op_qtys):
         self.ensure_one()
--- a/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
+++ b/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
@@ -71,7 +71,7 @@ class MakeProcurementOrderpoint(models.TransientModel):
                 self.env['procurement.group'].run(
                     item.orderpoint_id.product_id,
                     item.qty,
-                    item.orderpoint_id.product_uom,
+                    item.uom_id,
                     item.orderpoint_id.location_id,
                     item.orderpoint_id.name,
                     item.orderpoint_id.name,


stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -68,7 +68,7 @@ def make_procurement(self):
             values = item.orderpoint_id._prepare_procurement_values(item.qty)
             # Run procurement
             try:
-                self.env['procurement.group'].run(
+                self.env['procurement.group'].with_context(by_pass=True).run(
```
rousseldenis: @bodedra It is not recommended to pass parameters through context. You could face performance problems as whole environment will be recreated.

You can add parameter 'by_pass' to run function.

rousseldenis (changes requested):

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
bodedra: @rousseldenis Corresponded. Thank you.

rousseldenis (approved):

LoisRForgeFlow:
@bodedra why is this double calculation happening? Seems a little bit ugly to me to pass that by_pass parameter on the dict with the values to procure, not sure if there is any other option though...

bodedra:
@lreficent 

Following two modules are independed and third one is depends on previous two.
    
    i)   stock_orderpoint_uom (override run() method) and
    ii)  stock_orderpoint_manual_procurement (wizard button click will fire procurement with the help of run() method)
    iii) stock_orderpoint_manual_procurement_uom (will display converted procure quantity to wizard line)    


Twice procure quantity calculation raise when procure uom come into picture.

Run procurement from wizard with uom quantity. https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py#L71

This line of code convert quantity to procure uom. https://github.com/ursais/stock-logistics-warehouse/blob/c8496279a686357979cdf7206e1987c3cc9646c1/stock_orderpoint_manual_procurement_uom/models/stock_warehouse_orderpoint.py#L12 so no need to covert quantity again with procure uom.

This line of code will avoid twice calculation. https://github.com/ursais/stock-logistics-warehouse/blob/13d7d32cfa1ff83796ad2c43909e0ae5c1030c17/stock_orderpoint_uom/models/procurement_group.py#L14

LoisRForgeFlow:
@bodedra thanks for the explanation I see now.

But what confuses me is the following:
If you reach this point [stock_orderpoint_uom-check procure uom and orderpoint uom](https://github.com/ursais/stock-logistics-warehouse/blob/13d7d32cfa1ff83796ad2c43909e0ae5c1030c17/stock_orderpoint_uom/models/procurement_group.py#L16)
There are two options:
1. Automatically created: then it will convert correctly to the proure uom and qty. Only once through the code, nothing to do with stock_orderpoint_manual_procurement_uom.
2. Manually created through stock_orderpoint_manual_procurement_uom: then the procure should've been done with the correct uom ([stock_orderpoint_manual_procurement_uom - procure uom passed](https://github.com/ursais/stock-logistics-warehouse/blob/c8496279a686357979cdf7206e1987c3cc9646c1/stock_orderpoint_manual_procurement_uom/wizards/make_procurement_orderpoint.py#L15)) and qty ([what you already pointed](https://github.com/ursais/stock-logistics-warehouse/blob/c8496279a686357979cdf7206e1987c3cc9646c1/stock_orderpoint_manual_procurement_uom/models/stock_warehouse_orderpoint.py#L12)). In this case the [above mentioned if](https://github.com/ursais/stock-logistics-warehouse/blob/13d7d32cfa1ff83796ad2c43909e0ae5c1030c17/stock_orderpoint_uom/models/procurement_group.py#L16) will not be true so no second recomputation of the qty.

What am I missing?

LoisRForgeFlow (changes requested):
just checked locally, this is not the way. the probelm is here:

https://github.com/ursais/stock-logistics-warehouse/blob/0c203c34b64250d82ec551452d77a99a1fc76b5d/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py#L75

you have to modify that properly to be able to pass the `item.uom_id` instead of `item.orderpoint_id.product_uom` once you install `stock_orderpoint_manual_procurement_uom`

LoisRForgeFlow:
Actually not even needed to do anything more than change that line I just pointed with `item.uom_id` because uom_id is on the wizard.

bodedra:
@lreficent thanks for your feedback. I have corresponded as per your suggestion. Would you please update your review ?

LoisRForgeFlow:
@bodedra thanks! remove all the other changes of the PR, that line is all you need. Squash it on only one commit so we can merge

stock_orderpoint_manual_procurement/models/stock_warehouse_orderpoint.py
```diff
@@ -24,14 +24,6 @@ class StockWarehouseOrderpoint(models.Model):
         compute="_compute_procure_recommended",
     )
 
-    def _prepare_procurement_values(self, product_qty, date=False,
```
LoisRForgeFlow, line 27: You don't have to remove this either. Please, readd it

LoisRForgeFlow (changes requested):

stock_orderpoint_manual_procurement/models/stock_warehouse_orderpoint.py
bodedra: @lreficent We have added that method to use of product quantity. Afterwards, we improve it at https://github.com/OCA/stock-logistics-warehouse/commit/bcccda5bd4bc4ee60e0d12b81fcb309a231d3af3

So now no need of these method to get product quantity. Can you guide me to further use of it ?

LoisRForgeFlow: Ok, not related to this fix, I got you now. Thanks!

LoisRForgeFlow (approved):

LoisRForgeFlow:
Finally merged, thanks @bodedra

PR #397, opened 2018-03-21
[10.0][IMP] purchase_packaging: optimise def subtract_procurements_fr…

tfossoulw:
…om_orderpoints

This is a bit more performant when we have several orderpoint to process.

### DIFF
--- a/purchase_packaging/models/procurement_order.py
+++ b/purchase_packaging/models/procurement_order.py
@@ -50,10 +50,18 @@ class Orderpoint(models.Model):
         # This override will return 12 and no additionnal procurement will be
         # created
         res = super(Orderpoint, self).subtract_procurements_from_orderpoints()
+        procurements = self.env['procurement.order'].search(
+            [('orderpoint_id', 'in', self.ids),
+             ('state', 'not in', ['cancel', 'done']),
+             ('purchase_line_id.state', '=', ['draft', 'sent', 'to approve'])])
+        procurements.mapped('product_uom.rounding')
+        procurements.mapped('purchase_line_id.state')
+        procs_by_orderpoint = dict.fromkeys(
+            self.ids, self.env['procurement.order'])
+        for proc in procurements:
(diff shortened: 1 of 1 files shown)


gurneyalex (approved):

lmignon (approved):

purchase_packaging/models/procurement_order.py
```diff
@@ -50,10 +50,18 @@ def subtract_procurements_from_orderpoints(self):
         # This override will return 12 and no additionnal procurement will be
         # created
         res = super(Orderpoint, self).subtract_procurements_from_orderpoints()
+        procurements = self.env['procurement.order'].search(
+            [('orderpoint_id', 'in', self.ids),
+             ('state', 'not in', ['cancel', 'done']),
+             ('purchase_line_id.state', '=', ['draft', 'sent', 'to approve'])])
+        procurements.mapped('product_uom.rounding')
+        procurements.mapped('purchase_line_id.state')
```
LoisRForgeFlow, line 58: Why these two `mapped`??

rousseldenis: @lreficent That's done to be sure all data being fetched in recordset before entering the loop (that's the case on functions that don't depends on ...'depends' - which makes pre-fetching).

That's a good way to avoid too much queries.

tfossoulw: It prefetch data and put them in memory. Then the following code run must faster.

LoisRForgeFlow: @tfossoulw @rousseldenis Thanks for the explanation, good to know it!

LoisRForgeFlow (approved):
LGTM

rousseldenis:
@tfossoulw @lreficent Module version upgrade missing

LoisRForgeFlow:
@rousseldenis @pedrobaeza thanks for pointing it and fixing it.

PR #401, opened 2018-03-29
[11.0] Fix computation of stock available unreserved

guewen:
The computation of the unreserved available amount using the
StockQuant._get_available_quantity was wrong as soon as more than one
quant was found for the same product. It can easily happen when you have
sublocations and a quant in each location.

The reason is that the algorithm was:

1. searching for all the quants for a given product
2. calling StockQuant._get_available_quantity for each quant
3. _get_available_quantity is an @api.model method, which itself will
search for all quants for the product and the given location and
children

Which means that if you have these locations:

Stock
Stock > Bin A
Stock > Bin B

And these quants:

1. Product: Product A
   Location: Bin A
   Quantity: 60
   Reserved: 0

2. Product: Product A
   Location: Bin B
   Quantity: 10
   Reserved: 0

Instead of 70, the result was 140. (One loop for each quant, each loop
recomputing the total quantity in _get_available_quantity, all summed
togethed, for each new quant, an additional sum would be added).

Ultimately, the _get_available_quantity method does the sum of (quantity
- quantity reserved). This commit uses the same logic than the 10.0
branch, it finds the quants contextually using
ProductProduct._get_domain_locations and get the available quantity as
the sum of (quantity - quantity reserved).

We can't really use StockQuant._get_available_quantity because this one
expects a location, while here we don't necessarily know it.

### DIFF
--- a/stock_available_unreserved/models/product.py
+++ b/stock_available_unreserved/models/product.py
@@ -6,6 +6,7 @@
 
 from odoo import api, fields, models
 from odoo.addons import decimal_precision as dp
+from odoo.tools.float_utils import float_round
 
 UNIT = dp.get_precision('Product Unit of Measure')
 
@@ -57,15 +58,9 @@ class ProductProduct(models.Model):
     )
 
     @api.multi
-    def _product_available_not_res_hook(self, quants):
-        """Hook used to introduce possible variations"""
-        return False
-
-    @api.multi
     def _prepare_domain_available_not_reserved(self):
         domain_quant = [
             ('product_id', 'in', self.ids),
-            ('contains_unreserved', '=', True),
         ]
         domain_quant_locations = self._get_domain_locations()[0]
         domain_quant.extend(domain_quant_locations)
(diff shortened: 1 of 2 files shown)


jgrandguillaume (dismissed):
Oups Travis needs some love... Strange the test supposed to pass fail :(

stock_available_unreserved/models/product.py
```diff
@@ -56,16 +57,10 @@ class ProductProduct(models.Model):
...
-        """Hook used to introduce possible variations"""
-        return False
-
     @api.multi
     def _prepare_domain_available_not_reserved(self):
         domain_quant = [
             ('product_id', 'in', self.ids),
-            ('contains_unreserved', '=', True),
         ]
         domain_quant_locations = self._get_domain_locations()[0]
```
jgrandguillaume, line 65: May be the issue comes from here [0] remove the children ?

guewen:
Fixed the test. When run with pytest, it runs smoothly, but when run with `--test-enable`, we have to explicitly update the parent store of the new locations with `cls.env['stock.location']._parent_store_compute()`

JordiBForgeFlow:
@guewen Ohh I recall that. I also applied it in ddmrp https://github.com/Eficent/ddmrp/blob/9.0/ddmrp/tests/test_ddmrp.py#L81

jgrandguillaume (approved):
LGTM !

JordiBForgeFlow (approved):
:+1: Tested on runbot.

JordiBForgeFlow:
@guewen Thanks for the fix! I am now merging this one as we have 2 approvers and thorough testing

JordiBForgeFlow:
Oh @jgrandguillaume @max3903 and @guewen related to the topic of reservations, be very careful with module procurement_jit, because it makes automatic stock reservations. And ddmrp will not take into account reserved inventory for the purposes of computing the net flow equation. 

Users of ddmrp should not use procurement_jit and should make manual reservations just before shipping, because the buffers are the ones responsible to make sure that inventory is going to be available.

stock_available_unreserved/models/product.py
guewen: For the record, `_get_domain_locations` generate a list of domain fragments for each product in the recordset, domain which contains the filter for the locations. But the domain is the same for each product as they have the same context, so only one on them is required.

max3903:
@jbeficent procurement_jit is not installed ;)

rousseldenis:
@guewen @jbeficent @jgrandguillaume @max3903 I didn't follow the conversation but @Cedric-Pigeon has already done improvements in v10 for every stock_available modules as he has a project with very huge data in stock. See https://github.com/OCA/stock-logistics-warehouse/pull/305

@Cedric-Pigeon you've already analyzed such processes, so could you give a review ?

PR #403, opened 2018-04-04
[11.0][fix] stock_request and stock_request_purchase for multicompany environments

JordiBForgeFlow:
Ensures consistency across all models company-wise.

### DIFF
--- a/stock_request_purchase/models/purchase_order_line.py
+++ b/stock_request_purchase/models/purchase_order_line.py
@@ -1,7 +1,8 @@
 # Copyright 2017 Eficent Business and IT Consulting Services, S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
-from odoo import api, fields, models
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
 
 
 class PurchaseOrderLine(models.Model):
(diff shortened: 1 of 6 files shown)


JordiBForgeFlow:
cc @etobella @mreficent

MiquelRForgeFlow (approved):

stock_request_purchase/models/purchase_order_line.py
```diff
@@ -10,6 +11,16 @@ class PurchaseOrderLine(models.Model):
     stock_request_ids = fields.Many2many(comodel_name='stock.request',
                                          string='Stock Requests', copy=False)
 
+    @api.constrains('stock_request_ids')
+    def _check_purchase_company_constrains(self):
+        for rec in self:
+            for request in rec.stock_request_ids:
```
Garamotte: You can replace the two for loops by a single loop: `for request in self.mapped('stock_request_ids'):`.

stock_request_purchase/models/stock_request.py
```diff
@@ -17,6 +18,15 @@ class StockRequest(models.Model):
                                          string='Purchase Order Lines',
                                          readonly=True, copy=False)
 
+    @api.constrains('purchase_line_ids')
+    def _check_purchase_company_constrains(self):
+        for rec in self:
+            for line in rec.purchase_line_ids:
```
Garamotte: Same about the two for loops.

stock_request_purchase/models/stock_request.py
```diff
@@ -17,6 +18,15 @@ class StockRequest(models.Model):
                                          string='Purchase Order Lines',
                                          readonly=True, copy=False)
 
+    @api.constrains('purchase_line_ids')
```
MiquelRForgeFlow: should has also `'company_id'`? 🤔

stock_request_purchase/models/purchase_order_line.py
JordiBForgeFlow: But then you loose the ability to do the check 'if request.company_id != rec.company_id', because you do not have rec.

stock_request_purchase/models/stock_request.py
JordiBForgeFlow: Fixed.

JordiBForgeFlow:
Superseded by https://github.com/OCA/stock-logistics-warehouse/pull/409

PR #406, opened 2018-04-09
[IMP] stock_request: Add stock request order

etobella:
This PR creates Stock Request Orders as a set of Stock Requests done by the same user for the same location/warehouse.
It is useful when the same user must ask for several products.
Usage of Stock Request orders is defined in configuration.
Also, icon has been modified in order to be more similar to v11 icons.
@jbeficent


stock_request/models/stock_request.py
```diff
@@ -196,6 +201,61 @@ def _check_product_uom(self):
...
+            ))
+
+    @api.constrains('order_id', 'picking_policy')
+    def check_order_picking_policy(self):
+        if (
+            self.order_id and
+            self.order_id.picking_policy != self.picking_policy
+        ):
+            raise ValidationError(_(
+                'Location must be equal to the order'
```
JordiBForgeFlow: The picking policy must be equal to the order.

JordiBForgeFlow:
@aheficent @lreficent

stock_request/views/stock_request_order_views.xml
```diff
@@ -0,0 +1,126 @@
...
+                            string="Confirm" type="object"
+                            attrs="{'invisible': [('state', 'not in', ['draft'])]}"/>
+                    <button name="action_cancel" states="draft,open"
+                            type="object" string="Cancel"/>
+                    <button name="action_draft" states="cancel" type="object"
+                            string="Set to Draft"/>
+                    <field name="state" widget="statusbar"/>
+                </header>
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
```
JordiBForgeFlow, line 35: I suggest to add here a link to the stock requests associated to the order. It makes it easier to view than the pop-up window of the lines within the stock request order.

stock_request/views/stock_request_views.xml
```diff
@@ -128,16 +128,4 @@
         </field>
     </record>
```
JordiBForgeFlow, line 132: I suggest to add to the stock.request form view the order_id field, so that a user can navigate to it.

JordiBForgeFlow (changes requested):

stock_request/views/stock_request_order_views.xml
etobella: Well, the lines are useful in order to add all the elements.
Do you think that is a good option to add the button once it is confirmed and hide the table then?

stock_request/views/stock_request_views.xml
etobella: Done!

stock_request/views/stock_request_order_views.xml
JordiBForgeFlow: @etobella Agreed. I think the lines need to be there for sure. I was asking myself if additionally the button would be useful, because you have full management capabilities in stock requests (which is different from PO lines or invoice lines). So, it boils down to usability.

etobella: Done!

JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:

AaronHForgeFlow (changes requested):
cannot confirm Request Order with more than one line:
image

stock_request/views/stock_request_order_views.xml
```diff
@@ -0,0 +1,136 @@
...
+                                   groups="stock.group_stock_multi_locations"/>
+                            <field name="procurement_group_id"
+                                   groups="stock.group_adv_location"/>
+                            <field name="company_id"
+                                   groups="base.group_multi_company"
+                                   options="{'no_create': True}"/>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page name="Items" string="items">
```
AaronHForgeFlow: I suggest to use high caps for the string and low caps for the name.

stock_request/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,48 @@
...
+                    <div class="row mt16 o_settings_container" id="medical_administration">
+                        <div class="col-xs-12 col-md-6 o_setting_box">
+                            <div class="o_setting_left_pane">
+                                <field name="group_stock_request_order"/>
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label string="Enable orders"
+                                       for="group_stock_request_order"/>
+                                <div class="text-muted">
+                                    Activates Stock request orders
```
AaronHForgeFlow: Minor thing, but I suggest to use high caps either for all the words or for the first word only.

AaronHForgeFlow (approved):
:+1: It works well for orders with one and with more than one stock request.

A nice improvement for the future would be to allow to add more requests to a request order after it's approved.  Regards.

stock_request/security/ir.model.access.csv
```diff
@@ -5,9 +5,9 @@ access_stock_request_stock_user,stock.request stock user,model_stock_request,sto
...
 access_stock_request_allocation_manager,stock request allocation manager,model_stock_request_allocation,group_stock_request_manager,1,1,1,1
 access_stock_request_allocation_stock_user,stock.request.allocation stock user,model_stock_request_allocation,stock.group_stock_user,1,0,0,0
-access_stock_request_manager,stock request manager,model_stock_request,group_stock_request_manager,1,1,1,1
-access_stock_warehouse_user,stock.warehouse.user,stock.model_stock_warehouse,group_stock_request_user,1,0,0,0
 access_stock_location_user,stock.location.user,stock.model_stock_location,group_stock_request_user,1,0,0,0
 access_stock_location_request_manager,stock.location request manager,stock.model_stock_location,group_stock_request_manager,1,0,0,0
 access_procurement_rule_request_manager,procurement_rule request_manager,stock.model_procurement_rule,group_stock_request_manager,1,0,0,0
 access_procurement_rule,procurement.rule.flow,stock.model_procurement_rule,group_stock_request_user,1,0,0,0
+access_stock_request_order_user,stock request user,model_stock_request_order,stock_request.group_stock_request_user,1,1,1,
```
LoisRForgeFlow: You missed the last 0 here.

LoisRForgeFlow (commented):
Code looks good.

Functional test is good, but some remarks/questions:

* You can delete Stock Request from the SR group (even if the request are in progress or done). I would remove the option to delete lines in any state but draft.
* When trying to force the route, I had to first save and then edit again to be able to see all the routes available. * Also, all routes are displayed, this is a smaller issue but some filtering there would be more user friendly.

Nice improvement!

etobella:
The routes are filtered when you change the product, however, I have made a modification so you can only access routes available for the product, location & warehouse.

https://github.com/OCA/stock-logistics-warehouse/pull/406/files#diff-cb7ac1537b7c039cfc237796ef052a90R205

I think it is better to filter it this way.
Unlink of requests is now disabled if order is not in draft state.

LoisRForgeFlow (approved):
Great! LGTM now :+1:

stock_request/models/stock_request.py
```diff
@@ -196,26 +224,65 @@ def _check_product_uom(self):
...
-                [('warehouse_ids', 'in', self.warehouse_id.ids)])
-        parents = self.get_parents().ids
-        routes = routes.filtered(lambda r: any(
-            p.location_id.id in parents for p in r.pull_ids))
-        return routes
+    @api.constrains('order_id', 'requested_by')
+    def check_order_requested_by(self):
+        if self.order_id and self.order_id.requested_by != self.requested_by:
+            raise ValidationError(_(
+                'Requested by must be equal to the order'
```
elicoidal, line 231: must be the same as in the order
(multiple)

PR #409, opened 2018-04-11
[11.0][FIX] Number Sequence and some uppercases

MiquelRForgeFlow:
This PR fixes:

- [x]  Error with the sequence number.
- [x]  Visible texts that should be in uppercases.
- [x]  `order_id` should only be visible if `group_stock_request_order` option is enabled.
- [x]  adds more tests
- [x]  adds consistency between models company-wise

@jbeficent @etobella @lreficent

### DIFF
--- a/stock_request/__manifest__.py
+++ b/stock_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "11.0.2.0.0",
+    "version": "11.0.2.1.0",
     "license": "LGPL-3",
     "website": "https://github.com/stock-logistics-warehouse",
     "author": "Eficent, "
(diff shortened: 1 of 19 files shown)


JordiBForgeFlow:
@mreficent @etobella I have added significant changes in my last commit. Fixes the problem with the sequence,  adds consistency between models company-wise and adds more tests.

JordiBForgeFlow:
BTW the problem with the sequence was caused because you are not supposed to change directly the vals in the create method, as vals is passed by reference and changes will afect subsequent uses of vals outside of the scope of the create method. Instead use copy() to copy vals to a new variable, and then change the new variable.

JordiBForgeFlow:
@lreficent @mreficent @etobella This one is ready to review & merge.

etobella (approved):
LGTM :+1:

MiquelRForgeFlow (commented):
LGTM :+1:

PR #410, opened 2018-04-12
[11.0][IMP] stock_request: create request order from products

aisopuro:
This commit allows the user to select multiple products from
a product list view and to create a Stock Request Order with
an action. If the user selects templates, all of those templates'
variants will be added to the order.

This is useful as it allows the user to filter products beforehand
based on their stocks, and to create a request order for all
those products whose stock is too low.

### DIFF
--- a/stock_request/__manifest__.py
+++ b/stock_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "11.0.2.1.0",
+    "version": "11.0.2.2.0",
     "license": "LGPL-3",
     "website": "https://github.com/stock-logistics-warehouse",
     "author": "Eficent, "
@@ -16,6 +16,7 @@
     "data": [
         "security/stock_request_security.xml",
         "security/ir.model.access.csv",
+        "views/product.xml",
         "views/stock_request_views.xml",
         "views/stock_request_allocation_views.xml",
         "views/stock_move_views.xml",
--- a/stock_request/README.rst
+++ b/stock_request/README.rst
@@ -72,6 +72,7 @@ Contributors
 
 * Jordi Ballester (EFICENT) <jordi.ballester@eficent.com>.
 * Enric Tobella <etobella@creublanca.es>
+* Atte Isopuro <atte.isopuro@avoin.systems>
 
 Maintainer
 ----------
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -2,7 +2,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
 from odoo import api, fields, models, _
-from odoo.exceptions import UserError, ValidationError
+from odoo.exceptions import UserError, ValidationError, AccessError
 
 REQUEST_STATES = [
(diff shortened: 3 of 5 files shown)


stock_request/__manifest__.py
```diff
@@ -8,6 +8,7 @@
     "license": "LGPL-3",
     "website": "https://github.com/stock-logistics-warehouse",
     "author": "Eficent, "
+              "Avoin.Systems, "
```
JordiBForgeFlow: Please add yourself to the contributors section, not to authorship. Authorship is granted when your contribution to the module is significant. Good pratices in OCA.

JordiBForgeFlow (approved):
OK, with small comment.

stock_request/__manifest__.py
aisopuro: Whoops, sorry. Yeah, I'll fix that.

aisopuro: OK @jbeficent, fixed.

LoisRForgeFlow (approved):
LGTM :+1: 

Could you please bump the version number?

Thanks!

aisopuro:
I bumped the minor version number @lreficent .

LoisRForgeFlow:
Great, thanks @aisopuro

JordiBForgeFlow:
cc @etobella FYI

aisopuro:
Thanks for the merge @lreficent. I noticed in the commit history that the Transifex translations seem to be updated quite seldom. Is that because the translations are changed so rarely, or is the schedule that slow? I updated the Finnish translations of this repository just now, is there some specific date I can expect them to be updated to GitHub?

aisopuro:
Sorry for the ping-trash @lreficent @jbeficent @etobella but I can't seem to find any info about the Transifex update frequency by Googling. @pedrobaeza told me that pull requests should not include translations, and that they should be handled in Transifex. I have updated the finnish translation of the `stock_request`-module in Transifex around the time this PR was merged, but it seems the translations have not been merged back into this repository. What kind of schedule is there for updating the translations from Transifex to GitHub? We need to deploy this module at some point, and I would like to know if I need to manually add a `.po` file if the Transifex translation won't arrive in time.

Thanks, and sorry again for the spam.

JordiBForgeFlow:
I will ckeck..

aisopuro:
Hi again @jbeficent . It seems there is still no update to the translations in the commit history, at least not in the 11.0 branch. Is there a problem with the translation sync, or am I looking in the wrong place?

aisopuro:
Hi again @jbeficent . I still can't see any new translations in the 11.0 branch. Is this some kind of bigger problem with the Transifex sync? Is there some place where I could track the problem?

JordiBForgeFlow:
Sorry, I will try to resolve ASAP.

JordiBForgeFlow:
@aisopuro Transifex has been applying some limitations to the number of translations being processed and this might be causing this issue. OCA is transitioning to a new translation platform, Weblate, and we'd like to work on it during the next Sprint in Barcelona.

Today I made some translations of stock_request to spanish. Let's see if tomorrow the translation is done.

aisopuro:
It looks like nothing's come in from Transifex @jbeficent. So does this mean that translations will not be updated until OCA has transitioned to Weblate?

PR #412, opened 2018-04-16
[11.0][FIX] stock_putaway_product: return the model.

LoisRForgeFlow:
The [putaway_apply method](https://github.com/odoo/odoo/blob/11.0/addons/stock/models/product_strategy.py#L24) should return a stock.location model or record not the id.


rousseldenis (approved):
Yes!

JordiBForgeFlow (approved):

PR #413, opened 2018-04-19
Wrong MTS source location when the warehouse has two or more Stock

ougc27:
**Impacted Versions:**
10.0

**Steps to Reproduce**
- Create 2 stock locations to have products stock. 
-  Create a route with 2 rules of  MTS - MTO rules.
-  Create a product with 2 units in one location and 3 units in another location.
-  Create a Sale Order with the product created with 10 of product quantity and confirm it 
  
**Current behavior before PR:**
- The stock location routes create wrong stock pickings because in the method "get_mto_qty_to_order" always takes the quantity available of the main stock location per warehouse (self.warehouse_id.lot_stock_id.id), so if I have many stock locations with products the method returns wrong quantities.

captura realizada el 2018-04-19 08 46 58

**Desired behavior after PR is merged:**
The method must return the available quantity of the source stock location configured on the MTS rule.

Hello @pedrobaeza, What can I do to have someone review this PR?
@alan196  Could you review this code?.

Best Regards!

### DIFF
--- a/stock_mts_mto_rule/model/procurement.py
+++ b/stock_mts_mto_rule/model/procurement.py
@@ -19,7 +19,7 @@ class ProcurementOrder(models.Model):
     @api.multi
     def get_mto_qty_to_order(self):
         self.ensure_one()
-        stock_location = self.warehouse_id.lot_stock_id.id
+        stock_location = self.rule_id.mts_rule_id.location_src_id.id
         proc_warehouse = self.with_context(location=stock_location)
         virtual_available = proc_warehouse.product_id.virtual_available
         qty_available = self.product_id.uom_id._compute_quantity(
--- a/stock_mts_mto_rule/__manifest__.py
+++ b/stock_mts_mto_rule/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {'name': 'Stock MTS+MTO Rule',
- 'version': '10.0.1.0.0',
+ 'version': '10.0.1.0.1',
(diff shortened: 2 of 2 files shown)


florian-dacosta (approved):
Hello,
Thanks for the PR.
It makes sense. It has already been changed in version 11 during migration : 
https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_mts_mto_rule/model/procurement_rule.py#L35

stock_mts_mto_rule/model/procurement.py
```diff
@@ -19,7 +19,7 @@ class ProcurementOrder(models.Model):
     @api.multi
     def get_mto_qty_to_order(self):
         self.ensure_one()
-        stock_location = self.warehouse_id.lot_stock_id.id
+        stock_location = self.rule_id.mts_rule_id.location_src_id.id
```
rousseldenis, line 22: Seems coherent. Is there situation 'self.rule_id.mts_rule_id.location_src_id.id' returns False ?

Can you add a unit test for this change?

florian-dacosta: location_src_id is mandatory in view for this type of procurement rule (action=move).
So no risk to return False. If it does, the configuration is badly wrong anyway

rousseldenis: Ok. Since module integrates this in v11, this can be fast one.

rousseldenis (approved):

rousseldenis:
@ougc27 Please update module version

ougc27:
@rousseldenis the changes are done :+1:

PR #417, opened 2018-04-20
11.0 mig stock change qty reason

andreasdsp:
[MIG] stock_change_qty_reason: Migration to 11.0


pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

gurneyalex (approved):
Code review

stock_change_qty_reason/README.rst
```diff
@@ -0,0 +1,60 @@
...
+============
+
+To install this module, simply click on the installation button
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/10.0
```
pedrobaeza: 11.0

stock_change_qty_reason/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/stock-logistics-warehouse/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/stock-logistics-warehouse/issues/new?body=module:%20stock_change_qty_reason%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Use latest template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pedrobaeza (changes requested):
You can also squash together the first 4 commits from the first author, as they are all a logical group.

rousseldenis:
@andreasdsp

stock_change_qty_reason/wizard/stock_change_product_qty.py
```diff
@@ -1,11 +1,13 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016-2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+"""stock_change_product_qty"""
```
rousseldenis, line 3: Why this ?

rousseldenis (approved):
Minor change but LGTM

rousseldenis:
@pedrobaeza Can you update your review?

rousseldenis:
@pedrobaeza IMHO the transbot commits don't block this (only 2).

pedrobaeza:
I have rebased and added README by fragments. Merging when CIs finish.

pedrobaeza (approved):

JordiBForgeFlow:
@andreasdsp I do not really agree on this module. In my opinion the reasons should not be free-form, but tabulated. It should be possible to indicate the reason for the change in quantity in both the product wizard, and also when you do an inventory adjustment.

If you use stock accounting, you should be able to associate an account.account to a reason code, so that posting an inventory adjustment using a reason would post to that account instead of the default account used by the virtual location of inventory loss.

We will propose a separate module that introduces an inventory.change.reason as a separate model.

pedrobaeza:
@jbeficent this is only a migration...

JordiBForgeFlow:
@pedrobaeza OK then. We will propose soon 2 new modules:
stock_inventory_change_reason
stock_account_inventory_change_reason

pedrobaeza:
But if they are overlapping features, then it shouldn't be both. Why don't you expand this one allowing to have predefined texts? You have 2 fields: the many2one and the text, filling the text one with the name of the many2one record and making the many2one not required.

About `stock_account_inventory_change_reason`, I don't get the goal.

JordiBForgeFlow:
@pedrobaeza The goal is that, for example, an inventory adjustment made with reason 'Expenses by marketing' should post the inventory loss to the account 'Marketing expenses', instead of expensing to the generic 'Inventory loss' account.

If you do not manage perpetual inventory this won't apply, of course.

JordiBForgeFlow:
> But if they are overlapping features, then it shouldn't be both. Why don't you expand this one allowing to have predefined texts? You have 2 fields: the many2one and the text, filling the text one with the name of the many2one record and making the many2one not required.
> 
> About `stock_account_inventory_change_reason`, I don't get the goal.

Putting an encoded reason is important, because it allows the organization to control why people make changes to inventory. It is a best practice to have a catalog of reason codes:

Microsoft: https://docs.microsoft.com/en-us/dynamics365/unified-operations/supply-chain/warehousing/reason-codes-for-counting-journals

SAP: https://help.sap.com/SAPhelp_47x200/helpdata/en/68/f51f8a12692a4d9a5c0304e4cca87a/content.htm?no_cache=true

JordiBForgeFlow:
Entering a reason as a many2one, with reasons being a separate model, allow also to add additional behaviour, such as what account.account to use as contra-account for the inventory adjustment. This is also a best practive in inventory management

pedrobaeza:
@jbeficent I'm not against the feature, but I'm talking about integrating the feature in current module the way I told.

JordiBForgeFlow:
The problem is that, as it is now, it is based on a Char, which makes it hard to extend. But we will analyze if we can find a way without making the module too complex.

pedrobaeza:
Please re-read my explanations about how to make it work with both fields.

JordiBForgeFlow:
@pedrobaeza Ok, I follow you now.

HviorForgeFlow (approved):

stock_change_qty_reason/wizard/stock_change_product_qty.py
```diff
@@ -0,0 +1,23 @@
...
+
+    reason = fields.Char('Reason',
+                         help='Type in a reason for the '
+                         'product quantity change')
+
+    @api.multi
+    def change_product_qty(self):
+        """Function to super change_product_qty"""
+        if self.reason:
+            this = self.with_context(change_quantity_reason=self.reason)
```
JordiBForgeFlow, line 21: sending the info using context is not a good idea, as I discussed. It is better to extend the stock.inventory.line with a field 'reason' and pass the value from the wizard using https://github.com/odoo/odoo/blob/11.0/addons/stock/wizard/stock_change_product_qty.py#L50

stock_change_qty_reason/models/stock_inventory_line.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockInventoryLine(models.Model):
+    """Class to inherit model stock.inventory.line"""
+    _inherit = "stock.inventory.line"
+
+    def _get_move_values(self, qty, location_id, location_dest_id, out):
```
JordiBForgeFlow, line 11: define a field 'reason' in stock.inventory.line, and fetch the value, to then send it to the stock move.

JordiBForgeFlow (changes requested):
We need an approach to indicate the justification for changes that can be updated both in the wizard 'stock.change.product.qty' and in the stock inventory adjustment. We are working on that already in https://github.com/OCA/stock-logistics-warehouse/pull/524

stock_change_qty_reason/wizard/stock_change_product_qty.py
rousseldenis: I agree despite if I did that in the original module. Bad implementation. :smiley:

JordiBForgeFlow:
@rousseldenis @gurneyalex @andreasdsp @pedrobaeza we have created the proposed PR here: #524 . I think that it's better that we review the newer PR before we merge. But happy also to merge now this one and continue with #524.

rousseldenis:
@jbeficent Ok, let's merge this and continue with #524 

Do I consider your request changes to be applied on new PR ?

LoisRForgeFlow:
@rousseldenis yes, you can review the other PR, it is also ready.

PR #418, opened 2018-04-20
11.0 mig stock inventory discrepancy

andreasdsp:


stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,81 @@
...
+    _inherit = 'stock.inventory'
+
+    INVENTORY_STATE_SELECTION = [
+        ('draft', 'Draft'),
+        ('cancel', 'Cancelled'),
+        ('confirm', 'In Progress'),
+        ('pending', 'Pending to Approve'),
+        ('done', 'Validated')]
+
+    @api.one
```
rousseldenis: Deprecated

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,81 @@
...
+        help="States of the Inventory Adjustment:\n"
+             "- Draft: Inventory not started.\n"
+             "- In Progress: Inventory in execution.\n"
+             "- Pending to Approve: Inventory have some discrepancies "
+             "greater than the predefined threshold and it's waiting for the "
+             "Control Manager approval.\n"
+             "- Validated: Inventory Approved.")
+    over_discrepancy_line_count = fields.Integer(
+        string='Number of Discrepancies Over Threshold',
+        compute=_compute_over_discrepancy_line_count,
```
rousseldenis: You can use ' to surround function name

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,81 @@
...
+        store=True)
+
+    @api.model
```
rousseldenis: This is not api.model as self is used

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,81 @@
...
+        if grp_inv_val in self.env.user.groups_id:
+            return True
+        else:
+            raise UserError(
+                _('The Qty Update is over the Discrepancy Threshold.\n '
+                  'Please, contact a user with rights to perform '
+                  'this action.')
+            )
+
+    @api.one
```
rousseldenis: Deprecated

stock_inventory_discrepancy/models/stock_inventory_line.py
```diff
@@ -0,0 +1,55 @@
...
+"""stock_inventory_line"""
+
+from odoo import api, fields, models
+
+
+class StockInventoryLine(models.Model):
+    """Class to inherit stock.inventory.line"""
+    _inherit = 'stock.inventory.line'
+
+    @api.multi
```
rousseldenis: It's better to declare a depends here with the good fields

stock_inventory_discrepancy/models/stock_inventory_line.py
```diff
@@ -0,0 +1,55 @@
...
+            whs = line.location_id.get_warehouse()
+            if line.location_id.discrepancy_threshold > 0.0:
+                line.discrepancy_threshold = line.location_id.\
+                    discrepancy_threshold
+            elif whs.discrepancy_threshold > 0.0:
+                line.discrepancy_threshold = whs.discrepancy_threshold
+            else:
+                line.discrepancy_threshold = False
+
+    discrepancy_qty = fields.Float(
```
rousseldenis: Please put fields on top and declare compute functions with ''.

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

stock_inventory_discrepancy/models/stock_location.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=import-error,protected-access,too-few-public-methods
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+"""stock_location"""
```
rousseldenis: What's this ? (and in other files)

rousseldenis (changes requested):
Code review

stock_inventory_discrepancy/models/stock_location.py
andreasdsp: Thank you for the review. About the code, which one? 'pylint: disable...'? i add this because of the plugin in my script writter (atom) that automatically check the pylint error always mark 'from odoo import ...' as import-error. That is irritating. But i forget to delete it before push it. I'm sorry for that one and i'll delete it in the next push.

rousseldenis: @andreasdsp No, it was just the """stock_location""". But that's detail

rousseldenis (approved):
Code review

LoisRForgeFlow:
@andreasdsp thanks for the migration. Have a look at the history you've generated, some of the changes done in the migration are in the transifex commit https://github.com/OCA/stock-logistics-warehouse/pull/418/commits/f6a3c5d47e0fff271f2b99df8eeac86c47a83744. You probably would want to fix that.

andreasdsp:
@lreficent i already rebase and push it, maybe there is failure that i made when rebase it. what if i fixup the transifex commit into my migration commit, so it doesn't have transifex commit anymore?

LoisRForgeFlow:
I prefer if you respect history as it is, it shouldn't be difficult to fix. Let me know if you struggle or need any help.

andreasdsp:
done. i reset it and then rebase it again.

stock_inventory_discrepancy/models/__init__.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+"""Init"""
```
LoisRForgeFlow: Which is the benefit of these lines with just the title of the file?

stock_inventory_discrepancy/models/__init__.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
```
LoisRForgeFlow: remove the encoding line for every file as it is not need anymore on python3

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -1,13 +1,16 @@
 # -*- coding: utf-8 -*-
+# pylint: disable=import-error,protected-access,too-few-public-methods
```
LoisRForgeFlow: remove this

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -70,4 +78,5 @@ def action_done(self):
 
     @api.multi
     def action_force_done(self):
+        """Function to super action_force_done"""
```
LoisRForgeFlow: IMHO these lines are not very useful, it is good to add comments on the code but always to add some value, you are just saying "what" is being done which is obvious looking at the code, for me comments should be about "why" or any extra info that cannot be obtained reading the code itself.

stock_inventory_discrepancy/models/stock_location.py
```diff
@@ -2,11 +2,13 @@
...
 #   (http://www.eficent.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+"""stock_location"""
 
 from odoo import fields, models
 
 
 class StockLocation(models.Model):
+    """Class to inherit stock.location"""
     _inherit = 'stock.location'
```
LoisRForgeFlow: same for comments on classes

LoisRForgeFlow (changes requested):
Thank you for fixing the history.

I did the code review of the changes, sorry for being a little bit nitpicking. See my comments inline

andreasdsp:
Thank you for the review. I already remove unnecessary code.

stock_inventory_discrepancy/models/stock_inventory.py
LoisRForgeFlow: This can be improved to:

```
@api.multi
def action_over_discrepancies(self):
    self.write({'state': 'pending'})

```

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -57,7 +57,7 @@ def _check_group_inventory_validation_always(self):
                   'this action.')
             )
 
-    @api.one
+    @api.multi
```
LoisRForgeFlow: if you change from api.one to api.multi you have to adapt the method too.

LoisRForgeFlow (changes requested):
Some more issues I found

stock_inventory_discrepancy/README.rst
```diff
@@ -0,0 +1,82 @@
...
+  and show the quantity of lines that exceed the threshold.
+* If both WH and location thresholds are configured, the location one has
+  preference.
+* The user with "Validate All inventory Adjustments" rights can force the
+  validation of an inventory pending to approve.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/10.0
```
RoelAdriaans: Runbot link to 11.0

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,73 @@
...
+    _inherit = 'stock.inventory'
+
+    INVENTORY_STATE_SELECTION = [
+        ('draft', 'Draft'),
+        ('cancel', 'Cancelled'),
+        ('confirm', 'In Progress'),
+        ('pending', 'Pending to Approve'),
+        ('done', 'Validated')]
+
+    state = fields.Selection(
```
RoelAdriaans, line 19: Why not fields.Selection(selection_add= [...]) ?
Using the selection_add with not break functionality with other modules extending the state.

stock_inventory_discrepancy/models/stock_inventory_line.py
```diff
@@ -0,0 +1,50 @@
...
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class StockInventoryLine(models.Model):
+    _inherit = 'stock.inventory.line'
+
+    discrepancy_qty = fields.Float(
```
RoelAdriaans, line 12: Maybe use the decimal precision with the product for the qty, as used in the product qty: 
https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_inventory.py#L320 ?
Difference can be too small to see

RoelAdriaans (changes requested):

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -0,0 +1,73 @@
...
+        store=True)
+
+    @api.multi
+    @api.depends('line_ids.product_qty', 'line_ids.theoretical_qty')
+    def _compute_over_discrepancy_line_count(self):
+        for inventory in self:
+            lines = inventory.line_ids
+            inventory.over_discrepancy_line_count = sum(
+                d.discrepancy_percent > d.discrepancy_threshold
+                for d in lines)
```
gurneyalex: I'm not too fond of this use of the automatic conversion of booleans to integers, since it is not obvious at first read how this works (for a `_count` method, I expect to see a `len()` not a `sum()`)

Can this be rewritten for instance as:

```
def _compute_over_discrepancy_line_count(self):
        for inventory in self:
            lines = inventory.line_ids.filtered(
                lambda line: line.discrepancy_percent > line.discrepancy_threshold
            )
            inventory.over_discrepancy_line_count = len(lines)
```

stock_inventory_discrepancy/models/stock_inventory_line.py
gurneyalex: yes

gurneyalex (commented):
Code review

andreasdsp:
done

RoelAdriaans (approved):

stock_inventory_discrepancy/models/stock_inventory.py
LoisRForgeFlow: 2 reasons:

* With the selection_add the state added is always displayed add the end of the list on the status bar and as far as I know there is no way to modify that order without overriding the field.

image

* I added help to the field which is displayed when you hover the statusbar, which you have removed completely.

stock_inventory_discrepancy/models/stock_inventory.py
LoisRForgeFlow: again, this must be `api.multi`.

LoisRForgeFlow (changes requested):

stock_inventory_discrepancy/models/stock_inventory.py
andreasdsp: yes, i agree with that. the only way is overriding the field.

LoisRForgeFlow:
@andreasdsp Could you change [this line](https://github.com/OCA/stock-logistics-warehouse/pull/418/commits/ad1bfe55dacb66c4d527e44d017ec020f4135e8a#diff-385ded1bdbd9cfc91016ea8cd9767444R44) to `api.multi`?

Thanks!

LoisRForgeFlow (approved):
Functional and technical review :+1:

PR #420, opened 2018-04-20
Backport stock_inventory_discrepancy from 9.0 to 8.0

daramousk:


daramousk:
@NL66278 Not sure why this fails, the button is obviously there.
https://github.com/OCA/OCB/blob/8.0/addons/stock/stock_view.xml#L92

NL66278:
@daramousk The button is renamed here:
https://github.com/OCA/stock-logistics-warehouse/blob/8.0/stock_inventory_exhaustive/views/stock_inventory_view.xml#L20

Note: this is a link to the latest part of the code. In our (pinned) version of stock-logistics-warehouse, the module stock_inventory_exhaustive is called stock_inventory_location.

Renaming buttons is a terribly bad idea. You will now understand why. But seems we have to live with the fact.

I think we should update the pin on stock-logistics-warehouse and test very well everything after that still works, but otherwise we will have a growing discrepancy with the oca code.

LoisRForgeFlow:
@daramousk Don't forget to change the version number on the manifest file.

@NL66278 Indeed renaming buttons is a very bad idea, the module you pointed might need to be refactored to avoid it.

LoisRForgeFlow:
Pleas check travis, tests are failing.

daramousk:
@lreficent I fixed the issue with the tests but it seems to be crashing for different reasons now, not related to this module.

daramousk:
@hbrunn Any ideas on travis?

hbrunn:
lint checks are indeed not caused by this. You probably can speedup reviews however by fixing those two in a separate commit - people have a tendency to prefer spanding their time reviewing green PRs

The failure in the actual test running is some travis glitch, I restarted the build, we'll see what happens there.

hbrunn:
this looks better now. You might want to fix https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/397732531#L1299 however. The other test failure is unrelated.

daramousk:
@hbrunn The failure of this test doesn't have to do anything with this module. I installed the module `stock_traceability_operation` and run the tests without `stock_inventory_discrepancy` installed. The first time the tests fail, while all the other times they succeed. This is a rather odd situation which requires drilling down to the quants code which I am very happy to do but still I do not think this module is involved into this.

hbrunn:
that's what `The other test failure is unrelated.` was meant to say. The problem I linked to is the warning `WARNING openerp_test openerp.models: stock.change.product.qty.create() with unknown fields: product_tmpl_id`

daramousk:
Ah my mistake, It is fixed now.

rousseldenis:
@daramousk Please fix flake8 error
https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/400970837#L575

daramousk:
@rousseldenis Fixed the pylint issue. This next one does not seem to come from this module.

rousseldenis:
@daramousk You can try to rebase? Fix was done by @pedrobaeza here : https://github.com/OCA/OCB/commit/087733682d0dee7297b2e645db95943cf2c4e941

pedrobaeza:
Just restarting the build should be enough. I have done it. Thanks for the follow-up.

rousseldenis:
@daramousk

daramousk:
@rousseldenis I have rebased this PR but I am not sure why it still fails. 
I have been unable to:
1) reproduce the error locally
2) Find an offending rule in the testing instance (http://3340757-420-c5cb50.runbot2.odoo-community.org)
3) Trigger the error myself from the web client.

Any ideas are welcome.

rousseldenis:
@daramousk This is due to security update in orm. You cannot use a wizard created by another user. Fix your test. See https://github.com/OCA/stock-logistics-warehouse/commit/02e9084a95d80630fb4cc5986fdcfba60b17a611

daramousk:
@rousseldenis Thanks, this looks fine now.

rousseldenis:
Ok, remove the WIP

rousseldenis (approved):
Code review

LoisRForgeFlow (approved):
Code review LGTM, let's move forward.

PR #425, opened 2018-04-25
[ADD] stock_request_kanban

etobella:
Stock Request Kanban
====================

On most companies there is products that must be purchased often but cannot be
stored as a usual product because no consumption moves are made.
Usually, they are stored as consumables or putaway rules are defined.
In both cases, reordering rules cannot be used.
This module allows to use stock request as reordering rules for this kind of
products.

It is created following the concept of lean kanban cards.

Usage
=====

Creation
--------
* Go to 'Stock Requests / Stock Requests Kanban' and create a new Kanban.
* Indicate a product, quantity and location.
* Press 'Save'.
* Print the kanban and put it in the storage of the product

Request kanban
--------------

This should be used if you want to create the kanban when the card is consumed.

* Once the product is consumed, take the card
* Go to 'Stock Requests / Order Kanban Card'
* Scan the card
* The stock request is created

Request kanban batch
--------------------

This should be used when you will store the cards and create request orders
for kanbans later.

* Once the product is consumed, take the card and store it
* Create a store request order
* Press the scan button
* Scan all the pending kanban cards

@jbeficent 
In order to simplify the code, I have modified `stock.request` and split it's code into `stock.request.abstract` and `stock.request`. Also, this module can be installed from stock_request


JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
Tested and code review. This is a great one, thanks @etobella :clap:

stock_request/models/stock_request_abstract.py
```diff
@@ -0,0 +1,197 @@
...
+        if warehouse:
+            res['warehouse_id'] = warehouse.id
+            res['location_id'] = warehouse.lot_stock_id.id
+        return res
+
+    @api.depends('product_id', 'product_uom_id', 'product_uom_qty')
+    def _compute_product_qty(self):
+        for rec in self:
+            rec.product_qty = rec.product_uom_id._compute_quantity(
+                rec.product_uom_qty, rec.product_id.uom_id)
```
rousseldenis: @etobella I haven't checked if it was changed in v11, but there is a performance problem accessing related fields of 'inherits' models. Here, use rec.product_id.product_tmpl_id.uom_id instead. And add it in the depends.

stock_request/models/stock_request_abstract.py
```diff
@@ -0,0 +1,197 @@
...
+        required=True,
+        default=lambda self: self._context.get('product_uom_id', False),
+    )
+    product_uom_qty = fields.Float(
+        'Quantity', digits=dp.get_precision('Product Unit of Measure'),
+        required=True,
+        help="Quantity, specified in the unit of measure indicated in the "
+             "request.",
+    )
+    product_qty = fields.Float(
```
rousseldenis, line 61: @etobella I think it misses decimal precision.

stock_request/models/stock_request_abstract.py
```diff
@@ -0,0 +1,197 @@
...
+    @api.depends('product_id', 'warehouse_id', 'location_id')
+    def _compute_route_ids(self):
+        for record in self:
+            routes = self.env['stock.location.route']
+            if record.product_id:
+                routes += record.product_id.mapped(
+                    'route_ids') | record.product_id.mapped(
+                    'categ_id').mapped('total_route_ids')
+            if record.warehouse_id:
+                routes |= self.env['stock.location.route'].search(
```
rousseldenis: @etobella I think for performance reasons, you should do the search on self.mapped('warehouse_id').ids before the loop.

stock_request/models/stock_request_abstract.py
```diff
@@ -0,0 +1,197 @@
...
+                _('You have to select a product unit of measure in the '
+                  'same category than the default unit '
+                  'of measure of the product'))
+
+    @api.onchange('warehouse_id')
+    def onchange_warehouse_id(self):
+        """ Finds location id for changed warehouse. """
+        res = {'domain': {}}
+        if self.warehouse_id:
+            # search with sudo because the user may not have permissions
```
rousseldenis: @etobella I don't think this is right specifically in onchange as if user has no right on warehouse, he has no right on locations, but that field is displayed...

Am I right ?

rousseldenis (changes requested):
Code review.

Please also change stock_request and stock_request_purchase version

stock_request/models/stock_request_abstract.py
etobella: Fixed

stock_request/models/stock_request_abstract.py
etobella: Well, the sudo is used when the warehouse has a company different than the location. In this case, the location is no longer visible, and we cannot check that the warehouse of the location is different and we must change it to a location of the warehouse. The only way I found to get the warehouse is to pass the sudo.

stock_request/models/stock_request_abstract.py
etobella: I am not sure, the warehouse could be different between two records. Would it be correct to use the mapped in this case?

stock_request/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "11.0.2.2.0",
+    "version": "11.0.2.3.0",
```
MiquelRForgeFlow: Not blocking, but I would put `11.0.3.0.0` instead 🤔

MiquelRForgeFlow (approved):

stock_request/models/stock_request_abstract.py
rousseldenis: Yes, of course. You have to build a dict grouped by warehouse.
And then in loop, do a dict.get(record.warehouse_id) and you get the right locations, but no more query.

etobella: I did a different approach, but I think that it also solves your concern. @rousseldenis

JordiBForgeFlow:
@rousseldenis are you good with the fixes?

mreficent:
Travis :red_circle:

stock_request/models/stock_request_abstract.py
rousseldenis: @etobella I don't think so or I missed it ?

stock_request/models/stock_request_abstract.py
rousseldenis: @etobella Ok, but it's a little bit a weird solution. In some cases, you can have many warehouses and queries are n-dependant to this number.

You would have done :
```python
routes = route_obj.search([('warehouse_ids', '=', self.mapped('warehouse_id')
routes_by_wh = {}
for warehouse in self.mapped('warehouse_id'):
    routes_by_wh.update(warehouse.id: routes.filtered(lambda r: warehouse in r.warehouse_ids))
for record in self:
    # Your code for products and categories
    routes |= routes_by_wh.get(record.warehouse_id.id, self.env['stock.location.route'].browse())
```

And that's it!

rousseldenis:
@jbeficent Not yet.

stock_request/models/stock_request_abstract.py
etobella: Sorry, I had a mistake and put the precision in another field :( I think it's ok now.

etobella:
I think that all expected changes have been applied. What do you think?
@rousseldenis @jbeficent

JordiBForgeFlow:
@etobella I'm good. cc @rousseldenis are you good?

PR #280, opened 2017-05-09
[10.0][MIG] port stock_available_mrp to 10.0

Cedric-Pigeon:
- [x] depends on https://github.com/OCA/stock-logistics-warehouse/pull/279
- [x] depends on https://github.com/OCA/stock-logistics-warehouse/pull/305
- [x] depends on https://github.com/OCA/stock-logistics-warehouse/pull/474

PR #428, opened 2018-04-26
 [MIG] stock_available_mrp: Migration to 11.0 

tschanzt:


stock_available_mrp/models/product_product.py
```diff
@@ -30,64 +20,66 @@ class ProductProduct(models.Model):
     )
 
     @api.multi
-    @api.depends('potential_qty')
-    def _immediately_usable_qty(self):
+    def _compute_available_quantities_dict(self):
         """Add the potential quantity to the quantity available to promise.
 
         This is the same implementation as for templates."""
-        super(ProductProduct, self)._immediately_usable_qty()
```
grindtildeath: to respect the inheritance, it would be better to call super here, and change the results in your for loop, because other modules might be using the same function to compute some other fields, and the inheritance would be broken.

stock_available_mrp/models/product_product.py
```diff
@@ -30,64 +20,66 @@ class ProductProduct(models.Model):
...
     @api.multi
-    @api.depends('component_ids.potential_qty')
-    def _get_potential_qty(self):
+    @api.depends('virtual_available',
+                 'component_ids.potential_qty',
+                 'component_ids.virtual_available')
+    def _compute_available_quantities(self):
+        super()._compute_available_quantities()
+
+    def _get_potential_qty(self, product):
```
grindtildeath: IMO this should be `@api.multi` without the product arg.
You can add `self.ensure_one()` if it's meant to be called only for one product.

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -196,16 +199,16 @@ def test_group_mrp_missing(self):
...
+        p2 = self.product_model.create({'name': 'Test P2', 'type': 'product'})
 
         self.create_simple_bom(p1, p2,
                                routing_id=self.ref('mrp.mrp_routing_0'))
         self.create_inventory(p2.id, 1)
 
         test_user_p1 = p1.sudo(test_user)
         # Test user doesn't have access to mrp_routing, can't compute potential
-        self.assertEqual(0, test_user_p1.potential_qty)
+        self.assertEqual(1, test_user_p1.potential_qty)
```
grindtildeath: If he can't compute the potential, shouldn't we have `0` here ? :confused:

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -284,14 +287,14 @@ def test_potential_qty(self):
             "Receiving variant 2's component should not change "
             "variant 1's potential")
         self.assertPotentialQty(
-            self.var2, 42.0,
+            self.var2, 313.0,
```
grindtildeath, line 260: Can you explain these change of 42 to 313 ? If the quantities don't change in the setup, I find it weird it's changed in the result of the test.

stock_available_mrp/models/product_template.py
```diff
@@ -1,46 +1,35 @@
...
+                 'product_variant_ids.potential_qty')
+    def _compute_available_quantities_dict(self):
         """Compute the potential as the max of all the variants's potential.
 
         We can't add the potential of variants: if they share components we
         may not be able to make all the variants.
         So we set the arbitrary rule that we can promise up to the biggest
         variant's potential.
         """
+        res = {}
```
grindtildeath: Same as on `product.product`, please respect the inheritance and call super().

grindtildeath (changes requested):
Please attend these comments and fix the error encountered on runbot trying to create a new product (might need a new PR to fix stock_available though) : 
```
  File "/.repo_requirements/odoo/odoo/fields.py", line 937, in __get__
    value = record.env.cache.get(record, self)
  File "/.repo_requirements/odoo/odoo/api.py", line 960, in get
    value = self._data[field][record.id][key]
KeyError: <odoo.api.Environment object at 0x7f8a9a05ce80>

During handling of the above exception, another exception occurred:

  File "/.repo_requirements/odoo/odoo/http.py", line 650, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 692, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 342, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 335, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 936, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 931, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 923, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/models.py", line 5091, in onchange
    record.mapped(dotname)
  File "/.repo_requirements/odoo/odoo/models.py", line 4504, in mapped
    recs = recs._mapped_func(operator.itemgetter(name))
  File "/.repo_requirements/odoo/odoo/models.py", line 4483, in _mapped_func
    vals = [func(rec) for rec in self]
  File "/.repo_requirements/odoo/odoo/models.py", line 4483, in <listcomp>
    vals = [func(rec) for rec in self]
  File "/.repo_requirements/odoo/odoo/models.py", line 4750, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/.repo_requirements/odoo/odoo/fields.py", line 943, in __get__
    self.determine_draft_value(record)
  File "/.repo_requirements/odoo/odoo/fields.py", line 1063, in determine_draft_value
    self._compute_value(record)
  File "/.repo_requirements/odoo/odoo/fields.py", line 999, in _compute_value
    getattr(records, self.compute)()
  File "/home/odoo/build/OCA/stock-logistics-warehouse/stock_available/models/product_template.py", line 28, in _compute_available_quantities
    for key, value in res[product.id].items():
KeyError: <odoo.models.NewId object at 0x7f8a9aab7ac8>
```

stock_available_mrp/tests/test_potential_qty.py
tschanzt: @grindtildeath He expects an error to raise in _get_components_needs which is caught and sets the value to 0 but in v11 the error didn't get raised anymore. I don't think that i removed any conditions, but there were some changes to mrp. We can add a manual check or maybe we can get rid of the test.

stock_available_mrp/tests/test_potential_qty.py
tschanzt: @grindtildeath The Tests depended on Demodata of product, mrp. etc which has changed or was removed since v9. So i tried to recreate the data. Here there probably is a product_qty missing. Since its Likely that the original bom line had a efficiency on it, I ditdn't try to mach it exactly.

stock_available_mrp/README.rst
```diff
@@ -0,0 +1,87 @@
...
+quantity that can be manufactured with the components immediately at hand.
+By configuration, the "Potential quantity" can be computed based on other product field.
+For example, "Potential quantity" can be the quantity that can be manufactured
+with the components available to promise.
+
+Usage
+=====
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/8.0
```
simahawk, line 21: :fire:

stock_available_mrp/README.rst
```diff
@@ -0,0 +1,87 @@
...
+* add an option (probably as a sub-module) to consider all raw materials as available if they can be bought from the suppliers in time for the manufacturing.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
```
simahawk, line 58: :fire:

stock_available_mrp/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2014 Numérigraphe SARL, Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Consider the production potential is available to promise',
+    'version': '11.0.1.0.0',
+    "author": "Numérigraphe,"
+              "Odoo Community Association (OCA)",
+    'category': 'Hidden',
```
simahawk, line 9: `'website': 'https://github.com/OCA/stock-logistics-warehouse',`

stock_available_mrp/demo/mrp_data.xml
```diff
@@ -0,0 +1,101 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <data noupdate="1">
```
simahawk: `noupdate` can stay on `odoo` tag and you can remove `data`

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,126 @@
+# Copyright 2014 Numérigraphe SARL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from collections import Counter
+
+from odoo import models, fields, api
+
```
simahawk: too many empty lines

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,126 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    # Needed for fields dependencies
+    # When self.potential_qty is compute, we want to force the ORM
+    # to compute all the components potential_qty too.
+    component_ids = fields.Many2many(
+        comodel_name='product.product',
+        compute='_get_component_ids',
```
simahawk: `_compute_component_ids`

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            component_needs = self._get_components_needs(product, bom)
+        except AccessError:
+            # If user doesn't have access to BOM
+            # he can't see potential_qty
+            component_needs = None
+        if not component_needs:
+            # The BoM has no line we can use
+            product.potential_qty = 0.0
+
```
simahawk, line 67: too many empty lines

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,126 @@
...
+        """ Compute component_ids by getting all the components for
+        this product.
+        """
+        bom_obj = self.env['mrp.bom']
+
+        bom_id = bom_obj._bom_find(product_id=self.id)
+        if bom_id:
+            bom = bom_obj.browse(bom_id)
+            for bom_component in bom_obj._bom_explode(bom, self, 1.0)[0]:
+                self.component_ids |= self.browse(bom_component['product_id'])
```
simahawk: should be @api.multi + loop and collect all the product ids and browse only once at the end.

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,553 @@
...
+        self.assertPotentialQty(
+            test_user_tmpl, 0,
+            "The bom should not be visible to non members of the bom's "
+            "company or company child of the bom's company")
+        bom.company_id = chicago_id
+        self.assertPotentialQty(
+            test_user_tmpl, 1000.0, '')
+
+    def test_group_mrp_missing(self):
+        test_user = self.env['res.users'].create({
```
simahawk: tracking_disable, no_reset_password

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,553 @@
...
+            "Wrong potential quantity in main WH location")
+        self.assertPotentialQty(
+            self.tmpl.with_context(
+                location=self.wh_ch.lot_stock_id.id),
+            313.0,
+            "Wrong potential quantity in Chicago WH location")
+
+    def test_multi_unit_recursive_bom(self):
+        # Test multi-level and multi-units BOM
+
```
simahawk: too many empty lines

stock_available_mrp/tests/test_potential_qty.py
```diff
@@ -0,0 +1,553 @@
...
+from odoo.tests.common import TransactionCase
+from odoo.osv.expression import TRUE_LEAF
+
+
+class TestPotentialQty(TransactionCase):
+    """Test the potential quantity on a product with a multi-line BoM"""
+
+    def setUp(self):
+        super(TestPotentialQty, self).setUp()
+        self.product_model = self.env["product.product"]
```
simahawk, line 14: tracking_disable

simahawk (changes requested):

stock_available_mrp/models/product_product.py
tschanzt: I think we can drop this compute function since it is currently never called since component_ids is never accessed directly. component_ids was used for an api.depends so we would update the quantity if it changed on one of the components. But since the potential_qty isn't stored we would recalculate it anyways if accessed. Do you agree or am i missing something? /cc @grindtildeath

grindtildeath: @tschanzt Seems to make sense although I fear performance could be an issue not storing `potential_qty`...
Let us have a look together on monday :wink:

tschanzt:
@grindtildeath I couldn't reproduce the error locally and with the new runbot build it seems to work.

stock_available_mrp/models/product_product.py
grindtildeath: @tschanzt ~Allright let's remove this function and `_compute_available_quantities` and we'll be good !~ :+1:

grindtildeath: @tschanzt As we've seen together, we need a relation to `potential_qty` in the `api.depends` of the compute function to ensure it's calculated correctly as recursive. So I suppose we still need to compute the components of each product to use this relation and ensure the components are computed in the right order.
IMO would be great to add a second test function based on `test_potential_qty__list` but with the creation of the components and finished product in the reverse order, to ensure we don't get the right result by a random order in the test.

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,124 @@
...
+        # Need by product (same product can be in many BOM lines/levels)
+        try:
+            component_needs = self._get_components_needs(self, bom)
+        except AccessError:
+            # If user doesn't have access to BOM
+            # he can't see potential_qty
+            component_needs = None
+        if not component_needs:
+            # The BoM has no line we can use
+            self.potential_qty = 0.0
```
grindtildeath: Should be returned and not assigned to potential_qty.

grindtildeath (changes requested):
I will open a PR on your branch to fix this :wink:

rafaelbn:
Hi @tschanzt ,

I think you should start migration from https://github.com/OCA/stock-logistics-warehouse/pull/280

Don you think so?

:thinking:

grindtildeath:
Hi @rafaelbn ,

Well, it would have been done, but when @tschanzt opened this PR, the PR you mentioned was not working and stalled without activity for months although reviews were done for months as well.
So as we didn't have the need to migrate this module on v10, we migrated it straight from v9.

rafaelbn:
Thanks @grindtildeath . I'm just trying to help. It looks that https://github.com/OCA/stock-logistics-warehouse/pull/280  should be merged ( @Cedric-Pigeon is not sure about why travis is :red_circle: ) by any @OCA/logistics-maintainers . Also @Cedric-Pigeon could be interested in reviewing this PR :smile: 

I can make a functional review here once is fixed  :smiley:

rafaelbn:
Nice functionally. @simahawk @grindtildeath maybe you would like to take a look again :smile:

rafaelbn (approved):
Tested in runbot :+1:

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from collections import Counter
+from odoo import models, fields, api
+from odoo.exceptions import AccessError, UserError
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+# Needed for fields dependencies
```
simahawk: indent this comment

stock_available_mrp/README.rst
```diff
@@ -0,0 +1,87 @@
...
+* add an option (probably as a sub-module) to consider all raw materials as available if they can be bought from the suppliers in time for the manufacturing.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
```
simahawk, line 58: we don't have this link anymore as it's hard to maintain https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

stock_available_mrp/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        We can't add the potential of variants: if they share components we
+        may not be able to make all the variants.
+        So we set the arbitrary rule that we can promise up to the biggest
+        variant's potential.
+        """
+        res = super()._compute_available_quantities_dict()
+        for tmpl in self:
+            if not tmpl.product_variant_ids:
+                continue
+            # immediately_usable_qty of the product iincludes the potential
```
simahawk: s/iincludes/includes

simahawk (approved):
still some minor glitches. Approving to not block when ready ;)

tschanzt:
@simahawk i fixed the mentioned issues.

grindtildeath (approved):

simahawk:
@tschanzt good. Squash your commits pls, then we are ready to merge when it gets green ;)

tschanzt:
@simahawk squashed and its green.

lmignon:
@tschanzt #280 includes a lot of improvement especially in terms of performance.. I think it would be necessary to complete the review of #280 and include its improvements into the current one.

lmignon (changes requested):
@tschanzt #280 includes a lot of improvement especially in terms of performance.. I think it would be necessary to complete the review of #280 and include its improvements into the current one.

rousseldenis:
@tschanzt Can you attend @lmignon comments ?

tschanzt:
@rousseldenis yeah sure.

tschanzt:
@rousseldenis Done. Its based on #280 now. Could you take a look and Review?

rafaelbn:
@tschanzt check why Travis is :red_circle:

rousseldenis:
@lmignon Could you update your review ?

rafaelbn:
I think this PR should be merged :+1: don't you think so @lmignon ?

rafaelbn:
I'm not PSC but I think we could merge as https://github.com/OCA/stock-logistics-warehouse/pull/280

cc @OCA/logistics-maintainers

rafaelbn (approved):
:+1:

rousseldenis:
@tschanzt You will squash last two commits or let it as is ?

tschanzt:
@rousseldenis the last two commits are squashed now.

stock_available_mrp/models/product_product.py
```diff
@@ -104,12 +103,20 @@ def _compute_available_quantities_dict(self):
...
-                                 components_potential_qty)
+                bom_id = product.bom_id
+                potential_qty = (bom_id.product_qty * components_potential_qty)
+
+                # We want to respect the rounding factor of the potential_qty
+                # Rounding down as we want to be pesimistic.
+                potential_qty = bom_id.product_uom_id._compute_quantity(
+                    potential_qty,
+                    product.bom_id.product_tmpl_id.uom_id,
+                    rounding_method='DOWN'
```
rousseldenis, line 118: @tschanzt Just a question about this. The standard option is 'UP'. So, why rounding down as it could be weird for users if behaviour is not the same on other places?

tschanzt: @rousseldenis The point of it is to be pesimistic about how much you can deliver. The original problem was, that we didn't respect the rounding of the uom. E.G. if you want to make a skateboard (requiring the board and 4 wheels) and you had the board and 2 wheels it would return 0.5 instead of 0 so i added this to apply the rounding, but if we round up we then get 1 as the potential quantity which isn't what we can produce.

rousseldenis: Makes sense ! So, let's go

PR #431, opened 2018-05-03
11.0 mig stock quant merge

reynaldiyosfino:


pedrobaeza:
This module has no sense now, as there's no quant fragmentation

rousseldenis:
@reynaldiyosfino Please confirm @pedrobaeza comments.

This contains modification on other modules too. Dirty.

rousseldenis:
@pedrobaeza The tests are showing that is not the case. You know why ?

pedrobaeza:
This is because you can still create several quants for same values, but this is only done on a rare concurrency case, and amended on core on next touch, so definitively not needed. I close it as this even contains other unwanted files.

Issue #430, opened 2018-05-02
stock_orderpoint_purchase_link


max3903:
https://github.com/Eficent/ddmrp/pull/60

# Stock Order Point Purchase Link

This module allows to link the stock order point and the generated PO line.

# Usage

* Create a MTS purchased product with a reordering rule
* Run the scheduler
* Check the new PO line is linked to the reordering rule

JordiBForgeFlow:
Good! I'll check asap

max3903:
@bodedra should have something for us tomorrow. Meanwhile @jbeficent can you help us here:
https://github.com/Eficent/stock-rma/pull/30 and there:
https://github.com/Eficent/stock-rma/pull/31

bodedra:
We are ready to go.

CC @max3903 @jbeficent

max3903:
Implemented in #435

PR #432, opened 2018-05-04
[ADD][11.0] stock_orderpoint_purchase_link

bodedra:
Review description: https://github.com/OCA/stock-logistics-warehouse/issues/430


max3903 (approved):

stock_orderpoint_purchase_link/models/purchase.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.multi
+    @api.depends("product_id", "product_qty", "move_dest_ids")
+    def _compute_orderpoint_id(self):
+        for rec in self:
+            orderpoint_id = rec.orderpoint_id
+            warehouse = rec.order_id.picking_type_id.warehouse_id
+            # Check warehouse configuration steps for incoming shipments
+            if warehouse and not orderpoint_id:
+                # Set orderpoint for 1 steps
+                if warehouse.reception_steps == 'one_step':
```
rousseldenis, line 19: IMHO, it's quite weird to base your computation on this as, even if this exists, it can be bypassed and configured to receive in 4 or n-steps...

As I understand, as procurement order has disappeared, you want to get back the purchase line link.

Odoo described (I don't remember where) that in their simplification of stock flow, the procurement order is 'quite' replaced directly by stock move.

So, why not storing (as Odoo don't do that for the moment, maybe in another version) on the stock move level when the orderpoint generates it?

It's what I've in mind at this moment.

bodedra: Thanks for your review. It seems https://github.com/OCA/stock-logistics-workflow/pull/425 will be good option for achieve it. Would you please review it.

CC @max3903

rousseldenis: @bodedra Done.

So, what's your plan for this?

bodedra: Thank you. @rousseldenis We are on it. If everything goes well with that module (*stock_picking_purchase_propagate*) then will close these.

CC @max3903

JordiBForgeFlow:
I think that this is the wrong approach. I have proposed https://github.com/OCA/stock-logistics-warehouse/pull/435. This PR should be closed IMHO.

Issue #433, opened 2018-05-11
[stock_putaway_product] - exception in MPS


morphace:
Odoo 11: if you try to set the min/max values from the MPS with the module installed, you will receive an exception:
```
2018-05-11 14:39:41,390 35896 ERROR odoo11_bct odoo.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 650, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/Users/markus/projects/odoo/v11/ce/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 692, in dispatch
    result = self._call_function(**self.params)
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 342, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/Users/markus/projects/odoo/v11/ce/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 335, in checked_call
    result = self.endpoint(*a, **kw)
  File "/Users/markus/projects/odoo/v11/addons/thirdparty/project_agile/web_syncer/http.py", line 9, in __call__
    result = self.method(*args, **kw)
  File "/Users/markus/projects/odoo/v11/ce/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/Users/markus/projects/odoo/v11/ce/addons/web/controllers/main.py", line 931, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/Users/markus/projects/odoo/v11/ce/addons/web/controllers/main.py", line 923, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/Users/markus/projects/odoo/v11/ce/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/Users/markus/projects/odoo/v11/ce/odoo/api.py", line 672, in call_kw_model
    result = method(recs, *args, **kwargs)
[...]

PR #434, opened 2018-05-14
[FIX] Make stock_putaway_product more robust

asaunier:
Check the current form view actually has a 'product_putaway_ids' field
before trying to manipulate it.

Related to https://github.com/OCA/stock-logistics-warehouse/issues/433


pedrobaeza (approved):
Please change module version number

asaunier:
@pedrobaeza Thanks, I have amended my commit to increment the version number.

pedrobaeza:
OK, let's fast-track it for avoiding the problem

PR #448, opened 2018-06-19
[11.0][ADD] stock_picking_procure_method: New module

chienandalu:
This module adds the posibility to set the supply method to *Apply Procurement Rules* in the pickings, which is normally set to *Take From Stock* by default.

This way, you can configure MTO flows triggered from the picking itself.

cc @Tecnativa


stock_picking_mto/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def _set_move_procure_method(self):
+        self.filtered('procure_method').mapped('move_lines').write({
+            'procure_method': self.procure_method,
+        })
+
+    procure_method = fields.Selection(
+        selection=_get_selection_procure_method,
+        compute=_compute_procure_method,
+        inverse=_set_move_procure_method,
```
pedrobaeza: This should be called `_inverse_procure_method`

stock_picking_mto/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2018 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Stock Picking MTO',
+    'version': '11.0.1.0.0',
+    "development_status": "Alpha",
```
pedrobaeza: Alpha?

stock_picking_mto/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* David Vidal <david.vidal@tecnativa.com>
+* Pedro Baeza <pedro.baeza@tecnativa.com>
```
pedrobaeza: `Pedro M. Baeza`, without emails, and with the company name and link, please :angel:

stock_picking_mto/views/stock_picking_views.xml
```diff
@@ -0,0 +1,16 @@
...
+<odoo>
+
+    <record model="ir.ui.view" id="stock_picking_form_view">
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <field name="scheduled_date" position="after">
+                <field name="procure_method"
+                       attrs="{'readonly': [('state', '!=', 'draft')], 'invisible': [('state', '!=', 'draft')]}"
+                       groups="stock.group_adv_location"/>
```
pedrobaeza: Advanced location? I think it should be the group for advanced routes

pedrobaeza (changes requested):

stock_picking_mto/views/stock_picking_views.xml
chienandalu: Is the same Odoo uses for its stock.move pair: https://github.com/odoo/odoo/blob/11.0/addons/stock/views/stock_move_views.xml#L242

In any case, is just an unfortunate permission naming: https://github.com/odoo/odoo/blob/11.0/addons/stock/security/stock_security.xml#L38 The description is more accurate: *Manage Push and Pull inventory flows*

pedrobaeza: Yeah, indeed the name is misleading, because I though it's the "Manage multiple locations (advanced)", but it's the correct one

stock_picking_mto/__manifest__.py
chienandalu: It was... anyway I'm not very convinced with this manifest entry, maybe I should just drop it

stock_picking_mto/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def _get_selection_procure_method(self):
+        return self.env['stock.move'].fields_get(
+            allfields=['procure_method'])['procure_method']['selection']
+
+    def _compute_procure_method(self):
```
rousseldenis: @chienandalu Even in pure computed method, add a depends on move_lines.procure_method to fetch the recordset before entering the loop.

stock_picking_mto/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+        for picking in self:
+            procure_method = False
+            for move in picking.move_lines:
+                if not procure_method:
+                    procure_method = move.procure_method
+                elif procure_method != move.procure_method:
```
rousseldenis: @chienandalu Why this kind of logic? If you want to get the first move procure method in picking:
```python
for picking in self:
    for move in picking.move_lines:
        picking.procure_method = move.procure_method
        break
```
As procure_method in moves is required, you won't have False values.

stock_picking_mto/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+                    procure_method = False
+                break
+            picking.procure_method = procure_method
+
+    def _inverse_procure_method(self):
+        self.filtered('procure_method').mapped('move_lines').write({
```
rousseldenis: @chienandalu Never do a write in an inverse. This will trigger a new recomputation. Simply affect recordset fields.

rousseldenis (changes requested):
Code review

stock_picking_mto/models/stock_picking.py
pedrobaeza: That's not the intention: this is going to be a tri-valued field: empty if there are different values across moves, or one of the 2 possible values if they are all the same.

stock_picking_mto/models/stock_picking.py
pedrobaeza: Do `update` instead of `write`

stock_picking_mto/models/stock_picking.py
rousseldenis: @pedrobaeza Ok, but the _get_selection_procure_method is missing the empty value.

Moreover, written like that, picking.procure_method will always have the first move value...

I would have written it like this :

```python
# ensure you have at least one move line - I'm not sure a picking could be void
for picking in self.filtered('move_lines'):
    first_move = picking.move_lines[0]
    procure_method = first_move.procure_method
    if any(move.procure_method != procure_method for move in picking.move_lines - first_move):
        procure_method = ''  # Not False, empty is '' in selection
    picking.procure_method = procure_method
```
It's clearer to read and the use of any is more efficent

pedrobaeza: @rousseldenis No, the empty value is False, and it's obtained if no moves, or if there's any discrepancy. The problem is one break that is not correctly indented.

stock_picking_mto/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def _compute_procure_method(self):
+        for picking in self:
+            procure_method = False
+            for move in picking.move_lines:
+                if not procure_method:
+                    procure_method = move.procure_method
+                elif procure_method != move.procure_method:
+                    procure_method = False
+                break
```
pedrobaeza: @chienandalu you have to indent this one more.

stock_picking_mto/models/stock_picking.py
rousseldenis: Ok understand

rousseldenis:
@chienandalu I suppose you will add some tests?

chienandalu:
@rousseldenis Sure

stock_picking_mto/models/stock_picking.py
rousseldenis: But my method is correct as if there is no moves, the value is initialized with False.

pedrobaeza: Your method is less efficient and I see it less readable.

rousseldenis: Ok, that was a little to optimize as we are used to huge pickings, so we make attention to such computations.

But another way (:innocent: ):
```python
for picking in self.filtered('move_lines'):
    procure_methods = picking.mapped('move_lines.procure_method')
    if len(procure_methods) == 1:
        picking.procure_method = procure_methods[0]
```
More readable :smiley:

rousseldenis:
@chienandalu @pedrobaeza I'm not convinced by the module name as MTO is just a use case but the module can do it for every procure_method no ?

chienandalu:
@rousseldenis Could be. What do you think, @pedrobaeza ?

stock_picking_mto/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2018 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Stock Picking MTO',
```
LoisRForgeFlow: I'm missing a summary key here and I agree that the name is a little bit misleading

stock_picking_mto/models/stock_picking.py
pedrobaeza: This is not going to work, as these aren't recordsets, and values are not "joined". You can get `['mto', 'mto', 'mto',...]`

stock_picking_procure_method/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,38 @@
...
+   resupply this warehouse* or both set.
+#. Create **WH2** setting off *Manufacture in this Warehouse* and *Purchase to
+   resupply this warehouse*. Set **WH1** as the *Resupply Warehouse*.
+#. Go to *Inventory > Configuration > Warehouse Management > Routes* and click
+   on the *Make To Order* one.
+#. Add a new *Procurement Rule* with these settings and save:
+
+   - Name: *WH1 -> WH2-MTO*
+   - Action: *Move From Another Location*
+   - Procurment Location: *WH2/Stock*
```
cubells: Typo: `Procurement`

stock_picking_procure_method/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,38 @@
...
+   - Move Supply Method: *Create Procurement*
+   - Operation Type: *WH1: Internal Transfers*
+   - Propagation of Procurement Group: *Propagate*
+   - Propagate cancel and split: `True`
+   - Warehouse to Propagate: *WH1*
+
+Now if you want to trigger a manufacture:
+
+   - Create a stockable product product with a BoM list.
+   - In the product's *Invetory > Routes section* set *Make To Order* and
```
cubells: Typo: `Inventory`

stock_picking_procure_method/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,38 @@
...
+     *Manufacture* on.
+
+Or if you want to trigger a purchase:
+
+   - Create a stockable product product with a supplier.
+   - In the product's *Invetory > Routes section* set *Make To Order* and
```
cubells: Typo: `Inventory`

stock_picking_procure_method/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module adds the posibility to set the supply method to *Apply Procurement
```
cubells: Typo: `possibility`

cubells (approved):
MInor changes

LGTM

Code review and test

chienandalu:
Thanks @cubells . Changes done

stock_picking_procure_method/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+                    break
+            picking.procure_method = procure_method
+
+    def _inverse_procure_method(self):
+        self.filtered('procure_method').mapped('move_lines').update({
+            'procure_method': self.procure_method,
+        })
+
+    procure_method = fields.Selection(
+        selection=_get_selection_procure_method,
```
rousseldenis: Please put field definition on top and use lambda to define selection and strings to define compute and inverse to allow easy override

rousseldenis (changes requested):
Minor comment.

Is this can be considered as ready ?

chienandalu:
All done and comments reviewed

stock_picking_procure_method/README.rst
```diff
@@ -0,0 +1,142 @@
...
+Now if you want to trigger a manufacture:
+
+   - Create a stockable product product with a BoM list.
+   - In the product's *Invetory > Routes section* set *Make To Order* and
+     *Manufacture* on.
+
+Or if you want to trigger a purchase:
+
+   - Create a stockable product product with a supplier.
+   - In the product's *Invetory > Routes section* set *Make To Order* and
```
pedrobaeza: s/Invetory/Inventory

stock_picking_procure_method/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+        string='Supply Method',
+        help='By default, the system will take from the stock in the source '
+             'location and passively wait for availability. The other '
+             'possibility allows you to directly create a procurement on the '
+             'source location (and thus ignore its current stock) to gather '
+             'products. If we want to chain moves and have this one to wait '
+             'for the previous, this second option should be chosen.',
+    )
+
+    def _get_selection_procure_method(self):
```
pedrobaeza: `_selection_procure_method` by guidelines

stock_picking_procure_method/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+            for move in picking.move_lines:
+                if not procure_method:
+                    procure_method = move.procure_method
+                elif procure_method != move.procure_method:
+                    procure_method = False
+                    break
+            picking.procure_method = procure_method
+
+    def _inverse_procure_method(self):
+        self.filtered('procure_method').mapped('move_lines').update({
```
pedrobaeza, line 41: Shouldn't you apply the procurement method to all lines?

rousseldenis (approved):
LGTM apart @pedrobaeza remarks. Code review

rousseldenis:
@chienandalu @pedrobaeza

stock_picking_procure_method/models/stock_picking.py
chienandalu: Am I not?

pedrobaeza: No, you are filtering first those with `procure_method` field filled.

chienandalu: self is a stock.picking object. If the picking procure_method is not set, we should not force that stock.move property and leave it as it is by default, isn't it?

pedrobaeza: I don't think so. If we are forcing this value to be set, we are setting it for the whole selection.

chienandalu: I mean, if we set the picking procure_method to False, we shouldn't set it's moves' procure_method to False (we can't anyway, since it's required in the stock.move model)

pedrobaeza: But the contrary should happen

chienandalu: There's a compute method for that

pedrobaeza: OK, I got your point.

pedrobaeza (approved):

PR #452, opened 2018-06-28
[11.0][MIG] stock_inventory_preparation_filter

xavierjimenez:
Migration of the module **stock_inventory_preparation_filter** to version 11.0.


HaraldPanten (approved):
It works as expected :+1: 

I think it's ready to merge

rousseldenis:
@xavierjimenez Travis is red and is related to this module

stock_inventory_preparation_filter/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "name": "Extended Inventory Preparation Filters",
+    "version": "11.0.1.0.0",
+    "depends": [
+        "stock",
+    ],
+    "author": "AvanzOSC,"
+              "Tecnativa,"
+              "Odoo Community Association (OCA)",
+    "category": "Inventory, Logistic, Storage",
+    "website": "http://github.com/OCA/stock-logistics-workflow",
```
sergio-teruel:
```suggestion
    "website": "http://github.com/OCA/stock-logistics-warehouse",
```

stock_inventory_preparation_filter/views/stock_inventory_view.xml
```diff
@@ -0,0 +1,44 @@
...
+                        <attribute name="attrs" />
+                    </notebook>
+                    <xpath expr="//button[@name='action_reset_product_qty']/ancestor::*[position()=1]" position="attributes">
+                        <attribute name="attrs">{'invisible':[('state','=','draft')]}</attribute>
+                    </xpath>
+                    <notebook position="before">
+                        <group>
+                            <field name="categ_ids"
+                                   nolabel="1"
+                                   attrs="{'invisible':[('filter','!=','categories')]}" />
```
sergio-teruel: Put every field in separated groups, by this way the field full all page and is visible in the same place. And add a string to group tag and attrs="{'invisible...
Example:
`<notebook position="before">`
`    <group string="Filter by categories" attrs="{'invisible':[('filter','!=','categories')]}" >`
`        <field name="categ_ids" nolabel="1"/>`
`     </group>`

sergio-teruel (changes requested):

rousseldenis:
@xavierjimenez Do you plan to achieve this PR and attend comments ?

pedrobaeza (approved):
I have fixed the failing test + attend remaining comments.

@rousseldenis @sergio-teruel can you please check?

HaraldPanten:
Hi all,

How's this PR going?

I think it's ready to merge after @pedrobaeza 's changes.

pedrobaeza:
@HaraldPanten please do a review and if correct by your part, I'll merge it.

stock_inventory_preparation_filter/models/stock_inventory.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class StockInventoryEmptyLines(models.Model):
+    _name = 'stock.inventory.line.empty'
+
+    product_code = fields.Char(
+        string='Product Code',
+        required=True,
+    )
+    product_qty = fields.Float(
```
rousseldenis, line 16: @xavierjimenez I know this is a migration but shouldn't this implement decimal precision ?

rousseldenis (changes requested):
Code review. Minor comment

stock_inventory_preparation_filter/models/stock_inventory.py
pedrobaeza: The idea behind this when we did it is to not put any kind of precision as this a "free table" to input data. Then when this is moved to the inventory lines, precision is applied.

rousseldenis: IMHO, that is against the principle of decimal precision and can be weird for user... Do you want to say that if a user enter 1.333333333 in field, precision is 2 decimals, so inventory get 1.33 but in empty lines (in notebook page 'Capture Lines') he will see 1.333333333?

For me, that's inconsistent

pedrobaeza: Well, in fact the user will see 1.33 as well because by default in screen they are presented with 2 decimals, but the rest will be stored with full precision. The idea behind this is to save as is the numbers introduced, but rip them when validating. Why? Because you can follow the trace in the DB of original values vs used values.

rousseldenis: That's inconsistent for me. I know about what is stored in DB is different from what is displayed.

But if 'Product Unit of Measure' precision is 3, you have to display 1.333 !

Tested on runbot

pedrobaeza: OK, I have put that precision as I don't have any strong feeling for defending this.

rousseldenis (approved):

rousseldenis:
@sergio-teruel Could you update your review ?

HaraldPanten (commented):
Hi @pedrobaeza @rousseldenis ,

I think there's a lack in this V11 module. I'm not sure if I missed that issue in my previous review (sorry for that). Let me give you an example.

Issue:

If you select to inventory by "selected products", the product list field is a little bit "limited" in my opinion.

In previous versions of that module, users were able to open the product list (tree view), apply several filters and then select (let's say) 250 products in just a couple of clicks.

Case of use:

Imagine that I have lots of products and I want to inventory only the ones that satisfy 3-4 filters.

I could have maybe 2k of these products. How would you manage this situation? I think that inventorying by category could be a little bit limited and tedious if I have to deal with lots of filters, don't you?

See attached screenshots. V10 vs V11.

**V10**
v10

**V11**
v11

pedrobaeza:
@HaraldPanten install https://github.com/OCA/web/tree/11.0/web_widget_many2many_tags_multi_selection

HaraldPanten (approved):
You're right,

Functional review 👍

PR #458, opened 2018-07-15
[FIX][9.0]  pylint warning and errors

gurneyalex:
and hopefully make travis happy (see https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/395945550)


stock_cycle_count/models/stock_cycle_count_rule.py
```diff
@@ -25,23 +26,23 @@ def _selection_rule_types(self):
...
+                    _('Zero confirmation rules can only have one warehouse '
+                      'assigned.')
                 )
+            if rec.rule_type == 'zero':
+                zero_rule = self.search([
+                    ('rule_type', '=', 'zero'),
+                    ('warehouse_ids', '=', rec.warehouse_ids.id)])
+                if len(zero_rule) > 1:
+                    raise UserError(
+                        _('You can only have one zero confirmation rule per '
```
elicoidal: Warehouses can only have one zero-confirmation rule

stock_inventory_exclude_sublocation/models/stock_inventory.py
```diff
@@ -35,6 +35,9 @@ def _get_inventory_lines(self, inventory):
                 domain += ' and package_id = %s'
                 args += (inventory.package_id.id,)
 
+            # disable error about SQL injection as the code here is generating
+            # a vulnerability
+            # pylint: disable = E8103
```
elicoidal, line 40: not sure why you add the comment but do not alter code

elicoidal (approved):
LGTM

stock_cycle_count/models/stock_cycle_count_rule.py
gurneyalex: I'm reluctant to break the translations here, when what I'm really doing is fixing flake8 / lint issues.

stock_inventory_exclude_sublocation/models/stock_inventory.py
gurneyalex: because without the comment I still get an error by pylint on this.

stock_cycle_count/models/stock_location.py
```diff
@@ -20,19 +20,23 @@
...
-            wh = self.env['stock.warehouse'].browse(wh_id)
-            if len(history) > wh.counts_for_accuracy_qty:
-                self.loc_accuracy = mean(history[:wh.counts_for_accuracy_qty].
-                                         mapped('inventory_accuracy'))
-            else:
-                self.loc_accuracy = mean(history.mapped('inventory_accuracy'))
+        for rec in self:
+            history = self.env['stock.inventory'].search([
+                ('location_id', '=', rec.id), ('state', '=', 'done')])
+            history = history.sorted(key=lambda r: r.write_date, reverse=True)
```
simahawk: why not `order='write_date desc'` on search?

stock_inventory_discrepancy/models/stock_inventory.py
```diff
@@ -17,13 +17,13 @@ class StockInventory(models.Model):
...
     @api.depends('line_ids.product_qty', 'line_ids.theoretical_qty')
     def _compute_over_discrepancy_line_count(self):
-        lines = self.line_ids
-        self.over_discrepancy_line_count = sum(
-            d.discrepancy_percent > d.discrepancy_threshold
-            for d in lines)
+        for rec in self:
+            lines = rec.line_ids
+            rec.over_discrepancy_line_count = sum(
+                d.discrepancy_percent > d.discrepancy_threshold
```
simahawk: unless this record has different comparison black-magic methods this sum is going to fail -> summing booleans.
Am I wrong?

stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -81,23 +83,24 @@ def _get_involved_lines_and_locations(self):
...
-    @api.one
+    @api.multi
     def action_confirm(self):
-        self.state = 'open'
-        involved_moves, involved_lines = \
-            self._get_involved_lines_and_locations()
-        self.involved_move_ids = involved_moves
-        self.involved_inv_line_ids = involved_lines
+        for rec in self:
+            rec.state = 'open'
```
simahawk, line 86: wouldn't be better a final `.write({...})`?

simahawk (changes requested):

stock_inventory_discrepancy/models/stock_inventory.py
gurneyalex: I don't know, I only fixed a warning about using  `api.one` by adding a loop on `self` in the body of the method

stock_inventory_verification_request/models/stock_slot_verification_request.py
gurneyalex: I don't think this changes a lot. Most of the time this method is going to be called on a single record.

stock_cycle_count/models/stock_location.py
gurneyalex: mainly because this is not something flake8 / pylint flagged :smile_cat: The goal of the PR was originally to remove all the errors causing a red build and as many warnings as convenient. 

But I agree this one has to be fixed, and I just pushed the fix. Thanks

stock_inventory_discrepancy/models/stock_inventory.py
gurneyalex: And I just checked: Python boolean can be promoted to integers. https://docs.python.org/2.4/lib/typesnumeric.html : "Booleans are a subtype of plain integers" and True + True == 2

simahawk: TIL tnx :smile:

simahawk (approved):

stock_inventory_discrepancy/models/stock_inventory.py
gurneyalex: not that I think it is a cool idea to rely on this, from a maintainability pov, now that I think of it. 

@simahawk https://github.com/OCA/stock-logistics-warehouse/pull/458/commits/d78e0a8efdae4601fbeb07bcf2b183cabb947d68 is ok with you?

simahawk: :+1:

LoisRForgeFlow (approved):
This looks good! Thanks a lot for taking the time! :+1:

PR #463, opened 2018-07-31
[FIX] 10.0 putaway_product : display variant on template

PierrickBrun:
The variant name does not appear on the template's tree view, it's confusing. This PR fixes it.

From what I understand, the field should be invisible if there is not a default_product_product_id.

### DIFF
--- a/stock_putaway_product/views/product.xml
+++ b/stock_putaway_product/views/product.xml
@@ -21,8 +21,8 @@
                             <field name="putaway_id"/>
                             <field name="product_tmpl_id" invisible="1"/>
                             <field name="product_product_id"
-                                   invisible="context.get('default_product_product_id', False)"
-                                   domain="[('product_tmpl_id', '=', product_tmpl_id)]"/>
+                                domain="['|', ('product_tmpl_id', '=', product_tmpl_id),
+                                    ('id', '=', context.get('default_product_tmpl_id'))]"/>
                             <field name="fixed_location_id"/>
                         </tree>
                     </field>
--- a/stock_putaway_product/views/product.xml
+++ b/stock_putaway_product/views/product.xml
@@ -21,7 +21,7 @@
                             <field name="putaway_id"/>
                             <field name="product_tmpl_id" invisible="1"/>
                             <field name="product_product_id"
-                                   invisible="context.get('default_product_product_id', True)"
+                                   invisible="context.get('default_product_product_id', False)"
                                    domain="[('product_tmpl_id', '=', product_tmpl_id)]"/>
                             <field name="fixed_location_id"/>
                         </tree>


rousseldenis:
@PierrickBrun I think this is wanted. If the form is loaded on product.template model, the product putaway is applied on itself. So, the value in the strategy is the reference to product template id.

If the form is loaded through product.product model, the filled in field is the product.product id.

I need to test it in the interface because I think the field should not appear at any time as the aim is to fill in the default value automatically either you are on product.product or product.template (eventually displaying the product name is not needed - as you know on which product you are).

rousseldenis:
@PierrickBrun I've just tested and I see why you've done this.

Indeed, if you activate product variants and you define a putaway on each, you'll get as many lines on product template form as putaway strategies you defined. So, it's easier to have variant name on lines.

But that introduces a new problem : in product variant forms, you can now choose any variant in strategies. That's unwanted behaviour. I think there is something to do with domain.

PierrickBrun:
Sorry If forgot to answer.
You got it exactly, only showing variant related putaway stategies should do it.
I'm going to look at it.

rousseldenis:
@PierrickBrun Any news on this ?

rousseldenis:
@PierrickBrun

PierrickBrun:
@rousseldenis Sorry for that delayed (!) answer, I think I have it right now. Every variant is shown on the template and only the lines about the specific variant are shown on the variant. Also if you have a line about the template, this isn't shown on the variant, I don't know if this is a problem

rousseldenis:
@PierrickBrun I relaunched travis as there was an error on 10.0 branch. I will test it.

Tonow-c2c (approved):
LGTM

Tonow-c2c:
Can you add screen-shoot for non technical people ?

tbaden:
@rousseldenis what's your call on this?

rousseldenis:
I've tested and it's weird I have a different product in the combo box (on runbot) :
image

rousseldenis:
@PierrickBrun Do you have the same behaviour ?

PierrickBrun:
@rousseldenis I have the same behaviour from the runbot, yes. I don't remember seeing it on my dev env. I'm investigating

PierrickBrun:
I've reproduced it locally and found a way to prevent it.
I've realized too that I can't display both the variant and template's line from the variant: the template's line is not listed in ProductProduct's `product_putaway_ids`.
IDK how to bypass this limitation, which will be very confusing for the user: you can create a line on the template but it disappear from your screen when you save it (but it still exists) !

rousseldenis:
@PierrickBrun I think there is still something to do here. Do you plan to do it ?

rousseldenis:
@PierrickBrun This is very old. Any plan to finish this ?

PierrickBrun:
> @PierrickBrun This is very old. Any plan to finish this ?

I don't think I will

PR #466, opened 2018-08-23
[FIX] fix a bug when iterate the data with call create

nicomacr:
Fix a bug when call the create method with data dict, in each iteration modify data dict.

### DIFF
--- a/stock_orderpoint_generator/models/orderpoint_template.py
+++ b/stock_orderpoint_generator/models/orderpoint_template.py
@@ -60,8 +60,9 @@ class OrderpointTemplate(models.Model):
             data.pop('auto_product_ids', None)
             data.pop('auto_last_generation', None)
             for product_id in product_ids:
-                data['product_id'] = product_id
-                orderpoint_model.create(data)
+                vals = data.copy()
+                vals['product_id'] = product_id
+                orderpoint_model.create(vals)
 
     @api.multi
     def create_orderpoints(self, product_ids):


rousseldenis (approved):
Makes sense.

rousseldenis:
@nicomacr Can you put maybe a stack trace you had or a little bit explain what happen.

Because data is a copy yet. And just the product_id is modified on each itearation.

nicomacr:
@rousseldenis what is happening is that when you call the create in each iteration the value "data" is overwritten somewhere in the create method and brings values ​​that then give error, like for example the followers of the models.

rousseldenis:
@nicomacr Ok, that makes sense obviously!

jjscarafia (approved):

pedrobaeza:
Do you know why Travis is red?

zaoral (approved):
+1
Thanks @nicomacr 

@pedrobaeza  It seems like this error is in 11.0 branch already, is not introduced by this change. You can see this in [travis](https://travis-ci.org/OCA/stock-logistics-warehouse/builds/419619256?utm_source=github_status&utm_medium=notification) 

Best Regards

zaoral:
Who has permissions to re run travis 9.0 branch? 
the last 9.0 run was a month ago. There is a change that this error was introduced in this month and that we do not notice?

It seems like this PR solve it https://github.com/OCA/stock-logistics-warehouse/pull/467
Thanks @rousseldenis

pedrobaeza:
Yes, it's OK, thanks.

PR #467, opened 2018-08-24
[10.0][FIX] stock_inventory_discrepancy: Launch wizard as current user

rousseldenis:
This has to be investigated but in test, user that haven't created wizard don't have access to it.

@lreficent @jbeficent


MiquelRForgeFlow (approved):

rousseldenis:
The line that launches error :
https://github.com/odoo/odoo/blob/b50958c40f943df44bd2cf9fe4d460a1438b17c5/odoo/models.py#L3187

rousseldenis:
Ok, we can merge. There is a change on read access on transient models.

https://github.com/odoo/odoo/commit/e93fa785629500e997330731d84853eb2edac299#diff-9af8b31d921dcccb7bcf230271fd3ef4

sbidoul (approved):

PR #474, opened 2018-09-03
[10.0][CHG]stock_available...: improve stock methods computation

Cedric-Pigeon:
By using Odoo core dictionnaries of Odoo in place of using record set I reduced computation time by 3 on a big database0


rousseldenis (approved):
Makes sense.
Runbot error due to other module

rousseldenis:
cc/ @acsone @tschanzt @OCA/logistics-maintainers

LoisRForgeFlow (approved):
Looks like a good improvement, thanks!

PR #475, opened 2018-09-05
[10.0] stock_quant_merge: add ability to disable it on some calls

alexis-via:
Now possible to disable it on some particular calls via the context key 'disable_stock_quant_merge'

Used to fix a compatibility issue with the module "stock_quant_packages_moving_wizard".

### DIFF
--- a/stock_quant_merge/models/stock_move.py
+++ b/stock_quant_merge/models/stock_move.py
@@ -12,5 +12,7 @@ class StockMove(models.Model):
         for move in self:
             quants = move.reserved_quant_ids
             super(StockMove, move).quants_unreserve()
-            if quants and not self._context.get('disable_stock_quant_merge'):
+            if (
+                    quants and
+                    not self.env.context.get('disable_stock_quant_merge')):
                 quants.merge_stock_quants()
--- a/stock_quant_merge/models/stock_move.py
+++ b/stock_quant_merge/models/stock_move.py
@@ -12,4 +12,5 @@ class StockMove(models.Model):
         for move in self:
             quants = move.reserved_quant_ids
             super(StockMove, move).quants_unreserve()
-            quants.merge_stock_quants()
+            if quants and not self._context.get('disable_stock_quant_merge'):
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
Fast one

stock_quant_merge/models/stock_move.py
```diff
@@ -12,4 +12,5 @@ def quants_unreserve(self):
         for move in self:
             quants = move.reserved_quant_ids
             super(StockMove, move).quants_unreserve()
-            quants.merge_stock_quants()
+            if quants and not self._context.get('disable_stock_quant_merge'):
```
sbidoul: nitpicking maybe but I think `self._context` is deprecated in favor of `self.env.context`.

Otherwise LGTM

alexis-via: I would be curious to know where you read that self._context was deprecated.

I had a look on odoo master:
% grep -ri "self._context.get(" *|wc -l
361
% grep -ri "self.env.context.get(" *|wc -l 
303

rousseldenis: @alexis-via I think @sbidoul wanted to say that using self.env.context is the good manner and the usage in OCA modules.

pedrobaeza: Well, it's not OCA itself, but Python conventions: if you put a `_` in the beginning, this means this variable/method is internal, and it might change to unexpected values. You should use public variables instead, as they should be preserved as public API.

rousseldenis: @pedrobaeza @alexis-via I've found this : https://github.com/odoo/odoo/blob/1d0940d96f3f5de45b38a9ddad9af65087dd04f3/odoo/models.py#L4836

So @sbidoul was right. Use of _context is deprecated!

alexis-via:
This one is ready to be merged.

rousseldenis:
@alexis-via Could you upgrade module version ?

PR #476, opened 2018-09-06
[9.0][ADD] module stock_location_lockdown for version 9

florian-dacosta:
Setting view type for location belonging to a warehouse, in order to avoid stock entrance is no possible, as the view type is not done for this purpose and will lead to different bugs.
So this module add a boolean field on locations, to avoid any stock entrance in this location. The final goal is to have an internal location which can't receive stock, this location is just used to create the hierarchy inside a warehouse.

@pedrobaeza If you have a better idea for the module's name, let me know.


florian-dacosta:
@sbidoul I did add a translation file in this new module, I know there is a tool to translate, but from what I understand, you can use it only once the module is merged.
As I plan to use this module in production very soon, and won't wait for it to be merged, how should I process. Is adding the po file like this ok?

rousseldenis:
Seems good to me. But to be complete, shouldn't we warn user on onchange location destination on stock move/stock picking form level ?

florian-dacosta:
@rousseldenis No, I don't think so, as it is normal to put these location on the stock moves contained in a picking. When we create the picking, we don't always know where will be the exact source or destination location the operations (`stock.pack.operation`) are used to hold this information.
Let's take an example : 

```
                                                         stock
             Asile 1                                     Aisle 2                                  Aisle 3
shelf1   shelf2   sehlf3                shelf 1 shelf 2 shelf 3                   shelf 1  shelf 2   shelf 3
```

If I have a warehouse with these hierarchy, when a custom order some products, Odoo will create a delivery order with the "stock" parent location on all stock moves.
Then, after the reservation process, the delivery order will become available and the `stock.pack.operation` will be generated, with the final location of the product reserved (one of the 9 shelves).
During the transfer, the products will be moved and there won't be any blocking, even though the `location_dest_id` of the stock moves is configured not to receive stock.

pedrobaeza:
We can call it `stock_location_lockdown`, being inline with other module in this repo and the meaning.

florian-dacosta:
Seems good to me, I'll make the change.

rousseldenis:
@florian-dacosta Sorry I was not precise. I had in mind the manual move or picking creation (through interface - not the automated one)

florian-dacosta:
@rousseldenis 
I don't know, I guess it is a possible and normal case to create a manual picking with a zone as location_destination_id
I don't see how it is different from automatic ones (coming from sale/purchase order).

sbidoul:
@florian-dacosta for new modules it's totally ok to add .po files, as there is no risk of merge conflicts with weblate commits.

alexis-via:
FYI, the starting point of this module was this odoo bug report https://github.com/odoo/odoo/issues/26679 and the answer of jco-odoo at the end of the bug report.

stock_location_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class StockLocation(models.Model):
+    _inherit = 'stock.location'
+
+    block_stock_entrance = fields.Boolean(
+        help="if this box is checked, put stock in this location won't be "
```
alexis-via: put -> putting

alexis-via (approved):

stock_location_lockdown/tests/test_block_stock_location_entrance.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
```
alexis-via: It is really copyright Eficent ?

florian-dacosta: mmh no, not really...
I'll fix it

florian-dacosta:
@rousseldenis, would you consider approve this PR so it can be merged, as it is done in version 10 already?
Thanks

rousseldenis (approved):

florian-dacosta:
How quick, thanks!

PR #477, opened 2018-09-11
[11.0][NEW] stock_secondary_unit: New module for get stock in a secondary unit

sergio-teruel:
cc @Tecnativa


rousseldenis:
cc @Tecnativa @sergio-teruel Please take a look at https://github.com/OCA/stock-logistics-warehouse/tree/10.0/packaging_uom.

There are also sale_packaging and purchase_packaging modules.

You are introducing duplicate functionalities.

If you need some explanation or if you are missing feature , please let us know. These modules have been deployed in production for several years (v7, v8, v10) in multiple environments with success.

sergio-teruel:
@rousseldenis we continue the conversation here https://github.com/OCA/product-attribute/pull/377.
Thanks

rousseldenis:
@sergio-teruel Can you maybe rebase this ?

rousseldenis:
@sergio-teruel Any news on this ?

pedrobaeza:
I think the problem is the missing `product-attribute` line in oca_dependencies.txt. I have added to see.

pedrobaeza:
I have added the repo and seems to be correct, although Travis is still running.

pedrobaeza (approved):

rafaelbn (approved):
:+1:  

@sergio-teruel we can merge after fixing travis because now is :red_circle:

pedrobaeza:
The error is in `stock_request_purchase` module, and it's not directly related to this PR, so I merge this old one.

stock_secondary_unit/models/procurement_rule.py
```diff
@@ -0,0 +1,23 @@
...
+class ProcurementRule(models.Model):
+    _inherit = 'procurement.rule'
+
+    def _get_stock_move_values(self, product_id, product_qty, product_uom,
+                               location_id, name, origin, values, group_id):
+        res = super(ProcurementRule, self)._get_stock_move_values(
+            product_id, product_qty, product_uom, location_id, name, origin,
+            values, group_id
+        )
+        if values.get('sale_line_id', False):
```
rousseldenis, line 15: @sergio-teruel I think this part of code is out of scope of this module.

Could you fix this quickly or reverting the merge of this module as it breaks 11.0 branch ?

stock_secondary_unit/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'version': '11.0.1.0.0',
+    'development_status': 'Beta',
+    'category': 'stock',
+    'website': 'https://github.com/OCA/stock-logistics-warehouse',
+    'author': 'Tecnativa, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'depends': [
+        'sale_stock',
```
rousseldenis, line 15: @sergio-teruel This should be 'stock'

stock_secondary_unit/tests/test_stock_secondary_unit.py
```diff
@@ -0,0 +1,85 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase
+
+
+class TestProductSecondaryUnit(SavepointCase):
+    at_install = False
+    post_install = True
+
+    @classmethod
+    def setUpClass(cls):
```
rousseldenis, line 11: @sergio-teruel Could you add tests on stock move flows?

rousseldenis:
> The error is in `stock_request_purchase` module, and it's not directly related to this PR, so I merge this old one.

@pedrobaeza Did you check that really? Seems weird as there was no change on that module since September 2018...

pedrobaeza:
@rousseldenis the branch is broken not by this module, but the other one mentioned already.

stock_secondary_unit/models/procurement_rule.py
pedrobaeza: This is why this depends on `sale_stock`, not only on `stock`.

rousseldenis: Nope. The only aim of this part of code is to fill in fields depending on sale_secondary_unit... That's why the code fails!

pedrobaeza: This part doesn't activate if the value is not present on the dictionary, so no possible fail here.

rousseldenis: This kind of code that depends partially on other modules that are 'possibly' installed is not safe as it is impossible to write test for them. This should be avoided at all!

rousseldenis: Moreover, if you just install sale and this module, this part of code will be triggered. See #544

pedrobaeza: I see now the problem. It's not on this line, but the next one. Fixing with a glue module.

rousseldenis:
@pedrobaeza Did you debug that ?

pedrobaeza:
We debugged it with both installed modules. That's why we didn't detect the problem in our CI. Sorry again.

PR #478, opened 2018-09-12
[11.0][MIG] stock_inventory_exclude_sublocation

LoisRForgeFlow:
Migration to v11.

Stock Inventory Exclude Sublocation
============================
This module extends the functionality of Inventory Adjustment to allow you to
exclude all the sublocations when doing an inventory adjustment for a
given location.

Sometimes we just want to make an inventory adjustment of just one shelf, or
space and forget about extra subdivisions in the location. In other cases we
do inventories of smaller locations contained in our stock, so we don't want
to count them again when doing an inventory adjustment of the parent location.
E.g. if we apply a cycle count strategy.

Usage
=====

To use this module, you simply need to:

1. Create a new inventory adjustment.
2. Select the option inventory of all products.
3. Check the box "Exclude Sublocations".


JordiBForgeFlow (approved):

MiquelRForgeFlow (approved):
LGTM

PR #479, opened 2018-09-12
[11.0][MIG] stock_inventory_lockdown

LoisRForgeFlow:
Migration to v11.


LoisRForgeFlow:
@sbidoul FYI, I ran oca-gen-addon-readme in my local for this module and I got an error due to an author with a non-ascii name ("Numérigraphe"). I don't know whether this was only a local issue or it can happen after merge.

pedrobaeza:
That's because you have installed maintainer-tools for Python 2 instead of Python 3.

LoisRForgeFlow:
@pedrobaeza You are right! Thanks. pre-generated README added.

sbidoul:
As Lucky @pedrobaeza Luke says :)

image

rousseldenis (approved):
Code review

gurneyalex (approved):
code looks good, tested ok

PR #484, opened 2018-09-19
[11.0][MIG] stock_cycle_count

LoisRForgeFlow:
Migration to v11.


stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -11,6 +10,7 @@ class StockCycleCount(models.Model):
     _name = 'stock.cycle.count'
     _description = "Stock Cycle Counts"
     _inherit = 'mail.thread'
+    _order = "id desc"
```
rousseldenis, line 13: @lreficent Is this change can break behaviours with existing data (e.g. : when migrating)?

rousseldenis (approved):
Code review.

stock_cycle_count/models/stock_cycle_count.py
LoisRForgeFlow: @rousseldenis How? 

FYI I'm currently working on a migration that includes this module, if it generates any issue I will probably see it soon.

rousseldenis: I don't know the flow exactly but sometimes changing something small as the order can lead to weird behaviours usually when manipulating multiple recordsets (as it changes recordset ordering).

LoisRForgeFlow: interesting, I will bear it in mind and try to see if anything strange happen.

rousseldenis:
@jbeficent @lreficent What's the status of this ?

LoisRForgeFlow:
@rousseldenis I think it is just fine, no problem until now.

JordiBForgeFlow (approved):

PR #486, opened 2018-09-25
[10.0][FIX] purchase_packaging: related field product_tmpl_id on purchase order line must be readonly

Cedric-Pigeon:


Cedric-Pigeon:
@rousseldenis @lmignon

rousseldenis (approved):
Fast one

bodedra (approved):
LGTM :+1:

pedrobaeza (commented):
This name will choke with other developments like purchase_product_configurator, that needs a real field. One of both should be renamed to a unique field name (like `product_tmpl_id_purchase_packaging` for example). It can be a good moment for doing it.

Cedric-Pigeon:
@pedrobaeza Sorry but your remark is out of scope. I just wanted to fix a bug, not doing a refactoring. Feel free to do it if you want.

pedrobaeza:
Well, it's very simple and without implications as it's not stored, but OK, as you wish. I'm not a user of this module, but as you are a user of both mentioned, you can have a conflict in the future. I was just giving an advice about my experience with this (I had the same problem with `sale_stock` + `sale_product_configurator` which leads to the creation of `sale_stock_product_configurator`).

Cedric-Pigeon:
@pedrobaeza Thanks for your advice but I do not even know the module you mentioned

pedrobaeza:
Your colleagues should know them (at least @sbidoul and @zakiuu), as they did the porting to OCA, but OK, I don't insist more.

rousseldenis:
@pedrobaeza You're right but this one exists from v7 (and it's not a migration). Following the conversation I had with @sergio-teruel , we'll do brainstorming on this module (and packaging_uom) on sprint days. We'll take into account

pedrobaeza:
@rousseldenis yeah, I know that you are trying to converge :smiley:, and that's another reason for taking the occasion to change this, as this module is very easy to be done (not stored, and only few references).

rousseldenis:
@pedrobaeza We'll do it for further versions. This is a bug fix for the actual version. The module is on v10 branch already.

PR #493, opened 2018-10-01
[10.0][FIX] purchase_packaging: purchase line product qty should be s…

rousseldenis:
…tored


rousseldenis:
@Cedric-Pigeon

Cedric-Pigeon (approved):

pedrobaeza:
If you have an inverse, I think you have to change the compute method for taking that store value for being consistent.

purchase_packaging/migrations/10.0.1.0.5/post-migrate.py
```diff
@@ -0,0 +1,10 @@
+import odoo
```
Cedric-Pigeon: This file should be named post-migration.py

Cedric-Pigeon (changes requested):

purchase_packaging/migrations/10.0.1.0.5/post-migrate.py
Cedric-Pigeon: Sorry post-migrate is also valid.

rousseldenis:
> If you have an inverse, I think you have to change the compute method for taking that store value for being consistent.

I re read this and there is no change in the logic of compute/inverse here. Just store the value.

I will explore travis errors

rousseldenis:
This one is greenifying!

This change is quite wanted as the product_qty field on order lines was computed one not stored. That would be annoying if we want to print some reports as the value was void.

As the tests lines amount is quite covering much cases (and are not changed here), we would think this is acceptable.

rousseldenis:
@Cedric-Pigeon Could you udpate your review ?

rousseldenis:
@lmignon

purchase_packaging/models/purchase_order_line.py
```diff
@@ -186,8 +187,20 @@ def create(self, vals):
...
 
     @api.multi
     def write(self, vals):
-        return super(PurchaseOrderLine, self).write(self.update_vals(vals))
+        res = super(PurchaseOrderLine, self).write(self.update_vals(vals))
+        # Required if we want to avoid recursion as there is nothing in
+        # environment that would help distinguish write call from a
+        # compute
+        if 'product_qty' in vals and not \
+                self.env.context.get('no_recompute_product_qty'):
```
lmignon:
```suggestion
                self.env.context.get('recomputing_product_qty'):
```

purchase_packaging/models/purchase_order_line.py
```diff
@@ -186,8 +187,20 @@ def create(self, vals):
...
+            self.with_context(
+                no_recompute_product_qty=True)._compute_product_qty()
```
lmignon:
```suggestion
                recomputing_product_qty=True)._compute_product_qty()
```

lmignon (changes requested):
@rousseldenis  What do you think about my naming proposal?

purchase_packaging/models/purchase_order_line.py
rousseldenis: Done!

rousseldenis:
@lmignon Could you update your review ? This one seems ready

purchase_packaging/migrations/10.0.1.0.8/post-migrate.py
```diff
@@ -0,0 +1,9 @@
+import odoo
+
+
+def migrate(cr, version):
+    # We recompute possible void values
+    env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+    void_purchase_lines = env['purchase.order.line'].search([
+        ('product_qty', '=', False)])
+    void_purchase_lines._compute_product_qty()
```
Cedric-Pigeon: you have to set this value in context to prevent recursivity: with_context(recomputing_product_qty=True)
see: https://github.com/OCA/stock-logistics-warehouse/pull/493/files#diff-0b094619c99f4fabbe358caff73c8626R205

sbidoul:
@rousseldenis @lmignon @Cedric-Pigeon is this one ready to merge?

lmignon (approved):

lmignon:
@sbidoul Yes it's ready.

PR #498, opened 2018-10-17
[9.0] stock_cycle_count: fix issue with nan while doing the mean calculation

JordiBForgeFlow:
Fixes problem by which 'loc_accuracy' in stock.location was being set with value NaN when 'counts_for_accuracy_qty' value of the warehouse was 0. This produced that the form view of stock location could not be opened in those cases.

So, if you do not set the parameter 'Inventories for location accuracy calculation' the accuracy of the location should be 0 by default, as no inventory is going to be considered.


rousseldenis (approved):

rousseldenis:
@jbeficent is it happens in other versions too?

LoisRForgeFlow (approved):
:+1:

PR #500, opened 2018-10-29
[FIX] stock_mts_mto_rule: fix routes to supply

umiphos:
### This MR is an effort to fix a bug
To replicate the bug, follow this steps:

1.  Create a product that it's routes are **Manufacture** and **Make to Order**(found in Inventory tab)
2. Add materials to that product:
 a. To material A, add a vendor, add **Buy** and **Make to Order**
 b. To material B, add a vendor, add **Buy** and **Make To Order + Make To Stock**
3. Create a SO with that product(the first one)

To check the bug, go to **Purchase** and you will see that one of the materials hasn't had a purchase order. 

### Desired behavior after PR is merged:
That after the test from before, there are 2 PO lines or orders, one for each product material

### Current behavior before PR:
Click on this video to see what it does
[Video about this PR](https://youtu.be/eYoisV5v7RI "Video description")


stock_mts_mto_rule/model/mrp_production.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
rousseldenis: Please remove this and put licensing following this format https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py

stock_mts_mto_rule/model/mrp_production.py
```diff
@@ -0,0 +1,35 @@
...
+        except UserError:
+            mto_route = False
+        # This new route is for taking in count this module, the rest is the same as odoo # noqa
+        mto_route |= self.env['stock.location.route'].search([
+            ('name', 'like', 'Make To Order + Make To Stock')], limit=1)
+        for move in self.move_raw_ids:
+            product = move.product_id
+            routes = product.route_ids + product.route_from_categ_ids
+            # TODO: optimize with read_group?
+            pull = self.env['procurement.rule'].search([
```
rousseldenis: The better is to do search before the loop and parse result inside.

stock_mts_mto_rule/view/warehouse.xml
```diff
@@ -6,7 +6,7 @@
         <field name="model">stock.warehouse</field>
         <field name="inherit_id" ref="stock.view_warehouse"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='delivery_steps']" position="after">
+            <xpath expr="//field[@name='partner_id']" position="after">
```
rousseldenis: Why this change ?

rousseldenis (changes requested):
Code review.

Could you please insert tests for your method change ?

stock_mts_mto_rule/model/mrp_production.py
rousseldenis: And you can remove utf-8 header

stock_mts_mto_rule/view/warehouse.xml
umiphos: if you install this all by itself, the checkbox for the field is not displayed, that's why I saw this modification as needed

umiphos:
@moylop260 can you give me your opinion about this change?
Context for the solution:
We need that this module takes the buy route if we select this mdule MTO+MTS, odoo, by default only takes MTS routes

moylop260:
I think that you should use a splitted module for mrp in order to avoid install mrp if you dont likenit

rousseldenis:
> I think that you should use a splitted module for mrp in order to avoid install mrp if you dont likenit

Agree. More coherent

stock_mts_mto_mrp/README.rst
```diff
@@ -0,0 +1,92 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+==================
+Stock MTS+MTO Rule
```
umiphos: Please, check the Reame and make improvements on how to use it, if there is a way to add a video, please do it.

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,38 @@
...
+            mto_route = False
+        # This new route is for taking in count this module, the rest is the same as odoo # noqa
+        mto_route |= self.env['stock.location.route'].search([
+            ('name', 'like', 'Make To Order + Make To Stock')], limit=1)
+        for move in self.move_raw_ids:
+            product = move.product_id
+            routes = product.route_ids + product.route_from_categ_ids
+            # TODO: optimize with read_group?
+            pull = self.env['procurement.rule'].search([
+                ('route_id', 'in', [x.id for x in routes]),
```
moylop260: routes.ids

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,38 @@
...
+                ('location_src_id', '=', move.location_id.id),
+                ('location_id', '=', move.location_dest_id.id)], limit=1)
+            if pull and (pull.procure_method == 'make_to_order'):
+                move.procure_method = pull.procure_method
+            elif not pull:  # If there is no make_to_stock rule either
+                intersection = set(mto_route.ids).intersection(
```
moylop260: `mto_route & routes`

stock_mts_mto_mrp/oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+addons-vauxoo git@github.com:OCA/stock-logistics-warehouse.git
```
moylop260: Use `stock-logistics-warehouse`

stock_mts_mto_rule/__manifest__.py
```diff
@@ -8,6 +8,7 @@
  'category': 'Warehouse',
  'summary': 'Add a MTS+MTO route',
  'depends': ['stock',
+             'mrp',
```
moylop260: What is the focus to split the modules if you still need mrp here

stock_mts_mto_rule/tests/test_mto_mts_route.py
```diff
@@ -142,6 +142,38 @@ def test_rename_warehouse(self):
         self.warehouse.name = new_warehouse_name
         self.assertEqual(new_rule_name, self.warehouse.mts_mto_rule_id.name)
 
+    def test_standard_mto_mts_buy_route(self):
```
moylop260: Shouldn't be a tests for mrp?

stock_mts_mto_mrp/README.rst
```diff
@@ -0,0 +1,72 @@
...
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+
+Contributors
+------------
+
+* Arturo Flores <umiphos@gmail.com>
```
moylop260: vx email

stock_mts_mto_rule/tests/test_mto_mts_route.py
umiphos: yes, but also I added the buy route, that's an important part from the base mto+mts, that it does not make the buy from the vendor

stock_mts_mto_rule/__manifest__.py
umiphos: my :bat:

stock_mts_mto_mrp/model/mrp_production.py
umiphos: I've so much to learn....

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+            # TODO: optimize with read_group?
+            pull = self.env['procurement.rule'].search([
+                ('route_id', 'in', [x for x in routes.ids]),
+                ('location_src_id', '=', move.location_id.id),
+                ('location_id', '=', move.location_dest_id.id)], limit=1)
+            if pull and (pull.procure_method == 'make_to_order'):
+                move.procure_method = pull.procure_method
+            elif not pull:  # If there is no make_to_stock rule either
+                intersection = mto_route & routes
+                if mto_route and intersection:
```
moylop260: Is not redundant validate `mto_route and intersection` considering that intersection has mto_route too?
In fact, you can re-use the `elif not pull` for `elif not pull and mto_route & routes`

I mean,

```python
            if pull and (pull.procure_method == 'make_to_order'):
                move.procure_method = pull.procure_method
            elif not pull and mto_route & routes:  # If there is no make_to_stock rule either
                move.procure_method = 'make_to_order'
```

stock_mts_mto_mrp/oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+stock-logistics-warehouse git@github.com:OCA/stock-logistics-warehouse.git
```
moylop260: Use just `stock-logistics-warehouse`
The URL to OCA is auto-completed

stock_mts_mto_rule/tests/test_mto_mts_route.py
```diff
@@ -142,6 +142,38 @@ def test_rename_warehouse(self):
...
+        mto_route = self.env.ref('stock.route_warehouse0_mto')
+        manufacture_route = self.env['stock.location.route'].search([
+            ('name', 'ilike', 'Manufacture')])
+        vendor = self.env['res.partner'].search([], limit=1)
+
+        product.write({'seller_ids': [(6, 0, [vendor.id])]})
+        product.route_ids = [(6, 0, [mto_route.id, manufacture_route.id])]
+
+        bom_bom = self.env.ref('product.product_product_3')
+        bom_material = self.env['mrp.bom'].create({
```
moylop260: The module `stock_mts_mto_rule` doesn't depend of `mrp` anymore

If you run the tests of this module without install another one you will raise an error because mrp.bom model is not defined.

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+    def _adjust_procure_method(self):
+        """This method is taken from odoo as is from the module mrp. This is
+        done because there was no way to interfiere in the decision making
+        about what routes to take"""
+        try:
+            mto_route = self.env['stock.warehouse']._get_mto_route()
+        except UserError:
+            mto_route = False
+        # This new route is for taking in count this module, the rest is the same as odoo # noqa
+        mto_route |= self.env['stock.location.route'].search([
```
moylop260: If there is a UserError exception then you will raise a error here of 
`False |= browse`

Better use:
```python
        try:
            mto_route = self.env['stock.warehouse']._get_mto_route()
        except UserError:
            mto_route = self.env['stock.warehouse']
```

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+            ('name', 'like', 'Make To Order + Make To Stock')], limit=1)
+        for move in self.move_raw_ids:
+            product = move.product_id
+            routes = product.route_ids + product.route_from_categ_ids
```
moylop260: | instead of + since that could be duplicated

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+            # TODO: optimize with read_group?
+            pull = self.env['procurement.rule'].search([
+                ('route_id', 'in', [x for x in routes.ids]),
```
moylop260: ('route_id', 'in', routes.ids)

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,37 @@
...
+                ('location_src_id', '=', move.location_id.id),
+                ('location_id', '=', move.location_dest_id.id)], limit=1)
+            if pull and (pull.procure_method == 'make_to_order'):
```
moylop260: If pull is a BrowseNull then you will have `pull.procure_method` with result Null
Then you don't need validate 2 times `if pull and pull.`

You can use just: `if pull.procure_method == 'make_to_order'`

stock_mts_mto_mrp/model/mrp_production.py
umiphos: thanks, should I propose(later in my free time) a fix to the odoo too?

stock_mts_mto_rule/tests/test_mto_mts_route.py
umiphos: removed everything I was editing from mts+mto module

moylop260 (approved):

moylop260:
@rousseldenis 
Could you review it, please?

stock_mts_mto_mrp/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2018 Vauxoo, www.vauxoo.com
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import model
```
rousseldenis: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#14directories

rousseldenis (changes requested):
Code review. Minor comment about directory name.

Even if the inside loop search inside the loop is limited to one record, finding a solution to retrieve all records at once is better. Have you try with huge records in DB (move_raw_ids)?

stock_mts_mto_mrp/model/mrp_production.py
```diff
@@ -0,0 +1,35 @@
...
+    def _adjust_procure_method(self):
+        """This method is taken from odoo as is from the module mrp. This is
+        done because there was no way to interfiere in the decision making
+        about what routes to take"""
+        try:
+            mto_route = self.env['stock.warehouse']._get_mto_route()
+        except UserError:
+            mto_route = self.env['stock.location.route']
+        # This new route is for taking in count this module, the rest is the same as odoo # noqa
+        mto_route |= self.env['stock.location.route'].search([
```
moylop260: The unique part that you are adding for this custom method is concatenating a new mto_route so you should chagne the method `_get_mto_route` instead.

I mean,

```python
...
def _get_mto_route(self):
	mto_route = self.env['stock.location.route'].search(
		[('name', 'like', 'Make To Order + Make To Stock')], limit=1)
	return mto_route or super()._get_mto_route()
```

moylop260: another important point, use the translation like the original method:
https://github.com/odoo/odoo/blob/717b7249d5d47421894052e8ad8dc67bc6f3603f/addons/stock/models/stock_warehouse.py#L479

umiphos: I will do(try to make) the fix so adding another another route similar to mto will work, I think that your reviews will make an point in that PR... but I will not make an enhance to their efficiency...

umiphos: I will do(try to make) the fix so adding another route similar to mto will work, I think that your reviews will make a good point in that PR... but I will not make an enhance to their efficiency(at the moment)...

umiphos: Just an update, didn't make that PR because we have deeper problems with the logic of MTS+MTO, odoo is not fit(yet) to store this method, so, after adding this fix to one of our modules we found that we got another problem:

The materials with MTS+MTO in a product aren't going to ever touch our module, so, the decision between stock and purchase is never triggered. If we added the logic that this action is triggered in MRP we enter into a loop...right now I'm out of ideas and the team is looking forward a solution.

saxomanu: Hi, What's the difference between this PR and this : https://github.com/OCA/manufacture/tree/11.0/mrp_mto_with_stock/
In my mind, with mrp_mto_with_stock the job is done, but you need to clic on "check avaibility" on a MO, and it's very hard when you have a lot of MO.
Why don't you improve mrp_mto_with_stock ?

saxomanu:
Hi, What's the difference between this PR and this : https://github.com/OCA/manufacture/tree/11.0/mrp_mto_with_stock/
In my mind, with mrp_mto_with_stock the job is done, but you need to clic on "check avaibility" on a MO, and it's very hard when you have a lot of MO.
Why don't you improve mrp_mto_with_stock ?

bodedra:
@saxomanu IMO these module https://github.com/OCA/manufacture/tree/11.0/mrp_auto_assign will reserve the raw material moves automatically at creation. If you want, you may improve that module and submit PR.

florian-dacosta:
IMO, there is problem with module stock_mts_mto_rule and mrp_mto_with_stock.
As they are 2 separated modules trying to fix the same problem, the first one for bought products or top level manufactured product and the second for raw materials of a MO...
I guess the module mrp_mto_with_stock should be rethinked, to depends on the module stock_mts_mto_rule. 
Because today, people are kind of lost and do not understand which modules install and why and they have to make 2 configuration on 1 product for the same thing.

It is more or less what you do in this PR by making another module, but
- I think it should be auto-installed
- I think it should be proposed only in version 12, as a replacement of module mrp_mto_with_stock from manufacture REPO, else it will be even worse to have one more conccurent module...
- And mainly, the present implementation of this module do not solve the problem of 'mts+mto' for raw material, as it always force make to order, which is not what a user expects at all !
So today, using stock_mto_mts_rule + mrp_mto_with_stock better answer the need even if it is more configuration.

Actually without this module raw material products configured with mts+mto rule behave like mts products. 
With this module, they behave like mto products.
But neither situation is the one we want!

Maybe this module could be kind of merged with mrp_mto_with_stock in version 12... So it would really do the job, and people would not be lost with 2 different systems to attend the same need.
But until this module does the job (really make mts+mto for raw material), it should not be here...

That said, I think making mrp_mto_with_stock depend on stock_mts_mto_rule is not an easy task and one need to study the feasability seriously...If not possible, then maybe both modules should be rethinked totally to attend to this problem! Sadly, this take a lot of time...

umiphos:
Closing this because we didn't solved all the problems in this version, in future versions we might try to find a better solution for the MTS+MTO

PR #503, opened 2018-11-02
[11.0][FIX] stock_mts_mto_rule: fix view to show check for module use 

umiphos:
This fix if due to the check is hidden because the field it's based their xpath is not in list, we are using a field that actually works out.

Without this change
image

With this change
image

### DIFF
--- a/stock_mts_mto_rule/view/warehouse.xml
+++ b/stock_mts_mto_rule/view/warehouse.xml
@@ -6,7 +6,7 @@
         <field name="model">stock.warehouse</field>
         <field name="inherit_id" ref="stock.view_warehouse"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='delivery_steps']" position="after">
+            <xpath expr="//field[@name='partner_id']" position="after">
                 <field name="mto_mts_management" />
             </xpath>
         </field>


stock_mts_mto_rule/view/warehouse.xml
```diff
@@ -6,7 +6,7 @@
         <field name="model">stock.warehouse</field>
         <field name="inherit_id" ref="stock.view_warehouse"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='delivery_steps']" position="after">
+            <xpath expr="//field[@name='partner_id']" position="after">
```
rousseldenis, line 9: @umiphos Ok for this. But to be more coherent, shouldn't have also been located on Inventory > Settings with a related field on  res.config.settings ?

umiphos: I don't follow what you are saying, also, this field is only placed inside the warehouse and nowhere else.

rousseldenis: @umiphos Ok. I see. Believed it was set on company level ('Your Company' in your picture):smile:

rousseldenis (approved):
Fast one

rousseldenis:
@umiphos Just tested on runbot. I see the field. I think this is normal behaviour. You have to activate advanced warehouse management. See:

image

PR #508, opened 2018-11-12
[MIG] stock_quant_manual_assign: Migration to 11.0

cubells:
Migration standard of module.

cc @Tecnativa

Supersede https://github.com/OCA/stock-logistics-warehouse/pull/361


cubells:
@lreficent , @rousseldenis , @pedrobaeza can you test again this?

fanha99:
Hi @cubells , in my pull request, some auto test case was removed.
Is it better to check and fix them, also?

cubells:
@fanha99 I've updated tests. I think all cases are covered.

pedrobaeza:
Works very well. Only 2 minor things:

* Change the widget of the check for being toggle widget, so that you don't have to click 2 times for selecting a line.
* If you have already selected all the quantity needed, when you press on other line the "Select" check, if turns selected, but no quantity is added, so this is weird. I think in 2 possibilities:

  * To turn to readonly the fields where quantity = 0 and parent.remaning quantity = 0, so when you select last possible line, the rest is not possible to be selected. This one is more UX friendly.
  * On the onchange of the check, go back to unchecked if the quantity is 0.

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -0,0 +1,149 @@
...
+from odoo.exceptions import UserError
+from odoo.tools.float_utils import float_compare
+
+
+class AssignManualQuants(models.TransientModel):
+    _name = 'assign.manual.quants'
+
+    @api.multi
+    @api.constrains('quants_lines')
+    def check_qty(self):
```
rousseldenis: Please prefix with '_'

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -0,0 +1,149 @@
...
+        move = self.env['stock.move'].browse(self.env.context['active_id'])
+        for record in self.filtered('quants_lines'):
+            if float_compare(record.lines_qty, move.product_qty,
+                             precision_digits=precision_digits) > 0:
+                raise UserError(
+                    _('Quantity is higher than the needed one'))
+
+    @api.depends('quants_lines', 'quants_lines.qty')
+    def _compute_qties(self):
+        move = self.env['stock.move'].browse(self.env.context['active_id'])
```
rousseldenis: I think checking active_id in a compute method is not correct

LoisRForgeFlow (approved):
Issues I saw in the other PR are solved. Thanks :+1:

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -0,0 +1,149 @@
...
+        digits=dp.get_precision('Product Unit of Measure'))
+    move_qty = fields.Float(string='Remaining qty', compute='_compute_qties',
+                            digits=dp.get_precision('Product Unit of Measure'))
+    quants_lines = fields.One2many('assign.manual.quants.lines',
+                                   'assign_wizard', string='Quants')
+
+    @api.multi
+    def assign_quants(self):
+        quant = self.env['stock.quant']
+        move = self.env['stock.move'].browse(self.env.context['active_id'])
```
rousseldenis: Try to make an attribute in the model that will be filled in with active_id (with a default_get). It will be accessible everywhere

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -0,0 +1,149 @@
...
+            for ml in move.move_line_ids:
+                ml.qty_done = ml.product_qty
+        move._recompute_state()
+        move.mapped('picking_id')._compute_state()
+        return {}
+
+    @api.model
+    def default_get(self, fields):
+        res = super(AssignManualQuants, self).default_get(fields)
+        move = self.env['stock.move'].browse(self.env.context['active_id'])
```
rousseldenis, line 79: See my remark above

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -0,0 +1,149 @@
...
+        available_quants = self.env['stock.quant'].search([
+            ('location_id', 'child_of', move.location_id.id),
+            ('product_id', '=', move.product_id.id),
+            ('quantity', '>', 0)
+        ])
+
+        quants_lines = []
+        for x in available_quants:
```
rousseldenis: Maybe 'quant' is more readable but let to your appreciation

rousseldenis (changes requested):
Code review.

Main part is about move link which can be done through a field.

stock_quant_manual_assign/wizard/assign_manual_quants.py
pedrobaeza: Well, as this is always executed on a wizard popup, is tolerable, but yes, you can add a helper field `move_id` for storing the move reference, which is filled by default with the active_id.

cubells: @rousseldenis 
If I change this line for self.move_id, I get move = stock.move() and quantities are calculated incorrectly. So I let this line as is.

pedrobaeza: @cubells the idea is that `move_id` is filled on `default_get` method (or a direct default method).

pedrobaeza:
@cubells it seems the toggle button is working even worst. Have you tried? I remember that Odoo fixed for not needing 2 clicks for using it, but it seems it's not true. We can see this together tomorrow.

cubells:
@pedrobaeza 
I've tested your two options proposed and if I change readonly attribute of selected fields in view, readonly attribute is not setted to selected in any moment.
And if I change onchange method of selected field, user has to click on checkbox twice to enter quantity.
So I leave code as is.

fanha99 (approved):
LGTM

LoisRForgeFlow:
@pedrobaeza @rousseldenis Could you guys check if this is good to go for you now?

pedrobaeza:
@rousseldenis I have attended your comments and also improve the module with:

* Better layout
* Remove active_id dependency in some computed fields
* Clean code
* Refine constraint
* Take into account if the current line is previously reserved before clicking on the button.

so all your comments have been honored. If you don't answer in a couple of days, I'll merge this.

rousseldenis (approved):
Code review. LGTM

PR #510, opened 2018-11-14
9.0 mig stock inventory preparation filter

chienandalu:
cc @Tecnativa


stock_inventory_preparation_filter/models/stock_inventory.py
```diff
@@ -46,6 +47,11 @@ def _selection_filter(self):
         res_filters.append(('empty', _('Empty list')))
         return res_filters
 
+    filter = fields.Selection(
```
rousseldenis, line 50: isn't this a reserved name ?

rousseldenis (commented):
LGTM. Just a question

stock_inventory_preparation_filter/models/stock_inventory.py
pedrobaeza: Well, yes, but blame Odoo that puts that name to the field...

rousseldenis: Ok didn't realised it was an overridden one.

rousseldenis (approved):

pedrobaeza (approved):

stock_inventory_preparation_filter/models/stock_inventory.py
chienandalu: I don't think we're shadowing the `filter` function since we're just defining an object property, don't you think?

PR #511, opened 2018-11-24
[11.0][stock_available_unreserved][FIX] change computation method to check for unreserved quantity and avoid useless long compute method.

JordiBForgeFlow:
This fix implements a better mechanism to look for quants that have unreserved quantity.
The previous approach was unnecessarily defining a compute method in the quant that takes forever to compute during a migration from v10 to v11.

### DIFF
--- a/stock_available_unreserved/models/product.py
+++ b/stock_available_unreserved/models/product.py
@@ -81,7 +81,8 @@ class ProductProduct(models.Model):
         res = {}
 
         domain_quant = self._prepare_domain_available_not_reserved()
-        quants = self.env['stock.quant'].with_context(lang=False).read_group(
+        ctx = {'lang': False, 'has_unreserved_quantity': True}
+        quants = self.env['stock.quant'].with_context(ctx).read_group(
             domain_quant,
             ['product_id', 'location_id', 'quantity', 'reserved_quantity'],
             ['product_id', 'location_id'],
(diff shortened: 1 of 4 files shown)


rousseldenis (changes requested):
Code review. LGTM. 

@jbeficent Could you update module version ?

stock_available_unreserved/models/product.py
```diff
@@ -111,12 +112,29 @@ def _compute_qty_available_not_reserved(self):
...
+            rec += (domain_quant,)
+        return rec
+
     def _search_quantity_unreserved(self, operator, value):
         if operator not in OPERATORS:
             raise UserError(_('Invalid domain operator %s') % operator)
         if not isinstance(value, (float, int)):
             raise UserError(_('Invalid domain right operand %s') % value)
-
+        if value and operator == '>' and not (
```
aisopuro, line 129: Could you add a comment going through the reasoning why this is OK? I'll admit it seems to me that this isn't true: if you have a quant with `unreserved_quantity` 5 and a `quantity` of 10, and ask "which quants have `unreserved_quantity > 7`, this will say that the quant in question fits the criteria, despite not having the requested quantity.

I think the tests don't cover this situation, could you add a test (or a section to the existing test) to cover the scenario I described? Or just explain the obvious thing I missed that makes this clearly an Ok thing to do :laughing:

stock_available_unreserved/views/stock_quant_view.xml
```diff
@@ -1,28 +1,11 @@
...
-        <field name="inherit_id" ref="stock.quant_search_view"/>
-        <field eval="10" name="priority"/>
-        <field name="arch" type="xml">
-            <field name="owner_id" position="after">
-                <field name="contains_unreserved"/>
-            </field>
-            <filter name="internal_loc" position="after">
-                <filter name='internal_unreserved'
-                        string="Internal Unreserved"
-                        domain="[('contains_unreserved','=', True), ('location_id.usage','=', 'internal')]"/>
```
aisopuro, line 16: Would it be possible to re-engineer this filter to work with these new fields, instead of removing it entirely? Or is it agreed that this filter is useless?

aisopuro (changes requested):
Code review done.

I agree with @rousseldenis that this needs a version update. I would also like to see that the version be changed `11.0.1.0.0` -> `11.0.2.0.0`.

The rationale is that I think this is a breaking change: we introduce a new field and remove an old one. If anyone has built any modification on top of this module, we need to make it as clear as possible that they do not want to just blindly update to this newest version. As such, I feel we should increment the major version.

As a corollary I think the commit message should be tagged `[FEAT]` instead of `[FIX]`.

rousseldenis:
@jbeficent

stock_available_unreserved/views/stock_quant_view.xml
bealdav: Probably useless because you may specify which quantity you search for with advanced filter. It should not blocking, I think

bealdav:
@jbeficent could you have a look here please because #526 need it to be merged ?

Thanks a lot

rousseldenis (approved):

rousseldenis:
@JordiBForgeFlow Is this still applicable ?

PR #513, opened 2018-11-28
[11.0][FIX] stock_request_kanban

AdriaGForgeFlow:
When printing the stock request kanban, in the case the name of the location contained a long word (E.g. Manufacturing) the resulting pdf file format was not looking good:
selection_003
The proposed fix is to remove the two column layout and instead use two rows to display the product code and its location:
selection_005

### DIFF
--- a/stock_request_kanban/report/stock_request_kanban_templates.xml
+++ b/stock_request_kanban/report/stock_request_kanban_templates.xml
@@ -3,49 +3,64 @@
      License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
     <template id="report_simple_label">
-        <div class="col-xs-3" style="padding:0;height:350px;">
-            <div class="col-xs-6"
-                 style="height:30%;border:2px solid black;text-align:center;vertical-align:middle;display:table;">
-                <div style="display: table-cell; vertical-align: middle;">
-                    <strong t-field="o.product_id.default_code"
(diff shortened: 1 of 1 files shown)


rousseldenis:
@ageficent Tested on runbot. This seems not to work with a long product name.

JordiBForgeFlow:
cc  @etobella

HviorForgeFlow:
@ageficent I not sure about it, but you could wrap that rows in a container class to avoid it.

rousseldenis:
Nearly.

image

rousseldenis:
@ageficent Good if we have few spaces. So, I think this is acceptable

rousseldenis (approved):

AdriaGForgeFlow:
cc @lreficent

HviorForgeFlow:
@ageficent Main issue is related with "display: table-cell" css property, once is removed text can jump lines.

LoisRForgeFlow:
It would be also nice to wrap long strings without spaces, but this is much more a corner case than the original problem that motivated the PR. I let the final word to @etobella.

etobella:
Mmmh! At the end, everyone will define how their cards should be shown (Usually, it should be customized)
The changes are good for me.

etobella (approved):

JordiBForgeFlow:
@ageficent Can you please squash the commits? We'll then merge.

LoisRForgeFlow (approved):

PR #517, opened 2018-12-21
[11.0][IMP] stock_orderpoint_manual_procurement

AdriaGForgeFlow:
Improve module stock_orderpoint_manual_procurement in order to update Responsible field in a Manufacturing Order to show the user that has originated it and avoid displaying always the Administrator.

### DIFF
--- a/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
+++ b/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
@@ -69,6 +69,7 @@ class MakeProcurementOrderpoint(models.TransientModel):
             values = item.orderpoint_id._prepare_procurement_values(item.qty)
             values['date_planned'] = fields.Datetime.to_string(
                 fields.Date.from_string(item.date_planned))
+            values['requested_uid'] = self.env.user
             # Run procurement
             try:
                 self.env['procurement.group'].run(


stock_orderpoint_manual_procurement/models/procurement.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class ProcurementRule(models.Model):
+    _inherit = 'procurement.rule'
+
+    def _prepare_mo_vals(self, product_id, product_qty,
```
LoisRForgeFlow: I just realized that this is a method added in `mrp` module. Maybe we need to move this part to a glue module, something like `mrp_orderpoint_manual_procurement`. Opinions? @jbeficent

LoisRForgeFlow (approved):

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -69,6 +69,7 @@ def make_procurement(self):
             values = item.orderpoint_id._prepare_procurement_values(item.qty)
             values['date_planned'] = fields.Datetime.to_string(
                 fields.Date.from_string(item.date_planned))
+            values['requested_uid'] = self.env.user
```
rousseldenis, line 72: @ageficent @lreficent  Is that field introduced through an OCA module ?

LoisRForgeFlow: it is not a field itself, the purpose is to have available the information of who have made the procurement. This can be used in the destination model (see for MO: https://github.com/OCA/manufacture/pull/328).

rousseldenis: Ok, wasn't clear

JordiBForgeFlow (approved):

PR #518, opened 2018-12-24
[11.0] [ADD] stock_request_non_internal

AaronHForgeFlow:
The idea of the module is to use effectively stock requests to consume products by transferring them to a location out of the company.

With the standard module we can create stock request for internal locations and no company assigned. Thus, we can use internal locations with no company assigned for consuming products using stock requests.

However, in this case. the company in the location is empty and it does not match with the company in the stock request. Thus, we can potentially create 2 stock_requests for different companies and same location. To solve this this module adds a constraint.

### DIFF
--- a/stock_request_non_internal/models/stock_request_abstract.py
+++ b/stock_request_non_internal/models/stock_request_abstract.py
@@ -1,0 +1,19 @@
+# Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class StockRequest(models.Model):
+    _inherit = "stock.request"
+
+    @api.constrains('location_id', 'company_id')
+    def check_order_company(self):
+        if not self.location_id.company_id:
+            others = self.search([('location_id', '=', self.location_id.id),
+                                  ('company_id', '!=', self.company_id.id)])
+            if others:
+                raise ValidationError(_(
+                    'There are requests for the same location but other'
+                    ' company'))
--- a/stock_request_non_internal/README.rst
+++ b/stock_request_non_internal/README.rst
@@ -6,20 +6,21 @@
 Stock Request Non Internal
 ==========================
 
-Allow to request stock for non internal locations
+Allow to request stock for internal locations and no company assigned.
+Constrains make sure to not create stock_request for the same location and
+different companies.
 
 
 Configuration
 =============
 
-There must be at least a route with a procurement rule for the wanted virtual
-location.
+There must be at least a route with a procurement rule for the wanted location
 
 
 Usage
 =====
 
-Create a stock request with no warehouse and the wanted virtual location
+Create a stock request for a location without company.
 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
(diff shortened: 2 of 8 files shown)


stock_request_non_internal/models/stock_request_abstract.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, fields, models
+
+
+class StockRequest(models.AbstractModel):
+    _inherit = "stock.request.abstract"
+
+    warehouse_id = fields.Many2one(required=False)
```
rousseldenis: Why are you changing this ?

stock_request_non_internal/models/stock_request_abstract.py
```diff
@@ -0,0 +1,44 @@
...
+    location_id = fields.Many2one(domain=[])
```
rousseldenis: I suppose that the domain in stock_request does not match your module. Why not changing to a lambda method (in stock_request) to compute it and override it here ?

stock_request_non_internal/models/stock_request_abstract.py
```diff
@@ -0,0 +1,44 @@
...
+    company_id = fields.Many2one(required=False)
```
rousseldenis: There is a default method. Why changing ?

stock_request/models/res_config_settings.py
```diff
@@ -15,3 +15,6 @@ class ResConfigSettings(models.TransientModel):
 
     module_stock_request_kanban = fields.Boolean(
         string='Stock Requests Kanban integration')
+
```
rousseldenis: Please update also the module version

stock_request_non_internal/models/stock_request_abstract.py
rousseldenis: Ok, just read the readme. The problem is that this changes the default behaviour too.
Maybe is there a more elegant way to do so? Introducing a 'type' maybe.

rousseldenis (changes requested):
Code review.

stock_request_non_internal/models/stock_request_abstract.py
AaronHForgeFlow: Good idea :+1:

stock_request_non_internal/models/stock_request_abstract.py
AaronHForgeFlow: I'll try to find a way. It is difficult because virtual locations are not linked to any company, and any warehouse. The ultimate purpose is to post expenses when moving products to those virtual locations. That is only possible if the locations are out of the company.

Maybe the same can be achieved by changing the constrains and the domains on the fields. I'll try.

stock_request_non_internal/models/stock_request_abstract.py
AaronHForgeFlow: The default is still valid, isn't it?. In the new use case we need the company to be empty because virtual locations are out of the company, no company assigned. To keep consistency stock requests should not be assigned to any company. Perhaps, the same can be achieved by changing the constrains and the domains on the fields. I'll try.

AaronHForgeFlow:
Hi @rousseldenis  Thank you for the review. After thinking this twice I think the approach I followed is not the proper one. The thing is, with the standard module we can create stock request for internal locations and no company assigned. Thus, we can use internal locations with no company assigned for consuming products using stock requests.

However, in this case. the company in the location is empty and it does not match with the company in the stock request. Thus, we can potentially create 2 stock_requests for different companies and same location. To solve this I added a constraint. That way we don't need to change the whole module.

@jbeficent

AaronHForgeFlow:
I don't think we don't need a new module for the new constraint. Moreover, as long as the warehouse is required, it is not possible to create two stock requests for the same location (without company) and different company (the stock request). Closing.

PR #519, opened 2018-12-24
[11.0] [ADD] stock_request_analytic

AaronHForgeFlow:
This means to integrate stock_request to analytics. The analytic account is passed from the stock_request to the stock moves. That way we can account for stock_requests and classify them.

### Roadmap
The integration with stock_request_purchase. Done in #520


AaronHForgeFlow:
This is ready to review. It seems there is some load issue in runbot, I'm restarting the build.

stock_request_analytic/models/stock_request_order.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2018 Creu Blanca
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, fields, models
+
+
+class StockRequestOrder(models.Model):
+    _inherit = 'stock.request.order'
+
+    analytic_count = fields.Integer(string='analytic count',
```
rousseldenis: No need of the string

stock_request_analytic/models/stock_request_order.py
```diff
@@ -0,0 +1,41 @@
...
+                                    compute='_compute_analytic_ids',
+                                    readonly=True)
+    analytic_account_ids = fields.One2many(
+        'account.analytic.account',
+        compute='_compute_analytic_ids',
+        string='Analytic Accounts',
+        readonly=True, copy=False)
```
rousseldenis: IMHO copy is not needed

stock_request_analytic/models/stock_request_order.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @api.depends('stock_request_ids')
+    def _compute_analytic_ids(self):
+        for req in self.sudo():
+            req.analytic_ids = req.stock_request_ids.mapped(
```
rousseldenis: I think this field does not exist

rousseldenis (changes requested):
Code review

LoisRForgeFlow:
@rousseldenis I attended your review and dis some minor fixes. Could you have a look?

LoisRForgeFlow (approved):
IMO this is ready :+1:

rousseldenis (approved):
Code review. @aheficent Maybe you could use README fragments?

PR #522, opened 2019-01-02
11.0 mig stock move location WIP

mpanarin:
WIP of migration and refactoring for the [old module](https://github.com/OCA/stock-logistics-warehouse/tree/6.1/stock_move_location) with additional functionality

- [x] fix known issue with reservations
- [x] improve readme info
- [ ] add docs
- [x] get tests up


stock_move_location/models/stock_inventory.py
```diff
@@ -0,0 +1,54 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class StockInventory(models.Model):
+    _inherit = "stock.inventory"
+
+    inventory_type = fields.Selection(
+        string='Type',
+        selection=lambda self: self._get_inventory_type_selection(),
```
simahawk: lambda not needed: pass a string, change its name to `_select_*`

stock_move_location/tests/test_common.py
```diff
@@ -0,0 +1,71 @@
...
+
+from odoo.tests import common
+
+
+class TestsCommon(common.SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestsCommon, cls).setUpClass()
+        cls.location_obj = cls.env["stock.location"]
```
simahawk, line 14: set tracking_disable globally as you know ;)

stock_move_location/tests/test_move_location.py
```diff
@@ -0,0 +1,42 @@
...
+            lot_id=self.lot2,
+        )
+        self.set_product_amount(
+            self.product_lots,
+            self.internal_loc_1,
+            1,
+            lot_id=self.lot3,
+        )
+
+        wizard = self._create_wizard(self.internal_loc_1, self.internal_loc_2)
```
simahawk, line 92: there's no test here...

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _get_collected_inventory_name(self):
+        sequence = self.env['ir.sequence'].next_by_code(
+            'stock.inventory.move') or '/'
+        res = "{sequence}:{location_from}:{location_to}".format(
+            sequence=sequence,
+            location_from=self.origin_location_id.display_name,
+            location_to=self.destination_location_id.display_name,
+        )
+        return res or "/"
```
simahawk: this OR is useless

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+        grouped_quants_data = quants_obj.read_group(
+            domain=domain,
+            fields=['product_id', 'lot_id'],
+            groupby=['product_id', 'lot_id'],
+            lazy=False,
+        )
+
+        product_data = []
+        for group in grouped_quants_data:
+            quants = quants_obj.search(group.get("__domain"))
```
simahawk: seems that `quants` is needed only for the sum: can't you select it straight instead of using sum+mapped?

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
+# Copyright (C) 2011 Julius Network Solutions SARL <contact@julius.fr>
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockMoveLocationWizard(models.TransientModel):
+    _name = "stock.move.location"
```
simahawk: I'd prefix or suffix this model w/ `wiz`

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+                    'lot_id': lot,
+                    'product_uom_id': product.uom_id.id,
+                    'move_location_wizard_id': self.id,
+                }
+            )
+        return product_data
+
+    def add_lines(self):
+        self.ensure_one()
+        if not self.stock_move_locaiton_line_ids:
```
simahawk: s/locaiton/location

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+            for line_val in self._get_stock_move_location_lines_values():
```
simahawk, line 185: how many records do we expect to create potentially? This can be reaaaaally heavy w/ all the writes and the computations after hand. We could do a multiple insert via SQL then trigger recompute on newly created records.

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+                'product_qty': self._get_available_quantity(wizard_line),
+                'inventory_id': inventory.id,
+                'location_id': self.origin_location_id.id,
+            })
+        return lines
+
+    def set_inventory_lines(self, inventory):
+        inventory_line_obj = self.env["stock.inventory.line"]
+        for line_vals in self._get_inventory_lines_values(inventory):
+            inventory_line_obj.create(line_vals)
```
simahawk: same

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,178 @@
...
+        We don't care about the reservations but we do care about not moving
+        more than exists."""
+        if not line.product_id:
+            return 0
+        available_qty = sum([x.quantity for x in self._get_quants(line)])
+        if available_qty < line.move_quantity:
+            return available_qty
+        return line.move_quantity
+
+    def _get_quants(self, line):
```
simahawk: again: can't you just select the sum?

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,57 @@
...
+        readonly=True,
+    )
+    product_uom_id = fields.Many2one(
+        string='Product Unit of Measure',
+        comodel_name='product.uom',
+    )
+    lot_id = fields.Many2one(
+        string='Lot/Serial Number',
+        comodel_name='stock.production.lot',
+        domain="[('product_id','=',product_id)]")
```
simahawk: for consistency move the closing parenthesis on next line

simahawk (changes requested):

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,128 @@
...
+        return self.env['stock.picking'].create({
+            'picking_type_id': self.env.ref('stock.picking_type_internal').id,
+            'location_id': self.origin_location_id.id,
+            'location_dest_id': self.destination_location_id.id,
+        })
+
+    def _create_moves(self, picking):
+        return self.stock_move_location_line_ids.create_move_lines(picking)
+
+    def action_move_location(self):
```
simahawk, line 123: `@api.multi` + ensure_one

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,128 @@
...
+        product_data = []
+        for group in self.env.cr.dictfetchall():
+            product = product_obj.browse(group.get("product_id")).exists()
+            product_data.append({
+                'product_id': product.id,
+                'move_quantity': group.get("sum"),
+                'max_quantity': group.get("sum"),
+                'origin_location_id': self.origin_location_id.id,
+                'destination_location_id': self.destination_location_id.id,
+                # cursor returns None instead of False
```
simahawk, line 174: None is fine anyway

stock_move_location/wizard/stock_move_location.py
simahawk: @mpanarin any answer to this?

stock_move_location/tests/test_common.py
```diff
@@ -0,0 +1,93 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo.tests import common
+
+
+class TestsCommon(common.SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestsCommon, cls).setUpClass()
```
Tonow-c2c: no need `TestsCommon, cls` any-more in Python3 but it's not a problem :wink:

Tonow-c2c (commented):
Just answer to @simahawk question and remove the README.rst and for me it's ok

stock_move_location/wizard/stock_move_location.py
mpanarin: To be honest - I don't know about the number of lines. That depends on the workflow of the end user. 

I minimized the amount of creates. So here - only lines of the wizards are created and nothing more. As well as i switched from inventories to pickings, which are less heavy. 

Switching to sql inserts is not worth it in my opinion anyway :\

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,130 @@
...
+        GROUP BY product_id, lot_id
+        """.format(
+            location_id=location_id,
+            company_id=company.id,
+        )
+
+    def _get_stock_move_location_lines_values(self):
+        product_obj = self.env['product.product']
+
+        # Using sql as search_group doesn't support aggregation functions
```
rousseldenis, line 162: @mpanarin You can use the read_group function. The aggregation function by default for float fields is 'sum'. See https://github.com/odoo/odoo/blob/9246375b25f4e90000bb6e578aceb8a116dfc28a/odoo/fields.py#L1182

stock_move_location/wizard/stock_move_location.py
rousseldenis: @simahawk I think this operation is not an usual one. If your stock locations are fine tuned (just some product or just one in a same location), the solution is quite acceptable

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,130 @@
...
+        location_id = self.origin_location_id.id
+        company = self.env['res.company']._company_default_get(
+            'stock.inventory',
+        )
+        return """
+        SELECT product_id, lot_id, SUM(quantity)
+        FROM stock_quant
+        WHERE location_id = {location_id} AND company_id = {company_id}
+        GROUP BY product_id, lot_id
+        """.format(
```
rousseldenis, line 154: @mpanarin If you use cr.execute, pass parameters as second argument of the function.

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,109 @@
...
+    max_quantity = fields.Float(
+        string="Maximum available quantity",
+        digits=dp.get_precision('Product Unit of Measure'),
+        readonly=True,
+    )
+
+    @api.constrains("max_quantity", "move_quantity")
+    def _contraints_max_move_quantity(self):
+        for record in self:
+            if (record.move_quantity > record.max_quantity or
```
rousseldenis: @mpanarin You have to use float_compare function

rousseldenis:
@mpanarin I'm not sure if the implementation is the right one.

I would have done something different like :

- A new parameter on stock picking types : 'Product Change Location' (with a little help).
- With this, you can go to the dashboard, create a picking with that type.
- Add a button which is visible with that type that fill in the picking as you did.

So, the flow respect the standard one and maybe is more intuitive for the user.

Moreover, you can add a magic button on locations that with context creates a new picking of that type with the origin location already filled in.

What do you think about ?

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,115 @@
...
+        digits=dp.get_precision('Product Unit of Measure'),
+        readonly=True,
+    )
+
+    @api.constrains("max_quantity", "move_quantity")
+    def _contraints_max_move_quantity(self):
+        for record in self:
+            if (float_compare(
+                    record.move_quantity,
+                    record.max_quantity, 3
```
rousseldenis: @mpanarin  You cannot hardcode rounding. As it concerns moves, I think you should use product uom rounding.

sebalix:
@rousseldenis I continue the work started by @mpanarin . It has been stated that your proposal is the good approach, but we are lacking of time to do it, so we think to keep it as it is for now, I added a roadmap with your suggestions.
I made another PR by rebasing/squashing current work: #535

rousseldenis:
@mpanarin Can we close this ?

sebalix:
@rousseldenis you can, @mpanarin is on leave for a few days :)

PR #527, opened 2019-01-09
[11.0][FIX] stock_request: solve unique constraint problems with product-attribute/p…

HviorForgeFlow:
…roduct_sequence
When these models are tested together system crash because os unique constraint:
```
odoo.addons.stock_request.tests.test_stock_request.TestStockRequest)
2019-01-09 16:45:06,143 43 ERROR test odoo.sql_db: bad query: b'INSERT INTO "product_product" ("id", "product_tmpl_id", "default_code", "active", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval(\'product_product_id_seq\'), 59, \'SH\', true, 1, 1, (now() at time zone \'UTC\'), (now() at time zone \'UTC\')) RETURNING id'
ERROR: duplicate key value violates unique constraint "product_product_uniq_default_code"
DETAIL:  Key (default_code)=(SH) already exists.
```

### DIFF
--- a/stock_request/tests/test_stock_request.py
+++ b/stock_request/tests/test_stock_request.py
@@ -39,7 +39,7 @@ class TestStockRequest(common.TransactionCase):
             [self.stock_request_manager_group.id],
             [self.main_company.id, self.company_2.id])
         self.product = self._create_product('SH', 'Shoes', False)
-        self.product_company_2 = self._create_product('SH', 'Shoes',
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:

rousseldenis (approved):

PR #532, opened 2019-01-15
[11.0][FIX] stock_orderpoint_manual_procurement: requested_uid

LoisRForgeFlow:
pass requested_uid in context to be able to have it through a chain of procuremets.

### DIFF
--- a/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
+++ b/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
@@ -61,6 +61,10 @@ class MakeProcurementOrderpoint(models.TransientModel):
     def make_procurement(self):
         self.ensure_one()
         errors = []
+        # User requesting the procurement is passed by context to be able to
+        # update final MO, PO or trasfer with that information.
+        pg_obj = self.env['procurement.group'].with_context(
+            requested_uid=self.env.user)
         for item in self.item_ids:
             if not item.qty:
                 raise ValidationError(_("Quantity must be positive."))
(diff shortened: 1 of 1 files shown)


stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -61,6 +61,10 @@ def default_get(self, fields):
     def make_procurement(self):
         self.ensure_one()
         errors = []
+        # User requesting the procurement is passed by context to be able to
+        # update final MO, PO or trasfer with that information.
+        pg_obj = self.env['procurement.group'].with_context(
```
rousseldenis, line 66: @lreficent Is that cannot be done without requiring context change ? This can decrease performances.

LoisRForgeFlow: It would require several hooks into odoo code.

LoisRForgeFlow: Note that I only add the context at the beggining then the loop uses `pg_obj` variable unmodified.

rousseldenis: Yes that depends only on how many times 'make_procurement' is called

LoisRForgeFlow: Well, it is something manually triggered so it will only create a new environment with that context each time a user uses the wizard. I think it is not something critical in this case.

rousseldenis (approved):

JordiBForgeFlow (approved):

PR #533, opened 2019-01-21
[11.0][FIX] Skipping potential error when move has been deleted

BT-rmartin:
If a move was deleted on this line (https://github.com/brain-tec/odoo/blob/11.0/addons/stock/models/stock_move_line.py#L413), this line would lead to an error about the move was already deleted so we cannot iterate on that. 

This fix allows to avoid this problem and working on the moves still existing on self.

### DIFF
--- a/stock_request/__manifest__.py
+++ b/stock_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "11.0.3.2.0",
+    "version": "11.0.3.1.1",
     "license": "LGPL-3",
     "website": "https://github.com/stock-logistics-warehouse",
     "author": "Eficent, "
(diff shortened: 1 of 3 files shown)


BT-rmartin:
Could you please check this @jbeficent ? Thanks

etobella (approved):
LGTM :+1: 
Change version of the module.

BT-rmartin:
> LGTM 
> Change version of the module.

Why do we have to change the version number? It's just a python change, not -u is needed.

rousseldenis:
@BT-rmartin That updates also package on pypi.

BT-rmartin:
@rousseldenis  Done

stock_request/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "11.0.3.1.0",
+    "version": "11.0.3.2.0",
```
rousseldenis: @BT-rmartin You could also change the patch part (as the modification is not so great)

BT-rmartin:
@rousseldenis Done

rousseldenis (approved):
Code review. Fast one

PR #535, opened 2019-01-25
[11.0][MIG] stock_move_location

sebalix:
Supersedes #522

ping @rousseldenis @simahawk @mpanarin


stock_move_location/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    location_move = fields.Boolean(
+        string="Part of move location",
+        help="Wether this move is a part of stock_location moves",
```
simahawk: Whether

stock_move_location/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+
+    @api.depends("location_move")
+    def _compute_show_details_visible(self):
+        super()._compute_show_details_visible()
+        for move in self:
+            if move.location_move:
```
simahawk:
```py
for move in self:
    move.show_details_visible = move.location_mode
```

stock_move_location/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+* A new menuitem Stock > Move from location... opens a wizard
+  where 2 location ca be specified.
+* Select origin and destination locations and press "IMMEDIATE TRANSFER" or "PLANNED TRANSFER"
+* Press `ADD ALL` button to add all products available
+* Those lines can be edited. Move quantity can't be more than a max available quantity
+* Move doesn't care about the reservations and will move stuff anyway
+* If during you operation with the wizard the real quantity will change
```
simahawk: *your

stock_move_location/tests/test_move_location.py
```diff
@@ -0,0 +1,111 @@
...
+    def _create_wizard(self, origin_location, destination_location):
+        return self.wizard_obj.create({
+            "origin_location_id": origin_location.id,
+            "destination_location_id": destination_location.id,
+        })
+
+    def test_move_location_wizard(self):
+        """Test a simple move.
+        """
+        self.setup_product_amounts()
```
simahawk: if you need `setup_product_amounts` to run every time, just move it to `setUp`

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,186 @@
...
+        location_id = self.origin_location_id.id
+        company = self.env['res.company']._company_default_get(
+            'stock.inventory',
+        )
+        return """
+        SELECT product_id, lot_id, SUM(quantity)
+        FROM stock_quant
+        WHERE location_id = {location_id} AND company_id = {company_id}
+        GROUP BY product_id, lot_id
+        """.format(
```
simahawk: * rename to `_get_group_quants`
* move the call to execute + fetchall here
* do not use format and rely on `execute` args interpolation

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,143 @@
...
+        string="Maximum available quantity",
+        digits=dp.get_precision('Product Unit of Measure'),
+    )
+    custom = fields.Boolean(
+        string="Custom line",
+        default=True,
+    )
+
+    @api.model
+    def get_rounding(self):
```
simahawk: _get_rounding

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,143 @@
...
+        return self.env.ref("product.decimal_product_uom").digits or 3
+
+    @api.constrains("max_quantity", "move_quantity")
+    def _constraint_max_move_quantity(self):
+        for record in self:
+            if (float_compare(
```
simahawk: can we move the 1st compare to a static method and reuse it where needed?

simahawk (changes requested):
minor glitches

simahawk:
build is red since `_get_available_quantity` gives you back a float and not a list

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -53,17 +53,20 @@ class StockMoveLocationWizardLine(models.TransientModel):
     )
 
     @api.model
-    def get_rounding(self):
+    def _get_rounding(self):
```
rousseldenis: @sebalix Shouldn't rounding come from product_uom_id field ?

sebalix: @rousseldenis indeed it should, I'm looking how to convert the rounding precision to decimal precision, if there is a method in Odoo for that.

sebalix: ok, I need to use the `precision_rounding` parameter of `float_compare`, as simple as that.

fclementic2c (approved):
Functional test ok

stock_move_location/README.rst
```diff
@@ -0,0 +1,21 @@
+**This file is going to be generated by oca-gen-addon-readme.**
```
Tonow-c2c, line 1: I'am not totally sure but I things this file is useless

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,143 @@
...
+    max_quantity = fields.Float(
+        string="Maximum available quantity",
+        digits=dp.get_precision('Product Unit of Measure'),
+    )
+    custom = fields.Boolean(
+        string="Custom line",
+        default=True,
+    )
+
+    @api.model
```
simahawk: this is not api.model as you are using `self.product_uom_id`
I'd make it a static method and pass rounding factor:
```py
@staticmethod
def _compare(qty1, qty2, precision):
```
Alternative is to use @api.multi + ensure_one or... 
do not use any decorator at all as it's unlikely that you gonna call this via rpc.
In any case, is going to be broken in a very clear way if you call it on non recordset or multiple records.

stock_move_location/i18n/stock_move_location.pot
```diff
@@ -0,0 +1,168 @@
+# Translation of Odoo Server.
```
gurneyalex, line 1: I think the PRs for new modules must not include a pot file these days as this is generated by weblate. 

@sbidoul @pedrobaeza do you confirm?

(BTW maybe we should have an automated test for this)

gurneyalex (approved):

stock_move_location/README.rst
sebalix: It'll be regenerated by a bot, no? I'm not sure too

stock_move_location/wizard/stock_move_location_line.py
sebalix: PR updated (I used the @staticmethod, I'm not sure to understand the link with/without decorator and RPC interface, I though that by default @api.multi was applied by Odoo).

simahawk (approved):

stock_move_location/README.rst
simahawk: it's fine, and in the ctx of a pr like this it tells reviewers "don't check here, go to /readme" and tells bots/travis/etc to not complain about missing files

stock_move_location/i18n/stock_move_location.pot
sbidoul: .pot will be (re)generated by travis after merge. It does not harm if one is provided in the PR.

The only thing we try to avoid is PRs to **existing** modules that update **.po**, because that can generate merge conflicts with weblate.

sebalix:
Fixup commits squashed.

rousseldenis (approved):

sebalix:
Thank you all!

PR #537, opened 2019-01-31
[add][stock_orderpoint_route]

JordiBForgeFlow:
This module allows to restrict a specific route to be used in the
reordering rules. This route will be used instead of the default determined by
default.

Usage
=====

Go to the reordering rule and change the route. Only the routes applicable
for this location or locations above it can be selected.


stock_orderpoint_route/views/stock_warehouse_orderpoint_views.xml
```diff
@@ -0,0 +1,21 @@
...
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record id="view_warehouse_orderpoint_form" model="ir.ui.view">
+        <field name="name">stock.warehouse.orderpoint.form</field>
+        <field name="model">stock.warehouse.orderpoint</field>
+        <field name="inherit_id" ref="stock.view_warehouse_orderpoint_form"/>
+        <field name="arch" type="xml">
+            <field name="group_id" position="before">
+                <field name="location_id" groups="stock.group_stock_multi_locations"
```
LoisRForgeFlow: this fields is already in the original view

stock_orderpoint_route/views/stock_warehouse_orderpoint_views.xml
```diff
@@ -0,0 +1,21 @@
...
+                       domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]"/>
+                <field name="route_id"
+                       options="{'no_create': True}" groups="stock.group_stock_multi_locations"/>
```
LoisRForgeFlow: groups="stock.group_adv_location"

stock_orderpoint_route/tests/test_stock_orderpoint_route.py
```diff
@@ -0,0 +1,154 @@
...
+        return self.env['product.product'].create(dict(
+            name=name,
+            default_code=default_code,
+            uom_id=self.env.ref('product.product_uom_unit').id,
+            company_id=company_id,
+            type='product',
+            **vals
+        ))
+
+    def test_create_orderpoint_01(self):
```
LoisRForgeFlow: test_orderpoint_route_01?

stock_orderpoint_route/tests/test_stock_orderpoint_route.py
```diff
@@ -0,0 +1,154 @@
...
+        self.assertIn(self.route, orderpoint.route_ids)
+        self.assertIn(self.route2, orderpoint.route_ids)
+        orderpoint.route_id = self.route.id
+        self.procurement_group_model.run_scheduler()
+        move = self.env['stock.move'].search(
+            [('product_id', '=', self.product.id),
+             ('location_id', '=', self.ressuply_loc.id)], limit=1)
+        self.assertEqual(len(move), 1)
+
+    def test_create_orderpoint_02(self):
```
LoisRForgeFlow: test_orderpoint_route_02?

LoisRForgeFlow (commented):
Functional review :+1: 

Code review, some minor comments inline

stock_orderpoint_route/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo import api, fields, models
+
+
+class StockWarehouseOrderpoint(models.Model):
+    _inherit = 'stock.warehouse.orderpoint'
+
+    route_ids = fields.Many2many(
+        'stock.location.route', string='Allowed routes',
+        compute='_compute_route_ids',
+        readonly=True,
```
Nikul-Chaudhary: @jbeficent No need `readonly` here

Nikul-Chaudhary (commented):
Improve Code

rousseldenis:
@jbeficent Could you attend comments ?

CasVissers-360ERP (approved):
Functional test, again a great module..

HviorForgeFlow (approved):
Comments attended, not sure about remove readonly True on M2M fields. @Nikul-Chaudhary why is needed to do that?

stock_orderpoint_route/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+    route_id = fields.Many2one('stock.location.route', string='Route',
+                               domain="[('id', 'in', route_ids)]",
+                               ondelete='restrict')
+
+    @api.depends('product_id', 'warehouse_id', 'location_id')
+    def _compute_route_ids(self):
+        route_obj = self.env['stock.location.route']
+        for wh in self.mapped('warehouse_id'):
+            wh_routes = route_obj.search(
```
rousseldenis, line 22: @jbeficent Why not using the field 'route_ids'  on warehouse (and adding it in depends) ?

rousseldenis (changes requested):
Code review.

stock_orderpoint_route/models/stock_warehouse_orderpoint.py
rousseldenis: @jbeficent Maybe this could avoid searches.

guewen:
Do you think an `orderpoint_selectable` as in https://github.com/OCA/stock-logistics-warehouse/pull/783/files#diff-370e5fe8458a06eb66c2f15fba555006R10-R15 makes sense?

In any case, we have the same need, I'm going to port your PR to 12.0 right now.

guewen (approved):

stock_orderpoint_route/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,53 @@
...
+                [('warehouse_ids', '=', wh.id)])
```
guewen, line 23: Is there a reason to filter on the warehouse if anyway we filter on the location afterwards?

guewen: Instead of filtering on "warehouse selectable" + sublocations, do you think we can filter on a new flag "orderpoint selectable" + sublocations?

The end result will be the same as anyway we filter on the location of the orderpoint, but it seems more understandable to me.

I'm moving forward on the migration to 12.0, so it's the best time for me to make a change, that I can backport if you agree.

cc @jgrandguillaume

jgrandguillaume: @jbeficent I must say I agree with @guewen here, do you?

guewen: I overlooked the "look in the routes of the product, look in the routes of the product categories, look in the routes of the warehouse" part. Still, I would be interested to have your thoughts on this @jbeficent.

stock_orderpoint_route/models/stock_warehouse_orderpoint.py
rousseldenis: @JordiBForgeFlow Any news on this ?

PR #540, opened 2019-02-07
12.0 mig stock available

devang-dreambits:


stock_available/README.rst
```diff
@@ -0,0 +1,77 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
rousseldenis: Could you please follow migration guide about readme fragments : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0 ?

stock_available/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2014 Numérigraphe
+# Copyright 2016 Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
```
rousseldenis, line 5: Could you please add some missing entries like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L9

rousseldenis (changes requested):
Code review. Minor comments

stock_available/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'name': 'Stock available to promise',
+    'version': '12.0.1.0.0',
```
Nikul-Chaudhary: @devang-dreambits Add website key and URL like `https://github.com/OCA/stock-logistics-warehouse`

stock_available/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2014 Numérigraphe
```
Nikul-Chaudhary: remove Copyright from all `init` file

stock_available/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Usage
+=====
+
+This module adds a field named `Available for sale` on the Product form.
+Various additional fields may be added, depending on which information you
+chose to base the computation on.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/11.0
```
Nikul-Chaudhary: 11.0/12.0

Nikul-Chaudhary (commented):
Improve Code

ray-odoo:
I will have the opportunity to test this in the next few days.  Woo hoo!!

ray-odoo:
When testing this, I noticed that I was unable to set the "Exclude incoming goods" boolean via Inventory --> Settings.

I looked into the module code, and it seems that the configuration option "Exclude incoming goods" is designed to install "stock_available_immediately" when set to TRUE as the help for this boolean is "This installs the module stock_available_immediately.".

I don't see that module in this PR, and I don't see it available for testing at v12.  

The User Experience when stock_available_immediately isn't available is to appear to do nothing, as the boolean is not set to True, but the user is provided no warning.

It appears that this module also optionally installs:  stock_available_sale, sale_delivery_date, stock_available_mrp but I also don't see these modules available for testing at v12.

So I can't really test until these other modules are available.  Do I misunderstand something?

I am going to take a look at stock_available_unreserved in an effort to understand the differences between this module and that one.

rousseldenis:
@devang-dreambits Could you add recent improvements from v10.0 like https://github.com/OCA/stock-logistics-warehouse/commit/29b6bcdd0a61f2a9a19f5ddc3e50ba345fcd2890#diff-2c6955de963025ce90888f7e13128f22 ?

rousseldenis:
> When testing this, I noticed that I was unable to set the "Exclude incoming goods" boolean via Inventory --> Settings.
> 
> I looked into the module code, and it seems that the configuration option "Exclude incoming goods" is designed to install "stock_available_immediately" when set to TRUE as the help for this boolean is "This installs the module stock_available_immediately.".
> 
> I don't see that module in this PR, and I don't see it available for testing at v12.
> 
> The User Experience when stock_available_immediately isn't available is to appear to do nothing, as the boolean is not set to True, but the user is provided no warning.
> 
> It appears that this module also optionally installs: stock_available_sale, sale_delivery_date, stock_available_mrp but I also don't see these modules available for testing at v12.
> 
> So I can't really test until these other modules are available. Do I misunderstand something?
> 
> I am going to take a look at stock_available_unreserved in an effort to understand the differences between this module and that one.


Yes indeed, stock_available module was designed like that and options to install other modules that depend on this are present anyway even if modules are not present.

The good solution should be removing all those options here and adding them in each module. It's clearer for user.

@lreficent @jbeficent @Cedric-Pigeon @devang-dreambits What do you think of this? That could be a good improvement for user experience

rafaelbn:
Please just notice about performance issues with this module:

- https://github.com/odoo/odoo/pull/33967
- https://github.com/OCA/stock-logistics-warehouse/pull/611

pedrobaeza:
@karan-dreambits are you interested in continuing this PR? There's another PR for the module in #632 that has an interesting change in the smart-button style, but yours have priority as first in arrive.

karan-dreambits:
@pedrobaeza I see #632 is much better so lets close this and let it supersede this :)
@devang-dreambits  Please update our project setup to use the #632 PR instead of this one

sergio-teruel:
@karan-dreambits Please, can you review #632 ??

pedrobaeza:
Thanks for your comprehension, so closing.

PR #545, opened 2019-02-19
[11.0][FIX] stock_secondary_unit: Remove procurement_rule move. Fix dependencies

sergio-teruel:
cc @Tecnativa


stock_secondary_unit/__manifest__.py
```diff
@@ -12,7 +12,6 @@
     'application': False,
     'installable': True,
     'depends': [
-        'sale_stock',
```
pedrobaeza, line 15: Depend on `stock`

pedrobaeza (changes requested):

rousseldenis (approved):

pedrobaeza (approved):

PR #546, opened 2019-02-20
[11.0][FIX] stock_request: sudoing

HviorForgeFlow:
At some point using previous sudos, some context are lost.

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -107,12 +107,12 @@ class StockRequest(models.Model):
 
     @api.depends('allocation_ids')
     def _compute_move_ids(self):
-        for request in self.sudo():
+        for request in self:
             request.move_ids = request.allocation_ids.mapped('stock_move_id')
 
     @api.depends('allocation_ids')
     def _compute_picking_ids(self):
-        for request in self.sudo():
+        for request in self:
             request.picking_count = 0
             request.picking_ids = self.env['stock.picking']
             request.picking_ids = request.move_ids.filtered(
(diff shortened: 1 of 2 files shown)


HviorForgeFlow:
hey, @etobella @jbeficent with this fix stock request users are able to see generated transfers once they are confirmed. It seems it is not breaking any other workflows. Could you review it?

HviorForgeFlow:
Having this error if user has only stock_request_user. Should Stock Request User read from stock_move ??
image

Event if it is forced to read, users won't be able to see stock move info

JordiBForgeFlow:
The stock request user should have the minimum permissions to create and
view stock.move, stock.move.line, stock picking,... So that this kind of
error does not occur.

Sudoing prevents this from being needed, but is also generating the side
effect that you noticed.

etobella:
@hveficent , could you improve the permissions of stock request user in order to fix the issue?

JordiBForgeFlow (approved):

etobella (approved):

chienandalu (approved):
Looks good :+1:

chienandalu (changes requested):
Sorry, I didn't notice that there's a travis error. Could you check?

pedrobaeza:
Can you check Travis?

HviorForgeFlow:
Travis errors are not related to this PR.
See https://github.com/OCA/stock-logistics-warehouse/pull/552

rousseldenis (approved):
Seems good to me

rousseldenis:
@hveficent Is this complete ?

rousseldenis:
@chienandalu Could you update your review?

chienandalu (approved):

PR #547, opened 2019-02-20
[10.0] [FIX] purchase_packaging avoid conversion error when using uom not in units category

Cedric-Pigeon:
Creating a po through a procurement for a product with a unit not in 'unit' category is impossible.
The method preparing the po line values does not take care of `product_purchase_uom_id` value and triggers the default: https://github.com/OCA/stock-logistics-warehouse/blob/10.0/purchase_packaging/models/purchase_order_line.py#L13
So it leads to a failing conversion error.

This PR intends to fix that.

### DIFF
--- a/purchase_packaging/models/procurement_order.py
+++ b/purchase_packaging/models/procurement_order.py
@@ -21,6 +21,8 @@ class ProcurementOrder(models.Model):
             date=po.date_order and fields.Date.to_string(
                 fields.Date.from_string(po.date_order)) or None,
             uom_id=self.product_id.uom_po_id)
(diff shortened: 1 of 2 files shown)


hparfr (approved):
LGTM

ThomasBinsfeld (approved):

rafaelbn (approved):
Hi @Cedric-Pigeon could you review travis in order to merge?

We must not loose this in v12

cc @rousseldenis

rousseldenis (approved):

Cedric-Pigeon:
@rafaelbn Done.

PR #555, opened 2019-03-11
[10.0] Fix mts+mto rules creation on warehouse creation

florian-dacosta:
With the current version of the module, when you try to create a warehouse with the `mto_mts_management` flag, it fails.
Odoo raise an error here https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_mts_mto_rule/model/warehouse.py#L35 because it did not yet write the data coming from `create_routes` mehod on the warehouse.
The same happens with mts_rules after.
I just move the logic from `create_routes` to `create` method in order to be sure the fields used in `_get_mts_mto_rule` are already filled.

### DIFF
--- a/stock_mts_mto_rule/model/warehouse.py
+++ b/stock_mts_mto_rule/model/warehouse.py
@@ -88,15 +88,15 @@ class Warehouse(models.Model):
                     })
         return res
 
-    @api.multi
-    def create_routes(self):
-        pull_model = self.env['procurement.rule']
-        res = super(Warehouse, self).create_routes()
-        if self.mto_mts_management:
-            mts_mto_pull_vals = self._get_mts_mto_rule()
+    @api.model
+    def create(self, vals):
+        wh = super(Warehouse, self).create(vals)
+        if wh.mto_mts_management:
+            pull_model = self.env['procurement.rule']
+            mts_mto_pull_vals = wh._get_mts_mto_rule()
             mts_mto_pull = pull_model.create(mts_mto_pull_vals)
-            res['mts_mto_rule_id'] = mts_mto_pull.id
-        return res
+            wh.mts_mto_rule_id = mts_mto_pull.id
+        return wh
 
     @api.multi
     def write(self, vals):
--- a/stock_mts_mto_rule/model/warehouse.py
+++ b/stock_mts_mto_rule/model/warehouse.py
@@ -88,14 +88,19 @@ class Warehouse(models.Model):
                     })
         return res
 
+    @api.multi
+    def _set_mts_mto_rule(self):
+        self.ensure_one()
+        if self.mto_mts_management:
+            pull_model = self.env['procurement.rule']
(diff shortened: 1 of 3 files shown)


stock_mts_mto_rule/model/warehouse.py
```diff
@@ -88,15 +88,15 @@ def _get_push_pull_rules_values(
...
-    @api.multi
-    def create_routes(self):
-        pull_model = self.env['procurement.rule']
-        res = super(Warehouse, self).create_routes()
-        if self.mto_mts_management:
-            mts_mto_pull_vals = self._get_mts_mto_rule()
+    @api.model
+    def create(self, vals):
+        wh = super(Warehouse, self).create(vals)
+        if wh.mto_mts_management:
```
rousseldenis: @florian-dacosta Maybe I would have put code here under into separate method (to easy override).

rousseldenis (changes requested):
Code review.

At least, change module version

hparfr (approved):
+1 on minor version number.
This fix looks good to me.

florian-dacosta:
@rousseldenis 
I have attended your comments.
I thought the version change was done during the merge.

rousseldenis:
@florian-dacosta Nope, that depends on developer will.

Could you check Travis ?

florian-dacosta:
I fixed a test that was failing on stock_mts_mto_rule.
Unfortunatly, it seems there is still an error, but coming from another module I do not know.

rousseldenis:
@florian-dacosta Could you rebase ? 10.0 branch seems to be green, so...

florian-dacosta:
Strange indeed, I can't rebase as it is already up to date!

rousseldenis:
So, the error is due to this. In the log, they appear after installation of this module. Could you test those two modules with and without stock_mts_mto_rule one ?

florian-dacosta:
@rousseldenis 
I can't reproduce the problem, and I don't see how my fix could generate such an error.
The 10.0 branch seems red : https://travis-ci.org/OCA/stock-logistics-warehouse/builds/490502048
With the same problem, so it means the problem do not come from my PR, right?
Or am I missing something?

rousseldenis:
@florian-dacosta Yes, it seems that last good build is dated from 6th february on 10.0. I will explore log and keep you in touch

rousseldenis:
@florian-dacosta Ok. I know this. I found recently the solution on another repo.

Due to https://github.com/odoo/odoo/commit/2b1d3ff82dd23fb0372b3b73ecb84841a0d593d3 you have to raise ValidationError and not another exception in constrains.

I will propose a fix for this repo

rousseldenis:
https://github.com/OCA/stock-logistics-warehouse/pull/567

rousseldenis:
@florian-dacosta Green!

florian-dacosta:
Nice, thanks a lot @rousseldenis !

rousseldenis (approved):
Seems ok for me

PR #568, opened 2019-04-02
[12.0][MIG] stock_orderpoint_uom

kittiu:
Migrate to v12


rousseldenis:
@kittiu Is it normal you have deleted some .po files ?

kittiu:
> @kittiu Is it normal you have deleted some .po files ?

Yes, this area I am still unsure what to do. We have to recreated the .pot for v12, right? Can we still keep the old .po files?

kittiu:
@rousseldenis I found old OCA mail Weblate.
So, from now we do not need to create .pot or .po file anymore, same for migration? So, we should delete them all and wait for Weblate to recreate?

> Important: always translate using Weblate, and never push changes to .po files manually. This simple rule will avoid all merge conflicts, as all .pot file changes are done by Travis, and Weblate is the master of all .po files.
> Weblate has an upload feature: use it in case you need to translate outside the Weblate interface for any reason.
>

rousseldenis:
@kittiu Indeed, Weblate is the tool used to translate OCA modules. But, this is true for new translations. Old po files should be kept on migration flow (as you are not aware easily if those po files were generated by Weblate or through another mean).

kittiu:
> @kittiu Indeed, Weblate is the tool used to translate OCA modules. But, this is true for new translations. Old po files should be kept on migration flow (as you are not aware easily if those po files were generated by Weblate or through another mean).

I have moved the translation files from 11.0 to 12.0. The .pot file also as of 11.0.

rousseldenis (approved):
Code review. LGTM.
@kittiu Could you amend your commit message to keep only '[12.0][MIG] stock_orderpoint_uom' string ?

kittiu:
> Code review. LGTM.
> @kittiu Could you amend your commit message to keep only '[12.0][MIG] stock_orderpoint_uom' string ?

Done

LoisRForgeFlow (approved):
LGTM, Thanks for migrating!

PR #574, opened 2019-04-16
[11.0][IMP] stock_move_location: Allow move products from one location to another directly from selected quants

sergio-teruel:
cc @Tecnativa
This PR adds various improvements to this module:

1.  When in the wizard the user changes the location_destination_id all lines are erased, so the user loses the quantities to transfer to another location. With this PR only change the location_destination_id for all lines.

2. Add improvements that allow to do a transfer selecting quants directly from various origin locations. In this case you can not change origin location in the wizard. I try to no update the original behavior.

Thanks

### DIFF
--- a/stock_move_location/README.rst
+++ b/stock_move_location/README.rst
@@ -25,7 +25,8 @@ Move Stock Location
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module allows to move entire location of products from one place to another
+This module allows to move entire location of products from one place to
+another and move only selected quants.
 
 **Table of contents**
 
@@ -45,6 +46,14 @@ Usage
   it will move only the available quantity at the button press
 * Products will be moved and a form view of picking that did that will show up
 * If "PLANNED TRANSFER" is used - the picking won't be validated automatically
+
+If you want to transfer a full quant:
+
+*  Go to `Inventory > Master Data > Products` and click "On hand" smart button
+   or `Inventory > Reporting > Inventory`, the quants view will be
+   opened.
+
+*  Select the quants which you want move to another location
 
 Known issues / Roadmap
 ======================
@@ -79,6 +88,7 @@ Contributors
 
 * Mathieu Vatel <mathieu@julius.fr>
 * Mykhailo Panarin <m.panarin@mobilunity.com>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
 
 Maintainers
 ~~~~~~~~~~~
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -1,8 +1,10 @@
 # Copyright (C) 2011 Julius Network Solutions SARL <contact@julius.fr>
 # Copyright 2018 Camptocamp SA
+# Copyright 2019 Sergio Teruel - Tecnativa <sergio.teruel@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import api, fields, models
+from odoo.fields import first
 
 
 class StockMoveLocationWizard(models.TransientModel):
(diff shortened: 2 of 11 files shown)


stock_move_location/README.rst
```diff
@@ -45,6 +46,7 @@ Usage
   it will move only the available quantity at the button press
 * Products will be moved and a form view of picking that did that will show up
 * If "PLANNED TRANSFER" is used - the picking won't be validated automatically
+* In quants view you has an action to move selected quants to another location
```
pedrobaeza: Please put as a different  paragraph in the section:
```rst
* If "PLANNED...

If you want to transfer a full quant:

#. Go to ...
```

stock_move_location/data/stock_move_location.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: This should be named `stock_quant_views.xml`

pedrobaeza (approved):
Little details

victor-ditay (approved):
just the @pedrobaeza review others LGTM

sergio-teruel:
@pedrobaeza @victormmtorres Changes done!!

stock_move_location/wizard/stock_move_location.py
```diff
@@ -30,19 +31,38 @@ class StockMoveLocationWizard(models.TransientModel):
...
+        res['stock_move_location_line_ids'] = [(0, 0, {
+            'product_id': quant.product_id.id,
+            'move_quantity': quant.quantity,
+            'max_quantity': quant.quantity,
+            'origin_location_id': quant.location_id.id,
+            'lot_id': quant.lot_id.id,
+            'product_uom_id': quant.product_uom_id.id,
+            'custom': False,
+        }) for quant in quants]
+        res['origin_location_id'] = quants[:1].location_id.id
```
rousseldenis: Prefer using 'first' method:
```suggestion
        from odoo.fields import first
        res['origin_location_id'] = first(quants).location_id.id
```

rousseldenis (changes requested):
LGTM. Minor change

stock_move_location/wizard/stock_move_location.py
sergio-teruel: Hi Denis, I didn't know this function, but... ¿Can you explain me in this case why is better use first(with prefetch) than directly records[:1]?
I have done a profile for two cases and I do not see a big difference, even records[:1] is faster.. :smile: 
You have to take into account that I already access to records so the data is in cache.

**Profiler with records[:1]**
calls     queries   ms
1         0         0.02          @profile
                                  @api.model
                                  def default_get(self, fields):
1         1         1.46              res = super().default_get(fields)
1         0         0.01              if self.env.context.get('active_model', False) != 'stock.quant':
                                          return res
                                      # Load data directly from quants
1         0         0.03              quants = self.env['stock.quant'].browse(
1         0         0.02                  self.env.context.get('active_ids', False))
1         0         0.0               res['stock_move_location_line_ids'] = [(0, 0, {
                                          'product_id': quant.product_id.id,
                                          'move_quantity': quant.quantity,
                                          'max_quantity': quant.quantity,
                                          'origin_location_id': quant.location_id.id,
                                          'lot_id': quant.lot_id.id,
                                          'product_uom_id': quant.product_uom_id.id,
                                          'custom': False,
1         6         70.74             }) for quant in quants]
**_1         0         0.04              res['origin_location_id'] = quants[:1].location_id.id_**
1         0         0.0               return res

Total:
1         7         72.32   

**Profiler with method firts**
calls     queries   ms
1         0         0.01          @profile
                                  @api.model
                                  def default_get(self, fields):
1         1         1.25              res = super().default_get(fields)
1         0         0.01              if self.env.context.get('active_model', False) != 'stock.quant':
                                          return res
                                      # Load data directly from quants
1         0         0.02              quants = self.env['stock.quant'].browse(
1         0         0.02                  self.env.context.get('active_ids', False))
1         0         0.0               res['stock_move_location_line_ids'] = [(0, 0, {
                                          'product_id': quant.product_id.id,
                                          'move_quantity': quant.quantity,
                                          'max_quantity': quant.quantity,
                                          'origin_location_id': quant.location_id.id,
                                          'lot_id': quant.lot_id.id,
                                          'product_uom_id': quant.product_uom_id.id,
                                          'custom': False,
1         6         70.37             }) for quant in quants]
**_1         0         0.05              res['origin_location_id'] = first(quants).location_id.id_**
1         0         0.0               return res

Total:
1         7         71.74

rousseldenis: I agree, but as Odoo provides the helper method, it's safer to use it than indexes (and more logic as indexes are for lists not really for recordsets).

sergio-teruel: Ok, Thanks!! :smile:

pedrobaeza: @rousseldenis not really, as that notation is Python standard, and recordsets adopt that standards (the same as set logic operations).

sergio-teruel:
@rousseldenis change done!

rousseldenis (approved):
Code review

HviorForgeFlow (approved):
Functional review :+1: Nice feature :smile:

HviorForgeFlow:
@sergio-teruel I see this feature is available for Inventory Users, it could be dangerous that a large amount of users in company could move freely quants between location without restriction. 

This is not a blocking request, but creating a group to enable this feature could be a good practice.

pedrobaeza:
@hveficent IMO, a warehouse user is already able to do the same, but manually, so adding more layers is overkilling. This can't be done accidentally, as you have to accept the wizard popup, so for me it's OK, although other opinions are welcome.

rousseldenis:
> @hveficent IMO, a warehouse user is already able to do the same, but manually, so adding more layers is overkilling. This can't be done accidentally, as you have to accept the wizard popup, so for me it's OK, although other opinions are welcome.

Yes, but it's an open door to massive changes. I agree with @hveficent, but the inventory user could be added by default (and let admin change it if required)

pedrobaeza:
Then it's OK as is?

sergio-teruel:
Ok, I have added the stock.group_stock_user group to quants action

PR #579, opened 2019-04-24
[11.0][IMP] stock_orderpoint_move_link

HviorForgeFlow:
Improve usability:
- [x] Add some test
- [x] Ensure orderpoint name on origin
- [x] Add Purchase like right-top button for stock pickings on Orderpoint views.

@Eficent


stock_orderpoint_move_link/models/procurement_group.py
```diff
@@ -0,0 +1,23 @@
...
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def run(self, product_id, product_qty, product_uom, location_id, name,
+            origin, values):
+        if 'orderpoint_id' in values:
+            if values['orderpoint_id'].name not in origin:
+                origin += ', %s' % values.get('orderpoint_id').name
```
LoisRForgeFlow: it is possible for origin to be empty? If so this could end being something like `, OP/NNN` with an initial comma that may be unwanted.

HviorForgeFlow: It's hard to get it, action who triggers this is commonly https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_move.py#L762,
I can add more code to avoid this if required, I don't think so.

HviorForgeFlow: comments attended

LoisRForgeFlow (approved):

rousseldenis:
@hveficent Do you plan to add test about 'origin' field (see coverage) ?

HviorForgeFlow:
@rousseldenis I can't reproduce it, these lines where added to get more consistency with third party addons.

https://github.com/OCA/stock-logistics-warehouse/pull/579#discussion_r281641913

HviorForgeFlow:
@lreficent I changed from stock move to pickings on orderpoint top right buttons, take a look.

MiquelRForgeFlow (approved):
LGTM

LoisRForgeFlow:
Tested on runbot, looks good! :+1: 

/ocabot merge minor

PR #580, opened 2019-04-24
[11.0][IMP] stock_orderpoint_mrp_link

HviorForgeFlow:
Improve usability:
- [x] Add Purchase like right-top button for MO on Orderpoint views.

https://github.com/OCA/stock-logistics-warehouse/pull/579

@Eficent


stock_orderpoint_mrp_link/tests/test_stock_orderpoint_mrp_link.py
```diff
@@ -104,3 +104,6 @@ def test_02_stock_orderpoint_mrp_link_indirect_routing(self):
         self.assertTrue(mo2)
         self.assertEqual(mo2.orderpoint_id, self.orderpoint_secondary_loc)
         self.assertEqual(mo2.orderpoint_id, self.orderpoint_secondary_loc)
+
+    def test_03_stock_orderpoint_mrp_link_action_view(self):
+        self.orderpoint_secondary_loc.action_view_mrp_productions()
```
etobella: The test could more extended? Maybe check that if you use the domain returned and search, the result is the expected?

etobella (approved):
Just a comment on the test, just testing the function is not the best option, check that the result is the expected somehow

stock_orderpoint_mrp_link/tests/test_stock_orderpoint_mrp_link.py
HviorForgeFlow: @etobella I tried something...

etobella: Perfectly fine!

LoisRForgeFlow (commented):
Looks good, some small comments:

1. Check travis
2. Could you change the button icon.
image

it is used already for purchases, and we also add it in DDMPR :sweat_smile:

HviorForgeFlow:
@lreficent comments attendend

LoisRForgeFlow (approved):
Runbot error is unrelated. :+1:

PR #585, opened 2019-05-06
[9.0] Upgrade stock_inventory_preparation_filter from V8 to V9

BenjaHe:


pedrobaeza:
This is already migrated, so closing.

PR #593, opened 2019-05-16
[11.0][FIX] stock_orderpoint_manual_procurement_(uom)

HviorForgeFlow:
Before this PR once you made a procurement request and wizard was open all was right. But when you execute the procurement, `default_get` couldn't find `location_id, display_name, warehouse, product_id` for any item. 

That was solved using orderpoint location and product information when you run the procurement.

There was a lack of extendibility.

That PR tries to fix this issue across `stock_orderpoint_manual_procurement` and `stock_orderpoint_manual_procurement_uom` modules

@Eficent

### DIFF
--- a/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
+++ b/stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
@@ -118,16 +118,20 @@ class MakeProcurementOrderpointItem(models.TransientModel):
     def default_get(self, fields):
         res = super(MakeProcurementOrderpointItem, self).default_get(fields)
         orderpoint_obj = self.env['stock.warehouse.orderpoint']
-        orderpoint_id = self.env.context['active_ids'] or []
+        orderpoint_ids = self.env.context['active_ids'] or []
         active_model = self.env.context['active_model']
 
-        if not orderpoint_id:
+        if not orderpoint_ids:
             return res
         assert active_model == 'stock.warehouse.orderpoint', \
-            'Bad context propagation'
+            UserError('Bad context propagation')
 
-        # On each item we only must find 1 orderpoint
-        res = self._prepare_item(orderpoint_obj.browse(orderpoint_id))
+        # When active_id field is taken from context it does not change on
+        # every iteration on default_get, so the only data it can be trusted
+        # is active_ids. Poping records from active_ids is the only way I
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (approved):
This works and improves extensibility, maybe is not the best apprach, but I cannot come up with any better idea either.

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -125,3 +113,34 @@ def onchange_uom_id(self):
...
+    @api.model
+    def default_get(self, fields):
+        res = super(MakeProcurementOrderpointItem, self).default_get(fields)
+        orderpoint_obj = self.env['stock.warehouse.orderpoint']
+        orderpoint_ids = self.env.context['active_ids'] or []
+        active_model = self.env.context['active_model']
+
+        if not orderpoint_ids:
+            return res
+        assert active_model == 'stock.warehouse.orderpoint', \
```
rousseldenis, line 126: Please change to UserError here

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -125,3 +113,34 @@ def onchange_uom_id(self):
...
+            'Bad context propagation'
+
+        # Ugly hack to ensure that all active ids are used as active_id
+        # is always the same
+        res = self._prepare_item(orderpoint_obj.browse(orderpoint_ids[0]))
```
rousseldenis, line 133: Do you have more explanation on this ?

Why poping context values ?

rousseldenis (changes requested):
I'm not really comfortable with this.

Maybe I'll test it.

Change at least UserError

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
HviorForgeFlow: Comment attended, not sure if they are right.

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
HviorForgeFlow: I didn't find another way to make default_get work correctly. I added some more explanation in code comment

PR #600, opened 2019-05-27
 [12.0][MIG] packaging_uom

rousseldenis:


rousseldenis:
All packaging stuffs should go to each good repo. Migrating this to product-attribute

Issue #595, opened 2019-05-20
[IMP] stock_request: Add a "Submitted" state


max3903:
# Requirement

Users from other departments than the Warehouse uses stock requests to specify where products needs to be and when. This can be done from a project task, help desk ticket or field service order. Users may make modifications before the requests get processed by the warehouse team.

# Current module

The current module provides the following states: Draft, In Progress, Done, Cancelled.

# Change

We would like to add a new state "Submitted" between Draft and In Progress so the other teams can make modifications to their requests before it is submitted and showing up on the radar of the Warehouse team.

What do you think?

max3903:
Ping @jbeficent @patrickrwilson @wolfhall

etobella:
It's a good option, but could you add a parameter in order to allow this new state? This way we can maintain the old functionality with the new one

rousseldenis:
@max3903 And what about https://github.com/OCA/stock-logistics-warehouse/tree/11.0/stock_request_tier_validation

That is more compliant to your business flow requirement no ?

max3903:
@rousseldenis How would you configure it using stock_request_tier_validation? What happens when the user create the stock request?

We would like to hide the draft records to the Warehouse users.

rousseldenis:
> We would like to hide the draft records to the Warehouse users.

Ok, maybe misunderstood your requirements.

Do your flow first create request, secondly review request and then submit to warehouse users ?

Is that correct?

max3903:
Internal User:
* create stock requests
* come back later and modify it.
* when ready, submit to Warehouse user

Warehouse User:
* Review submitted requests
* Make modifications
* Confirm

There is no review or validation step.

JordiBForgeFlow:
I would add it as a new state, and through the settings decide if you want to use it.

Something like:
[ ] Allow user to submit stock requests.

In the line that @etobella suggests.

max3903:
:+1: Thank you. We will work on it. Please expect the pull request to review next week.

PR #605, opened 2019-05-28
[IMP] stock_request: Add setting/group to skip submitted state (#595)

max3903:
#595


rousseldenis (changes requested):
LGTM.

Please change module version

rousseldenis (approved):

max3903:
@rousseldenis Done!

etobella (changes requested):
I don't know if this happens to all people, but if you submit a stock request you cannot do anything more. It is right?

stock_request/views/stock_request_views.xml
```diff
@@ -50,16 +50,23 @@
...
                     <button name="action_confirm"
-                        string="Confirm" type="object"
-                        attrs="{'invisible': [('state', 'not in', ['draft'])]}"/>
-                    <button name="action_cancel" states="draft,open"
-                            type="object" string="Cancel"/>
-                    <button name="action_draft" states="cancel" type="object"
-                            string="Set to Draft"/>
+                            string="Confirm" type="object"
+                            class="oe_highlight"
+                            states="draft,submitted"/>
```
max3903, line 56: @etobella The confirm button is still there to confirm it...

etobella: Yes, but it gives a Validation Error...

max3903: @etobella Can you provide the error?

etobella: imagen

As admin I checked the flag on settings for submitting. I created a SR. I was not able to confirm. I submitted it and I tried to confirm it again. But it kept giving the same error again and again

HviorForgeFlow:
It is not working at all. I think better approach could be to allow stock request users only to submit and stock managers to confirm.

etobella:
Also, It could be better to reuse the same button and confirm if you are allowed and submit if you are not.
Maybe you need two buttons in order to make it work, but it is the best in order to show just the expected buttons.

stock_request/views/stock_request_views.xml
max3903: If you are in the group to submit, it means someone else not in that group needs to confirm... This is the expected behavior.

max3903: Internal Users can create stock requests and submit them. The users in the Warehouse User group can then review and confirm.

max3903:
@hveficent Please review the issue. The scenario is described there.

max3903:
@etobella Odoo does not allow to hide a button to a specific group.

etobella:
I know, maybe you should try to reuse the same then. If there is both buttons it is confusing for users.

max3903:
@etobella Clicking on Confirm does nothing and says nothing now. Not sure if it's better, but if you like it like that, I am fine.

HviorForgeFlow:
@max3903  I didn't test your last commit.
stock_request_manager_cant_confirm

max3903:
@hveficent You cannot do the whole process with the same user. This is on purpose and by design.

patrickrwilson (approved):

etobella:
I will check the changes tomorrow. Could you add the tests with the new functionalities

etobella:
Still not working... Process to show it:

* Access as admin
* Go to 'Stock request > Settings'
* Check 'Allow to submit'
* Access as demo
* Create a Stock Request
* It transform to Submitted
* As demo (Who is Stock Users) you are unable to transform it to In progress
* Access as admin
* Access the same Stock request
* As admin you are unable to transform it to in progress

As I said yesterday. I think two tests are necessary with the expected flows in order to be sure that the new flow is working and the old one has not been broken.

stock_request/models/stock_request.py
```diff
@@ -97,7 +98,8 @@ def _get_default_requested_by(self):
         states={'draft': [('readonly', False)]}, readonly=True
     )
     route_id = fields.Many2one(
-        states={'draft': [('readonly', False)]}, readonly=True
+        states={'draft': [('readonly', False)],
+                'submitted': [('readonly', False)]}, readonly=True
```
etobella, line 108: You should add this change for all fields.

etobella (changes requested):
I am unable to check your functionality...
Demo user, that is a Stock request user is able to confirm a Stock Request when I checked the flag on configuration.
Also, I don't understand your code. If the user has the group submit it is unable, but I have to set it with debug? It is weird that a group removes permissions... Wouldn't it be easier to do it differently? For example, with your previous code, you could have added if user is group submitted and not request manager... or adding a group that allows confirmation. MAnager has it and you remove it from stock request users on configuration.
Also, submitted state should be invisible, in order to keep the old configuration.

max3903:
@etobella 

As admin,
* Check the box "Allow users to submit stock requests" in the settings.
* Add users to the group "Allow users to submit stock requests": Alice
* Add users to the group "Inventory / User": Bob

As Alice,
* Go to Stock Requests
* Create a stock request and submit it

As Bob,
* Go to Stock Requests
* Open the stock request submitted by Alice
* Set the route and confirm it

etobella:
How do you add the permissions? with debug?

max3903:
@etobella With the developer mode.

etobella:
It is a bad choice IMO.
Every time I want to configure a user that has to submit I have to enter en development mode. 

Just a question. Wouldn't be easier to create a group that allows to confirm? This way, manager group inherits this and you can configure it on the `res.config` in order to allow users. It would be for all users by default.

max3903:
@etobella As discussed on the issue, the idea was to keep the current process as the default one. I think the current implementation is the best we can achieve considering Odoo's current capabilities.

The other option is to have the submitted state by default and have a configuration option to allow a specific group "Allow users to confirm directly" to bypass the state. I am not sure @jbeficent and @rousseldenis  would agree with this.

etobella:
You might define that the base group inherits ii on installation, then everyone will have it (probably a migration script might be necessary)

etobella:
By the way, the changes you made are only for stock.request, but not for stock.request.order. They should have similar behavior, so review that all works for both please.

max3903:
@jbeficent and @rousseldenis What do you think of having an option to **bypass** the submitted state (instead of one to force it)?

rousseldenis:
> @jbeficent and @rousseldenis What do you think of having an option to **bypass** the submitted state (instead of one to force it)?

For me it's ok!

stock_request/models/res_config_settings.py
```diff
@@ -10,6 +10,14 @@ class ResConfigSettings(models.TransientModel):
...
         implied_group='stock_request.group_stock_request_order')
 
+    group_stock_request_submit = fields.Boolean(
+        group='stock.group_stock_user',
+        implied_group='stock_request.group_stock_request_manager')
+
+    group_bypass_submit_request = fields.Boolean(
+        group='stock.group_stock_user',
+        implied_group='stock_request.group_bypass_submit_request')
+
```
max3903, line 15: @osi-scampbell We only need the 2nd one with:

```
    group_bypass_submit_request = fields.Boolean(
￼	implied_group='stock_request.group_bypass_submit_request')
```

stock_request/models/stock_request.py
max3903: @etobella The warehouse user is not supposed to change fields in submitted state. Route is an exception.

stock_request/views/res_config_settings_views.xml
```diff
@@ -18,6 +18,19 @@
...
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label string="Allow users to submit stock requests"
+                                       for="group_stock_request_submit"/>
+                                <div class="text-muted">
+                                    Add users to the group "Allow users to submit stock requests".
+                                    This option also allows Inventory users to confirm all stock requests.
+                                </div>
+                            </div>
+                        </div>
```
max3903: @osi-scampbell To be removed. No need to add this.

stock_request/security/stock_request_security.xml
```diff
@@ -26,6 +26,16 @@
         <field name="category_id" ref="base.module_category_hidden"/>
     </record>
 
+    <record id="group_stock_request_submit" model="res.groups">
+        <field name="name">Allow users to submit stock requests</field>
+        <field name="category_id" ref="base.module_category_hidden"/>
+    </record>
```
max3903: @osi-scampbell To be removed.

stock_request/models/res_config_settings.py
```diff
@@ -10,6 +10,13 @@ class ResConfigSettings(models.TransientModel):
     group_stock_request_order = fields.Boolean(
         implied_group='stock_request.group_stock_request_order')
 
+    group_stock_request_submit = fields.Boolean(
+        group='stock.group_stock_user',
+        implied_group='stock_request.group_stock_request_manager')
```
max3903: @osi-scampbell To be removed

stock_request/models/stock_request.py
```diff
@@ -257,7 +269,7 @@ def _action_launch_procurement_rule(self):
         errors = []
         for request in self:
             if (
-                request.state != 'draft' or
+                request.state not in ('draft', 'submit') or
```
max3903:
```suggestion
                request.state not in ('draft', 'submitted') or
```

max3903:
@osi-scampbell Can you fix the Travis errors?

max3903:
@etobella, @rousseldenis @jbeficent @hveficent I believe we have a good compromise here. Please review.

etobella (changes requested):
as demo, you are unable to submit
Please, add a migration script in order to add the group automatically
Also, `stock.request.order` is not modified with this behavior. Change it also please.

max3903:
@osi-scampbell Can you implement @etobella changes please?

osi-scampbell:
@max3903 Yes I'll make the workflow on stock.request.order the same as stock.request and add the migration script

max3903:
@osi-scampbell Here is some doc for the migration script:
https://odoo-development.readthedocs.io/en/latest/maintenance/data-migration.html

stock_request/__manifest__.py
```diff
@@ -22,9 +22,14 @@
         "views/stock_move_views.xml",
         "views/stock_picking_views.xml",
         "views/stock_request_order_views.xml",
+        "data/default.xml",
```
max3903:
```suggestion
```

stock_request/data/default.xml
```diff
@@ -0,0 +1,5 @@
+<odoo>
+    <data noupdate="1">
+            <function model="res.config.settings" name="set_default_parameters"/>
+    </data>
+</odoo>
```
max3903: File to remove.

stock_request/models/res_config_settings.py
```diff
@@ -21,3 +24,18 @@ class ResConfigSettings(models.TransientModel):
...
+        if not settings:
+            settings = self.env['res.config.settings'].create({
+                'group_bypass_submit_request': True,
+            })
+            settings.execute()
+        else:
+            settings = self.env['res.config.settings'].write({
+                'group_bypass_submit_request': True,
+            })
+            settings.execute()
```
max3903: Method to remove

max3903:
@osi-scampbell In addition to my comments, the submitted state and the direction field still need to be added to `stock.request.order`. Thanks!

max3903:
@osi-scampbell Can you squash your commits and fix the Travis error?

Also the default value of the expected date needs to take into account the change of day:
https://github.com/OCA/stock-logistics-warehouse/pull/605/files#diff-cb7ac1537b7c039cfc237796ef052a90R28

stock_request/models/res_config_settings.py
```diff
@@ -21,3 +24,18 @@ class ResConfigSettings(models.TransientModel):
...
```
max3903: To be removed.

stock_request/models/res_config_settings.py
```diff
@@ -1,7 +1,7 @@
 # Copyright 2018 Creu Blanca
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
```
max3903:
```suggestion
from odoo import fields, models
```

stock_request/migrations/12.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    settings = env['res.config.settings'].\
+        search([('group_bypass_submit_request', '=', True)])
+    if not settings:
+        settings = env['res.config.settings'].create({
+            'group_bypass_submit_request': True})
+    settings.execute()
```
max3903: File to be removed

stock_request/data/default.xml
```diff
@@ -0,0 +1,5 @@
+<odoo>
+    <data noupdate="1">
+            <function model="res.config.settings" name="default_parameters"/>
+    </data>
+</odoo>
```
max3903: File to be removed

stock_request/__manifest__.py
max3903:
```suggestion
```

stock_request/views/stock_request_views.xml
```diff
@@ -90,8 +93,10 @@
                         </group>
                         <group>
                             <field name="warehouse_id" widget="selection" groups="stock.group_stock_multi_locations"/>
+                            <field name="direction"/>
                             <field name="location_id" groups="stock.group_stock_multi_locations"
                                    domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]"/>
+                            <field name="direction"/>
```
max3903:
```suggestion
```

stock_request/models/stock_request_order.py
```diff
@@ -111,6 +112,23 @@ def _get_default_requested_by(self):
          'Stock Request name must be unique'),
     ]
 
+    direction = fields.Selection([('outbound', 'Outbound'),
```
HviorForgeFlow: Have you thought to split this changes in a new module? It would be fine to keep stock_request as simple as it can be and delegate specific features to dependent modules.

Just wondering

max3903: @hveficent We have spent quite a bit of extra time on this pull request. Let's have this discussion for the v13 migration. Please create an issue.

HviorForgeFlow: Ok, but then, direction fields seems to be out of the scope of this PR. What is used for?

max3903: It allows the user to specify if the product needs to be shipped to the external location or brought back to the warehouse. Based on the direction, the inventory location is set.

osi-scampbell (approved):

max3903:
@etobella @rousseldenis @hveficent @jbeficent @lreficent This is ready to be reviewed.

JordiBForgeFlow:
There seems to be an issue with Travis. Can you check? Perhaps you need to rebase?

max3903:
@jbeficent This is an issue with stock_request_analytic...

max3903:
@etobella @rousseldenis @hveficent @jbeficent @lreficent This is ready to be reviewed.

@patrickrwilson has been using it in production for 2 months.

stock_request/models/stock_request.py
```diff
@@ -195,7 +221,17 @@ def _action_confirm(self):
 
     @api.multi
     def action_confirm(self):
-        self._action_confirm()
+        # Check if user allowed to skip Submit stage
+        if self.env.user.\
+                has_group('stock_request.group_bypass_submit_request'):
+            self._action_confirm()
+            return True
```
hveficent: Nitpicking: redundant return True

stock_request/models/stock_request.py
```diff
@@ -21,6 +22,15 @@ class StockRequest(models.Model):
     def _get_default_requested_by(self):
         return self.env['res.users'].browse(self.env.uid)
 
+    def _get_default_expected_date(self):
+        # Round time to the next hour
+        t = fields.Datetime.now()
+        if t.hour == 23:
```
etobella: Why do you use this hook? Why just an hour for expected date?

You could use: t + timedelta(hours=1). Just one line and simpler.

stock_request/models/stock_request_order.py
```diff
@@ -50,7 +51,7 @@ def _get_default_requested_by(self):
         states={'draft': [('readonly', False)]})
     location_id = fields.Many2one(
         'stock.location', 'Location', readonly=True,
-        domain=[('usage', 'in', ['internal', 'transit'])],
+        domain=[('usage', 'in', ['internal', 'transit', 'customer'])],
```
etobella: Why customer? Stock request was intended for internal usage

stock_request/models/stock_request.py
```diff
@@ -97,14 +108,29 @@ def _get_default_requested_by(self):
...
+                                 string='Direction',
+                                 states={'draft': [('readonly', False)]},
+                                 readonly=True)
 
     _sql_constraints = [
         ('name_uniq', 'unique(name, company_id)',
          'Stock Request name must be unique'),
     ]
 
+    @api.onchange('direction')
```
etobella: As I said, stock requests should be used for internal movements. If you think it is interesting for outside movements, maybe a stock_request_sale module should be necessary

stock_request/models/stock_request.py
```diff
@@ -195,7 +221,17 @@ def _action_confirm(self):
...
-        self._action_confirm()
+        # Check if user allowed to skip Submit stage
+        if self.env.user.\
+                has_group('stock_request.group_bypass_submit_request'):
+            self._action_confirm()
+            return True
+        else:
+            # Move one state
+            if self.state == 'draft':
+                self.state = 'submitted'
```
etobella: maybe a `self._action_submit()` in order to allow hooks?

stock_request/security/stock_request_security.xml
```diff
@@ -26,6 +26,11 @@
         <field name="category_id" ref="base.module_category_hidden"/>
     </record>
 
+    <record id="group_bypass_submit_request" model="res.groups">
```
etobella, line 29: A post-migration script is necessary in order to set that this group is set automatically on migrations in order to not change the current behavior. This should only apply on installed version

etobella (changes requested):
Just a technical review. When this is answered I will make the functional review.

Originally, stock requests were created to allow movements between internal locations. You are allowing to send products to customers, but, I don't think it is a right solution, as no account movements will be generated. If you want to add this, I think you should add a new module.

stock_request/models/stock_request_order.py
max3903: @etobella Because we use stock requests in field service and field service locations sometimes belong to the customer.

stock_request/models/stock_request.py
max3903: I don't understand why you want to limit stock requests to internal moves when it can apply to a broader scope...

stock_request/security/stock_request_security.xml
max3903: @etobella it's already part of this pull request:
https://github.com/OCA/stock-logistics-warehouse/pull/605/files#diff-1e2d0f1127b658c446c2268123be061e

stock_request/models/stock_request.py
etobella: Because it is not a sale. It will not generate invoices or sale orders.

stock_request/security/stock_request_security.xml
etobella: True, I missed the file. Sorry

stock_request/models/stock_request_order.py
etobella: But you cannot change the behavior of a generic module for an specific module. If you need this change, make it there.

JordiBForgeFlow: @max3903 IMHO this would cause mistakes from users that would incorrectly enter a customer location, leading to undesired results. 

I am sure that this makes complete sense to your use case, but perhaps the field 'direction' or the 'customer' location type  should be added in a field service specific module, or even should be part of your own customization.

etobella: As I said in other comments, If you need this kind of changes, you should make in specific modules. You are creating a hole for most companies because field service companies uses this behavior. I think it is a bad idea for a generic module as `stock_request`. What do you think @jbeficent as creator of this module?

max3903: @etobella I am sorry. I am changing the behavior of a specific module (that only supports internal moves) to be more generic (and supports external ones as well).

stock_request/models/stock_request.py
max3903: I am not sure to understand why it matters for the stock request module. Even an internal move can generate others documents, for example an accounting entry. I had a customer who tracked 2 internal locations in 2 different financial accounts.

stock_request/models/stock_request_order.py
etobella: @max3903 You should be able to see that a module that allows users to make external account moves without using sales or invoices can be a problem for a lot of companies. I think this a good reason to notice the problem and think if we need a new module.

I know that sometimes comments are not satisfying for everyone, but we should try to think for all the people that uses OCA, not only our customers or company.

JordiBForgeFlow: @max3903 we implemented a setting that allows users to enter any location, not only internal. You have to activate it here:
https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_request/models/res_config_settings.py#L19

And it's used here:
https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_request/models/stock_request_abstract.py#L188

So, a normal user would not see this 'special' behavior unless they activate the setting.

You could make use of this feature, or we can find a way that works for all.

stock_request/models/stock_request.py
JordiBForgeFlow: I think that the field 'direction' is very opinionated. A normal user would not understand why is this needed. IMHO this should be a customer-specific customization.

max3903:
@etobella @rousseldenis @hveficent @jbeficent @lreficent 

I believe we took care of all your comments. Are we good to merge this now?

JordiBForgeFlow:
@max3903 thank you very much for your patience on this. We identified a strong conflict with the stock_request_tier_validation, and we're proposing you https://github.com/OCA/stock-logistics-warehouse/pull/675.

PR #611, opened 2019-06-07
[IMP] stock_available_mrp: Easier and faster BoM finding

yajo:
Before this patch, the method was looping through found BoMs in each product, which produced an exponential double loop.

In a first attempt, I tried just avoiding the double loop, but then I noticed that with the inclusion of https://github.com/odoo/odoo/pull/33967, just calling upstream's method was easier, safer (takes more conditions into account, such as the company) and faster (due to cache, which also benefited further calls).

So here is the patch. See :fire: :chart_with_upwards_trend: here:

Captura de pantalla de 2019-06-07 10-06-43

This should be WIP until this one is merged, or possibly performance will be penalized:

- [ ] https://github.com/odoo/odoo/pull/33967

@Tecnativa

### DIFF
--- a/stock_available_mrp/models/product_product.py
+++ b/stock_available_mrp/models/product_product.py
@@ -22,35 +22,10 @@ class ProductProduct(models.Model):
         super(ProductProduct, self)._compute_available_quantities()
 
     @api.multi
-    def _get_bom_id_domain(self):
-        """
-        Real multi domain
-        :return:
-        """
-        return [
-            '|',
-            ('product_id', 'in', self.ids),
-            '&',
-            ('product_id', '=', False),
-            ('product_tmpl_id', 'in', self.mapped('product_tmpl_id.id'))
-        ]
-
-    @api.multi
     @api.depends('product_tmpl_id')
     def _compute_bom_id(self):
-        bom_obj = self.env['mrp.bom']
-        boms = bom_obj.search(
-            self._get_bom_id_domain(),
-            order='sequence, product_id',
-        )
-        for product in self:
-            product_boms = boms.filtered(
-                lambda b: b.product_id == product or
-                (not b.product_id and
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't an alternative patch without needing an Odoo patch?

yajo:
Yes, we could add cache to that method downstream. But I think we can wait a little bit and see what do they do.

rafaelbn:
Dear @Cedric-Pigeon @lmignon @rousseldenis are you experimenting this issues? Please check image to get a feedback from your customers please?

rousseldenis:
@rafaelbn @Yajo I did also a PR to improve _bom_find with a real multi implementation there : https://github.com/odoo/odoo/pull/26400

rafaelbn:
Hi @rousseldenis , thanks! We didn't notice that. And it looks like Odoo didn't answer :disappointed:  1 year performance improvement without answer. Did you open an OPW about that? It's in fact a problem in EE

rousseldenis:
> Did you open an OPW about that?

Nope, as it was for v10, I think they'll never change it. 

IMHO, implicit calls to clear_cache() should be avoided. It highlights maybe a problem inside cache implementation and should be fixed there.

Moreover, the code removed here was done to improve performances (it was worst before), see this https://github.com/OCA/stock-logistics-warehouse/pull/280/files.

Maybe a ping to @rco-odoo on https://github.com/odoo/odoo/pull/33967 should be done.

The problem is Boms inside boms.

rousseldenis:
@Cedric-Pigeon Do you have an opinion on this ?

Cedric-Pigeon (approved):
@rousseldenis From my point of view it seems healthy as a patch

rousseldenis (approved):
Code review

PR #612, opened 2019-06-09
[WIP] product cost evaluation

sergiocorato:
[WIP] this module add the ability to evaluate stock with different methodology (FIFO, LIFO, AVERAGE, STANDARD, LIST PRICE) for the same period, regardless of which method was chosen previously, storing in records that will be re-used for next period evaluation
[TODO]:
- use sql to reduce time of execution for large databases
- stop searching moves when previous period has already an evaluation
- create report for one single evaluation at a time


dcorio:
@sergiocorato shouldn't it be wiser to have those methods elsewhere rather than stock.inventory? You'll soon end up needing the very same on other models like manufacturing order & C.

For instance, you could move those on stock.move to be more reusable.

IMO.

sergiocorato:
> @sergiocorato shouldn't it be wiser to have those methods elsewhere rather than stock.inventory? You'll soon end up needing the very same on other models like manufacturing order & C.
> 
> For instance, you could move those on stock.move to be more reusable.
> 
> IMO.

@dcorio ok for move in the same new object created with this PR

rousseldenis:
@sergiocorato What's the status of this ?

sergiocorato:
@rousseldenis There is another module in testing in production, it will be proposed within 1 month probably.

rousseldenis:
> @rousseldenis There is another module in testing in production, it will be proposed within 1 month probably.

So, closing this ?

sergiocorato:
Yes, I close it.

PR #614, opened 2019-06-12
[11.0][IMP] procurement_auto_create_group: ensure that the procurement group is only created in the procurment process.

LoisRForgeFlow:
Add a context to be able to disable this moudule functionality on demand.

Sometimes the method _get_rule is just needed to obtain the information of the rule,
and it isn't followed by the run of the procurement.

@Eficent


AdriaGForgeFlow (approved):
:+1:

JordiBForgeFlow (approved):

rousseldenis:
@jbeficent Don't forget to add 'patch/minor/major' argument when PR concerns changes in existing module.

PR #616, opened 2019-06-13
[IMP] stock_request: Improve cancellation detection

etobella:
If you had a Stock Request with 2 units, and you served only one unit, the backorder was lost from the Stock REquest side. Also, if you cancelled the backorder (or the first picking), you should show it somehow on the stock request. Now it is setted as done when the picking is cancelled.


etobella:
@jbeficent

JordiBForgeFlow (approved):

JordiBForgeFlow:
@ageficent can you review?

HaraldPanten (approved):
@jbeficent  Functional review on runbot 👍

Issue #619, opened 2019-06-15
Migration of stock_reserve_sale to v12 but found bugs


hzh0292:
[Odoo module stock_reserve_sale migration to verssion 12.0](https://github.com/hzh0292/stock-logistics-warehouse)

1. The basic module stock_reserve is migrated from OCA/stock-logistics-warehouse branch 10.0.
2. The module stock_reserve_sale is migrated from OCA/stock-logistics-warehouse branch 9.0.Because the branch 10.0 of this module is unfinished yet and it was based version 8.0.
3. I replaced the old model 'procurement.rule' to 'stock.rule' in stock_reserve_sale/model/sale.py,and did some normal odoo 12.0 migration stuff.
4. Because I don't know much about odoo module development, I didn't modified the logical code.
5. This two modules can installed in odoo 12,and most of the functions work well.But I found a bug that when you create a stock reservation from a sale order which contain 2 or more lines,there will be a 'Odoo Server Error' show that Record does not exist or has been deleted.And the console will show a message like 'odoo.models.unlink: User #1 deleted stock.move records with IDs: [41]'.

![](https://i.bmp.ovh/imgs/2019/06/e3ec4a8e3bffe28f.png)

![](https://i.bmp.ovh/imgs/2019/06/c289c3ea59781563.png)

hzh0292:
Can somebody help me solve the bugs?

damendieta:
Hi, do the MR in order to have this migration there.

I'll try to install the modules and check the possible causes of the issue.

Daniel.

hzh0292:
> Hi, do the MR in order to have this migration there.
> 
> I'll try to install the modules and check the possible causes of the issue.
> 
> Daniel.

Thanks very much.It seems that I don't know how to do the MR.

hzh0292:
> Hi, do the MR in order to have this migration there.
> 
> I'll try to install the modules and check the possible causes of the issue.
> 
> Daniel.

I have created a new pull request to this repo.Is it the right way to do a MR?[#620](https://github.com/OCA/stock-logistics-warehouse/pull/620)

damendieta:
Hi, I'm checking the code and I have some comments:

- stock_reserve_sale should depend on sale, as you are inheriting sale.order
- you have except_orm but it's deprecated, replace it with an exception from the new api.
- The error is happening on line 140, of stock_reserve.py 
   self.mapped('move_id')._action_confirm()

After the move is confirmed, the move is unreacheable.

[...]

PR #620, opened 2019-06-16
[MIG] stock_reserve_sale: Migration to 12.0

hzh0292:
Migration of stock_reserve_sale to v12 but found bugs [#619](https://github.com/OCA/stock-logistics-warehouse/issues/619)
Ask the great community to help me with the migration.


stock_reserve_sale/__manifest__.py
```diff
@@ -0,0 +1,43 @@
...
+{'name': 'Stock Reserve Sales',
+ 'version': '12.0.1.0.0',
+ 'author': "Camptocamp,Odoo Community Association (OCA)",
+ 'category': 'Warehouse',
+ 'license': 'AGPL-3',
+ 'complexity': 'normal',
+ 'images': [],
+ 'website': "http://www.camptocamp.com",
+ 'depends': ['sale_stock',
+             'stock_reserve',
```
damendieta: add sale to the dependencies

stock_reserve/model/stock_reserve.py
```diff
@@ -0,0 +1,197 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from odoo import models, fields, api, _
+from odoo.exceptions import except_orm
```
damendieta: except_orm is deprecated, use UserError or Warning or any that suits your needs from the new api

stock_reserve/model/stock_reserve.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @api.multi
+    def reserve(self):
+        """ Confirm reservations
+
+        The reservation is done using the default UOM of the product.
+        A date until which the product is reserved can be specified.
+        """
+        self.write({'date_expected': fields.Datetime.now()})
+        self.mapped('move_id')._action_confirm()
```
damendieta: here you have the problem

I tested the following:
- If the product is storable and you have the qty to reserve, everything works.
- If the product is consumable it fails.
- If the product is storable but you have not enough qty available, it fails again.

So, you need to:
- Ad a filter in product to allow only storable products.
- Add a constrain in qty to reserve to avoid the user to ask a reservation to more products than available.
- Catch the error in order to avoid a problem is anyone else consumes the available qty before you confirm.

pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

stock_reserve/model/stock_reserve.py
```diff
@@ -0,0 +1,197 @@
...
+        self.mapped('move_id.picking_id').action_assign()
+        return True
+
+    @api.multi
+    def release(self):
+        """
+        Release moves from reservation
+        """
+        self.mapped('move_id').action_cancel()
```
damendieta: Here you have an error with action_cancel, it's _action_cancel, so you need to add the _

damendieta (changes requested):

stock_reserve/model/stock_reserve.py
hzh0292: Is ValidationError the suitable new api?

stock_reserve_sale/__manifest__.py
hzh0292: Module sale_stock depends sale,so it's not need to add sale dependency.

stock_reserve/model/stock_reserve.py
hzh0292: When the product is storable and you have the qty to reserve,not really everything works.When I create a sale order which contain two storable products,when I click the reserve stock button to create reversions and add a validity date,it will raise the error.

damendieta:
@hzh0292 before doing any other commit, please follow @pedrobaeza indications.

Your PR will not be accepted if you don't do it.

hzh0292:
> @hzh0292 before doing any other commit, please follow @pedrobaeza indications.
> 
> Your PR will not be accepted if you don't do it.

Yes, I know. Very sorry for that I am not good at this work. I will try.

stock_reserve/model/stock_reserve.py
damendieta: Yes, but is this a validation error?

stock_reserve/model/stock_reserve.py
damendieta: You are right, I didn't tested that as I thought the problem was the same.

The problem seems to be that you have several records and action confirm fails.

Thy doing a loop, instead of lines 140 and 141 try:
```

for m in self.mapped('move_id')
    m._action_confirm()
    m.picking_id.action_assign()

```
It's posible that it will fail because you may not have a picking on the move, if that's the case, try:

```
for m in self.mapped('move_id')
    m._action_confirm()
    if m.picking_id:
        m.picking_id.action_assign()
```

stock_reserve/model/stock_reserve.py
hzh0292: I an not sure. What's your advice?

stock_reserve/model/stock_reserve.py
damendieta: I was thinking about this and most of the time we will have one picking for all moves or no move at all, so, we can do:

```
for m in self.mapped('move_id')
    m._action_confirm()
for p in self.mapped('move_id.picking_id')
        p.action_assign()
```

hzh0292: OK,I'll try.Thank you very much.

hzh0292:
```python
for m in self.mapped('move_id'):
    m._action_confirm()
for p in self.mapped('move_id.picking_id'):
    p.action_assign()
```
This didn't solve the issue that create a stock reservation from a sale order with multi lines.

damendieta: Ok, I checked the problem. The problem is not because of multiple lines.

In v12 _action_confirm is defined like this _action_confirm(self, merge=True, merge_into=False), while in v10, it's def action_confirm(self).

As you can see, you have 2 new params in the definition, one of them is True by default.

For that reason, it merges the move wich deletes the original move.
Try this: `m._action_confirm(merge=False, merge_into=False)`

stock_reserve/model/stock_reserve.py
damendieta: you can remove the import line and leave only the ValueError on line 118
`except ValueError:`

Because your are not raising the error but catching it.

stock_reserve/model/stock_reserve.py
hzh0292: Great!!!It works.You are so helpful,thank you very much.

damendieta:
@hzh0292 I have seen a design flaw on this module, it's designed to "reserve" the quantities, but not for them to be moved to the reservation location. It meas that the stock can be reserved but not moved to a different location.

I't important to have the flow with the option for the move to be confirmed and moved, so, we have two options.

1. If the move is not confirmed, the actual workflow is fine.
2. If the move is done, you will be presented with a "You cannot cancel a stock move that has been set to 'Done'. " error, and you will not be able to undo the reservation.

My suggestion will be to add a condition to the code, like this:
First, add float_round and decimal_precision to your imports.

``` python
from odoo.addons import decimal_precision as dp
from odoo.tools.float_utils import float_round
```
Then, replace the release function:
``` python
    @api.multi
    def release(self):
        """
        Release moves from reservation
        """
        # First we cancel all reserved moves.
        self.mapped('move_id').filtered(
            lambda x: x.state not in ('done', 'draft', 'cancel')
        )._action_cancel()
        # If the move is done, we need to reverse it.
        done_moves = self.mapped('move_id').filtered(lambda x: x.state == 'done')
        # We group them by picking to have one revesed picking for all moves.
        for picking in done_moves.mapped('picking_id'):
            return_obj = self.env['stock.return.picking']
            product_return_moves = []

            if picking.state != 'done':
                raise UserError(_("You may only return Done pickings."))

            for move in done_moves.filtered(lambda x: x.picking_id == picking):
                if move.scrapped:
                    continue

                move_dest_exists = False
                if move.move_dest_ids:
                    move_dest_exists = True

                quantity = move.product_qty - sum(move.move_dest_ids.filtered(
                    lambda m: m.state in [
                        'partially_available', 'assigned', 'done'
                    ]).mapped('move_line_ids').mapped('product_qty')
                )
                quantity = float_round(
                    quantity,
                    precision_rounding=move.product_uom.rounding
                )
                product_return_moves.append(
                    (0, 0, {
                        'product_id': move.product_id.id,
                        'quantity': quantity,
                        'move_id': move.id,
                        'uom_id': move.product_id.uom_id.id
                    }))

            location_id = picking.location_id.id
            if picking.picking_type_id.return_picking_type_id \
                    .default_location_dest_id.return_location:

                location_id = picking.picking_type_id.return_picking_type_id \
                    .default_location_dest_id.id

            parent_location_id = picking.picking_type_id.warehouse_id \
                and picking.picking_type_id.warehouse_id.view_location_id.id \
                or picking.location_id.location_id.id

            wizard = return_obj.create({
                'picking_id': picking.id,
                'location_id': location_id,
                'product_return_moves': product_return_moves,
                'move_dest_exists': move_dest_exists,
                'parent_location_id': parent_location_id,
                'original_location_id': picking.location_id.id,
            })
            wizard._create_returns()
        return True
```
This code is working, but I didn't tested it too much.

I hope it helps.

hzh0292:
@damendieta I have updated the code.Please check it and give me some advice about the code adjustment and the work I should do following [odoo 12 migration guidline](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0) .

damendieta:
Hi, just follow the steps Pedro give you.

The module is almost complete with the modifications made.

hzh0292:
> Hi, this default has no sense.
> 
> If you are asking for a reservation, it's not for the same day you are asking for the reservation.
> 
> In case you want a default, you should have a configuration to have a default reservation period on the company or no default at all. Why do you need this?

OK

damendieta:
@hzh0292 please, don't forget to migrate the addons as Pedro told you, so we can merge this PR soon.

I know it's a little work, but it's important to merge the modules.

Thanks.

hzh0292:
> @hzh0292 please, don't forget to migrate the addons as Pedro told you, so we can merge this PR soon.
> 
> I know it's a little work, but it's important to merge the modules.
> 
> Thanks.

I am so sorry.But I am really don't know which part of work I should do next,so please tell me how can I do to finish it.I have seen the documentation of OCA,but it seems that I have problem of totally following this.

pedrobaeza:
@hzh0292, you must preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

You will need to start again from a fresh branch, doing that method, and then you can copy current code of this pull request and commit it with only the changes fro 12.0.

hzh0292:
> @hzh0292, you must preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0
> 
> You will need to start again from a fresh branch, doing that method, and then you can copy current code of this pull request and commit it with only the changes fro 12.0.

Thank you very much.I have started again.Should I make a new PR?

pedrobaeza:
If you preserve branch name and then forced push here, there's no need of closing the pull request.

hzh0292:
> If you preserve branch name and then forced push here, there's no need of closing the pull request.

But I have deleted the old repo and forked again.

pedrobaeza:
Yeah, in that case you'll need to open a new PR. Closing this one.

PR #623, opened 2019-06-21
[12.0][MIG] stock_inventory_exclude_sublocation: Migration to 12.0

bodedra:

PR #622, opened 2019-06-21
[12.0][MIG] stock_inventory_discrepancy: Migration to 12.0 

bodedra:

PR #625, opened 2019-06-24
[12.0][MIG] stock_cycle_count: Migration to 12.0

bodedra:
Depends on:
------------------

- [x] https://github.com/OCA/stock-logistics-warehouse/pull/622 stock_inventory_discrepancy
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/623 stock_inventory_exclude_sublocation


bodedra:
@jbeficent @lreficent @aheficent 

Would you please help me find out root cause to fail `('Zero confirmation not being created.')` test case ?

Screenshot_2019-06-25 Job #3890 3 - OCA stock-logistics-warehouse - Travis CI

LoisRForgeFlow:
@bodedra just opened a PR to fix the remaining issue: https://github.com/ursais/stock-logistics-warehouse/pull/2. Could you review?

bodedra:
@lreficent 

480b6216b16fb1a21acc318c3d805f53

Thanks for your fix.

Now all test case are passed. Still ci/runbot fail.

stock_cycle_count/reports/report_stock_location_accuracy.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class LocationAccuracyReport(models.AbstractModel):
+    _name = "report.stock_location_accuracy"
```
MiquelRForgeFlow, line 9:
```suggestion
    _name = "report.stock_location_accuracy"
    _description = "Location Accuracy Report"
```

bodedra: Done. @mreficent

stock_cycle_count/models/stock_warehouse.py
```diff
@@ -40,8 +39,7 @@ def get_horizon_date(self):
     @api.model
     def _get_cycle_count_locations_search_domain(
             self, parent):
-        domain = [('parent_left', '>=', parent.parent_left),
-                  ('parent_right', '<=', parent.parent_right),
+        domain = [('parent_path', '=like', parent.parent_path + '%'),
```
rousseldenis, line 42: I'm wondering why child of/parent of would not works here ?

See  https://github.com/odoo/odoo/blob/12.0/odoo/osv/expression.py#L755

bodedra: @rousseldenis  I use it as a reference of https://github.com/odoo/odoo/blob/12.0/addons/stock/models/product.py#L206 and https://github.com/odoo/odoo/blob/12.0/addons/stock/models/product.py#L211

rousseldenis: Yeah, that's weird because it's in the same commit that implements parent_path to optimize queries...

rousseldenis (approved):
Code review

rousseldenis:
@bodedra Could you squash your commits ?

stock_cycle_count/models/stock_warehouse.py
bodedra: Additionally, parent_left and parent_right fields are removed from stock.location object.

bodedra:
@rousseldenis Sure.

stock_cycle_count/models/stock_warehouse.py
rousseldenis: Yes, that was the aim of parent_path: to optimize child of/parent of search operators, so, normally, you should use them as before.

bodedra: so what do you recommend on these ?

MiquelRForgeFlow (approved):

bodedra:
@rousseldenis @mreficent Thanks for your review and suggestion.

@lreficent can you update your review ?

LoisRForgeFlow (approved):
LGTM (just fixed history a little to recover my contribution to the migration)

Let's merge after travis goes green.

PR #633, opened 2019-06-27
[WMS][12.0] Add vertical lift support - alpha version

guewen:
This is a draft pull request, absolutely not production-ready, and does not meet yet the standards in term of documentation. At the moment, this is for demo purpose, the Pick workflows more or less works for a demo, put and inventory modes are not implemented. It implements a poc for https://github.com/OCA/stock-logistics-warehouse/issues/641

The goal is to support Vertical Lift such as [Kardex](https://www.kardex-remstar.com/en/storage-retrieval-systems/vertical-lift-systems-en/xp250500.html) or similar systems.

It adds a stock location structure for the Vertical Lift "shelves", with different levels:
Vertical Lift → Shuttles → Trays → Cells. The Cells are automatically generated from Tray Types (example: 4 cols and 2 rows) selected on the trays.

Selection_649

A visual representation of a tray displays the selected cell and color the cells that contain stock.

Selection_650

The initial proposal assumes that each Shuttle will be driven by a screen displaying the operations to do (pick or put, inventory will need some work as well). Here's the (incomplete) dedicated UI, with the information about the product to pick + visualization of the tray and cell to pick.

Selection_651


damendieta:
This is a really nice idea, please, tag me if you want me to help with something.

Daniel.

jgrandguillaume:
Functional test, works well and as expected for this first version. Feedback will come after customer meeting.

jgrandguillaume:
Remark to no forget it, the vertical_lift_kind should use the "generic" kind from https://github.com/OCA/stock-logistics-warehouse/pull/653

damendieta:
Hi, amazing job.

I have done some test and I have a few ideas:

- When a location is a tray location, we can require a Z coordinate and propagate it to the cells. It will not use the coordinates as Odoo want's us to (corridor, shelf, height), but you are not doing it anyway with x or y. We can use the Z coordinate to display the position of the tray in the warehouse.
- We should not allow to store products on tray or shuttle locations, by setting a constraint or by forcing them to be view locations or we may end up having stock in the tray instead of the cell.
- In the data folder you have a default location and some default tray types. Even when they are on a noupdate tag, maybe it's better to have them as a separated module because those default values are not standard, so, it may not be the case for most users. A separated module will avoid having to install the module and remove the default values as the first step of configuration.

If you need my help please tell me what can I do.

Daniel.

damendieta:
Another annotation.

In the naming you have x#y# as the only option, but, most companies will want to customize the cell name, for example, 1-1 or x01y01 or X01/Y01.

Maybe it's impossible to allow all combinations required, but maybe we can do:

- Add a get_cell_name function to make the name easier to customize.
- Add some parameters in the shuttle (or tray) for naming conventions like: x_prefix, y_prefix, padding.

For example:

x_prefix = X
y_prefix = /Y
padding = 2

Will lead to a name: X01/Y01

Daniel.

damendieta:
Some of the ideas have been implemented here as a proof of concept.

https://github.com/guewen/stock-logistics-warehouse/pull/1

jgrandguillaume:
Hi @damendieta 


Thanks for your reviews.

>     * When a location is a tray location, we can require a Z coordinate and propagate it to the cells. It will not use the coordinates as Odoo want's us to (corridor, shelf, height), but you are not doing it anyway with x or y. We can use the Z coordinate to display the position of the tray in the warehouse.

-> I guess you mean Z will then be used to display the position of the TRAY in the Shuttle ? As you can see in this work (https://github.com/OCA/stock-logistics-warehouse/issues/654) we kind of redefine how will will actually manage the corridor, shelf, etc.. So I guess your proposal is feasible.
Question: Why do you need to know the Z of a try (because we don't need that over here) ?

>     * We should not allow to store products on tray or shuttle locations, by setting a constraint or by forcing them to be view locations or we may end up having stock in the tray instead of the cell.

-> I would tend to agree to that, though we had several bad experience adding location as view so far. May be @gurneyalex you have more infos ?

>     * In the data folder you have a default location and some default tray types. Even when they are on a noupdate tag, maybe it's better to have them as a separated module because those default values are not standard, so, it may not be the case for most users. A separated module will avoid having to install the module and remove the default values as the first step of configuration.

-> +1 for this proposal, or even move this to demo data, then add a paragraph in the configuration to ask people to create their tray type.

jgrandguillaume:
> In the naming you have x#y# as the only option, but, most companies will want to customize the cell name, for example, 1-1 or x01y01 or X01/Y01.
> 
> Maybe it's impossible to allow all combinations required, but maybe we can do:
> 
>     * Add a get_cell_name function to make the name easier to customize.
> 
>     * Add some parameters in the shuttle (or tray) for naming conventions like: x_prefix, y_prefix, padding.
> 
> 
> For example:
> 
> x_prefix = X
> y_prefix = /Y
> padding = 2
> 
> Will lead to a name: X01/Y01

Good suggestion of improvements. This one we may not be able to release our-self, but we'll happily review and accept any contribution through a PR.

Thanks again for your input.

damendieta:
>I guess you mean Z will then be used to display the position of the TRAY in the Shuttle ? As you can see in this work (#654) we kind of redefine how will will actually manage the corridor, shelf, etc.. So I guess your proposal is feasible. 
Question: Why do you need to know the Z of a try (because we don't need that over here) ?
>
Yes, that's the idea, to be able to plan a sequence for the picking. For example, if tray A and B share the same Z position, I can pick up products on A1 and B1 at the same time.

> I would tend to agree to that, though we had several bad experience adding location as view so far. May be @gurneyalex you have more infos ?
>

Me too, for that I think we should avoid using view but restrict products directly on some internal locations. So, we can Add products on A/x1y1 but not on A directly.

damendieta:
>Good suggestion of improvements. This one we may not be able to release our-self, but we'll happily review and accept any contribution through a PR.>

PR already done.

jgrandguillaume (approved):
LGTM, functional test ok. Needs the development status

guewen:
Extracted the tray part in https://github.com/OCA/stock-logistics-warehouse/pull/719
I'm about to rework this PR to use the new addon `stock_location_tray` as dependency.

guewen:
> I'm about to rework this PR to use the new addon `stock_location_tray` as dependency.

Done

stock_vertical_lift/models/vertical_lift_shuttle.py
```diff
@@ -0,0 +1,413 @@
...
+        readonly=True,
+    )
+
+    # TODO add a glue addon with product_expiry to add the field
+
+    _barcode_scanned = fields.Char(
+        "Barcode Scanned",
+        help="Value of the last barcode scanned.",
+        store=False,
+    )
```
sebalix: Is there a reason to redeclare this field as we're inheriting from [`barcodes.barcode_events_mixin`](https://github.com/odoo/odoo/blob/12.0/addons/barcodes/models/barcode_events_mixin.py#L15)?

stock_vertical_lift/models/stock_location.py
```diff
@@ -1,40 +1,131 @@
...
+        """
+        self.ensure_one()
+        if self.vertical_lift_kind == "cell":
+            if cell_location:
+                raise ValueError(
+                    "cell_location cannot be set when the location is "
+                    "a cell."
+                )
+            tray = self.location_id
+            tray.fetch_vertical_lift_tray(cell_location=self)
```
guewen, line 153: @gurneyalex I don't think we need to have an additional method `fetch_vertical_lift_tray_show_cell()` since we can call `fetch_vertical_lift_tray()` on the cell directly

guewen: let me know if you find it confusing though

stock_vertical_lift/models/vertical_lift_shuttle.py
guewen: hmm no

stock_vertical_lift/models/stock_location.py
gurneyalex: that's fine with me

stock_vertical_lift/static/src/js/vertical_lift.js
```diff
@@ -44,6 +45,53 @@ var ExitButton = FieldInteger.extend({
             clear_breadcrumbs: true,
         });
     },
+
+});
+
+
+FormController.include({
```
guewen, line 52: Note: I tried to use a widget to handle the notifications handling and insert it in the view (rather than using the `FormController` for this), but on `this.reload()`, the 'start' method of the new widget was called before the 'destroy' of the previous one, so the channel was de-registered of the `bus_service`.

guewen:
Work is continued in 13.0 in #797

PR #642, opened 2019-07-05
[FIX] calculate the inventory value based on total cost

JosDeGraeve:
When merging quants, the current behaviour was to calculate the
average cost for the merged quant's cost.

The problem with this that it does not keep the total inventory
value correct.  If we merge a quant with 100 units at a cost of 10
having a total value of 1000 with a quant with 2 units a a cost of 20
having a total value of 40, the old behaviour would generate a quant
with a cost of 15 resulting in a total value of 1530.

With this update the total value will stay at 1040.


rousseldenis:
@JosDeGraeve Could you add some lines in test to highlight this behaviour ?

rousseldenis:
@JosDeGraeve

JosDeGraeve:
Well,

I spent the better part of the day trying to come up with an easy way to test it, but i'm giving up.

PR #643, opened 2019-07-05
[FIX] fix negative quant in blocked location

sebastienbeau:
When forcing an outgoing move and then moving it. Odoo create a
negative quant. This quant should not be in the locked location
because no quand will go in it and so the negative quant will stay
here for ever

@florian-dacosta @alexis-via @bealdav

### DIFF
--- a/stock_location_lockdown/models/stock_quant.py
+++ b/stock_location_lockdown/models/stock_quant.py
@@ -9,18 +9,13 @@ from odoo.exceptions import UserError
 class StockQuant(models.Model):
     _inherit = 'stock.quant'
 
-    @api.model
-    def quants_move(
-            self, quants, move, location_to, location_from=False, lot_id=False,
-            owner_id=False, src_package_id=False, dest_package_id=False,
-            entire_pack=False):
-        if (
-                location_to.usage == 'internal' and
-                location_to.block_stock_entrance):
-            raise UserError(_(
-                "The location '%s' is not configured to receive stock.")
-                % location_to.display_name)
-        return super(StockQuant, self).quants_move(
-            quants, move, location_to, location_from=location_from,
-            lot_id=lot_id, owner_id=owner_id, src_package_id=src_package_id,
-            dest_package_id=dest_package_id, entire_pack=entire_pack)
+    @api.constrains('location_id')
+    def _check_location_blocked(self):
+        for record in self:
+            if record.location_id.block_stock_entrance:
+                raise UserError(
+                    _('The location %s is blocked and can '
+                      'not be used for moving the product %s')
+                    % (record.location_id.name, record.product_id.name)
+                    )
+        return True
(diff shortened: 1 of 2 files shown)


stock_location_lockdown/models/stock_quant.py
```diff
@@ -9,18 +9,13 @@
...
-                "The location '%s' is not configured to receive stock.")
-                % location_to.display_name)
-        return super(StockQuant, self).quants_move(
-            quants, move, location_to, location_from=location_from,
-            lot_id=lot_id, owner_id=owner_id, src_package_id=src_package_id,
-            dest_package_id=dest_package_id, entire_pack=entire_pack)
+    @api.constrains('location_id')
+    def _check_location_blocked(self):
+        for record in self:
+            if record.location_id.block_stock_entrance and record.location_id.usage == 'internal':
```
florian-dacosta: The second condition seems useless. If `block_stock_entrance` is checked, we do not want quant anyway, whatever the usage, no?

florian-dacosta (commented):
Just one minor comment, otherwise it seems fine!

stock_location_lockdown/models/stock_quant.py
sebastienbeau: your right

florian-dacosta (approved):

rousseldenis:
@sebastienbeau Could you check travis?

sebastienbeau:
@rousseldenis Travis is green

stock_location_lockdown/models/stock_quant.py
```diff
@@ -3,24 +3,19 @@
...
-                "The location '%s' is not configured to receive stock.")
-                % location_to.display_name)
-        return super(StockQuant, self).quants_move(
-            quants, move, location_to, location_from=location_from,
-            lot_id=lot_id, owner_id=owner_id, src_package_id=src_package_id,
-            dest_package_id=dest_package_id, entire_pack=entire_pack)
+    @api.constrains('location_id')
+    def _check_location_blocked(self):
+        for record in self:
+            if record.location_id.block_stock_entrance:
```
rousseldenis, line 15: @sebastienbeau You removed the check on 'usage' field. Why ?

florian-dacosta: @rousseldenis 
I suggested it in my review.
I think it is useless, the boolean `block_stock_entrance` is enought to know the location should not contain any quant.
When I made this module in v8, the case I wanted to resolve was really linked to internal location, and so I guess I put this condition, but actually, it shoud also concern view locations, and why not others...

rousseldenis (approved):
Code review

rousseldenis:
I will merge with minor bump if you don't have any objection

sebastienbeau:
ok for me

PR #644, opened 2019-07-06
[12.0][stock_available_mrp] Migration + make compatible with other stock_available_immediately

florian-dacosta:
Hi, This PR has 2 commits.
First one is standard migration, second one try to solve a problem I have detected.

If I understand well, the purpose code in product_template.py that I removed (https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_available_mrp/models/product_template.py#L14), was to be sure, in case a template with variants, that the immediately usable qty of our bom is the sum of all real planned qties of the variable (by default, virtual qty) + the potential quantity, which is the highest potential of its variants.

Indeed, we see in method `_compute_available_quantities_dict` of stock_available module :  https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_available/models/product_template.py#L29
that the computed immediately_usable_qty of our template is the sum of all quantities of the variants, but, the quantities of the variant, already include the potential quantity ! 
See : https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_available_mrp/models/product_product.py#L122

So, without this code (https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_available_mrp/models/product_template.py#L14) we add in template qty_immediately_usable the potential qty of all variants even though we want only to count the highest potential qty of the variant...

Anyway, the problem is that it makes it incompatible with others modules like stock_available_immediately. (and maybe others).
Indeed, stock_available_immediately remove the incoming quantity from the virtual quantity.
But the patch I removed (stock_available_immediately) go from virtual quantity once more...

That's what I have tried to correct in my second commit. For this I need to modify the logic in stock_available.
(+ make impossible to have a negative potential quantity, because I guess it makes no sense...but this have nothing to do with the problem described!)

It seems to work fine and the test passes without any problem.
But any opinion is welcome. @lmignon @Cedric-Pigeon @rafaelbn @tschanzt @simahawk @grindtildeath


rousseldenis:
@Cedric-Pigeon

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,150 @@
...
+                # We want to respect the rounding factor of the potential_qty
+                # Rounding down as we want to be pesimistic.
+                potential_qty = bom_id.product_uom_id._compute_quantity(
+                    potential_qty,
+                    product.bom_id.product_tmpl_id.uom_id,
+                    rounding_method='DOWN'
+                )
+
+            res[product.id]['potential_qty'] = potential_qty
+            res[product.id]['immediately_usable_qty'] += potential_qty
```
sergio-teruel, line 123: I not sure of this...
Selección_302
I have a bom for a glass cookcing table
Selección_301
Also, I create a work order like this
Selección_303
As yo can see... 
inmediately_usable_qty is 1 because is virtual_available and you increase all potential_qty so inmediately_usable_qty is 6.. Is correct?

florian-dacosta: hello @sergio-teruel 
Yes it is the behavior of the module, I did not change this.
The potential quantity is added to the immediately_usable_qty and it is intended.

rousseldenis:
@florian-dacosta I think you should take a look at what @Cedric-Pigeon has done in v10 : https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_available/models/product_template.py

Done from PRs:
https://github.com/OCA/stock-logistics-warehouse/pull/474
https://github.com/OCA/stock-logistics-warehouse/pull/305

rafaelbn:
Please @florian-dacosta take a quick read also in https://github.com/OCA/stock-logistics-warehouse/pull/611 Thank you for this great contribution! :smile:

florian-dacosta:
@rousseldenis 
It was not done in version 11? I'll take a look to include it in this PR then.

@rafaelbn Thanks for the pointer.
I won't include this here since it depends on a patch on Odoo...
But when it will be merged, if it is, we could be a improvment PR

florian-dacosta:
@rousseldenis Actually, both PR you pointed concern more the stock_available module, which has already been migrated to 12, and wich seems to contains all modifications.
But it does not concern this module (stock_available_mrp).
Unless I am missing something... 
So nothing to do I guess.

rafaelbn (approved):

stock_available_mrp/models/product_product.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class ProductProduct(models.Model):
+
+    _inherit = 'product.product'
+
+    bom_id = fields.Many2one(
+        'mrp.bom',
+        compute='_compute_bom_id',
+        string='Bill of Materials'
```
pedrobaeza: Change this string due to:
```
2019-08-16 15:08:49,323 175 WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (bom_ids, bom_id) of product.product() have the same label: Bill of Materials.
```

florian-dacosta: Done!

PR #646, opened 2019-07-09
[IMP] stock_quant_manual_assign: backorder moves maintain origin move quants

oihane:
WIP, I forgot to add the view :woman_facepalming:


pedrobaeza:
This new feature should be in a different module IMO, as this has nothing to do with the original intent.

oihane:
OK, I close this PR.

PR #648, opened 2019-07-10
[10.0][MIG] port stock_orderpoint_generator to 10.0

willmakenoise:
Ported from 11.0 to 10.0.

Nothing to do, besides changing version number.


willmakenoise:
I don't see any errors in the runbot logs, so I don't know why it failed.

PR #652, opened 2019-07-12
[IMP] stock_request: Allow to set to draft multiple records

etobella:
Without this change, you were unable to set an order with multiple requests back to draft.

@jbeficent


rousseldenis (approved):
Of course!

LoisRForgeFlow (approved):

PR #653, opened 2019-07-12
[WMS][12.0] - stock_location_zone - alpha version

jbaudoux:
This is a first prototype of the part "Warehouse map and location attributes" RFC: https://github.com/OCA/stock-logistics-warehouse/issues/654


stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class StockLocation(models.Model):
+    _inherit = 'stock.location'
+
+    # FIXME: add in selection: shuttle, tray (module vertical lift)
+    kind = fields.Selection([
+        ('zone', 'Picking Zone'),
+        ('area', 'Area'),
+        ('bin', 'Bin'),
```
guewen: Can't we compute this based on other properties? Example, a leaf is a bin, a record with a `picking_zone_id` is a zone, an area I don't know... Maybe not possible but worth to ask :)

guewen: BTW the vertical lift ones are computed, so a manual selection as here won't really fit together

stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+    @api.onchange('corridor', 'row', 'rack', 'level',
+                  'posx', 'posy', 'posz')
+    def _compute_name(self):
+        for location in self:
+            if not location.kind == 'bin':
+                continue
+            area = location
+            while not area.location_name_format:
+                if not area.location_id:
+                    continue
```
guewen:
```suggestion
                    break
```
`continue` will loop infinitely

stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+                area = area.location_id
+            location.name = area.location_name_format\
```
guewen: We have to check if `area.location_name_format` is not empty, in case no location in the hierarchy has one

stock_location_zone/models/picking_zone.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2017 Syvain Van Hoof (Okia sprl) <sylvainvh@okia.be>
+# Copyright 2017-2019 Jacques-Etienne Baudoux (BCIM sprl) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PickingZone(models.Model):
+    _name = 'picking.zone'
```
guewen: The model is bothering me a bit, I would expect `stock.picking.zone` to be coherent with `stock.location`, `stock.picking.type`, ...

setup/stock_location_zone/odoo/addons/stock_location_zone
```diff
@@ -0,0 +1 @@
+../../../stock_location_zone
```
guewen:
```suggestion
../../../../stock_location_zone
```

stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+    rack = fields.Char('Rack', oldname='shelf', help="House number")
+    level = fields.Char('Level', help="Height on the shelf")
+    posx = fields.Integer('Box (X)')
+    posy = fields.Integer('Box (Y)')
+    posz = fields.Integer('Box (Z)')
+
+    location_name_format = fields.Char(
+        'Location Name Format',
+        help="Format string that will compute the name of the location. "
+             "Use 'self' to access location object. Example: "
```
guewen: This is too unsafe, allowing too much python code, we should allow a safe list of keywords

stock_location_zone/models/stock_location.py
jbaudoux: I would like to be able to use the tray/cell concepts you implemented without having a vertical lift (shuttle). I could have a table or shelf with x/y coordinates that could use the matrix display.

stock_location_zone/models/stock_location.py
jbaudoux: then let's return instead of break

stock_location_zone/models/stock_location.py
jbaudoux: return

stock_location_zone/models/picking_zone.py
jbaudoux: me too. I agree. Should I then add migration file as it was named like that on v10 projects

stock_location_zone/models/stock_location.py
jbaudoux: Changed object argument to **object.read()

stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+                if not area.location_id:
+                    return
+                area = area.location_id
+            location.name = area.location_name_format\
+                .format(**location.read())
+
+    _sql_constraints = [
+        (
+            'unique_location_name',
+            'UNIQUE(name, location_id)',
```
guewen: name must be reset on copy or this constraint prevents duplicating a location

stock_location_zone/models/stock_location.py
guewen: I see, makes sense

jgrandguillaume:
```suggestion
        ('bin', 'Bin'), ],
```

stock_location_zone/models/stock_location.py
```diff
@@ -0,0 +1,78 @@
...
+
+class StockLocation(models.Model):
+    _inherit = 'stock.location'
+
+    # FIXME: add in selection: shuttle, tray (module vertical lift)
+    kind = fields.Selection([
+        ('zone', 'Picking Zone'),
+        ('area', 'Area'),
+        ('bin', 'Bin'),
+        ],
```
jgrandguillaume:
```suggestion
```

stock_location_zone/models/stock_location.py
guewen: Nope, return would skip the next records in `self`

guewen:
I'll push a commit

stock_location_zone/models/stock_location.py
guewen: It fails with the following error on a new record:

```
 The requested operation cannot be completed due to record rules: Document type: stock.location, Operation: read, Records: <odoo.models.NewId object at 0x7f659f6921c8>, User: 2
```

stock_location_zone/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'category': 'Stock Management',
+    'depends': [
+        'stock',
+    ],
+    'data': [
+        'views/stock_picking_zone.xml',
+        'views/stock_location.xml',
+        'security/ir.model.access.csv',
+    ],
+    'installable': True,
```
guewen, line 21:
```suggestion
    'installable': True,
    'development_status': 'Alpha',
```

jgrandguillaume (approved):
LGTM, functional test ok. To add the development status alpha and ready to merge

grindtildeath (approved):
Approved to add more changes later.

jgrandguillaume:
/ocabot merge

I try again as Travis was red because of unaccessibility...

rousseldenis:
@jbaudoux Sorry arriving late in this.

Just a comment about sql constraint with WHERE.

Actually, you can use _sql_constraints. See what I did there https://github.com/OCA/partner-contact/blob/10.0/partner_bank_sort_code/models/res_bank.py#L14

You can use sql clause EXCLUDE.

See https://www.postgresql.org/docs/11/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION

 or https://www.cybertec-postgresql.com/en/postgresql-exclude-beyond-unique/

grindtildeath:
@rousseldenis Thanks for the hint, anyway I took care of this in #711 where it's not needed after simplification 

EDIT: Was still needed and your solution works like a charm :)

PR #657, opened 2019-07-15
[12.0][IMP] stock_request: Allow to set to draft multiple records

LoisRForgeFlow:
Forward port of https://github.com/OCA/stock-logistics-warehouse/pull/652

cc @etobella @rousseldenis

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -199,7 +199,7 @@ class StockRequest(models.Model):
         return True
 
     def action_draft(self):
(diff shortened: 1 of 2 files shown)


HviorForgeFlow (approved):

etobella (approved):

LoisRForgeFlow:
There is an error with the test, but I don't have the time now to dig into it. It is likely unrelated to this commit but anyway probably some more action is needed to fix it.

PR #674, opened 2019-07-31
[12.0] [ADD] stock_package_nbr

jbaudoux:
Allows to register a number of packages used during the put in pack.

If you put your goods in 3 boxes, then you can make 1 pack with a packages_nbr
of 3. This way, you don't need to care of what is in which box during the
packing operation but you can still transfer to the transporter the amount of
packages to collect and deliver. So the transporter will deliver 1 package (and so you pay only 1 delivery cost) of 3 boxes.


JordiBForgeFlow:
cc @ageficent can you review?

stock_package_nbr/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def put_in_pack(self):
+        self.ensure_one()
+        if not self.move_line_ids.filtered(
+                lambda o: o.qty_done > 0 and not o.result_package_id):
+            # let's raise the standard error
+            return self.put_in_pack()
```
AdriaGForgeFlow:
```suggestion
            return super().put_in_pack()
```

AdriaGForgeFlow (changes requested):
Functional test looks good to me :+1: 

Code suggestion below:

stock_package_nbr/models/stock_picking.py
jbaudoux:
```suggestion
            return super(StockPicking, self).put_in_pack()
```

jbaudoux: Indeed. I have fixed this. Thanks

AdriaGForgeFlow (approved):
Fix runbot error by adding _description to the new model put.in.pack.nbr

> WARNING openerp_test odoo.models: The model put.in.pack.nbr has no _description

stock_package_nbr/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright 2017-2019 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Stock Package Nbr',
+    'version': '12.0.1.0.0',
+    'author': "BCIM, Okia, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/stock-logistics-warehouse",
+    'summary': "Add quantity of packages in a pack.",
+    'category': 'Stock Management',
```
rousseldenis: Change this to 'Warehouse' as it won't appear in the good category in OCA shop.

stock_package_nbr/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2017 Syvain Van Hoof (Okia) <sylvain@okia.be>
+# Copyright 2017-2019 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Stock Package Nbr',
```
rousseldenis: I don't really like the abbreviation at the end. Say 'number' explicitly

stock_package_nbr/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def put_in_pack(self):
+        self.ensure_one()
+        if not self.move_line_ids.filtered(
+                lambda o: o.qty_done > 0 and not o.result_package_id):
```
rousseldenis: You should use float_compare()

stock_package_nbr/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+            # let's raise the standard error
+            return super(StockPicking, self).put_in_pack()
+
+        res = self.check_destinations()
+        if res.get('type'):
+            return res
+
+        return self.env.ref('stock_package_nbr.put_in_pack_nbr').read()[0]
```
rousseldenis: I think this behaviour should be configurable. IMHO, an option on stock.picking.type level is a good way of doing this.

rousseldenis (changes requested):
Code review.

rousseldenis:
@jbaudoux What's the status of this ?

stock_package_nbr/__manifest__.py
jbaudoux: fixed

stock_package_nbr/__manifest__.py
jbaudoux: fixed

stock_package_nbr/models/stock_picking.py
jbaudoux: fixed

stock_package_nbr/models/stock_picking.py
jbaudoux: I completely agree but I leave it as an improvement

setup/stock_package_nbr/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
lmignon, line 1: `.eggs` directory should not be into git :smirk:

stock_package_nbr/models/stock_picking.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo import fields, models
+from odoo.tools import float_compare
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def put_in_pack(self):
+        self.ensure_one()
+        if not self.move_line_ids.filtered(
```
lmignon, line 14: This condition is very hard to read.... IMO it would be nice to add a comment to explain the process into this override of put_in_pack.

stock_package_nbr/wizards/put_in_pack_nbr.py
```diff
@@ -0,0 +1,28 @@
...
+        res = super(PutInPackNbr, self).default_get(fields)
+        res.update({'picking_id': self.env.context['active_id']})
+        return res
+
+    @api.multi
+    def do_put_in_pack(self):
+        self.ensure_one()
+
+        return self.picking_id.with_context(
+            default_nbr_packages=self.nbr_packages)._put_in_pack()
```
lmignon, line 28: In odoo < 13, the cache is reset when you put a new value into the context.  Why do you need this value into the context?

rousseldenis:
@jbaudoux What's the status of this as very old ?

jbaudoux:
I'm closing this PR. As delivery packages has been introduced in standard odoo, choosing the right delivery package is a better solution. You can create multi-parcels delivery packages to achieve the same result.

Thanks for the reviews.

PR #675, opened 2019-08-01
[12.0][ADD] stock_request_submit

HviorForgeFlow:
It has been noticed that PR #605 would break modules that are on the top of the stock_request like, for instance, stock_request_tier_validation.

Although submit state is a useful feature, we prefer not to be introduced on stock request core module.

This is why, after discussing internally, we think this is a non-breaking-dependencies approach in order to achieve submit state in stock request.

@max3903 @etobella @jbeficent


stock_request_submit/models/stock_request.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2019 Open Source Integrators
+# Copyright 2019 Eficent Business and IT Consulting Services, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo import api, fields, models
+
+REQUEST_STATES = [
```
etobella: You could add REQUEST_STATES on a model function in order to allow inheritance. What do you think?

etobella (changes requested):
I think this is a better approach.

I made a suggestion in order to minimize duplication of data similar to:
https://github.com/OCA/purchase-workflow/blob/11.0/purchase_request/models/purchase_request.py#L279

stock_request_submit/models/stock_request.py
HviorForgeFlow: From stock_request?

etobella: for example, and will be inherited by stock.request.order and you could add the submit state on `stock_request_submit` easily.

HviorForgeFlow: I should call the method and send all states to get submitted state the second. Maybe for v13 :smile_cat:

HviorForgeFlow:
@etobella I don't understand why it is failing. In my local environment sometimes they pass tests sometimes don't. It is weird.

etobella:
the problem seems related to stock_request_analytic
probably the module submit is not compatible with analytic.

HviorForgeFlow:
Look at https://github.com/OCA/stock-logistics-warehouse/pull/675/commits/957e98b19a4329907f7e98ba49178c4d3b13468a there is nothing that makes it incompatible.

JordiBForgeFlow:
@max3903 @bodedra we'll try to help and make travis green on this one.

max3903:
@jbeficent Thank you. I should be able to test and review next wednesday after we are done with our release.

max3903:
@hveficent Few feedbacks below:

1. Activating and saving the "Allow All Locations Types" setting does not work. The option remains unchecked. I haven't tried without `stock_request_submit` so that may be a bug in `stock_request` and should be fixed in a separate PR.
2. Can you prevent stock requests with a 0 quantity from being submitted? See
https://github.com/OCA/stock-logistics-warehouse/pull/605/commits/261bc2764e06f725de36b124691bf9b05e2c90ed
3. The expected date defaults to now when you creates a new order. First it is not very realistic (any request expects a delay to be processed or otherwise it is not a request anymore) and second if you create stock requests that you want to group on the same order, you get an error message that the dates between order and requests do not match. See https://github.com/OCA/stock-logistics-warehouse/pull/605/files#diff-cb7ac1537b7c039cfc237796ef052a90R26

HviorForgeFlow:
For 1 I'm going to review and open other PR
On 2 maybe we can add a constrain on abstract class
```
@api.constrains('product_qty')
    def _check_qty(self):
        if self.product_qty <= 0:
            raise ValueError(_(
                'Stock Request product quantity has to be strictly positive.'))
```
I can be proposed in the same PR for 1.
On 3. We need to find the better way, or let user decide how to manage expected date, that why I let easily inheritable for customization.

I think point 3 also is generating random errors on Travis when testing. @max3903 can you confirm it?

max3903:
@hveficent Which Travis errors?

HviorForgeFlow:
@max3903
ERROR on stock_request_analytic see here: https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/566982044 
First I found. It happened in this PR also, I rebuild and suddenly it passed.

max3903:
@hveficent That's why we should bump the expected date to the next hour by default.

HviorForgeFlow:
But this is some field that users can modify while is in draft on demand.
I think the constrain is so hard, maybe ensuring that stock request that come from an stock request orders must use expected date of SRO itself, but creating both by default is only generating troubles on testing randomly.

What do you think @etobella ?

max3903:
@hveficent FYI we can live without the default date. We implemented a different strategy:
https://github.com/OCA/field-service/pull/238/files#diff-463f52e9060fd927e486c9b99634617eR47

max3903 (approved):

JordiBForgeFlow (approved):

bodedra (approved):

HviorForgeFlow:
Uninstall hook required if submit feaute want to be uninstall properly. I'm not sure if submit state keeps on system when deactivating this module.
ping @mreficent

stock_request_submit/models/stock_request.py
MiquelRForgeFlow: I don't understand. Why the order of the strings in the selection is important?

max3903: It defines the sequence of states in the status bar.

MiquelRForgeFlow:
My contributions here are done and ready! 😉 😋

MiquelRForgeFlow (approved):

max3903:
@hveficent 

For Requests and Orders, the Submitted state is repeated various times in the status bar.

MiquelRForgeFlow:
@max3903 solved :)

max3903:
@hveficent @mreficent Let me know when the virtual location is fixed (1st point in https://github.com/OCA/stock-logistics-warehouse/pull/675#issuecomment-518354116) so I can close #605. Thanks!

PR #676, opened 2019-08-02
[12.0][MIG] stock_account_inventory_force_date

HviorForgeFlow:
Standard migration

CC @Eficent


stock_account_inventory_force_date/models/stock_inventory.py
```diff
@@ -0,0 +1,114 @@
...
+        help="Force the inventory moves and accounting entries to a given date"
+             "in the past.",
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
+
+    @api.onchange('date')
+    def _onchange_force_inventory_date(self):
+        if self.force_inventory_date:
+            self.accounting_date = self.date
```
daramousk: This results in a crash since `accounting_date` is `Date` while `date` is `Datetime` (unfortunate naming), calling tostring on `self.date` should suffice.

daramousk (changes requested):

stock_account_inventory_force_date/models/stock_inventory.py
HviorForgeFlow: You are right. New changes proposed.

rousseldenis (approved):
Code review

HviorForgeFlow:
@daramousk can you update your review? Thanks!

HviorForgeFlow:
ping @daramousk

daramousk (approved):
Looks great, although the code coverage needs to be raised so that stuff gets green.

rousseldenis:
@hveficent Do you plan to increase coverage ?

HviorForgeFlow:
@rousseldenis @daramousk, I think current tests cover most of the use cases. Not full coverage is due to doing an Inventory Adjustment of all product, so I can make it happen, but asserts can't be precise across changes in demo data, so I think it is not needed and it is ready to be merged.

PR #656, opened 2019-07-15
[FIX] stock_request_purchase: remove sudo

etobella:
There was an issue when a user was not admin but he was a purchase user and a stock request user.

It was not showing the button of purchase. Now it should work as expected.

@jbeficent @lreficent

PR #681, opened 2019-08-09
[11.0][FIX] stock_request: sudoing internal stock_request engine _compute_qty

HviorForgeFlow:
Inventory Users can't check availability on Transfers coming from a Stock Request.

Inventory Users have Stock request user role by default, but it also have a bunch of Record Rules that limit them to full edit other stock request that are not created by them.

You can do the test using demo data.
Steps to reproduce:
1. Log as admin create a SR, confirm it to generate an stock move.
2. Log with demo user go to that transfer and click check availability.

An error will be prompted when trying to compute allocation quantities with Stock Request User permissions.

Assuming this compute method is only used as an internal engine to keep allocations update, I'm proposing to sudo it.

### DIFF
--- a/stock_request/security/stock_request_security.xml
+++ b/stock_request/security/stock_request_security.xml
@@ -62,6 +62,16 @@
             <field name="perm_unlink" eval="True"/>
             <field name="domain_force">[('requested_by','=',user.id)]</field>
         </record>
+
+        <record id="stock_request_user_inventory_user_rule" model="ir.rule">
+            <field name="name">Stock Request User Inventory User</field>
+            <field name="model_id" ref="model_stock_request"/>
+            <field name="groups" eval="[(6,0, [ref('stock.group_stock_user')])]"/>
+            <field name="perm_read" eval="True"/>
+            <field name="perm_write" eval="True"/>
+            <field name="perm_create" eval="False"/>
+            <field name="perm_unlink" eval="False"/>
+        </record>
     
         <record id="stock_request_manager_rule" model="ir.rule">
             <field name="name">Stock Request Manager</field>


HviorForgeFlow:
Can you review @lreficent @etobella @max3903 ?

stock_request/models/stock_request.py
```diff
@@ -130,6 +130,7 @@ def _compute_picking_ids(self):
                  'allocation_ids.stock_move_id.move_line_ids.qty_done')
     def _compute_qty(self):
         for request in self:
```
bodedra:
```suggestion
         for request in self.sudo():
```

etobella:
Could we find another option?
I had a lot of errors with sudo as you can check in #656.
Maybe adding the read permissions for the inventory user is enough.

HviorForgeFlow:
@etobella But this is not the case, we are only using sudo to retrieve computed fields. I think there is no issue on that.

The way to solve it without sudo is allow Stock Request Users to read and write in any stock request.

HviorForgeFlow:
Still having an issue when Check availability with Demo User from an SR created by Admin in demo data.
image

HviorForgeFlow:
> Still having an issue when Check availability with Demo User from an SR created by Admin in demo data.
> image

The issue persist after this path due to Stock Request User ir.rule that is not allowing any operation if it is not requested by same user.

HviorForgeFlow:
Fixed avoiding sudo usage. 
@etobella

AdriaGForgeFlow (approved):
Code review :+1:

rousseldenis (approved):
LGTM.

bodedra (approved):

max3903 (approved):

PR #683, opened 2019-08-15
[WMS][12.0] Add stock_reserve_rule

guewen:
This module adds rules for advanced reservation / removal strategies.

Rules are applied on a location and its sub-locations.

A rule can exclude quants or locations based on configurable criteria,
and based on the selected quants, apply advanced removal strategies.

The rules have a sequence, which will be respected for the reservation.
So even without filter or advanced removal strategies, we can give a priority to
reserve in a location before another.

The advanced removal strategies are applied on top of the default one (fifo,
fefo, ...).

The included advanced removal strategies are:

* Default Removal Strategy: apply the default configured one (fifo, fefo, ...)
* Empty Bins: goods are removed from a bin only if the bin will be empty after
  the removal (favor smallest bins first, then apply the default removal
  strategy for equal quantities).
* Prefer Full Packaging: tries to remove full packaging (configured on the
  products) first, by largest to smallest package (default removal strategy is
  then applied for equal quantities).

Examples of scenario:

rules:

* location A: no filter, no advanced removal strategy
* location B: no filter, Empty Bins
* location C: no filter, no  advanced removal strategy

result:

* take what is available in location A
* then take in location B if available, only if bin(s) are emptied
* then take what is available in location C

The module is meant to be extensible, with a core mechanism on which new rules
and advanced removal strategies can be added.

It's part of the WMS addons to implement these specs:
https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#heading=h.9ad56twdce31

Demo data has been created by the module.

Product: Funky Socks

Locations:

 * Stock / Zone A / Bin A1: 200 Funky socks
 * Stock / Zone B / Bin B1: 100 Funky socks
 * Stock / Zone C / Bin C1: 100 Funky socks

Reservation Rules, in the following order

 * Zone A must have full quantities
 * Zone B
 * Zone C

Delivery Orders:

* Origin: Outgoing shipment (reservation rules demo 1)
* Origin: Outgoing shipment (reservation rules demo 2)

Test:

* Activate Storage Locations and Multi-Warehouses
* You can open Inventory > Configuration > Stock Reservation Rules to see the rules
* Open Transfer: Outgoing shipment (reservation rules demo 1)
* Check availability: it has 150 units, as it will not empty Zone A, it will not take products there, it should take 100 in B and 50 in C (following the rules order)
* Unreserve this transfer (to test the second case)
* Open Transfer: Outgoing shipment (reservation rules demo 2)
* Check availability: it has 250 units, it can empty Zone A, it will take 200 in Bin A1 and 50 in Bin B1.
* If you want to explore further, you can add a custom domain to exclude rules (for instance, a product category will not use Zone B).


guewen:
~As this one needs #653, I'll wait the alpha version of #653 is merged before removing the WIP flag and adding the alpha flag here.~

jgrandguillaume:
Hi @guewen ,

First test seems to be ok, I have only one issue with chained move. I'm wondering if the reserve rules must apply in such a case...

e.g. in a pick+pack+ship setup, if you haven't define rules on both packing and output location, you'll not be able to ship something. 

Imagine you made your setup for stock location, so here everything goes fine, but then, as you didn't configured any rule for packing, the pack operation will be waiting... It's ok it on stock I guess, but with chained move I would say they should be available.

What do you say ?

jgrandguillaume:
Hi @guewen 

Works fine now on this regards. My next issue is when used alongside with #695 .

Sometimes we want the following rules: 

1) Full case from Shelving
2) Full case from Shelving reserve
3) Partial from Shelving
4) Partial from Shelving reserve

Here if we do have the following scenario:

- Product A packaging if 50
- Stock in Shevling is 30
- Stock in Shelving reserve is 100

If I order let's say 50, it'll take:

- 30 from Shelving
- 20 from Shelving Reserve 

Expected:

- 50 from Shelving reserve (a full box)

I think we need to add in reserve rule a "Force full package" like for "Only full quantity", but for full package to deal with that case.

guewen:
> My next issue is when used alongside with #695 .

I'm actually not sure #695 and your issue are related. You are expressing 2 differents needs:

1. Your use case describes a constraint: the quants from a location must not be used if the quantity does not constitute a full packaging. The whole zone including all sublocations is discarded if the quantity does not match a packaging.
2. The module #695 indicates more a preference rather than a constraint for full packaging, as *within* sublocation of a location, it will first reserve the quants that match a packaging, then the rest, but eventually if nothing match, it reserves any quant.

At first look, adding an option "Use full packaging" seems to make sense.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -0,0 +1,233 @@
...
+        "and sub-locations.",
+    )
+
+    # advanced removal strategy
+    removal_strategy = fields.Selection(
+        string="Advanced Removal Strategy",
+        selection=[
+            ("default", "Default Removal Strategy"),
+            ("empty_bin", "Empty Bins"),
+            ("packaging", "Prefer Full Packaging"),
```
guewen: Rename to "Only Full Packaging" (everywhere)

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -0,0 +1,256 @@
...
+                    sum(quants.mapped("quantity"))
+                    - sum(quants.mapped("reserved_quantity")),
+                    quants,
+                    location,
+                )
+                for location, quants in quants_per_bin
+            ]
+        )
+
+        # Propose the smallest quants first, so we can empty as most bins
```
jbaudoux: This must be changed to choose the biggest quant (so invert the order). This makes less operational work. It is a bit stupid to start with the pallet with the few products on it and stack many products from other locations on it. You just want to start with the pallet with the most products and stack few additional products from other locations.

jbaudoux (changes requested):

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: reminder

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -0,0 +1,256 @@
...
+        # Group by location (in this removal strategies, we want to consider
+        # the total quantity held in a location).
+        quants_per_bin = quants._group_by_location()
+
+        product = fields.first(quants).product_id
+
+        # we'll walk the packagings from largest to smallest to have the
+        # largest containers as possible (1 pallet rather than 10 boxes)
+        packaging_quantities = sorted(
+            product.packaging_ids.mapped("qty"), reverse=True
```
jbaudoux: exclude lines where qty = 0

jbaudoux (changes requested):

guewen:
Work is continued in 13.0 in #790

PR #689, opened 2019-08-27
[11.0][FIX] stock_request: add expected_date implicit on tests to avoid errors

HviorForgeFlow:
It should avoid intermittent errors when running stock_request test on CI.

@lreficent @jbeficent @max3903

### DIFF
--- a/stock_request/tests/test_stock_request.py
+++ b/stock_request/tests/test_stock_request.py
@@ -157,10 +157,12 @@ class TestStockRequest(common.TransactionCase):
             order.location_id, self.warehouse.lot_stock_id)
 
     def test_onchanges_order(self):
+        expected_date = fields.Datetime.now()
         vals = {
             'company_id': self.main_company.id,
             'warehouse_id': self.warehouse.id,
             'location_id': self.warehouse.lot_stock_id.id,
+            'expected_date': expected_date,
             'stock_request_ids': [(0, 0, {
                 'product_id': self.product.id,
                 'product_uom_id': self.product.uom_id.id,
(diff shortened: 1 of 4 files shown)


max3903 (approved):

bodedra (approved):
Code review. LGTM :+1:

stock_request/tests/test_stock_request.py
```diff
@@ -157,17 +157,20 @@ def test_defaults_order(self):
             order.location_id, self.warehouse.lot_stock_id)
 
     def test_onchanges_order(self):
+        expected_date = fields.Date.today()
```
bodedra:
```suggestion
        expected_date = fields.Datetime.now()
```

stock_request/tests/test_stock_request.py
```diff
@@ -862,17 +865,20 @@ def test_allow_virtual_location(self):
         self.assertTrue(order.allow_virtual_location)
 
     def test_onchange_wh_no_effect_from_order(self):
+        expected_date = fields.Date.today()
```
bodedra:
```suggestion
        expected_date = fields.Datetime.now()
```

bodedra (changes requested):

etobella (approved):

rousseldenis (approved):
Code review. Makes sense apart @bodedra remark.

stock_request/tests/test_stock_request.py
HviorForgeFlow: There are more tests that relay on today date. As we only need to ensure that date expected is the same in SRO and SR it's fine.

stock_request/tests/test_stock_request.py
HviorForgeFlow: Same

stock_request/tests/test_stock_request.py
bodedra: @hveficent I am agree with your point. But `expected_date` field type is `datetime`.

HviorForgeFlow:
@bodedra done

bodedra (approved):

PR #696, opened 2019-09-05
[WMS][12.0] Add stock_move_location_dest_constraint_base - alpha version

grindtildeath:
This module provides a base mechanism to constrain the selection of destination
location on stock move lines.
It also adds a checkbox 'Bypass constraints' on stock location to deactivate
the checks implemented by installed modules.

Related to https://github.com/OCA/stock-logistics-warehouse/issues/691

Issue #691, opened 2019-08-28
[RFC][WMS] Put away rules and ABC classification support


jgrandguillaume:
Source : https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#
Part of: https://github.com/OCA/wms/issues/1

**General putaway behavior**

- No automatic 2 steps (or more) reception, instead we’ll use a reception to input location and from there create manual internal transfers (new picking type, from input to stock)
  - Refer to the known Odoo issue for 2 or more step reception: https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#heading=h.y4no98tmzocs
  - Handover of zones (if any) will be managed in the same way. A new manual operation is created from the handover Input to the zone stock
- When processing putaway, we have three layers to handle and those needs to be implemented and respected by any application of putaway:
  - Define routing: where should I put away goods. Those are described below in product putaway rules
  - Define constraints: Compute final destination in put away according to product/location attributes and constraints. This means, if no suitable location, we cannot put away the product.
  - Define preferences: Compute final destination according to preferences, where we try as much as possible to respect a rule. This is the case for chaotic storage below for example.
- Put away are recommended for every location mentioned as “default destination” in a pick operation type, but should at least be defined at stock location level
- The put away is currently computed once at assignation of the move. Therefore it is impossible to concatenate put away with push rules as they are applied earlier at confirmation time (e.g. not possible: put away on stock asks to place in zone A, then a push apply to move it to Area 2 where another putaway asks to go in Bin XY)

**This RCF to propose a POC implementation for the following**

- [ ] Base module to handle constraints on put away (when it's not a fixed location put away)
- [ ] Tag based constraint module on product and location
- [ ] Product and location ABC classification
- [ ] Chaotic put away based on ABC classification
- [ ] Add put away configuration in product form

grindtildeath:
ABC and restrictions PR are to be handled by the following modules:

 - stock_storage_type https://github.com/OCA/wms/pull/4
 - stock_putaway_storage_type_strategy https://github.com/OCA/wms/pull/5
 - stock_putaway_storage_type_abc https://github.com/OCA/wms/pull/6

So out of #707 that is a dependency of stock_storage_type, I guess we can close the rest.

jgrandguillaume:
[...]

PR #697, opened 2019-09-05
[WMS][12.0] Add stock_move_location_dest_constraint_empty - alpha version

grindtildeath:
This module implements destination location constraint on stock move line.

When installed stock moves will only be allowed if the destination location is
empty.

Related to #691 

Requires:
 - [ ] #696


stock_move_location_dest_constraint_empty/models/stock_location.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo.exceptions import ValidationError
+
+
+class StockLocation(models.Model):
+
+    _inherit = 'stock.location'
+
+    def check_move_dest_constraint(self, line=None, product=None):
+        """Raise Validation error if not allowed"""
+        res = super().check_move_dest_constraint(line=line)
```
grindtildeath:
```suggestion
        res = super().check_move_dest_constraint(line=line, product=product)
```

jgrandguillaume:
Hi @grindtildeath ,

A remark here that I probably still have to dig a bit.

Probably the empty bin is something that we want to configure on locations, it's probably not a default for everywhere. 
e.g. for handover place, it's not optimal to forbid an internal move if a move is already planned, because probably the bin will be empty by the time the next move will get there...

We should be able to define the "Empty bin" in the location hierarchy tree at a certain level and it should be true for all the children. Let's add a checkbox in location "Moves allowed only in empty bin" with a tooltip: "Check this box to forbid to move (or put away) goods in a location that is not empty or for which a move is already planned".

PR #698, opened 2019-09-05
[WMS][12.0] Add stock_move_location_dest_constraint_tag - alpha version

grindtildeath:
This module implements destination location constraint on stock move line.

When installed stock moves will only be allowed if the destination location
includes the same tags as defined on the product or its product category.

N.B.: If tags are defined on the product, tags defined on its product category
will not be considered

Related to #691

Requires:
 - [ ] #696


stock_move_location_dest_constraint_tag/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+
+    _inherit = 'stock.location'
+
+    product_storage_tag_ids = fields.Many2many(
+        'product.storage.tag',
+        string='Storage Tags',
+    )
+
+    def check_move_dest_constraint(self, line=None, product=None):
+        res = super().check_move_dest_constraint(product)
```
grindtildeath:
```suggestion
        res = super().check_move_dest_constraint(line=line, product=product)
```

stock_move_location_dest_constraint_tag/demo/product.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
```
Trivedi-Vacha-SerpentCS, line 2: should use odoo noupdate="1"

stock_move_location_dest_constraint_tag/demo/product_category.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
```
Trivedi-Vacha-SerpentCS, line 2: noupdate="1" is missing for all demo files

stock_move_location_dest_constraint_tag/models/product_category.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models, fields
```
Trivedi-Vacha-SerpentCS, line 3: import should be in alphabetical order

stock_move_location_dest_constraint_tag/models/product_storage_tag.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models, fields
```
Trivedi-Vacha-SerpentCS, line 3: same as above

stock_move_location_dest_constraint_tag/models/product_template.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models, fields
```
Trivedi-Vacha-SerpentCS, line 3: same as above

stock_move_location_dest_constraint_tag/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models, fields, _
```
Trivedi-Vacha-SerpentCS, line 3: It should be from odoo import _, fields, models

gurneyalex (commented):
I'm concerned about a few issues here:

* with this implementation, the Customers location needs to be updated with every possible tag, otherwise you will never be able to deliver a SO
* same for PACK and OUT locations of a WH
* there is no domain update in the choice of a destination location to help the user choose an appropriate location.

jgrandguillaume:
@gurneyalex I agree with you. We need to review all this put away design at a whole. The POC prooved it works, but it's not optimal.

PR #699, opened 2019-09-05
[WMS][12.0] Add stock_putaway_abc - alpha version

grindtildeath:
This module implements ABC chaotic storage putaways.

Instead of defining putaway with a fixed location for product or product
category, this module allows to define an ABC preference for product or product
category on the product putaway form.

When such a putaway is applied, it will look recursively for a stock location
without children whose ABC classification matches the ABC priority of the
putaway.

e.g. for a putaway with B priority, it will look on children locations
for a B location. If no B location is found, it will look for a C. If no C
location is found, it will look for an A. Finally, if no location matches, it
will return the stock location where the putaway is applied.

Related to #691

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -4,3 +4,4 @@ account-analytic
 product-attribute
 server-ux
 web
+stock-logistics-warehouse https://github.com/grindtildeath/stock-logistics-warehouse 12.0_add_stock_putaway_rule
--- a/stock_putaway_abc/__init__.py
+++ b/stock_putaway_abc/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/stock_putaway_abc/__manifest__.py
+++ b/stock_putaway_abc/__manifest__.py
@@ -19,4 +19,8 @@
         "views/stock_location.xml",
         "views/product_strategy.xml",
     ],
+    "demo": [
+        "demo/stock_location.xml",
+        "demo/product_strategy.xml",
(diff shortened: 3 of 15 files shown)


jgrandguillaume:
Hi @grindtildeath 

Thanks for the work. I think either I've not understood something or we didn't understood each other.

What I want to do..

Location structure:

- Stock
  - HRL
      - Bin 1 Class A
      - Bin 2 Class B
      - Bin 3 Class C
  - FBG
      - Bin 1 Class A
      - Bin 2 Class B
      - Bin 3 Class C

Product put away:

- On Stock: Put away 1: Product A goes to HRL (set as a fixed location put away)
- On HRL: Put away 2: Product A is manage as a classification "A" (set as a ABC put away)
- On FBG: Put away 3: Product A is manage as a classification "C" (set as a ABC put away)

Expected result:

- If I receive product A in Stock (e.g. from a PO), it should process the put away and lead me to HRL/Bin 1 because put away 1 goes to HRL, then there the put away says apply ABC, so bin 1 is the match here
- If I receive product A in FBG (e.g. internal move), it should process the put away and lead me to FBG/Bin 3 because put away 3 says apply ABC, so bin 3 is the match here

I'm not sure how to achieve this with the current module. Should I merge put away 1 and 2 at stock level and create put away 3 for FBG ? How can I ask in the put away 1 that product should go to HRL then ?

Thanks for the headline, may be I'm just tired ;)

PR #701, opened 2019-09-05
[IMP] stock_request: Set Procurement Group

osi-scampbell:
When an order is confirmed, create a procurement group and set it on the lines

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -201,6 +201,11 @@ class StockRequestOrder(models.Model):
 
     @api.multi
     def action_confirm(self):
+        if not self.procurement_group_id:
+            self.procurement_group_id = self.env['procurement.group'].create(
+                {'name': self.name,
+                 'move_type': 'direct'})
+            self.change_childs()
         self.mapped('stock_request_ids').action_confirm()
         self.write({'state': 'open'})
         return True
--- a/stock_request/tests/test_stock_request.py
+++ b/stock_request/tests/test_stock_request.py
@@ -910,6 +910,7 @@ class TestStockRequestBase(TestStockRequest):
             'company_id': self.main_company.id,
             'warehouse_id': self.warehouse.id,
             'location_id': self.virtual_loc.id,
+            'expected_date': fields.Datetime.now(),
             'procurement_group_id': group.id,
             'stock_request_ids': [(0, 0, {
                 'product_id': self.product.id,


max3903 (approved):

rousseldenis (approved):
Makes sense.

LoisRForgeFlow (changes requested):
What if you don't want procurement group on purpose?

rousseldenis:
@max3903 @osi-scampbell Could you answer @lreficent question ?

max3903:
@lreficent We can add an option in the settings to auto-create the procurement group. Would that work for you?

CC: @osi-scampbell

LoisRForgeFlow:
@max3903 What do you think about using this module: [procurement_auto_create_group](https://github.com/OCA/stock-logistics-warehouse/tree/12.0/procurement_auto_create_group)? If the route used by your stock request has the "Auto-create Procurement Group" field set to true, I think you will get your desired behavior.

max3903:
@osi-scampbell Can you try using procurement_auto_create_group?

max3903:
@osi-scampbell Any progress on this?

max3903:
Ping @osi-scampbell

max3903:
@osi-scampbell Let's add the issue #897 in the roadmap.rst file and get this pull request closed...

etobella:
@max3903 I don't think this is an issue as both behaviors are good, the original one and the new one. Have you tested the option that @LoisRForgeFlow proposed you? Maybe that fills your needs without breaking the behavior of the original module.

max3903:
@etobella No we didn't test it and we are not planning to in the near future. Maybe we will revisit this later.

etobella:
then I think this issue does not need to go the roadmap of the module.

max3903:
@osi-scampbell Let's fix the Travis error and get this merged.

etobella (changes requested):
this is changing the behavior of the module when other options are possible...

max3903:
@etobella Like I said we'll get back to this when appropriate.

murtuzasaleh (approved):

etobella:
As said in previous comments, this is changing the current behavior of the module, this should go to a new module or reuse the options that @LoisRForgeFlow showed.

People is using this module with the current version and they are expecting that.

LoisRForgeFlow:
@max3903 I don't think this PR should be merged unilaterally, moreover you are not the main author and you are more than aware that this module is widely used by at least ForgeFlow and Creu Blanca... My suggestion takes 10 minutes to be tested, I don't think skipping it by adding to the roadmap while breaking existing implementations is the way to go here.

LoisRForgeFlow (changes requested):
Just refreshing my red cross here, I don't want this to be merged by any "mistake".

rousseldenis:
@max3903 @LoisRForgeFlow So, we agree not merging this ?

PR #706, opened 2019-09-09
[WMS][12.0] Add stock_putaway_recursive - alpha version

grindtildeath:
This module applies putaway strategies recursively.

By default Odoo applies a putaway strategy on the destination location of a
stock move and returns the first fixed location it finds on the destination
location of the move.

With this module, if another putaway is defined on the returned fixed location,
it will be applied as well, as will be any other putaway strategy that is
defined on the next destination location that is found.

Related to #691

### DIFF
--- a/stock_putaway_recursive/tests/test_stock_putaway_recursive.py
+++ b/stock_putaway_recursive/tests/test_stock_putaway_recursive.py
@@ -16,7 +16,7 @@ class TestStockPutawayRecursive(SavepointCase):
         cls.shelf2_location = cls.env.ref('stock.stock_location_14')
         cls.refri_location = cls.env.ref('stock.location_refrigerator_small')
 
-        # Define putaway from Stock to Stock/Shelf 1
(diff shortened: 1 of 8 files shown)


stock_putaway_recursive/tests/test_stock_putaway_recursive.py
```diff
@@ -0,0 +1,87 @@
...
+        shelf1_putaway = cls.env['product.putaway'].create({
+            'name': 'shelf1 putaway',
+        })
+        cls.shelf1_location.write({
+            'putaway_strategy_id': shelf1_putaway.id
+        })
+        shelf1_putaway.write({
+            'product_location_ids': [(0, 0, {
+                'product_id': cls.product_flipover.id,
+                'fixed_location_id': cls.shelf2_location.id,
```
grindtildeath, line 29: Domain in https://github.com/odoo/odoo/blob/12.0/addons/stock/views/product_strategy_views.xml#L18 should prohibit this.

PR #709, opened 2019-09-10
[WIP][WMS][12.0] Available to Promise Release

guewen:
Implements: #710


stock_virtual_reservation/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+    # Add store on the field, as it is quite used in the searches,
+    # and this is an easy-win to reduce the number of SQL queries.
+    picking_type_code = fields.Selection(store=True)
+    need_rule_pull = fields.Boolean(compute="_compute_need_rule_pull")
+
+    @api.depends("move_lines.need_rule_pull")
+    def _compute_need_rule_pull(self):
+        for picking in self:
+            picking.need_rule_pull = any(
+                move.need_rule_pull for move in picking
```
florian-dacosta: should be `move.need_rule_pull for move in picking.move_lines`

florian-dacosta (commented):
I know it is a wip no worries, just a small comment.
Also, I try to understand well the use of the module, and I wonder about the make to order process.
If you configure one product with make to order and buy route.
On sale order validation, the delivery order will be created and marked as `need_rule_pull`.
So the purchase order for the make to order product won't be created.
Then, the delivery order should be releases once the products are virtually reserved, which won't ever happen for the make to order product, right?

guewen:
@florian-dacosta spot on! I'll find a correction for the make to order issue, thanks :)

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+
+    def _should_compute_ordered_available_to_promise(self):
+        return (
+            self.picking_code == "outgoing"
+            and not self.product_id.type == "consu"
+            and not self.location_id.should_bypass_reservation()
+        )
+
+    def _ordered_available_to_promise(self):
```
jbaudoux, line 49: Why is this method not part of _compute_ordered_available_to_promise ? Retrieving the value can be done by reading the field.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -0,0 +1,154 @@
...
+            0.,
+        )
+
+    def _previous_promised_quantity_domain(self):
+        domain = [
+            ("need_release", "=", True),
+            ("product_id", "=", self.product_id.id),
+            ("date_priority", "<=", self.date_priority),
+            ("warehouse_id", "=", self.warehouse_id.id),
+        ]
```
jbaudoux, line 66: Ideally we should exclude all the moves planned in the future that you can resupply on time. For example, there is no reason a delivery planned 3 months is blocking quantity while you can resupply in 7 days.
Simple solution would be configuring an horizon for considering blocking moves. A more complex and accurate solution would be by looking at the warehouse product orderpoint planned date.
To support sales order commitment dates, something must be done

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -0,0 +1,154 @@
...
+            "Product Unit of Measure"
+        )
+        for move in self:
+            if not move.need_release:
+                continue
+            if move.state not in ("confirmed", "waiting"):
+                continue
+            # do not use the computed field, because it will keep
+            # a value in cache that we cannot invalidate declaratively
+            available_quantity = move._ordered_available_to_promise()
```
jbaudoux, line 121: use the field "move._ordered_available_to_promise" instead

guewen: There is the explanation in the comment just above why the field is not used ;)
I can rephrase if not understandable enough.

stock_available_to_promise_release/models/stock_move.py
guewen: As we can't have proper depends, if values change in the transaction, the field will not be recomputed.
We could add invalidations everywhere, not sure that's useful. The field is good for the view, but the code better has to recompute the value on every access. I should add a docstring about this.

stock_available_to_promise_release/models/stock_move.py
jgrandguillaume: We go for the horizon solution here

jgrandguillaume (approved):

guewen:
Work continued on 13.0 on #799

Issue #722, opened 2019-09-23
[RFC][WMS] Goods reservation - min shelf-life


jgrandguillaume:
Source : https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#
Part of: https://github.com/OCA/wms/issues/1

**This RCF to propose a POC implementation for the following**

Allow a minimum delay for the life date. This parameter should be defined on the Sales Order with a default value on the customer. The parameter is global for the whole order.

If the customer wants a shelf-life of min 6 month, then only the quants respecting this can be reserved.

This condition should be ideally based on the delivery date (and not the order date) which should be the picking date.

For now, we implement this limit only at stock level (no link with sale yet)

jgrandguillaume:
RFC is over, were entering the pilot phase. To clean up I'm closing related issues

PR #723, opened 2019-09-23
[WIP][WMS][12.0] stock_minimum_shelf_life

damdam-s:
implements #722

### DIFF
--- a/stock_minimum_shelf_life/__manifest__.py
+++ b/stock_minimum_shelf_life/__manifest__.py
@@ -16,6 +16,7 @@
         "product_expiry",
     ],
     "data": [
+        "views/res_config_settings.xml",
     ],
     "installable": True,
     "development_status": 'Alpha',
--- a/stock_minimum_shelf_life/__manifest__.py
+++ b/stock_minimum_shelf_life/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2019 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock minimum shelf life",
+    "version": "12.0.1.0.0",
+    "license": "AGPL-3",
+    "summary": "Allows to set a minimum date in procurement group to force the"
+               " reservation mecanism to take only product that expiry date is"
+               " after the minimum date.",
+    "author": "Camptocamp,"
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "category": "Warehouse",
+    "depends": [
+        "stock",
+        "product_expiry",
+    ],
+    "data": [
+    ],
+    "installable": True,
+    "development_status": 'Alpha',
+}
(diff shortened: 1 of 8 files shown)


stock_minimum_shelf_life/models/res_company.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    minimum_shelf_life = fields.Integer(
```
rousseldenis, line 10: @jgrandguillaume  In case of fresh products, isn't it a too big window (days instead) ?

stock_minimum_shelf_life/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "author": "Camptocamp,"
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "category": "Warehouse",
+    "depends": [
+        "stock",
+        "product_expiry",
+    ],
+    "data": [
+    ],
```
MeetKD, line 20: Please remove empty key.

damdam-s: PR marked as draft, work in progress.
This will obviously not be empty soon

stock_minimum_shelf_life/models/stock_move.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return res
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    def _action_assign(self):
+        proc_sales_moves = self.filtered(
+            lambda m: m.picking_id.group_id.sale_id
```
JordiBForgeFlow, line 39: @jgrandguillaume are you sure that this should be limited to stock moves for sales to customers? What about checking it also in production orders, so that you don't accept to consume products that have a certain shelf life limit? This is very important, for example, for food processing.

Al alternative could be that the user defines in settings when they want to check minimum shelf life:
1) Shipping to customers (destination location in the stock move is type customer)
2) Consuming products for production (destination location in the stock move is type production)

stock_minimum_shelf_life/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    minimum_shelf_life = fields.Integer(
```
JordiBForgeFlow, line 10: @jgrandguillaume  In SAP is documented as 'minimum_remaining_shelf_life'. That would be better understood? Also, why document it at company level? Why not maintain it at product and product category levels?

stock_minimum_shelf_life/models/stock_move.py
jgrandguillaume: @jbeficent I agree with you, this is a very good remark. 

Will take this into account.

jgrandguillaume: Not so easy because if in pick pack ship setup, then the pick make the reservation.. so it's not a customer one yet..

rousseldenis:
@damdam-s @jgrandguillaume What's the status of this ?

PR #730, opened 2019-09-30
[8.0][FIX] .travis: Update to latest template

rousseldenis:


.travis.yml
```diff
@@ -1,39 +1,72 @@
...
+  # https://github.com/wkhtmltopdf/wkhtmltopdf/releases
+  # if you need to install wkhtmltopdf
+  # - WKHTMLTOPDF_VERSION="0.12.5"
+  # Set the above to install a `wkhtmltopdf` version that is not the one provided
+  # by the `pov-wkhtmltopdf` repo.
+  - PHANTOMJS_VERSION="latest"
+  # The above line controls the PhantomJS version that is used for JS testing.
+  #   It is not necessary to include this value unless you are altering the default.
+  #   Use `OS` to skip the PhantomJS upgrade & use the system version instead.
+  - WEBSITE_REPO="1"
```
pedrobaeza: No need to install all of this stuff

pedrobaeza (changes requested):

.travis.yml
rousseldenis: Is it ok now?

pedrobaeza (approved):

PR #734, opened 2019-10-16
[Fix] use the same view id for differnt view using in product and pro…

digitalsatori:
the inherited tree view for `product.product` and `product.template` is using the same view id. This patch fixes it.


bodedra (approved):
LGTM :+1:

rousseldenis (approved):

LoisRForgeFlow (approved):

PR #737, opened 2019-10-29
[11.0][FIX] stock_inventory_exclude_sublocation: Use correct lot in test

### LINKED ISSUE #736: [BUG][11.0] stock_inventory_exclude_sublocation: change in Odoo
@mreficent @lreficent 
The 11.0 main branch is red due to a change in Odoo core :

https://github.com/odoo/odoo/commit/7b29911d0ce633998e30546230bb46abf7952518

This is a new constraint to check if products are equivalent on move line level and lot level.

Could you check or point someone else to explore for a fix ?

rousseldenis:
Fixes #736

### DIFF
--- a/stock_inventory_exclude_sublocation/tests/test_exclude_sublocation.py
+++ b/stock_inventory_exclude_sublocation/tests/test_exclude_sublocation.py
@@ -61,6 +61,10 @@ class TestStockInventoryExcludeSublocation(TransactionCase):
             'name': 'Lot for product1',
             'product_id': self.product1.id
         })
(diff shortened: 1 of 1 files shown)


rousseldenis:
Fast-tracking this as 11.0 main branch is red

MiquelRForgeFlow:
Shouldn't this be forwardported to v12? I suppose the same constrain was forwardported to v12, or not?

rousseldenis:
> Shouldn't this be forwardported to v12? I suppose the same constrain was forwardported to v12, or not?

I'll check if the code exists in v12

Issue #741, opened 2019-11-04
stock_request error when using pt_BR


augustodinizl:
Good evening

Perhaps is there some misconfig on my instance, but i'm getting that error at using translation
Can somebody help me?

`Erro:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/api.py", line 1039, in get
    value = self._data[key][field][record._ids[0]]
KeyError: 1005

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/fields.py", line 981, in __get__
    value = record.env.cache.get(record, self)
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/api.py", line 1041, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: ('ir.actions.act_window(1005,).search_view', None)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/http.py", line 656, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/http.py", line 314, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/http.py", line 698, in dispatch
    result = self._call_function(**self.params)
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/http.py", line 346, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/bitnami/apps/odoo/lib/odoo-12.0.post20190918-py3.7.egg/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
[...]

PR #742, opened 2019-11-04
Update stock_request.py

augustodinizl:
As pointed from @mileo on issue #741 states field don't need '_()'

### DIFF
--- a/stock_request_submit/models/stock_request.py
+++ b/stock_request_submit/models/stock_request.py
@@ -9,10 +9,10 @@ class StockRequest(models.Model):
 
     def __get_request_states(self):
         states = super().__get_request_states()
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
If it's true that selection don't need translations, then why don't delete them also from the other modules, firstly from stock_request? 🤔

Also fix flake8

rousseldenis:
> If it's true that selection don't need translations, then why don't delete them also from the other modules, firstly from stock_request? 

Agree with that. The fix should also go in base module.

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/stock-logistics-warehouse/pull/780 which is ready to merge.

PR #744, opened 2019-11-06
[IMP] stock_inventory_exclude_sublocation: No create new user for tests, only assign rights to demo user

sergio-teruel:
cc @Tecnativa
If you have password_security installed the test fails due to defaullt password restriction when user is created


pedrobaeza (approved):

pedrobaeza:
As aclaration, this goes against the current of not depending on demo data, but in this case, it's better to go for this workaround. Maybe assigning an strong password is a better approach (or no password - is that allowed?).

sergio-teruel:
Yes... only need remove password key

pedrobaeza:
Then we can fast-track this

/ocabot merge patch

PR #745, opened 2019-11-08
[12.0] [IMP] stock_inventory_preparation_filter

ivantodorovich:
This PR improves the stock_inventory_preparation_filter module by making it easier to filter on products.

**Replace the many2many_tags in product_ids**
Using the standard many2many widget allows the user to easily search and select a list of products
Without this PR the user is expected to search one by one

Screen record from 2019-11-08 10 48 43

**Added a new filter type: domain**
This enables more complex searches, like searching based on a related field's field.
Screen record from 2019-11-08 10 43 54

### DIFF
--- a/stock_inventory_preparation_filter/models/stock_inventory.py
+++ b/stock_inventory_preparation_filter/models/stock_inventory.py
@@ -4,6 +4,7 @@
 
 from odoo import _, api, fields, models
 from odoo.addons import decimal_precision as dp
+from odoo.tools.safe_eval import safe_eval
 
 
 class StockInventoryEmptyLines(models.Model):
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Don't introduce several things in the same PR: do separate PRs. About the m2m_tags change, I say no. I think you need to know: https://github.com/OCA/web/tree/12.0/web_widget_many2many_tags_multi_selection

rousseldenis:
@pedrobaeza This concerns the same module. Changes are in separate commits so review is easy. So, the PR can stay as is.

pedrobaeza:
@rousseldenis of course they are of the same module, but very different scopes, so having in separate PRs would lead to different merge times if one generates discussion and the other not. And this is the case, I don't want the change on the m2m_tags.

ivantodorovich:
Hi @pedrobaeza 

I'm closing this as it's replaced by these two PRS:
https://github.com/OCA/stock-logistics-warehouse/pull/747
https://github.com/OCA/stock-logistics-warehouse/pull/746

pedrobaeza:
Thanks. I will comment on them.

PR #749, opened 2019-11-12
Correction for translation issue on menu item

augustodinizl:
as posted on issue:#741 only aim to correct the pointed error

### DIFF
--- a/stock_request_submit/models/stock_request.py
+++ b/stock_request_submit/models/stock_request.py
@@ -9,10 +9,10 @@ class StockRequest(models.Model):
 
     def __get_request_states(self):
         states = super().__get_request_states()
-        if not ('submitted', _('Submitted')) in states:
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Thanks :+1: 

(note for whoever merges it, it would be would to squash the fix into one commit)

LoisRForgeFlow (changes requested):
Actually there is a small lint that you need to fix: 

`stock_request_submit/models/stock_request.py:4:1: F401 'odoo._' imported but unused`

rousseldenis:
@augustodinizl Could you change PR's title to mention module name ?

LoisRForgeFlow:
This issue has been fixed here: https://github.com/OCA/stock-logistics-warehouse/pull/780.

PR #752, opened 2019-11-13
[11.0][FIX] stock_secondary_unit: move line create round

enriquemartin:
On move line creation, the precision of secondary qty rounding is wrong; the unit of measure factor is being used instead of rounding.


sergio-teruel (approved):
Yes, of course!!!! Can do you do v12 PR too??
https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_secondary_unit/models/stock_move.py#L46

pedrobaeza (approved):
Fast-tracking this

enriquemartin:
> 
> 
> Yes, of course!!!! Can do you do v12 PR too??
> https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_secondary_unit/models/stock_move.py#L46

Done! #753

PR #753, opened 2019-11-13
[12.0][FIX] stock_secondary_unit: move line create round

enriquemartin:
On move line creation, the precision of secondary qty rounding is wrong; the unit of measure factor is being used instead of rounding.


pedrobaeza:
Fast-tracking as in previous version

/ocabot merge patch

PR #758, opened 2019-11-14
[12.0][IMP] stock_cycle_count: Create test user without password to pass tests with other modules like password_policy

sergio-teruel:
cc @Tecnativa
@pedrobaeza @chienandalu Can you review?


pedrobaeza (approved):
Let's fast track it

PR #762, opened 2019-11-18
[12.0][ADD] stock_picking_location_info

JoanSForgeFlow:
New module:

In a stock picking, it appears the final destination of each product line, taking into account put away strategies.

cc : ~ @Eficent @ageficent @NuriaMartinXifre


stock_picking_location_info/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019  Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Stock picking location info',
+    'version': '12.0.1.0.0',
+    'author': "Eficent, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/stock-logistics-warehouse",
+    'category': 'Stock',
```
rousseldenis: Set this to 'Warehouse'

stock_picking_location_info/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019  Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Stock picking location info',
```
rousseldenis: IMHO the module name is too generic according to the field introduced (location_putaway_dest_id).

Either the aim is to make a generic module (base one) to add further information on picking level, either the module has to be renamed

rousseldenis (changes requested):

PR #763, opened 2019-11-20
[11.0] Improve performance of stock available of BoMs

guewen:
Use an optimized method to explode the BoM. The explode methods
include data and a cycle check that we don't need here. Besides,
it calls '_bom_find' on every bom line of the graph, which generates
numerous SELECT queries on large BoMs (`current_line.product_id.bom_id` uses the prefetched one).


guewen:
I'm still investigating since I have the feeling we can still improve it.

dreispt:
<s>Superseded by #1267 </s> It isn't - they target different branches

PR #783, opened 2019-12-02
[12.0] Add stock_orderpoint_route

guewen:
Adds a selection of route for moves created for an orderpoint.

### DIFF
--- a/stock_orderpoint_route/__init__.py
+++ b/stock_orderpoint_route/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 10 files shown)


LoisRForgeFlow:
Isn't this very similar to https://github.com/OCA/stock-logistics-warehouse/pull/537?

guewen:
@lreficent Oh, you are totally right, I missed this. I'll port #573 to 12.0 then.
BTW, don't you think the `orderpoint_selectable` on the routes make sense to have the same mechanism as the others (sale order line selectable, ...)?

LoisRForgeFlow:
> BTW, don't you think the orderpoint_selectable on the routes make sense to have the same mechanism as the others (sale order line selectable, ...)?

@guewen yes, I agree, that's a good improvement!

PR #785, opened 2019-12-05
[ADD][MIG][12.0] stock_orderpoint_route

guewen:
Port to 12.0 of https://github.com/OCA/stock-logistics-warehouse/pull/537

This module allows to restrict a specific route to be used in the
reordering rules. This route will be used instead of the default determined by
default.


LoisRForgeFlow:
As you propsed in the other PR:

> BTW, don't you think the orderpoint_selectable on the routes make sense to have the same mechanism as the others (sale order line selectable, ...)?

I think that's a good idea if you want to include it here.

stock_orderpoint_route/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Stock Orderpoint Route",
+    "summary": "Allows to force a route to be used when procuring from "
+               "orderpoints",
+    "version": "12.0.1.0.0",
+    "license": "LGPL-3",
+    "website": "https://github.com/stock-logistics-warehouse",
+    "author": "Eficent, "
+              "Camptocamp, "
+              "Odoo Community Association (OCA)",
+    "category": "Warehouse Management",
```
rousseldenis: Please set this to 'Warehouse'

rousseldenis (approved):
LGTM. Minor comment

stock_orderpoint_route/__manifest__.py
guewen: Done

Issue #640, opened 2019-07-05
[RFC][WMS] Split warehouse operations in relevant picking type


jgrandguillaume:
Source : https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#
Part of: https://github.com/OCA/wms/issues/1

**This RCF to propose a POC implementation for the following**

- A "picking zone" represents a zone where pickers will work on their related operations. They will be linked to a picking type (therefor hold a default source and destination location)
- In important warehouses, operations are split according to several criteria. The goals here are:
  - to define/split operations in different picking and picking type according to the picking zone settings
  - sometimes change the destination of a move according to the zone default destination location (when a zone have a handover point for example). In such a case, an additional move will be created automatically to reach the original destination of the move (from handover point to pack zone for example).
- Route explain the steps you want to produce whereas the “picking zone” define how operations are grouped according to their final source and destination location.
  - This allows for example:
    - To parallelize picking operations in two main zone of a warehouse, splitting them in two different picking type
    - To define pre-picking (wave) in some sub-zones, then roundtrip picking of the sub-zone waves
    - **Additional note**: If your product **_are stored in one zone only_**, you can already achieve this in Odoo using **_properly configured route_** (and related picking type) on products (and categories)
- The logistic operations will be split among zones at "removal strategy" computation time (reservation time). This means route are followed up to create the different steps using the pull rules in place (e.g. pick + ship). At reservation time, operations are split in separate pickings among “picking zones” according to source and destination of the moves
  - -> Having this occurring after the application of the removal strategy make sense, because it is **_only when we know exactly which product/lot we will take that will want to classify the proper operation_** in the proper picking and picking type (opposite to the route that compute this all in advance before knowing that)
- This behavior must allow pre-picking (wave) operations and round trip picking using the proper configuration of location and picking type. Pre-picking will bring goods to handover points, round trip picking will gather them all up to the next operation (e.g. pack)

**Example: Split operation by zones**

- Setup WH pick + ship
- Desired operations:
  - Fridge and Normal stock is processed as pre-picking (waves)
  - Long goods and small goods as well as handover of the fridge and normal stock handover points are all gathered in a roundtrip picking
- Order goods stored in Long goods, small goods, fridge and normal stock
- Location
  - Output
  - Stock
    - Long goods
    - Small goods
    - Fridge
    - Normal stock
    - Output Fridge
[...]

PR #788, opened 2019-12-13
[WMS][13.0] Add stock_dynamic_routing - alpha version

guewen:
# Introduction

Port of #639 

Detailed commit history for the past developments
can be found in https://github.com/guewen/stock-logistics-warehouse/tree/12.0-add-stock_picking_zone

Additional changes in data model and features to expect, so the PR is a draft now.

# Description

Route explains the steps you want to produce whereas the “picking routing
operation” defines how operations are grouped according to their final source
and destination location.

This allows for example:

* To parallelize picking operations in two locations of a warehouse, splitting
  them in two different picking type
* To define pre-picking (wave) in some sub-locations, then roundtrip picking of
  the sub-location waves

Context for the use cases:

In the warehouse, you have a High-Bay which requires to place goods in a
handover when you move goods in or out of it. The High-Bay contains many
sub-locations.

A product can be stored either in the High-Bay, either in the Shelving zone.

When picking:

When there is enough stock in the Shelving, you expect the moves to have the
usual Pick(Highbay)-Pack-Ship steps. If the good is picked from the High-Bay, you will
need an extra operation: Pick(Highbay)-Handover-Pack-Ship.

This is what this feature is doing: on the High-Bay location, you define
a "routing operation". A routing operation is based on a picking type.
The extra operation will have the selected picking type, and the new move
will have the source destination of the picking type.

When putting away:

A put-away rule targets the High-Bay location.
An operation Input-Highbay is created. You expect Input-Handover-Highbay.

You can configure a routing operation for the put-away on the High-Bay Location.
The picking type of the new Handover move will the routing operation selected,
and its destination will be the destination of the picking type.

This is the implementation of "Warehouse operations by zones" RFC: https://github.com/OCA/stock-logistics-warehouse/issues/640 original document is the WMS document:

# Try on runbot

* In Inventory Settings, activate:

  * Storage Locations
  * Multi-Warehouses
  * Multi-Step Routes

The initial setup in the demo data contains locations:

* WH/Stock/Highbay
* WH/Stock/Highbay/Bin 1
* WH/Stock/Highbay/Bin 2
* WH/Stock/Handover

The "Highbay" location (and children) is configured to:

* create a source routing operation from Highbay to Handover when
  goods are taken from Highbay (using a new picking type Highbay → Handover)
* create a destination routing operation from Handover to Highbay when 
  goods are put to Highbay (using a new picking type Handover → Highbay)

Steps to try the Source Routing Operation:

* In the main Warehouse, configure outgoing shipments to "Send goods in output and then deliver (2 steps)"
* Inventory a product, for instance "[FURN_8999] Three-Seat Sofa", add 50 items in "WH/Stock/Highbay/Bay A/Bin 1", and nowhere else
* Create a sales order with 5 "[FURN_8999] Three-Seat Sofa", confirm
* You'll have 3 transfers; a new one has been created dynamically for Highbay -> Handover.

Steps to try the Destination Routing Operation:

* In the "WH/Stock" location, create a Put-Away Strategy with:

  * "[DESK0004] Customizable Desk (Aluminium, Black)" to location "WH/Stock/Highbay/Bay A/Bin 1"
  * "[E-COM06] Corner Desk Right Sit" to location "WH/Stock/Shelf 1"

* Create a new purchase order of:

  * 5 "[DESK0004] Customizable Desk (Aluminium, Black)"
  * 5 "[E-COM06] Corner Desk Right Sit"

* Confirm the purchase
* You'll have 2 transfers:

  * one to move DESK0004 from Supplier → Handover and E-COM06 from Supplier → Shelf 1
  * one waiting on the other to move DESK0004 from Handover → WH/Stock/Highbay/Bay A/Bin 1 (the final location of the put-away)


guewen:
Check this: https://github.com/OCA/stock-logistics-warehouse/pull/639#discussion_r356151335

stock_routing_operation/models/stock_move.py
```diff
@@ -0,0 +1,317 @@
...
+                current_picking_type, move.location_id, destination
+            )
+
+            picking = move.picking_id
+            move._assign_picking()
+            if not picking.move_lines:
+                # When the picking type changes, it will create a new picking
+                # for the move. If the previous picking has no other move,
+                # we have to drop it.
+                picking.unlink()
```
florian-dacosta: Hi @guewen 
I know everything in the module may change drastically, but this problematic will probably continues (if this is done at reservation time at least), so I share my thoughts already.
I think deleting the picking is important, because as you say, it may be empty, but doing it here is an issue.

Indeed, the `_action_assign `of the `stock.move` is normally called by `action_assign` on `stock.picking`
And, in case we call `action_assign` on the pickings, the risk is that the picking itself is deleted during the call.
Meaning that any action on the picking done after the call of `moves._action_assign()` may fail.

I am not sure it happens with Odoo natives modules only, but I guess it is a big risk to conflict with multiple OCA modules or custom modules I guess.
Any module overriding `action_assign` of `stock.picking` and doing stuff after the super has big chances to fail because the record won't exist anymore.

Well, Unfortunately, I don't really have a good solution about it... I guess we could delete pickings afterward with a cron...(or a job but we don't want to depend on queue job for this module of course...)

What do you think?
Another minor issue is that if a user click on check availability on the button present in the picking form, it may be deleted. It may be a bit disturbing.

guewen: > Well, Unfortunately, I don't really have a good solution about it... I guess we could delete pickings afterward with a cron...(or a job but we don't want to depend on queue job for this module of course...)

And maybe it wouldn't solve everything, because if a module *does something* with the picking, the *does something* part is probably lost, or doing unwanted behavior.
@jgrandguillaume thoughts? (tldr: if we have a picking with a single move and the move is assigned to a new picking type, it means the original picking is empty (or several moves but all moves are moved). We should remove it, but removing it can make fail other parts of the code using the picking.)

florian-dacosta: > And maybe it wouldn't solve everything, because if a module does something with the picking, the does something part is probably lost, or doing unwanted behavior.

I agree, it would not solve everything. Something may apply for instance to the new created picking (the replacement(s)).
At least it would not fail.

Another issue with the change of picking is that some relevant information present in the original picking, may be relevant (or not) for the newly created picking, but the `_assign_picking` won't take care of it. Because it just put some basic info, but it is not a "copy" of the initial picking.
Like a data manually put on the initial picking, that would still be relevant for the new picking replacing it. (I don't specially have an exemple, could be a simple note relative to the expedition...)
In the same way, depending on the information/case, we would want or not to manage it...
Well, I guess this may be a specific part to manage if necessary that do not need to be attended in the module.

jgrandguillaume (approved):
Functional test ok. I suggest to merge it as Alpha. We have a roadmap for the next steps and wants to extract this routing things in another separate model.

stock_routing_operation/models/stock_move.py
guewen: we can cancel the picking, a bit better than deleting

florian-dacosta: Yes, I think it is a good solution. 

I would do this part in a separated method, in which we cancel the picking.
like `picking.post_process...()`
So anyone could override it easily to do another processing for this kind of picking.  Like if I want to delete it because I know it is safe depending on the other module installed on my instance, I still can do it. (or just delay the deletion).

Ok for me to merge then!

guewen: Not ready for merge though :D

florian-dacosta: What is missing? (I mean beside this post_process matter)

stock_routing_operation/models/stock_move.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def _apply_src_move_routing_operation(self):
+        src_moves = self._split_per_src_routing_operation()
+        src_moves._apply_move_location_src_routing_operation()
+
+    def _apply_dest_move_routing_operation(self):
+        dest_moves = self._split_per_dest_routing_operation()
+        dest_moves._apply_move_location_dest_routing_operation()
+
+    def _bypass_routing_operation_application(self, routing_type):
```
guewen: @florian-dacosta FYI, I'll remove this in the next evolutions of the module: it's possible to filter the moves by inheriting `_apply_dest_move_routing_operation` and `_apply_src_move_routing_operation` and calling super only for the records you want, and it's more optimized because you can filter all the moves at once (`_bypass_routing_operation_application` is called record by record).

Also, I'll add an optional domain on the routing operation itself: if the domain returns no result for a move, the routing is not applied.

With these 2 possibilities, I'm sure you can handle your use case.

florian-dacosta: @guewen 
Ok sure, I'll be able to manage it this way for sure, thanks.

I am actually working on it once more (still v8 thought !) and I have to manage another case...(a lot of exception you will tell me), but it still for the same problematic that, I want to apply the routing operation depending on the carrier (if any), that is my one big requirement that really make me differ on the cases handled by the module.

And so, I have a problem on this part : https://github.com/guewen/stock-logistics-warehouse/blob/13.0-add-stock_routing_operation/stock_routing_operation/models/stock_location.py#L90
You stop the loop on the first picking type found returning it.
In my case, the first one found may not be relevant, depending on my carrier. Then, I need to find the next one...etc...
There is not way for me to do it with the present module.
I see kind of 3 solutions
1) passing the move to the `_find_picking_type_for_routing` on location and add a hook in the loop to check if the found picking type is ok
2) just add a hook in the loop like `if match._is_allowed():` and manage my case with context and overriding `_is_allowed`
3) Add more logic in `_find_picking_type_for_routing` on `stock.move` and less on location side, but I don't think you will like that.

But anyway, I don't think it is worth proposing something about it now, because you say that the routing operation finding will change drastically in the neart future.
Beside, since I will definitly need the module for version 14 in the relatively near future, and I guess (hope) it won't be very different from v13, I can help with this refactore in v13 if needed.

florian-dacosta: > @florian-dacosta FYI, I'll remove this in the next evolutions of the module: it's possible to filter the moves by inheriting _apply_dest_move_routing_operation and _apply_src_move_routing_operation and calling super only for the records you want, and it's more optimized because you can filter all the moves at once (_bypass_routing_operation_application is called record by record).

Following this logic, the check on move state (https://github.com/guewen/stock-logistics-warehouse/blob/13.0-add-stock_routing_operation/stock_routing_operation/models/stock_move.py#L49) should also be done in `_apply_src_move_routing_operation` right?

stock_routing_operation/models/stock_move.py
guewen: As I said earlier, I have a lot of things to change in this module, notably the datamodel will have dedicated models. 
The change I'm working on currently is a change of the rules:

* if the move's dest location is a child of the routing's dest location: it's more precise so change only the picking type
* if the move's dest location is a parent of the routing's dest location: change the dest location to the routing's dest location and change the picking type
* if the move's dest location is outside of the routing's dest location: add a routing operation before

Which means that even if your destinations were the right ones, the move will be classified in the expected picking type.

stock_routing_operation/models/stock_move.py
guewen: > [](https://github.com/guewen/stock-logistics-warehouse/blob/13.0-add-stock_routing_operation/stock_routing_operation/models/stock_location.py#L90)
> You stop the loop on the first picking type found returning it.
> In my case, the first one found may not be relevant, depending on my carrier. Then, I need to find the next one...etc...

Should be supported with the change of data model we plan:

We'll have a model for routing with a location (M2o) field and routing rules, each rule having:

* sequence
* applied on source or destination
* picking type to apply
* domain

The first routing rule to match being applied.

stock_routing_operation/models/stock_move.py
florian-dacosta: > Which means that even if your destinations were the right ones, the move will be classified in the expected picking type.

That's nice, that is something I wanted to change. (change at least picking type, even if location dest is the same for instance)

stock_routing_operation/models/stock_move.py
florian-dacosta: Well, that's nice, I look forward to see these  changes !
Once again, if I can be of any help, let me know. Else, I'll test and review when it is done

stock_routing_operation/models/stock_move.py
```diff
@@ -0,0 +1,403 @@
...
+        super()._action_assign()
+
+        moves_routing = self._routing_compute_rules()
+
+        if not any(rule for routing in moves_routing.values() for rule in routing):
+            # no routing to apply, so the reservations done by _action_assign
+            # are valid and we can resolve to a normal flow
+            self.env["base"].flush()
+            # pylint: disable=sql-injection
+            self.env.cr.execute(
```
florian-dacosta: It will be assigned later with the _action_assign anyway right? so the moves will be assigned twice (not sure it has an incidence on the data thought, but if we could avoid it, it would be better)

stock_routing_operation/models/stock_move.py
```diff
@@ -0,0 +1,403 @@
...
+                # a parent or a child) to the routing destination: in this case
+                # we have to add a routing move before the current move to
+                # route the goods in the correct place
+                move._routing_pull_insert_move(
+                    routing_rule, current_picking_type, original_destination
+                )
+
+            pickings_to_check_for_emptiness |= move.picking_id
+            move._assign_picking()
+            move._action_assign()
```
florian-dacosta: We should not assign the move here because it will be just after when the _action_assign's super will be called?

florian-dacosta (commented):
Thanks for the hard work, the module is really flexible and seems quite clean/well done to me!
I like the flexibility and it meets my use case now without need of adding several hooks and so on.

I see  you are still working on it, so I keep a complete review for later, but still, there is one thing I have a hard time to understand, why do you have to call an `_action_assign` (and then roll back it) during the process?
What do we really gain in comparison to not doing it and just call the whole routing process after the _action_assign super, instead of before? (like it was in the previous version)

I think it adds a lot of complexity, and in some case may lead to unwanted/not needed action.
I mean, we could, after a change in the configuration for example, want to call the `_split_and_apply_routing` method on already assigned moves. Then we do not need to reassign it (and rollback it!)
Also, I wonder if it could be a problem to only call the super and not the whole method. It could probably lead to unwanted behaviors depending on action_assign override by depending modules.

guewen:
todo: rename to stock_dynamic_routing and move to OCA/wms

guewen:
> I see you are still working on it, so I keep a complete review for later, but still, there is one thing I have a hard time to understand, why do you have to call an `_action_assign` (and then roll back it) during the process?


We have to "peek" at what will be reserved to know the available quantities in which source location, so we can split the move in the parts that will be reserved and the parts that cannot be reserved and apply different routing rules. The part of the move that will be reserved will have its routing applied, the other part will not change. We cannot do it after the assign is done, because applying the routing may change the source of the moves, change the picking of the move, ...


> What do we really gain in comparison to not doing it and just call the whole routing process after the _action_assign super, instead of before? (like it was in the previous version)

It wasn't. The previous version had to call action_assign and unreserve before applying the routing.

> I think it adds a lot of complexity, and in some case may lead to unwanted/not needed action.
> I mean, we could, after a change in the configuration for example, want to call the `_split_and_apply_routing` method on already assigned moves. Then we do not need to reassign it (and rollback it!)

It's much more simple and clean to release a savepoint that the previous version which was calling super on _action_assign and then calling unreserve -> we are guaranteed that anything done for the first "action_assign" is rollbacked, while in the previous version we could have leftovers, or any action done by another module could be kept.

If you think it adds complexity, I think you missed how complex it was before ;)
As for calling `_split_and_apply_routing` on an assigned move, well, you can't.




> Also, I wonder if it could be a problem to only call the super and not the whole method. It could probably lead to unwanted behaviors depending on action_assign override by depending modules.

As for any module overriding a method, the order of the calls to the method could lead to issues sometimes, but I don't see more risk here than any other override. 
We call super in every execution path, and the first call to action_assign is either kept if no routing, or rollbacked entirely if a routing has to be called (which means *anything* done by another module is rollbacked and will be done again when assign is called on the routed moves).


The annoying point is that we have to execute _action_assign twice (only when a routing is applied), which is not great for performance, but I reckon there is no other way.

stock_routing_operation/models/stock_move.py
guewen: In this execution path, there is no routing to apply, so the savepoint is released and we return from `_action_assign()` (see https://github.com/OCA/stock-logistics-warehouse/pull/788/files/f43ea35ac03a1a86b2e9b6549daa8426d5cd68b8#diff-0295a9825355a3ac40a9eac755697f9bR67-R68)

guewen: Oh, I see what you mean, I missed this.

stock_routing_operation/models/stock_move.py
guewen: It has to be reserved at this point. If the routing has to split a move because it was partially available, you want to control which move will be reserved: the one on which you applied the routing which is this one. So we assign it first. Note that the second call to `_action_assign()` will do nothing because we are guaranteed to have a state "assigned" on this move and the method excludes this state.

stock_routing_operation/models/stock_move.py
guewen: Fixed

stock_routing_operation/models/stock_move.py
guewen: Added a comment

jgrandguillaume:
@guewen Would be really nice to move it to WMS repo, I'm trying to update the issue to help people followup the overall work done

jbaudoux (approved):
This works now like a charm. This is not alpha version anymore.
This module is stunning !

guewen:
> It should therefore be ready to merge by a maintainer

Note:
@jgrandguillaume wants me to move it to OCA/wms, so should not be merged here.

guewen:
@jgrandguillaume Moved to https://github.com/OCA/wms/pull/31 and changed to Beta.

cc @florian-dacosta

PR #790, opened 2019-12-16
[WMS][13.0] Add stock_reserve_rule

guewen:
Port of #683 to 13.0

Requires https://github.com/OCA/product-attribute/pull/536 first.

This module adds rules for advanced reservation / removal strategies.

Rules are applied on a location and its sub-locations.

A rule can exclude quants or locations based on configurable criteria,
and based on the selected quants, apply advanced removal strategies.

The rules have a sequence, which will be respected for the reservation.
So even without filter or advanced removal strategies, we can give a priority to
reserve in a location before another.

The advanced removal strategies are applied on top of the default one (fifo,
fefo, ...).

The included advanced removal strategies are:

* Default Removal Strategy: apply the default configured one (fifo, fefo, ...)
* Empty Bins: goods are removed from a bin only if the bin will be empty after
  the removal (favor smallest bins first, then apply the default removal
  strategy for equal quantities).
* Prefer Full Packaging: tries to remove full packaging (configured on the
  products) first, by largest to smallest package (default removal strategy is
  then applied for equal quantities).

Examples of scenario:

rules:

* location A: no filter, no advanced removal strategy
* location B: no filter, Empty Bins
* location C: no filter, no  advanced removal strategy

result:

* take what is available in location A
* then take in location B if available, only if bin(s) are emptied
* then take what is available in location C

The module is meant to be extensible, with a core mechanism on which new rules
and advanced removal strategies can be added.

It's part of the WMS addons to implement these specs:
https://docs.google.com/document/d/1mct6bFFWJqW01wGFcjc-uQNEjyCxvh6Y9TuFdRhe-b0/edit#heading=h.9ad56twdce31

Demo data has been created by the module.

Product: Funky Socks

Locations:

 * Stock / Zone A / Bin A1: 200 Funky socks
 * Stock / Zone B / Bin B1: 100 Funky socks
 * Stock / Zone C / Bin C1: 100 Funky socks

Reservation Rules, in the following order

 * Zone A must have full quantities
 * Zone B
 * Zone C

Delivery Orders:

* Origin: Outgoing shipment (reservation rules demo 1)
* Origin: Outgoing shipment (reservation rules demo 2)

Test:

* Activate Storage Locations and Multi-Warehouses
* You can open Inventory > Configuration > Stock Reservation Rules to see the rules
* Open Transfer: Outgoing shipment (reservation rules demo 1)
* Check availability: it has 150 units, as it will not empty Zone A, it will not take products there, it should take 100 in B and 50 in C (following the rules order)
* Unreserve this transfer (to test the second case)
* Open Transfer: Outgoing shipment (reservation rules demo 2)
* Check availability: it has 250 units, it can empty Zone A, it will take 200 in Bin A1 and 50 in Bin B1.
* If you want to explore further, you can add a custom domain to exclude rules (for instance, a product category will not use Zone B).


jgrandguillaume (approved):
Functional tests ok ! Go ahead

guewen:
Dependency is merged, I amended the last commit to trigger new builds.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -0,0 +1,278 @@
...
+
+    rule_domain = fields.Char(
+        string="Rule Domain",
+        default=[],
+        help="Domain based on Stock Moves, to define if the "
+        "rule is applicable or not.",
+    )
+
+    def _rules_for_location(self, location):
+        return self.search([("location_id", "parent_of", location.id)])
```
jbaudoux, line 67: I think a move should never look for a rule that could reserve goods in a location that is not a child of the move source location.

jbaudoux (changes requested):
You inverted in this commit https://github.com/OCA/stock-logistics-warehouse/pull/790/commits/768f186fd23e876c6daa7ccbd1423fbaf8b8cee0 parent_of into child_of. I don't agree with the optimization. After having fixed rule.location_id.parent_path.startswith(location.parent_path) into location.parent_path.startswith(rule.location_id.parent_path)
I compared both method and parent_of is more optimal. Also, the parent location structure was already in cache, so only 1 select on rules was triggered.

guewen:
You are right, the optimization based on `parent_path` doesn't make sense here as the ORM will read `parent_path` in the location and get the parent ids by splitting the ids, it doesn't need more than one query on stock_location which is done based on its id and can reuse the cache, there is no lookup on parent path for `parent_of`. Updating.


```
>>> env["stock.reserve.rule"].search([("location_id", "parent_of", 3125)])
2020-05-27 05:36:59,938 1 DEBUG log_p odoo.sql_db: query: SELECT "stock_location"."id" as "id","stock_location"."name" as "name","stock_location"."complete_name" as "complete_name","stock_location"."active" as "active","stock_location"."usage" as "usage","stock_location"."location_id" as "location_id","stock_location"."comment" as "comment","stock_location"."parent_path" as "parent_path", <stripped>,"stock_location"."create_uid" as "create_uid","stock_location"."create_date" as "create_date","stock_location"."write_uid" as "write_uid","stock_location"."write_date" as "write_date" FROM "stock_location" WHERE "stock_location".id IN (3125)
2020-05-27 05:36:59,942 1 DEBUG log_p odoo.sql_db: query: SELECT "stock_reserve_rule".id FROM "stock_reserve_rule" WHERE (("stock_reserve_rule"."active" = true)  AND  ("stock_reserve_rule"."location_id" in (1,7,8,133,134,135,144,207,3125))) ORDER BY "stock_reserve_rule"."sequence" ,"stock_reserve_rule"."id"
```

jgrandguillaume (approved):
I agree with that change.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -0,0 +1,305 @@
...
+        rounding = product.uom_id.rounding
+
+        def is_greater_eq(value, other):
+            return float_compare(value, other, precision_rounding=rounding) >= 0
+
+        for pack_quantity in packaging_quantities:
+            # Get quants quantity on each loop because they may change.
+            # Sort by max quant first so we have more chance to take a full
+            # package. But keep the original ordering for equal quantities!
+            bins = sorted(
```
jbaudoux, line 285: @jgrandguillaume I fear FEFO is broken here

PR #791, opened 2019-12-16
[MIG] stock_orderpoint_generator: Migration to 12.0

chienandalu:
Standar migration

cc @Tecnativa TT20139


pedrobaeza (approved):

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,189 @@
...
+        string='Products',
+        help="A reordering rule will be automatically created by the "
+             "scheduled action for every product in this list."
+    )
+    auto_last_generation = fields.Datetime(string='Last Automatic Generation')
+
+    def _template_fields_to_discard(self):
+        """In order to create every orderpoint we should pop this template
+           customization fields """
+        return [
```
Tardo, line 84: I see that this is used for "search" purposes... Use 'set' can increase the performance.

stock_orderpoint_generator/models/product.py
```diff
@@ -0,0 +1,122 @@
...
+        # default the compute the stock value anyway to default the value
+        # for products with no moves for the given period
+        initial_stock = {}
+        initial_stock = self.with_context(
+            location=location)._compute_quantities_dict(
+                False, False, False, to_date=from_date or to_date)
+        product_moves_dict = {}
+        for move in moves:
+            product_moves_dict.setdefault(move['product_id'][0], {})
+            product_moves_dict[move['product_id'][0]].update({
```
Tardo, line 84: Why don't you do it on the previous line?

Tardo (approved):
LGTM only non-blocking suggestions

stock_orderpoint_generator/models/product.py
chienandalu: There could be no key yet

stock_orderpoint_generator/models/orderpoint_template.py
chienandalu: Well, this method it's just to discard the fields not present in the original model used for templating purposes.

pedrobaeza:
As the questions have been answered, let's merge it:

/ocabot merge

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -36,12 +36,8 @@ class OrderpointTemplate(models.Model):
         help="Auto compute minimum quantity "
              "per product for a given a date range",
     )
-    auto_min_date_start = fields.Datetime(
-        string="From",
-    )
-    auto_min_date_end = fields.Datetime(
-        string="To",
-    )
+    auto_min_date_start = fields.Datetime()
```
rousseldenis, line 39: @chienandalu Why removing string ?

pedrobaeza: He has removed them in favor of the automatic ones for avoiding the warning of "duplicated labels", which in fact is more correct (for searches for example). They are put however on the view instead, for better display.

rousseldenis: Ok, saw in form, but for exports or ... the name will be Auto Min Date Start which could lead to inconsistencies

pedrobaeza: Well, on export you don't have context, so better to have full name, although it's not the same as you see in screen.

chienandalu: Yes, that was the exact reason :)

PR #798, opened 2019-12-18
[13.0][MIG] stock_removal_location_by_priority

ps-tubtim:


LoisRForgeFlow:
@ps-tubtim Tests are failing, can you have a look?

stock_removal_location_by_priority/models/stock_quant.py
```diff
@@ -14,7 +14,8 @@ class StockQuant(models.Model):
     )
 
     @api.model
-    def _get_removal_strategy_order(self, removal_strategy=None):
+    def _get_removal_strategy_order(self, removal_strategy):
+        self.flush()
```
LoisRForgeFlow: Could you clarify why this is needed?

rousseldenis: I think this is because computed (or related) fields are not automatically recomputed (flush() is done after a write()).

@ps-tubtim If this is needed for tests purpose, I suggest to use it specifically in tests.

ps-tubtim: @LoisRForgeFlow  @rousseldenis Thank you for suggestion. Edited.

rousseldenis (approved):
Code review

stock_removal_location_by_priority/models/stock_quant.py
LoisRForgeFlow: @rousseldenis But if you look at the Odoo code, it is not there, so that made me think that it was not needed there and the only reason was to make test work (what was the case regarding the latest update)

@ps-tubtim Thanks for the fix!

LoisRForgeFlow (approved):

stock_removal_location_by_priority/models/stock_quant.py
rousseldenis: > @rousseldenis But if you look at the Odoo code, it is not there, so that made me think that it was not needed there and the only reason was to make test work (what was the case regarding the latest update)

Your feeling was good.

flush() will commit everything in cache, so only useful in tests (see tests code in Odoo). A contrario as I said, it is only used when calling load() (and in unlink()) in BaseModel as we want to have fresh data.

PR #693, opened 2019-09-02
[WMS][12.0] Add stock_picking_completion_info - alpha version

grindtildeath:
This module adds completion information on stock picking.

If activated on the picking type, completion information is computed according
to the next chained pickings related to the stock moves of the actual picking.

In other words, if all the previous moves linked to the destination pickings 
moves are done, the completion of the actual picking allows the destination
pickings to be processed. In such case, a ribbon will appear on the stock 
picking form view, to inform the stock operator.

Relates to https://github.com/OCA/stock-logistics-warehouse/issues/640#issuecomment-525705914

PR #808, opened 2019-12-19
[MIG][13.0] stock_picking_completion_info

guewen:
Port #693
Depends on https://github.com/OCA/stock-logistics-warehouse/pull/861

### DIFF
--- a/stock_picking_completion_info/models/stock_picking.py
+++ b/stock_picking_completion_info/models/stock_picking.py
@@ -1,0 +1,64 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, models, fields
+
+
+class PickingType(models.Model):
+
+    _inherit = 'stock.picking.type'
+
+    display_completion_info = fields.Boolean(
+        help='Inform operator of a completed operation at processing and at'
+        ' completion'
+    )
+
+
+class StockPicking(models.Model):
+
+    _inherit = 'stock.picking'
+
+    completion_info = fields.Selection(
+        [
+            ('no', 'No'),
+            (
+                'last_picking',
+                'Completion of this operation allows next operations to be '
+                'processed.',
+            ),
+            (
+                'next_picking_ready',
+                'Next operations are ready to be processed.',
+            ),
+        ],
+        compute='_compute_completion_info',
+    )
+
+    @api.depends(
+        'picking_type_id.display_completion_info',
+        'move_lines.move_dest_ids.move_orig_ids.state',
+    )
+    def _compute_completion_info(self):
+        for picking in self:
+            if (
+                picking.state == 'draft'
+                or not picking.picking_type_id.display_completion_info
+            ):
+                picking.completion_info = 'no'
+                continue
+            # Depending moves are all the origin moves linked to the
+            # destination pickings' moves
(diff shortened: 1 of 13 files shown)


grindtildeath:
We should probably rework this module to depend on https://github.com/OCA/stock-logistics-warehouse/pull/861

simahawk:
@guewen any feedback?

guewen:
@simahawk well, @grindtildeath knows better, he wrote this module :)

stock_picking_completion_info/models/stock_picking.py
```diff
@@ -0,0 +1,86 @@
...
+                "Full order picking: You are processing a full order picking "
+                "that will allow next operation to be processed",
+            ),
+        ],
+        compute="_compute_completion_info",
+    )
+
+    @api.depends(
+        "picking_type_id.display_completion_info",
+        "move_lines.move_dest_ids.move_orig_ids.state",
```
grindtildeath: Wrong dependency (cf line 53)

grindtildeath:
@guewen Can you please merge https://github.com/guewen/stock-logistics-warehouse/pull/8 ? 

Maybe also add module dependency https://github.com/OCA/stock-logistics-warehouse/pull/861 in the PR description?

Tonow-c2c:
@guewen did you enough right to restart travis ?

simahawk:
@grindtildeath if you have time and you want to finish this, please take it over because Guewen is on holidays for 1 week :wink:

grindtildeath:
@simahawk @guewen Seems we only need a rebase here, and it should be fine...

guewen:
Rebased

simahawk:
@guewen is :red_circle:

stock_picking_completion_info/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def write(self, vals):
+        super().write(vals)
+        if "state" in vals:
+            # invalidate cache, the api.depends do not allow to find all
+            # the conditions to invalidate the field
+            self.env["stock.picking"].invalidate_cache(fnames=["completion_info"])
```
rousseldenis, line 85: Is it real or is it the behaviour you have in tests ? If it is just in tests, do the invalidate_cache() there and not here.

guewen:
@grindtildeath tests fail after de4dcfd, can you have a look?

grindtildeath:
@guewen should be fixed with https://github.com/guewen/stock-logistics-warehouse/pull/10

stock_picking_completion_info/models/stock_picking.py
grindtildeath: I don't have an issue running the tests without this override, @guewen might know better.

grindtildeath (approved):

stock_picking_completion_info/models/stock_picking.py
guewen: I think it was not only in tests, IIRC the dependencies in `@api.depends()` of the computed field could not be triggered in every case in chained moves. But maybe now that we have `"move_lines.common_dest_move_ids.state",` in the dependencies it's no longer the case and this could be removed?

jgrandguillaume (approved):
LGTM

PR #811, opened 2019-12-20
[13.0][MIG] stock_inventory_lockdown

AdriaGForgeFlow:
Standard migration of module stock_inventory_lockdown to v13.

cc ~ @ForgeFlow


rousseldenis (approved):
Code and functional review

LoisRForgeFlow (approved):
Technical review, LGTM

stock_inventory_lockdown/models/stock_location.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class StockLocation(models.Model):
+    """Refuse changes during exhaustive Inventories"""
+
+    _inherit = "stock.location"
+    _order = "name"
```
guewen, line 12: @AdriaGForgeFlow @rousseldenis 
Do you know if there is a reason for changing the default order (default is `complete_name`)?
It seems pretty brutal. In my case, it makes tests fail in https://github.com/OCA/stock-logistics-warehouse/pull/997

rousseldenis: Indeed, I didn't notice it. This is obviously not advised to change _order parameter.

@LoisRForgeFlow @AdriaGForgeFlow Is this needed ?

rousseldenis: @AdriaGForgeFlow

AdriaGForgeFlow: This is coming from the creation of this module in v9. I guess it made sense back then but I'm not really sure it is needed anymore, probably not

PR #810, opened 2019-12-20
[13.0][MIG] stock_available_unreserved

LoisRForgeFlow:
Standard migration to v13.

@ForgeFlow

PR #812, opened 2019-12-22
Migrate to V13

ghost:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

LoisRForgeFlow:
There is already an open migration for this: https://github.com/OCA/stock-logistics-warehouse/pull/810

PR #813, opened 2019-12-23
[13.0][MIG] stock_change_qty_reason

AdriaGForgeFlow:
Standard migration of module stock_change_qty_reason to v13

**Major changes:**
Ability to add a reason when updating product quantity from Update Quantity button in Product Form View has been removed. This button (when Multiple Locations options is activated) no longer opens an Inventory Adjustment but the Stock Quant list view is opened instead.
It must be discussed how to implement the logic in v13 so it has been added to the ROADMAP.

cc ~ @ForgeFlow


LoisRForgeFlow:
@AdriaGForgeFlow in such a case, maybe a manual update of the qty in the quant could generate a log (something like "stock.qty.change.log") with the date and difference (new_qty - prev_qty). That would cover that basic usecase, but it would not even add a view/report for it and let it know that in the roadmap (anyone really using this functionallity with no multi-locations enable, could propose and improvement adding that report and/or view).

What do you think?

EDIT: or even just describing this idea in the README and letting anyone needing it to implement it.

LoisRForgeFlow (approved):
LGTM :+1:

HviorForgeFlow (approved):

PR #816, opened 2019-12-27
[12.0][REF] stock_demand_estimate: split into 2 modules

LoisRForgeFlow:
Split into:

* stock_demand_estimate: just the datamodel.
* stock_demand_estimate_matrix (any suggestion for a better name?): adds the wizard to assist on the estimates creation and update.

@ForgeFlow


JordiBForgeFlow (approved):

stock_demand_estimate_matrix/models/stock_demand_estimate.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.multi
+    @api.depends(
+        "date_range_id", "manual_duration", "manual_date_from", "manual_date_to",
+    )
+    def _compute_dates(self):
+        date_range_records = self.filtered(lambda r: r.date_range_id)
+        res = super(
+            StockDemandEstimate, self - date_range_records)._compute_dates()
+        for rec in date_range_records:
+            rec.date_from = rec.date_range_id.date_start
```
HviorForgeFlow, line 25: Instead of multiple assignation a write could be used

LoisRForgeFlow: This is a compute method, you cannot do that

LoisRForgeFlow: plus it depends on the date_range_id of each record

AdriaGForgeFlow (approved):
Functional review :+1:

PR #824, opened 2020-01-07
[13.0][MIG] stock_inventory_discrepancy

HviorForgeFlow:
Standard migration to v13

CC @ForgeFlow


AdriaGForgeFlow:
@HviorForgeFlow what do you think? Not the cleanest way but is the only way I found to do it.

rousseldenis:
@HviorForgeFlow Do you plan to add few tests for missing coverage (as this module becomes really mature) ?

HviorForgeFlow:
@rousseldenis I'm not sure about validity on last commit I pushed. It intends to recursive set child locations with the same threshold as the parent. 
Could you do a functional test to ensure it is valid, then I can add tests covering that case.

HviorForgeFlow:
@jimhoefnagels reviews are welcome

stock_inventory_discrepancy/views/stock_warehouse_view.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017-2020 ForgeFlow S.L.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record id="view_warehouse_form" model="ir.ui.view">
```
jimhoefnagels: Had to look it up in the contributing guidelines but it is preferred "naming the inherited view after the original over using a name which follows the guidelines. This eases looking up the original view and other inheritance if they all have the same name."
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#253inherited-xml

stock_inventory_discrepancy/views/stock_inventory_view.xml
```diff
@@ -0,0 +1,59 @@
...
+            <xpath expr="//button[@name='action_cancel_draft'][2]" position="attributes">
+                <attribute name="states">pending,confirm</attribute>
+            </xpath>
+            <field name="company_id" position="before">
+                <field name="over_discrepancy_line_count" attrs="{'invisible': [('state', '!=', 'pending')]}"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="Va" model="ir.ui.view">
```
jimhoefnagels: change id to parent id: stock_inventory_line_tree2

stock_inventory_discrepancy/models/stock_location.py
```diff
@@ -15,3 +15,22 @@ class StockLocation(models.Model):
         "an Inventory Adjustment. Thresholds defined in Locations have "
         "preference over Warehouse's ones.",
     )
+    update_child_location = fields.Boolean(
+        string="Propagate to child locations",
+        help="Propagate Maximum Discrepancy Rate Threshold to child locations",
+    )
```
LoisRForgeFlow: I see the naming very generic. IMO something like `propagate_discrepancy_threshold` and a string "Propagate discrepancy threshold" would be more appropriate.

LoisRForgeFlow (commented):
Code review and functional review. Nice work here, this was not straight forward :+1: 

Just check my comment below about naming.

stock_inventory_discrepancy/models/stock_location.py
HviorForgeFlow: Done!

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
@jimhoefnagels could you update your review?

stock_inventory_discrepancy/hooks.py
```diff
@@ -0,0 +1,79 @@
...
+                "stock_inventory_discrepancy.group_stock_inventory_validation"
+            )
+            and not self.user_has_groups(
+                "stock_inventory_discrepancy.group_stock_inventory_validation_always"
+            )
+        ):
+            raise UserError(
+                _("Only a stock manager can validate an inventory adjustment.")
+            )
+        if self.state not in ["confirm", "pending"]:
```
rousseldenis, line 32: In which case do you need the inventory pending at validation (As I understand, the only thing you change with this hook ) ?

Why not having something that make pending inventory automatically confirmed ?

A little bit too much code copy for that.

jimhoefnagels (approved):
LGTM

stock_inventory_discrepancy/hooks.py
HviorForgeFlow: It is required for the module to run properly

LoisRForgeFlow:
@rousseldenis is this good to merge for you?

rousseldenis (approved):

PR #828, opened 2020-01-13
[13.0] Add isort third party detection from requirements.txt

guewen:
The current setup used in the OCA for isort is done by using 2
pre-commit hooks:

1. seed-isort-config: find third-party libs and updates the '.isort.cfg' file's "known_third_party" with the list
2. isort: runs isort, which will use the list of third party libs provided by 'seed-isort-config'

This is [an issue](https://github.com/OCA/maintainer-quality-tools/issues/625) with the workflow adopted by many contributors of the OCA: as the pull requests may need some time to be merged, we use temporary branches where we aggregate the various PRs.

If several pull requests add a third party library, we'll have a conflict, and when the process of merging these pull requests is automated (e.g. with [git-aggregator](https://github.com/acsone/git-aggregator)), it becomes tedious.

The list of libs in the "known_third_party" variable is not even editable manually, as "seed-isort-config" overwrite the value on every commit: it doesn't really make sense to actually store this.

As [pointed out](https://github.com/OCA/maintainer-quality-tools/issues/625#issuecomment-573094697) by @sbidoul, isort mentions in their changelog:

> Support for using requirements files to auto determine third-paty
> section if pipreqs & requirementslib are installed.

Which is the goal of this change, whose details are:

* Remove 'seed-isort-config' from pre-commit as the list of libs will be
  provided by requirements.txt
* Replace the isort pre-commit repo by the official repo (the mirror says
  it is been deprecated in favor of the official repo)
* Adds [pipreqs and pip-api](https://github.com/timothycrosley/isort/blob/500bafabbd51a6005c11a00c4738a2438990e48a/pyproject.toml#L42) in additional dependencies to activate isort's feature to read requirements.txt
* Adds types: [python] in the pre-commit config: the mirror had it and the official repo doesn't. Without it, some files such as .pot are modified (different EOF).
* Add an union merge driver on 'requirements.txt' to resolve conflicts
on this file (on conflicts, both lines are kept, which works in most
cases on this file)

Alternatively, we could evaluate ['reorder-python-imports'](https://github.com/asottile/reorder_python_imports) in place of 'isort'

I propose to evaluate this configuration on this repository, since this is where I ran into the issue. If the result is good (my first test shows it should), then we can apply the same commit on https://github.com/OCA/maintainer-quality-tools

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+requirements.txt merge=union
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
     name: isort except __init__.py
     'types': [python]
     exclude: /__init__\.py$
-    additional_dependencies: [pipreqs, pip-api]
+    additional_dependencies: ["pipreqs==0.4.10", "pip-api==0.0.13"]
 - repo: https://github.com/pre-commit/mirrors-eslint
   rev: v6.5.1
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,16 +51,14 @@ repos:
   rev: v1.24.0
   hooks:
   - id: pyupgrade
-- repo: https://github.com/asottile/seed-isort-config
-  rev: v1.9.3
-  hooks:
-  - id: seed-isort-config
-- repo: https://github.com/pre-commit/mirrors-isort
-  rev: v4.3.21
+- repo: https://github.com/timothycrosley/isort
+  rev: 4.3.21-2
   hooks:
   - id: isort
     name: isort except __init__.py
+    'types': [python]
     exclude: /__init__\.py$
+    additional_dependencies: [pipreqs, pip-api]
 - repo: https://github.com/pre-commit/mirrors-eslint
   rev: v6.5.1
   hooks:
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -9,4 +9,3 @@ line_length=88
 known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
-known_third_party=setuptools


.pre-commit-config.yaml
```diff
@@ -51,16 +51,14 @@ repos:
...
-  hooks:
-  - id: seed-isort-config
-- repo: https://github.com/pre-commit/mirrors-isort
-  rev: v4.3.21
+- repo: https://github.com/timothycrosley/isort
+  rev: 4.3.21-2
   hooks:
   - id: isort
     name: isort except __init__.py
+    'types': [python]
```
sbidoul, line 59: nit: I think this line is not necessary because the isort hook defines it too:
https://github.com/timothycrosley/isort/blob/500bafabbd51a6005c11a00c4738a2438990e48a/.pre-commit-hooks.yaml#L6

.gitattributes
```diff
@@ -0,0 +1 @@
+requirements.txt merge=union
```
sbidoul, line 1: I think pip is going to complain when there are duplicate requirements, but these can then be resolved manually if and when it happens.

sbidoul (approved):
Hi @guewen, thanks for the detailed writeup. This looks like a great improvement. 

> Alternatively, we could evaluate 'reorder-python-imports' in place of 'isort'

I like the separate import groups we do for odoo and odoo.addons, but if isort becomes too much of a trouble, it's certainly not a blocking point.

.pre-commit-config.yaml
guewen: I started without and had to add it otherwise I get:

```
isort except __init__.py.................................................Failed
hookid: isort

Files were modified by this hook. Additional output:

Fixing /home/gbaconnier/.../stock-logistics-warehouse/stock_orderpoint_move_link/i18n/stock_orderpoint_move_link.pot
Fixing /home/gbaconnier/.../stock-logistics-warehouse/stock_location_zone/i18n/stock_location_zone.pot
Skipped 17 files
```

And the diff is the empty line at EOF removed in these 2 files.

.gitattributes
guewen: Yes, and it will complain only if 2 lines have a different version's requirement.

.pre-commit-config.yaml
guewen: The explanation is that `master` has the types declared in the repo, but the latest release doesn't (https://github.com/timothycrosley/isort/blob/cd9c2f6e70196bd429d0179635e0fac2d034ac5e/.pre-commit-hooks.yaml).

yajo: In such case, you can add a TODO note here indicating that types should be removed after isort is upgraded to a newer version.

.pre-commit-config.yaml
```diff
@@ -51,16 +51,14 @@ repos:
...
     exclude: /__init__\.py$
+    additional_dependencies: [pipreqs, pip-api]
```
yajo: You should pin these versions.

yajo (changes requested):

.pre-commit-config.yaml
guewen: done

guewen:
Closing, to replace by https://github.com/OCA/maintainer-quality-tools/pull/639

guewen:
Replaced by https://github.com/OCA/stock-logistics-warehouse/pull/856

PR #830, opened 2020-01-14
[11.0][ADD] stock_request_employee

jarroyomorales:
This module adds a new Stock Location `Employee's Location` which allows you to control which products are in possession of your employees.
To use it, in a stock request or in a stock request order, mark the field `To Employee` and select which employee you are making the stock move to.

cc @etobella


stock_request_employee/views/hr_employee.xml
```diff
@@ -0,0 +1,23 @@
...
+<!-- Copyright 2020 Creu Blanca
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="view_employee_form">
+        <field name="name">hr.employee.form (in stock_request_employee)</field>
+        <field name="model">hr.employee</field>
+        <field name="inherit_id" ref="hr.view_employee_form"/>
+        <field name="arch" type="xml">
```
etobella, line 12: This view a group, the group that can see requests

stock_request_employee/models/stock_request_order.py
```diff
@@ -0,0 +1,67 @@
...
+    def change_childs(self):
+        res = super().change_childs()
+        if not self._context.get('no_change_childs', False):
+            for line in self.stock_request_ids:
+                line.employee_id = self.employee_id
+                line.to_employee = self.to_employee
+        return res
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
```
etobella, line 45: This is crazy, but this is the only way to update the context dict of a view. :disappointed:

etobella (approved):
LGTM :+1: Just a minor comment

stock_request_employee/views/hr_employee.xml
jarroyomorales: Done!

JordiBForgeFlow:
@AaronHForgeFlow @JoanSForgeFlow

JoanSForgeFlow (approved):
Functional review done, LGTM! :+1:

stock_request_employee/data/stock_location.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!--
+    Copyright 2020 Creu Blanca
+    License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
```
rousseldenis: Could you fix licensing inconsistencies ?

stock_request_employee/models/stock_request.py
```diff
@@ -0,0 +1,55 @@
...
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class StockRequest(models.Model):
+
+    _inherit = 'stock.request'
+
+    to_employee = fields.Boolean()
```
rousseldenis: Maybe add a little help for these two fields.

stock_request_employee/models/stock_request_order.py
```diff
@@ -0,0 +1,67 @@
...
+
+from lxml import etree
+from odoo import api, fields, models
+
+
+class StockRequestOrder(models.Model):
+
+    _inherit = 'stock.request.order'
+
+    to_employee = fields.Boolean()
```
rousseldenis: Same as above

stock_request_employee/models/stock_request_order.py
```diff
@@ -0,0 +1,67 @@
...
+        if not self._context.get('no_change_childs', False):
+            for line in self.stock_request_ids:
+                line.employee_id = self.employee_id
+                line.to_employee = self.to_employee
+        return res
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        """Add to the existing context of the field `stock_request_ids`
```
rousseldenis, line 47: To avoid all this stuff, why not creating a new action with the good default_... context ? That could make sense.

rousseldenis (changes requested):
Code review. Some remarks.

At least fix the license

stock_request_employee/models/stock_request_order.py
jarroyomorales: The problem is that we want to fix a one2many field that already has a context. We cannot update the field using the xml and your proposal would fix the `to_employee` field, but not the `employee_id`, as it can be changed.

We can make two menus, but this code cannot be totally removed. What do you prefer?

rousseldenis (approved):

PR #835, opened 2020-01-27
12.0 mig stock putaway product

kevinkhao:


pedrobaeza:
AFAIK, this module is not needed anymore as Odoo core already allows an strategy per product

kevinkhao:
> AFAIK, this module is not needed anymore as Odoo core already allows an strategy per product

Hi Pedro, yes indeed, however the module is pretty much completely changed from before. Most functionality is gone, I only added the possibility to check and manipulate putaway strategies from the product form view (instead of having to go to locations->putaway strategies)

pedrobaeza:
Then I think you should start from scratch and name it other way.

josepmy:
Hi @kevinkhao, I would like help to test it. I agree with you, I have the same issue and is better add the location from the product itself and generate the putaway rule.

Salut,
Josep M

sergio-teruel:
Please, make a rebase from v12.0 to fix travis...

kevinkhao:
https://github.com/OCA/stock-logistics-warehouse/pull/840

PR #846, opened 2020-02-11
[13.0][ADD] product_quantity_update_force_inventory

HviorForgeFlow:
### Product Quantity Update Force Inventory

This module forces users to create inventory adjustments instead of updating quantities on product card.

stock_quant_editable_view_block

It could be considered as a work around to manage Stock Accounting properly combining:

- https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_change_qty_reason
- https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_account_change_qty_reason 

CC @ForgeFlow


stock_quant_editable_view_block/hooks.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2018-20 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo import _, api
+
+from odoo.addons.stock.models.stock_quant import StockQuant
+
+
+def post_load_hook():
+    @api.model
+    def _new_get_quants_action(self, domain=None, extend=False):
```
rousseldenis: I'm not really in favor of such things as you lose all the inheritance mechanism.

Isn't there any other mean doing the same thing ?

HviorForgeFlow: Create a new action and overwrite on views. What is preferable?

rousseldenis: And in this case, why not overriding original method directly with normal inheritance ???

HviorForgeFlow: At this point, it is a matter of design, or am I missing anything?

This way it will also work! :smiley_cat: 
```
# Copyright 2018-20 ForgeFlow <http://www.forgeflow.com>

from odoo import api, models


    _inherit = 'stock.quant'

    @api.model
    def _get_quants_action(self, domain=None, extend=False):
        action = super()._get_quants_action(domain=domain, extend=extend)
        action['view_id'] = self.env.ref('stock.view_stock_quant_tree').id
        # Enables form view in readonly list
        action.update({
            'view_mode': 'tree,form',
            'views': [
                (action['view_id'], 'list'),
                (self.env.ref('stock.view_stock_quant_form').id, 'form'),
            ],
        })
        return action
```

HviorForgeFlow: I changed the override for that code again, it should perform the same, could we move it forward?

AdriaGForgeFlow (approved):
LGTM :+1: 

Could you clean the commits, I think you have some unwanted commits there

rousseldenis:
This is great! Indeed I would change module name in order to let people understand what this do.

HviorForgeFlow:
Any suggestion for module name change?

rousseldenis:
> Any suggestion for module name change?

product_quantity_update_force_inventory or something like that. More functional oriented.

LoisRForgeFlow (approved):
LGTM :+1:

AaronHForgeFlow (approved):
:+1: It works as it says

HaraldPanten (approved):
Nice one. Functional review 👍

product_quantity_update_force_inventory/README.rst
```diff
@@ -0,0 +1,76 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/stock-logistics-warehouse-13-0/stock-logistics-warehouse-13-0-product_quantity_update_force_inventory
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/153/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module block the ability to update Inventory quantities using Stock Quant Model.
```
JordiBForgeFlow: Small rewording of this section.

With this module, when a user goes to a product and presses the button 'Update Quantity' they will be directed to a new Inventory Adjustment for this product, instead of driving the user to update the quantities in the stock quants.

By doing this users will be able to have full traceability on the inventory adjustment activities.
 
Consider using this module together with the module 'stock_change_qty_reason' (https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_change_qty_reason).

HviorForgeFlow: comment attended @JordiBForgeFlow

product_quantity_update_force_inventory/models/product.py
```diff
@@ -0,0 +1,47 @@
...
+        self.ensure_one()
+        view_form_id = self.env.ref("stock.view_inventory_form").id
+        action = self.env.ref("stock.action_inventory_form").read()[0]
+        action.update(
+            {
+                "views": [(view_form_id, "form")],
+                "view_mode": "form",
+                "context": {
+                    "default_product_ids": [(6, 0, [self.id])],
+                    "default_name": _("%s: Inventory Adjustment" % self.display_name),
```
MiquelRForgeFlow:
```suggestion
                    "default_name": _("%s: Inventory Adjustment") % self.display_name,
```

MiquelRForgeFlow: the same in the other cases

HviorForgeFlow: Thanks @MiquelRForgeFlow :) Comment attended

MiquelRForgeFlow (approved):

product_quantity_update_force_inventory/models/product.py
```diff
@@ -0,0 +1,47 @@
...
```
MiquelRForgeFlow:
```suggestion
                    "default_name": _("%s: Inventory Adjustment") % self.display_name,
```

PR #847, opened 2020-02-11
[MIG][13.0] stock_inventory_preparation_filter: Migration to v13.0

sergio-teruel:
cc @Tecnativa TT21682

- **Atention!!** I have removed "Search EAN13 for removed products". I think that there are other modules that help this.

- Migration scripts have been included

Ready for review


stock_inventory_preparation_filter/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+        ("product_id", None),
+        ("category_id", None),
+        ("lot_id", None),
+    ]
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_columns(env, column_renames)
```
jjscarafia: thans for the pr @sergio-teruel 
one comment, this method should send env.cr instead of env
check definition [here](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L543)

jjscarafia (changes requested):

HaraldPanten (changes requested):
Hi,

It's mostly working fine, but I think that there's an important UX "issue". Probably it's not coming from the module but Odoo. Even though, is it possible to improve that performance?

1- Start an inventory adjustment.
2- Select the "Filtered Products" option.
3- Try to change the field domain rule.

**You'll see that the drop-down menu is not properly shown, look:**

Captura de pantalla 2020-07-10 a las 10 15 39

As a solution, you can use the keyboard arrow keys for sure. But, could any "front-end magic" be done in this use case? Should we consider this as a known issue or similar? 🤔

@sergio-teruel @jjscarafia @rousseldenis

sergio-teruel:
@jjscarafia Changes done!!
@HaraldPanten Added to roadmap

stock_inventory_preparation_filter/README.rst
```diff
@@ -0,0 +1,91 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Known issues / Roadmap
+======================
+
+* The widget domain is not displayed correctly, but this issue is related to
+  Odoo.
```
HaraldPanten:
```suggestion
* The widget domain is not displayed correctly, but this issue is related to
  Odoo. To avoid this malfunction, use the keyboard arrows to properly work
  with the domain option.
```

What do you think?

HaraldPanten (approved):
Overall LGTM 👍

jjscarafia:
@sergio-teruel the actual migration scripts broke the update of the module if the database is migrated by odoo upgrade. Because [these columns ](https://github.com/OCA/stock-logistics-warehouse/pull/847/files?file-filters%5B%5D=.py#diff-66871c8de6411734ca4a9b3a300f7bc3R6-R13) don't exist anymore on v13 and the are already deleted.

Perhaps the migration script could check if columns exists? Any other idea?

sergio-teruel:
@HaraldPanten Changes done!!! Thanks...
@jjscarafia Can you check now?? I must pass table name instead of model name... Check please!!

jjscarafia:
hi @sergio-teruel that fix maybe ok. But the problem I share is different. Let said you upgrade database with odoo upgrade service (or openupgrade without this module on the addons path).
1. when updating odoo stock module, the fields `"location_id", "product_id", "category_id", "lot_id" ` are deleted from stock_inventory table because they are not anymore there, they are now on inventory lines
2. then, the `openupgrade.rename_columns(env.cr, column_renames)` raise and error an all the update of the database is stopped. 

I believe that pre-migration script should check if those fields are there (perhaps also on post migration..)

sergio-teruel:
@jjscarafia Sorry... I didn't understand you.-..
Mybe @pedrobaeza can tell us more info about it...

pedrobaeza:
@jjscarafia sorry, but the migration scripts are correct. If you upgrade your DB without the presence of the module, the module is still pending to upgrade, and when the module will be there, then scripts will be executed. Meanwhile, no column deletion should be performed, as any upgrade service can deduce that the columns are no longer there due to "missing modules". This is how OpenUpgrade works, and if Odoo upgrade service removes them, then the problem is in that side, but what can't be done is to try to prevent any incorrect behavior coming from other parts. And anyway, even putting the safeguard for checking the column existence, then the module won't work correctly as data won't be correctly transferred.

pedrobaeza (approved):

jjscarafia:
@pedrobaeza If I'm not wrong, those fields are from [odoo stock module](https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L58), so just upgrading stock module delete those columns. I believe this error could be replicated using openupgrade if you first upgrade without this module and later with it. 

- [Fields definition on v12 (they are on stock module on stock_inventory)](https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L58)
- [on v13 they are part of stock_inventory_line](https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L283)

pedrobaeza:
No, on OpenUpgrade, there's no such problem, as columns are not dropped ever. If Odoo upgrade service does this, then there's a problem in general, as you can lose data that are not able to recover later. I doubt such operation is done there, and if so, I don't want to use it.

jjscarafia:
ok @pedrobaeza . forgt it. I don't agree that a module update should only work using openupgrade, I believe it's to much opinionated. I also don't get why this module is migrating data that is not created by this module, for sure I'm missing something.

pedrobaeza:
@sergio-teruel doesn't that columns belong to this module? If not, please add a new PR copying column or using them directly.

@jjscarafia I'm not restricting the migration scripts to OpenUpgrade. It's just that the solution you are saying is not migrating the module at all AFAIK.

ivantodorovich:
Actually I'm encountering the same issue that @jjscarafia described.

IMO this is not correct. 

This script (or at least most of it) belongs to OpenUpgrade. As it's actually trying to migrate the core `stock` module.

- `location_id` was changed to `location_ids` in core

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L53
https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L45

- `product_id` was changed to `product_ids` in core

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L58
https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L50

- `lot_id` and `category_id` where removed by odoo, AFAIK. it's nice that this module restores them, but the migration script should fail silently at least, or even not exist here (could be covered by openupgrade, in case this module is installed.. )


It's not within the scope of this module to migrate these fields.
They should be moved to OpenUpgrade

ivantodorovich:
In fact, all of this is already identified in the openupgrade analysis:
https://github.com/OCA/OpenUpgrade/blob/13.0/addons/stock/migrations/13.0.1.1/openupgrade_analysis.txt#L24-L25

PR #851, opened 2020-02-24
[12.0] [WIP] [IMP] stock_quant_manual_assign set quantity done optional by picking…

angelmoya:
On module stock_quant_manual_assign, whe you set manually a quant on stock move, this quant is reserved and done.

We need to manually reserve, but do not update quantity done.

To be consistent with previous development we decided to do it configurable, depending on picking type, this way previous installation will work same way, but now we can configure by picking type if we want reserve or update quantity done.

### DIFF
--- a/stock_quant_manual_assign/README.rst
+++ b/stock_quant_manual_assign/README.rst
@@ -23,7 +23,7 @@ Stock - Manual Quant Assignment
     :target: https://runbot.odoo-community.org/runbot/153/12.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module allows you to manually change the automatic quant selection.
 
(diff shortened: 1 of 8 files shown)


pedrobaeza:
I don't get why don't you want to fill done quantity.

stock_quant_manual_assign/README.rst
```diff
@@ -32,6 +32,12 @@ This module allows you to manually change the automatic quant selection.
 .. contents::
    :local:
 
+
+Configuration
+=============
+
+On /Inventory/Configuration/Operation Types you can configure to set quantity done or not on manual quants assign.
```
pedrobaeza: You mustn't edit this file, but add the fragment `CONFIGURE.rst` on `readme` folder, and generate the new README or let the bot to generate it on merge.

angelmoya: FIXED

angelmoya:
> I don't get why don't you want to fill done quantity.

We need to reserve specific lot, but we don't want to operation now.

Any way we also are thinking about add a boolean in wizard, instead of configuration.

I thought to write an Issue, but as the code was easiest I did this way to get feedback.

pedrobaeza:
Writing on `qty_done` doesn't mean that you do that operation. You need to click on "Validate" for that, so I still don't understand it.

PR #853, opened 2020-02-26
[ADD][12.0] stock_generate_putaway_from_inventory: Add to 12.0

kevinkhao:


stock_generate_putaway_from_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,89 @@
...
+                msg += "%s ," % el.name
+            msg = msg[:-2]
+        return msg
+
+    def action_generate_putaway_strategy(self):
+        missing_putaway = self.env["stock.location"]
+        not_validated = self.env["stock.inventory"]
+        for inventory in self:
+            try:
+                inventory._generate_putaway_strategy()
```
sebastienbeau: I think we should keep this as simple a possible.
In the huge majority of case the user will generate putaway only for one inventory (most of customer create one inventory per year).
I think the code will be a lot simplify if we simplify raise directly the error.
So just do 
```
for inventory in self:
       inventory._generate_putaway_strategy()
```
And directly raise error in the method "_generate_putaway_strategy" then you can drop the handle_errors and so drop 16~20 line of code

sebastienbeau (changes requested):

stock_generate_putaway_from_inventory/models/stock_inventory.py
kevinkhao: fixed at https://github.com/akretion/stock-logistics-warehouse/commit/f3159eba294a08566e4c752fe41cb6c6c7b1df9b

rousseldenis (approved):
Code review. LGTM

sebastienbeau (approved):

sebastienbeau:
squash commit done let's do the last step /ocabot merge

PR #855, opened 2020-03-02
[13.0] Create module stock_location_children

grindtildeath:
This module adds a `children_ids` field on `stock.location` in order to compute
and store all the children for a `stock.location` and not only its first level
children as is the case for `child_ids`

### DIFF
--- a/stock_location_children/models/stock_location.py
+++ b/stock_location_children/models/stock_location.py
@@ -19,9 +19,26 @@ class StockLocation(models.Model):
 
     @api.depends('child_ids', 'child_ids.children_ids')
     def _compute_children_ids(self):
+        query = """SELECT sub.id, ARRAY_AGG(sl2.id) AS children
+            FROM stock_location sl2,
+            (
+            SELECT id, parent_path
+            FROM stock_location sl
+            ) sub
+            WHERE sl2.parent_path LIKE sub.parent_path || '%%'
+            AND sl2.id != sub.id
+            AND sub.id IN %s
+            GROUP BY sub.id;
+        """
+        self.env.cr.execute(query, (tuple(self.ids),))
+        rows = self.env.cr.dictfetchall()
         for loc in self:
-            if not loc.child_ids.mapped('child_ids'):
-                all_children = loc.child_ids
+            all_ids = []
+            for row in rows:
+                if row.get('id') == loc.id:
+                    all_ids = row.get('children')
+                    break
+            if all_ids:
+                loc.children_ids = [(6, 0, all_ids)]
             else:
-                all_children = loc.child_ids | loc.child_ids.children_ids
-            loc.children_ids = all_children
+                loc.children_ids = [(5, 0, 0)]
--- a/stock_location_children/models/stock_location.py
+++ b/stock_location_children/models/stock_location.py
@@ -5,19 +5,19 @@ from odoo import api, fields, models
 
 class StockLocation(models.Model):
 
-    _inherit = 'stock.location'
+    _inherit = "stock.location"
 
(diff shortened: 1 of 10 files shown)


stock_location_children/models/stock_location.py
```diff
@@ -0,0 +1,27 @@
...
+        'parent_id',
+        'children_id',
+        compute='_compute_children_ids',
+        store=True,
+        help='All the children (multi-level) stock location of this location',
+    )
+
+    @api.depends('child_ids', 'child_ids.children_ids')
+    def _compute_children_ids(self):
+        for loc in self:
```
simahawk, line 36: can't we use `parent_path` to make this more performant?
Can you check its usage in stock module?

stock_location_children/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+        "stock.location",
+        "stock_location_children_ids",
+        "parent_id",
+        "children_id",
+        compute="_compute_children_ids",
+        store=True,
+        help="All the children (multi-level) stock location of this location",
+    )
+
+    @api.depends("child_ids", "child_ids.children_ids")
```
simahawk: shouldn't be `child_ids.child_ids`? :thinking:

simahawk: or better `parent_path`

gurneyalex: Test code:

```
>>> Loc = self.env['stock.location']
>>> stock = Loc.browse(8)
>>> t1 = Loc.create({'name': 'test1', 'location_id': stock.id})
>>> t2 = Loc.create({'name': 'test2', 'location_id': t1.id})
>>> t3 = Loc.create({'name': 'test2', 'location_id': t2.id})
>>> self.env.cr.commit()
```
With api.depends('child_ids', 'child_ids.child_ids'), the computation method is called once with self = stock.location(7, 8, 26, 27, 28)

With api.depends('child_ids', 'child_ids.children_ids', the computation method is called 6 times with self containing a single location each time. 

@grindtildeath  My advice: prefer child_ids.child_ids if it does not break any known test case as it seems more efficient. 

@simahawk I'm not sure how to depend on parent_path. Can you suggest something?

simahawk: @gurneyalex thanks for testing, I suspected this behavior.
Regarding `parent_path`: never mind. I did a quick check and I don't think is going to work because it's handled in "magic" methods.

grindtildeath:
```suggestion
    @api.depends("child_ids", "child_ids.child_ids")
```

stock_location_children/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+            (
+            SELECT id, parent_path
+            FROM stock_location sl
+            ) sub
+            WHERE sl2.parent_path LIKE sub.parent_path || '%%'
+            AND sl2.id != sub.id
+            AND sub.id IN %s
+            GROUP BY sub.id;
+        """
+        self.env.cr.execute(query, (tuple(self.ids),))
```
grindtildeath, line 34:
```suggestion
        self.flush()
        self.env.cr.execute(query, (tuple(self.ids),))
```

guewen: It should not flush all the fields because it can flush too much (following relations and flushing other models, ...).
`flush("location_id", "child_ids")`?

guewen: Pushed a new commit

guewen (approved):

simahawk (approved):

PR #861, opened 2020-03-11
[13.0] Add module stock_move_common_dest

grindtildeath:
This module adds a M2m field `common_dest_move_ids` on `stock.move` in order
to compute all the moves having a chained destination move sharing the same
picking as the actual move's destination move.

PR #862, opened 2020-03-11
[13.0] Add module stock_move_common_dest_sync_location - alpha

grindtildeath:
Depends on: 
 - [x] #861


stock_move_common_dest_sync_location/models/stock_move_line.py
```diff
@@ -0,0 +1,30 @@
...
+        related='picking_id.picking_type_id.sync_common_move_dest_location',
+        readonly=True
+    )
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get('location_dest_id') and not self.env.context.get('_sync_common_dest_location'):
+            related_moves = self.env['stock.move']
+            for move_line in self:
+                if not self.picking_type_sync_common_move_dest_location:
```
guewen:
```suggestion
                if not move_line.picking_type_sync_common_move_dest_location:
```

jgrandguillaume (approved):
LGTM, though it'll need more extensive testing to see how it behave. I would be in favor of merging this as alpha because it works as expected. We just don't know now if the feature will be relevant in all context.

guewen (changes requested):
We shouldn't merge it. I'll soon propose a new solution to solve the issue.
I'll close this PR when I open the new one.

guewen:
The replacement for this feature is #907

PR #874, opened 2020-03-23
[IMP] stock_request_direction (Warehouse On Change)

patrickrwilson:
Since the location and routes depend on the warehouse that is selected on the stock request order, if the warehouse changes then so will the route options. This adds an on change event so if the warehouse changes, then the direction gets cleared and so do the routes on the product lines. This forces the user to properly reset the items using the correct routes for the newly selected warehouse.

### DIFF
--- a/stock_request_direction/models/stock_request_order.py
+++ b/stock_request_direction/models/stock_request_order.py
@@ -24,6 +24,14 @@ class StockRequestOrder(models.Model):
             self.location_id = \
                 self.warehouse_id.lot_stock_id.id
 
+    @api.onchange('warehouse_id')
+    def _onchange_warehouse_id(self):
+        if self.direction:
+            self.direction = False
+        for stock_request in self.stock_request_ids:
+            if stock_request.route_id:
+                stock_request.route_id = False
+
     def change_childs(self):
         super().change_childs()
         if not self._context.get('no_change_childs', False):
--- a/stock_request_direction/models/stock_request_order.py
+++ b/stock_request_direction/models/stock_request_order.py
@@ -27,6 +27,11 @@ class StockRequestOrder(models.Model):
             if stock_request.route_id:
                 stock_request.route_id = False
 
+    @api.onchange('warehouse_id')
+    def onchange_warehouse_id(self):
+        # Onchange no longer needed
+        pass
+
     def change_childs(self):
         super().change_childs()
         if not self._context.get('no_change_childs', False):
--- a/stock_request_direction/models/stock_request_order.py
+++ b/stock_request_direction/models/stock_request_order.py
@@ -13,7 +13,7 @@ class StockRequestOrder(models.Model):
                                  states={'draft': [('readonly', False)]},
                                  readonly=True)
 
-    @api.onchange('direction')
+    @api.onchange('warehouse_id', 'direction')
     def _onchange_location_id(self):
         if self.direction == 'outbound':
             # Stock Location set to Partner Locations/Customers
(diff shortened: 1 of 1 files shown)


patrickrwilson:
Ping @max3903 @osi-scampbell

stock_request_direction/models/stock_request_order.py
```diff
@@ -24,6 +24,14 @@ def _onchange_location_id(self):
             self.location_id = \
                 self.warehouse_id.lot_stock_id.id
 
+    @api.onchange('warehouse_id')
+    def _onchange_warehouse_id(self):
+        if self.direction:
+            self.direction = False
```
max3903: @patrickrwilson If you change the warehouse, the direction should remain the same, no?

patrickrwilson: @max3903 When the direction stays set, it doesn't seem to update the location and route information available on the line items so by clearing that and the routes, then forcing the user to set the direction again it then resets the location properly and also seems to 'reset' what shows for the line items. This is why I cleared the direction.

max3903: @patrickrwilson what about calling `self._onchange_location_id()`?

patrickrwilson: @max3903 having it call the other on-change doesn't seem to work for me. I have 2 WH, 'Main WH' and 'WH2'. If I have it call the other on-change method instead of making the direction blank, the location changes to the stock location of the warehouse like if it were an inbound. 

Example, I set 'Main WH', direction 'Outbound' and it sets correctly for an outgoing order. Then change the warehouse to 'WH2', now the location changes to 'WH2/Stock' and not the correct one going outbound. If i change the direction to inbound then outbound it resets location correctly. 

This is what i changed the code to:
    @api.onchange('warehouse_id')
    def _onchange_warehouse_id(self):
        if self.direction:
            self._onchange_location_id()
        for stock_request in self.stock_request_ids:
            if stock_request.route_id:
                stock_request.route_id = False

Maybe i'm missing something or my code is wrong. Thanks for your help.

patrickrwilson:
@max3903 the last commit includes joining the new onchange with the original. I'm not sure why this happens but if you create a stock request, setting the direction to outgoing, it properly sets the location to customer...
image

But then change the warehouse, the onchange seems to not see that direction is already set to outgoing, therefore putting the stock location in...
image

Any idea why? This is part of the original problem, if you change the direction to inbound, then back to outbound it resets the locations correctly to customer.

max3903 (approved):

osi-scampbell:
@patrickrwilson So we might have a conflict of onchange methods at work. In stock_request_direction we have an onchange that checks warehouse and direction and sets the Location_id to companies customer stock location which sounds like the desired outcome. 
image

However in the base stock request module, if you change the warehouse it will set the Location_id to the Stock Location of the Warehouse itself which is what appears to happening to you right now when you change the warehouse.
image

I believe a possible solution would be to inherit that second method from your module, check if the direction is set, if it is, then don't run the method, if it is not, proceed with super().

patrickrwilson:
@osi-scampbell that was a good thought but i'm not able to fully confirm. Just to test I removed the onchange method you are referring to in stock_request. With that removed, when I change the warehouse, it doesn't change the location at all. Its almost like it's not triggering at all when the warehouse changes. Changing the direction makes the location change correctly so it does trigger on that field. I feel like i'm missing something simple.

patrickrwilson:
@osi-scampbell thanks for your help on this. The latest commit has the fix we worked on.

osi-scampbell (approved):

PR #770, opened 2019-11-26
[11.0][ADD] Scrap reason code module

bodedra:
Usage
--------

In the Inventory module, open the Configuration menu and select Scrap Reason Codes.
Create the required scrap reason codes. Under Operations, select Scrap. Click the 
create button to create a new scrap order. You will see a reason code field on the
scrap form which will allow you to select any of the scrap codes you created previously.

PR #875, opened 2020-03-27
[13.0][CHG] model name from reason.code to scrap.reason.code

bodedra:
Ref: https://github.com/OCA/stock-logistics-warehouse/pull/770#discussion_r370558559

cc @rousseldenis

### DIFF
--- a/scrap_reason_code/__manifest__.py
+++ b/scrap_reason_code/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     "name": "Scrap Reason Code",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
     "license": "AGPL-3",
     "summary": "Reason code for scrapping",
     "author": "Open Source Integrators, Odoo Community Association (OCA)",
--- a/scrap_reason_code/migrations/13.0.1.1.0/pre-migration.py
+++ b/scrap_reason_code/migrations/13.0.1.1.0/pre-migration.py
@@ -1,0 +1,14 @@
+# Copyright (C) 2019 IBM Corp.
+# Copyright (C) 2019 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+_model_renames = [
+    ("reason.code", "scrap.reason.code"),
+]
+
+
+@openupgrade.migrate()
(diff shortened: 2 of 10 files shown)


bodedra:
@dreispt @max3903 @Chandresh-SerpentCS are you guys available for a review?

AaronHForgeFlow:
The problem here is current instances with this module installed will need a migration script.

bodedra:
@AaronHForgeFlow Would you please an example?

AaronHForgeFlow:
Sorry I did not explain myself correctly. I actually wanted to say that I am not in favor of such changes in a stable version. Because there is not much benefit. Perhaps the module rename is better to be done in version 14 if the module is migrated. 

Nevertheless, I am not a user of this module so I will not block this PR. You can go ahead with this PR. Just for your information, please notice that you may need migration scripts if you have customers currently using this module, just that.

Regards.

cc @rousseldenis

bodedra:
@AaronHForgeFlow Appreciate your feedback. @rousseldenis What's your call?

rousseldenis:
Indeed, I don't remember that 13.0 version was merged as is by me... IMHO, those modules should be the same. So, could you do migration scripts ?
There is a rename_models method here : https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L149

bodedra:
@rousseldenis Thanks for sharing method. Comment addressed! How about now?

scrap_reason_code/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2019 IBM Corp.
```
rousseldenis: To make this working for existing installs, please change to '13.0.1.1.0' and change it in manifest too.

bodedra:
@rousseldenis @AaronHForgeFlow It seems this is ready for your review. :smile:

rousseldenis (approved):

AaronHForgeFlow (approved):
:+1: Just code review

bodedra:
Ping @dreispt

rousseldenis:
@bodedra Do you keep all those commits ?

bodedra:
@rousseldenis Let's do same thing with this one.

dreispt (approved):

PR #881, opened 2020-04-08
[11.0][MIG]stock_inventory_revaluation

AaronHForgeFlow:
Standard migration


matt454357:
I don't understand how this PR passes the tests.  When I run the tests locally, they all fail.  Doesn't the runbot perform tests?

For example:
- Setting product.template.property_cost_method = 'real' fails... in v11, it should be 'fifo'
- Writing to stock.quant.cost field will fail, because that field is removed in v11

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,481 @@
...
+    def button_draft(self):
+        self.write({'state': 'draft'})
+        return True
+
+    @api.multi
+    def button_cancel(self):
+        for revaluation in self:
+            for reval_quant in revaluation.reval_quant_ids:
+                reval_quant.quant_id.sudo().write(
+                    {'cost': reval_quant.old_cost})
```
CasVissers-360ERP: This only works for realtime inventory valuation? If inventory valuation is average shouldn't the cost price be reset?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,481 @@
...
+                for reval_quant in revaluation.reval_quant_ids:
+                    reval_quant.old_cost = reval_quant.quant_id.cost
+                    reval_quant._write_new_cost()
+            else:
+                if revaluation.product_id.\
+                        cost_method in ['standard', 'average']:
+
+                    if revaluation.revaluation_type == 'inventory_value':
+                        if float_compare(revaluation.new_value, 0.0,
+                                         precision_rounding=account_prec) < 0:
```
CasVissers-360ERP: new_value can never be negative?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,481 @@
...
+                                 digits=dp.get_precision('Account'),
+                                 compute="_compute_calc_product_value",
+                                 readonly=True)
+
+    old_value = fields.Float('Old value',
+                             help='Displays the current value of the product.',
+                             digits=dp.get_precision('Account'),
+                             readonly=True)
+
+    new_value = fields.Float('Credit/Debit amount',
```
CasVissers-360ERP: Onchange revaluation_type this should be reset to 0 to avoid incorrect warnings

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,481 @@
...
+                            readonly=True)
+
+    current_cost = fields.Float('Current cost',
+                                help='Displays the current cost of the '
+                                     'product.',
+                                digits=dp.get_precision('Product Price'),
+                                compute="_compute_calc_current_cost",
+                                readonly=True)
+
+    new_cost = fields.Float('New cost',
```
CasVissers-360ERP: Onchange revaluation_type this should be reset to 0 to avoid incorrect warnings

stock_inventory_revaluation/models/product.py
```diff
@@ -0,0 +1,55 @@
...
+                or False
+
+            reval = reval_model.create({
+                'revaluation_type': 'price_change',
+                'product_id': product.id,
+                'new_cost': new_price,
+                'increase_account_id': increase_account_id,
+                'decrease_account_id': decrease_account_id
+            })
+            reval.button_post()
```
CasVissers-360ERP, line 56: Valuation can't be posted if increase_account_id  or decrease_account_id is not set. The warning raised from the valuation is a bit confusing. Maybe improve the warning?

CasVissers-360ERP (changes requested):
Tested for an AVCO structure.

matt454357:
This module needs major refactoring, even for standard and average cost.  For example, look at this comment in the stock_account module:

https://github.com/odoo/odoo/blob/a7c90a6d080a77316924cec9d3c64f9662ac7436/addons/stock_account/models/stock.py#L332-L333

Look at this PR (#894) for 12.0.  The module should be working with `stock.move` for all costing methods, including standard and average.

AaronHForgeFlow:
Sorry not working on this anymore. just feel free to supersede in case you need this module. IMHO the version proposed by  @matt454357  is the one needed in v11 too.

PR #884, opened 2020-04-16
[13.0][IMP] stock_cubiscan: Handle unit without packaging

grindtildeath:
Depends on :
 - [x] https://github.com/OCA/product-attribute/pull/597

### DIFF
--- a/stock_cubiscan/wizard/cubiscan_wizard.py
+++ b/stock_cubiscan/wizard/cubiscan_wizard.py
@@ -1,7 +1,7 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, fields, models
+from odoo import _, api, exceptions, fields, models
 
 
 class CubiscanWizard(models.TransientModel):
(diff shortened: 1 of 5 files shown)


stock_cubiscan/wizard/cubiscan_wizard.py
```diff
@@ -23,46 +23,90 @@ class CubiscanWizard(models.TransientModel):
...
+                    ),
+                }
+            )
+        return [vals]
+
+    def _prepare_packaging_lines(self):
+        vals_list = []
+        packaging_types = self.env["product.packaging.type"].search([])
+        for seq, pack_type in enumerate(packaging_types):
+            pack = self.env["product.packaging"].search(
```
simahawk, line 75: if this wizard is used very often, I'd drop the search and add a read_group by product filtering on all pkg types before the loop.
Then in the loop you can lookup by product id.

stock_cubiscan/wizard/cubiscan_wizard.py
```diff
@@ -23,46 +23,90 @@ class CubiscanWizard(models.TransientModel):
...
+                        self.product_id.product_width, cubiscan_mm_uom
+                    ),
+                    "height": product_dimension_uom._compute_quantity(
+                        self.product_id.product_height, cubiscan_mm_uom
+                    ),
+                }
+            )
+        return [vals]
+
+    def _prepare_packaging_lines(self):
```
simahawk, line 71: docstring?

stock_cubiscan/wizard/cubiscan_wizard.py
grindtildeath: This part wasn't changed in this PR but only moved to a dedicated function. Sorry but I don't want to mess with its behaviour now...

guewen:
Rebased and applied pre-commit

guewen (approved):

simahawk (approved):

PR #887, opened 2020-04-17
[12.0][ADD] sale_stock_info_popup: new module

ernestotejeda:
Backport of qty_at_date_widget widget of Odoo v13

Cc @Tecnativa TT21547

sale_stock_info_popup

PR #889, opened 2020-04-22
[ADD] sale_stock_available_info_popup: new module

ernestotejeda:
Adds  'Available to promise' quantity to qty_at_date_widget widget added in https://github.com/OCA/stock-logistics-warehouse/pull/887

Depends on:
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/887

Cc @Tecnativa TT21547


sale_stock_available_info_popup/static/src/xml/qty_at_date.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<templates>
+    <div t-extend="sale_stock.QtyDetailPopOver">
+        <t t-jquery="t[t-if*='!data.is_mto']" t-operation="append">
+            <tr>
+                <td><strong>Available to promise</strong></td>
+                <td><t t-esc='data.immediately_usable_qty_today'/>
+                <t t-esc='data.product_uom.data.display_name'/></td>
```
Tardo, line 15: I can't test it... but is this correct?

manuelcalerosolis:
There is no translations file in this module?

pedrobaeza:
Please rebase

sale_stock_available_info_popup/static/src/xml/qty_at_date.xml
ernestotejeda: Yes @Tardo It is in this way, because of the data structure in 'data'. Thanks.

ernestotejeda:
@pedrobaeza Done

rousseldenis (approved):
Code + functional review. Works as expected

pedrobaeza:
One question: the color of the info icon in which field is based? I think we should adjust the colors according this new field, or offer a configuration option.

rousseldenis:
> One question: the color of the info icon in which field is based? I think we should adjust the colors according this new field, or offer a configuration option.

It depends on multiple fields and is based on classes : https://github.com/OCA/stock-logistics-warehouse/blob/12.0/sale_stock_info_popup/static/src/xml/qty_at_date.xml#L8

pedrobaeza:
Yeah, and what do you think about changing its meaning when installing this module?

rousseldenis:
> Yeah, and what do you think about changing its meaning when installing this module?

What do you mean ?

pedrobaeza:
I mean that colors of the icon are set according "available to promise" field.

ernestotejeda:
I agree with @pedrobaeza that the color depends on the quantity to promise instead of the forecasted quantity. Take a look at how I've done and tell me what do you think.

chienandalu (commented):
Testing in runbot the popup doesn't pop

pedrobaeza:
@chienandalu please re-review

chienandalu (approved):
Now it works right. Is it correct to not show stock qtys on consumables (I guess it is on purpose)

pedrobaeza:
Yes, indeed consumables don't have stock control. Let's go then:

/ocabot merge nobump

PR #894, opened 2020-04-28
[12.0][MIG] stock_inventory_revaluation

matt454357:
Migration from 10.0 to 12.0


sale_stock_info_popup/README.rst
```diff
@@ -0,0 +1,96 @@
+=====================
+Sale Stock Info Popup
```
CasVissers-360ERP: Don't know why this module is included here?

stock_inventory_revaluation/README.rst
```diff
@@ -0,0 +1,170 @@
...
+
+  - Version 11 moved the cost from stock.quant to stock.move.  That
+    means there is no longer an individual cost assigned to specific
+    lots/serials.
+  - As a work around, you can internally transfer the lot/serial you want to
+    revalue.  This will separate it from the other lots/serials on the
+    original inbound stock move.
+
+- Known Issue: It is not possible to cancel inventory revaluations for products
+  set with average or standard price.
```
CasVissers-360ERP: Why is this not possible? Seems like the use case is supported?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,574 @@
...
+        if not all(x == self.product_id.id for x in move_products):
+            raise UserError(_("The selected moves are for a different "
+                              "product than the revaluation"))
+
+    @api.model
+    def _set_validate_accounting_parameters(self):
+        """Set and validate accounting fields, in preparation to post"""
+        if not self.decrease_account_id or not self.increase_account_id:
+            raise UserError(
+                _("Please add an Increase Account and a Decrease Account"))
```
CasVissers-360ERP: Can we improve this message maybe showing the product category?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,574 @@
...
+        help='Total current value of product in stock')
+
+    old_value = fields.Float(
+        string='Old value',
+        help='Total current value of the product, prior to revaluation',
+        digits=dp.get_precision('Account'),
+        readonly=True)
+
+    new_value = fields.Float(
+        string='Credit/Debit amount',
```
CasVissers-360ERP: Credit/debit amount feels as in incorrect string? Maybe new value is better?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,574 @@
...
+        help="Enter the amount you wish to credit or debit from the current "
+             "inventory value of the item. Enter credit as a negative value. "
+             "Relevant only if the selected revaluation type is Inventory "
+             "Credit/Debit.",
```
CasVissers-360ERP: Help message is incorrect since a negative amount is not allowed.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
```diff
@@ -0,0 +1,574 @@
...
+            reval._validate_product_parameters()
+
+            if reval.product_id.cost_method in ['standard', 'average']:
+                reval.old_cost = reval.current_cost
+                reval.old_value = reval.current_value
+                if reval.revaluation_type == 'price_change':
+                    reval.new_value = reval.qty_available * reval.new_cost
+                    reval.product_id.with_context(
+                        force_company=reval.company_id.id
+                    ).sudo().write({'standard_price': reval.new_cost})
```
CasVissers-360ERP, line 334: I have some rounding issue, example:
Input: Product AVCO / Stock qty = 147 / Current cost = 3,40 / Current value = 499,80
Parameters: New value = 750
Output: Current value = 749,70

CasVissers-360ERP (changes requested):
Initial functional review for AVCO costing, will do FiFo later.

CasVissers-360ERP:
I think a nice addition would be a new revaluation type "Based on Template", in which the stock user can choose for example repair or modification as template and stock accounts are automatically loaded. This allows a user without accounting knowledge to do revaluations.

sale_stock_info_popup/README.rst
matt454357: Well, I'm a newbie here, but it's not mine, it's just part of the repo.  @pedrobaeza merged it to the repo on April 24th.  Can you explain why you ask about that particular module?  It seems to have something to do with this PR failing the flake8 tests.

pedrobaeza: @matt454357 the problem is that you have screwed up git thing in your branch (probably doing a rebase over an outdated or incorrect branch). Please rebase over a proper 12.0 branch and all that stuff will disappear.

matt454357:
@CasVissers,
Adding a template model is a great idea, not only for users without accounting knowledge, but also for automating revaluations from other modules.

Making it a revaluation type doesn't make sense to me. The revaluation types should be mutually exclusive. After choosing a template, we still need to know whether this is a Unit Price Change, or a Total Value Change.  I think I would make the revaluation type another field on the template.

stock_inventory_revaluation/views/stock_inventory_revaluation_view.xml
```diff
@@ -113,7 +113,7 @@
                     <field name="name"/>
                     <field name="revaluation_type"/>
                     <field name="document_date"/>
-                    <field name="user_id"/>
+                    <field name="create_uid"/>
```
rousseldenis, line 117: @JordiBForgeFlow Do you have an advice on this ? Why did you use a specific field for this ?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: The problem is in a core module: stock_account.  If accounting precision is 2 digits, then the standard_price field on the product gets rounded from 5.102041 to 5.10.  So, for standard and average cost, it doesn't matter whether the stock.move.remaining_value is 749.70 or 750.00.
product.stock_value = product.standard_price * product.qty_available
In this case, 5.1 * 147 = 749.70

matt454357: I have the Product Price decimal precision set much higher than the Account precision. For example, set Product Price to 6 digits, and leave Account at 2 digits. Then, current_cost will be 750.000027, and the journal entry will be for 750.00 - 499.80 = 250.20

sale_stock_info_popup/README.rst
matt454357: Sorry @pedrobaeza, could you help me out? I don't know where I went wrong.

I followed the beginner instructions here:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#technical-method-to-migrate-a-module-from-110-to-120-branch

Then, I made my pull request.

When I saw there were new commits to the 12.0 branch, I did this:
$ git checkout 12.0
$ git pull
$ git checkout 12.0-mig-stock_inventory_revaluation
$ git merge 12.0

To fast forward, and get rid of the merge commit, I did this:
$ git rebase

To put my commit last in the log, I changed the commit order with this:
$ git rebase -i eff9c69

To update this PR, I did this:
$ git push --force $USER_ORG 12.0-mig-$MODULE --set-upstream

stock_inventory_revaluation/README.rst
matt454357: I agree.  I will remove the note.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: Yes, I will do that.

sale_stock_info_popup/README.rst
pedrobaeza: You only have to do
```
git fetch origin
git rebase -i origin/12.0
git push -f
```
being origin OCA repository. The git merge operation screwed up this. Try this over your branch now.

matt454357: I followed these instructions, and I believe the problem is resolved

stock_inventory_revaluation/README.rst
matt454357: Done.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: Done.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: Done.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: Done.

stock_inventory_revaluation/models/stock_inventory_revaluation.py
matt454357: Leaving it the way it is, unless you have a solution in mind.

matt454357:
Still getting flake8 FAIL.  It passes flake8 tests locally.  Other PRs are passing the flake8 test.  Can someone help me understand why mine is failing?

stock_inventory_revaluation/models/stock_inventory_revaluation.py
CasVissers-360ERP: I think new_value should also be read-only if state = posted? Maybe better to define in the view.

CasVissers-360ERP:
@matt454357 Nice work on the template functionality!  

Added one minor remark about new_value which should be readonly when posted. 

For the decimal rounding in the value change I'm not sure yet. I understand the issue, but the current workflow implies that the value will be updated to the entered value. It will be confusing for end-users. 

The only solution I have for now is to add a price difference or rounding difference account which is used to book the difference. This gives the user more flexibility and also explains and warns for possible rounding issues. What do you think?

matt454357:
@CasVissers thanks for all the feedback.

I made new_value readonly, as you suggested.

For the decimal rounding, I noted the behavior in some field help attributes.  I also added some info to the list of known issues.  The issue seems to have been present in previous versions, so I don't feel bad about leaving it.

stock_inventory_revaluation/views/stock_inventory_revaluation_view.xml
matt454357: @rousseldenis, @JordiBForgeFlow, Can we go ahead, and merge this PR?  This is the only unresolved comment.

stock_inventory_revaluation/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+{
+    "name": "Stock Inventory Revaluation",
+    "summary": "Introduces inventory revaluation as single point to change "
+               "the valuation of products.",
+    "version": "12.0.1.0.0",
+    "author": "Eficent Business and IT Consulting Services S.L., "
+              "Serpent Consulting Services Pvt. Ltd.,"
+              "Matt Taylor,"
+              "Odoo Community Association (OCA)",
+    "website": "http://www.eficent.com",
```
rousseldenis: Please change this according to https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10

rousseldenis (approved):
Minor remark. Apart of that, seems great

stock_inventory_revaluation/__manifest__.py
matt454357: Done.  Thank you.

CasVissers-360ERP (approved):
LGTM. Even-though I would have liked a better flow for rounding differences when updating the stock value. I think this is still confusing for end-users.

matt454357:
@rousseldenis, @CasVissers, @pedrobaeza,
Is there something else I need to do, or can this be merged?

pedrobaeza:
I let @rousseldenis to merge as I'm not a qualified reviewer for this module and he has merge rights.

HaraldPanten (approved):
Seems to be OK

CasVissers-360ERP:
@rousseldenis 
Think this one can be merged?

marcelsavegnago (approved):

PR #899, opened 2020-05-05
[IMP] stock_request: pivot view

chienandalu:
When there are lots of warehouses and requests per day, a pivot view can be a good help to have visibility on what's being requested
image

cc@Tecnativa TT20739

### DIFF
--- a/stock_request/views/stock_request_views.xml
+++ b/stock_request/views/stock_request_views.xml
@@ -37,6 +37,8 @@
                 <field name="company_id" groups="base.group_multi_company"/>
                 <field name="product_id"/>
                 <filter string="Archived" name="inactive" domain="[('active','=',False)]"/>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I would prefer to put requester in the columns part instead of request day (or both), don't you think? Also a filter of last month or so it's needed for not populating so much data initially.

chienandalu:
It's just a pivot over the model action, not an sql report.

pedrobaeza:
Well, what I ask is for already included fields and you only have to change default col in the view definition.

chienandalu:
Ok. added those defaults. I grouped by locations better than by requester as it's more relevant.

HviorForgeFlow (approved):
:+1:

etobella (approved):
LGTM :+1:

PR #905, opened 2020-05-15
[IMP] stock_request_picking_type: Synchronize Picking Type and Warehouse

osi-scampbell:
If a user has 2 warehouses for a single company then it is reasonable to assume they would have an Operation Type for each warehouse related to Stock Requests for traceability purposes. The problem I produced on Runbot is that changing the warehouse does not change the picking type and can create a little confusion for the user from the Kanban view as shown here:

Kanban View:
image

List View:
image

The 2 options available is we can expose the picking_type_id field on the view, or if we want to keep it hidden, then we can go with my onchange method if it is acceptable.

I believe it is also worth noting, if the user creates the SRO inside one of the Kanban cards the Stock Picking Type will be set properly, this issue is only related to SRO's created from the Stock Request App.

### DIFF
--- a/stock_request_picking_type/models/stock_request_order.py
+++ b/stock_request_picking_type/models/stock_request_order.py
@@ -19,3 +19,12 @@ class StockRequestOrder(models.Model):
     picking_type_id = fields.Many2one(
         'stock.picking.type', 'Operation Type',
         default=_get_default_picking_type, required=True)
+
+    @api.onchange('warehouse_id')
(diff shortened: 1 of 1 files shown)


max3903 (approved):

patrickrwilson (approved):

max3903:
@rousseldenis Thank you!

rousseldenis:
> @rousseldenis Thank you!

Welcome! Travis was stuck

PR #911, opened 2020-05-26
Add stock_packaging_calculator

simahawk:
"""
Basic module providing an helper method to calculate the quantity of product by packaging.

Imagine you have the following packagings:

* Pallet: 1000 Units
* Big box: 500 Units
* Box: 50 Units

and you have to pick from your warehouse 2860 Units.

Then you can do:

    >>> product.product_qty_by_packaging(2860)

    [(2, "Pallet"), (1, "Big Box"), (7, "Box"), (10, "Units")]

With this you can show a proper message to warehouse operators to quickly pick the quantity they need.
"""
I'm not sure about module name and repo :grin:

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -1,0 +1,51 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import models
+from odoo.tools import float_compare
+
+
+class Product(models.Model):
+    _inherit = "product.product"
+
+    def product_qty_by_packaging(self, prod_qty, min_unit=None):
+        """Calculate quantity by packaging.
+
+        Limitation: fractional quantities are lost.
+
+        :prod_qty: total qty to satisfy.
+        :min_unit: minimal unit of measure as a tuple (qty, name).
+                   Default: to UoM unit.
+        :returns: list of tuple in the form [(qty_per_package, package_name)]
+
+        """
+        packagings = [(x.qty, x.name) for x in self.packaging_ids]
+        if min_unit is None:
+            # You can pass `False` to skip it.
+            single_unit = self.uom_id
+            min_unit = (single_unit.factor, single_unit.name)
+        if min_unit:
+            packagings.append(min_unit)
+        return self._product_qty_by_packaging(
+            sorted(packagings, reverse=True), prod_qty
+        )
+
+    def _product_qty_by_packaging(self, pkg_by_qty, qty):
+        """Produce a list of tuple of packaging qty and packaging name."""
+        # TODO: refactor to handle fractional quantities (eg: 0.5 Kg)
+        res = []
+        for pkg_qty, pkg in pkg_by_qty:
+            qty_per_pkg, qty = self._qty_by_pkg(pkg_qty, qty)
+            if qty_per_pkg:
+                res.append((qty_per_pkg, pkg))
+            if not qty:
+                break
+        return res
+
+    def _qty_by_pkg(self, pkg_qty, qty):
+        """Calculate qty needed for given package qty."""
+        qty_per_pkg = 0
+        while float_compare(qty - pkg_qty, 0.0, precision_digits=3) >= 0.0:
+            qty -= pkg_qty
+            qty_per_pkg += 1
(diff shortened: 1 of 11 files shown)


jgrandguillaume (approved):
LGTM

stock_packaging_calculator/models/product.py
```diff
@@ -0,0 +1,46 @@
...
+
+        :prod_qty: total qty to satisfy.
+        :min_unit: minimal unit of measure as a tuple (qty, name).
+                   Default: to UoM unit.
+        :returns: list of tuple in the form [(qty_per_package, package_name)]
+        """
+        packagings = [(x.qty, x.name) for x in self.packaging_ids]
+        if min_unit is None:
+            # You can pass `False` to skip it.
+            single_unit = self.env.ref("uom.product_uom_unit")
```
guewen: Shouldn't 1 be the `uom_id` of the product?

jgrandguillaume: Oh good catch, if I sold a pair of gloves, then 1 pair is in fact 2 units. I want to display 1 pair in that case

stock_packaging_calculator/models/product.py
```diff
@@ -0,0 +1,46 @@
...
+            if qty_per_pkg:
+                res.append((qty_per_pkg, pkg))
+            if not qty:
+                break
+        return res
+
+    def _qty_by_pkg(self, pkg_qty, qty):
+        """Calculate qty needed for given package qty."""
+        qty_per_pkg = 0
+        while (qty - pkg_qty) > -1:
```
guewen: So if the unit is kg, you can take the last remaining 0.5 kg in the void at the bottom of the empty pallet? :stuck_out_tongue_winking_eye: 
I would compare with `>= 0`, ideally with float utils

stock_packaging_calculator/models/product.py
```diff
@@ -0,0 +1,46 @@
...
+                   Default: to UoM unit.
+        :returns: list of tuple in the form [(qty_per_package, package_name)]
+        """
+        packagings = [(x.qty, x.name) for x in self.packaging_ids]
+        if min_unit is None:
+            # You can pass `False` to skip it.
+            single_unit = self.env.ref("uom.product_uom_unit")
+            min_unit = (1.0, single_unit.name)
+        if min_unit:
+            packagings.append(min_unit)
```
sebalix, line 28: No need to check that `min_unit` isn't already in `packagings`? Or use a `set`?

sebalix: It was misleading, we are mixing packaging with uom here OK, so `min_unit` can't be in `packaging`.

simahawk: eh, the point is that very likely you don't have a packaging for the single unit, that's why I'm just using names and allow you to provide different unit if really needed.

stock_packaging_calculator/models/product.py
simahawk: but to do this really properly we should have UoM on tha packaging (https://github.com/OCA/product-attribute/pull/486) unless you tell me that by default the pkging qty refers always to the unit one

stock_packaging_calculator/models/product.py
simahawk: :thinking:

sebalix (approved):

stock_packaging_calculator/models/product.py
```diff
@@ -0,0 +1,51 @@
...
+            if qty_per_pkg:
+                res.append((qty_per_pkg, pkg))
+            if not qty:
+                break
+        return res
+
+    def _qty_by_pkg(self, pkg_qty, qty):
+        """Calculate qty needed for given package qty."""
+        qty_per_pkg = 0
+        while float_compare(qty - pkg_qty, 0.0, precision_digits=3) >= 0.0:
```
rousseldenis, line 48: Why arbitrary precision ?

HaraldPanten (approved):

stock_packaging_calculator/models/product.py
simahawk: @rousseldenis I missed a TODO but I'll fix it in the next round I'm about to start. 
Meanwhile, as this 1st alpha part has been approved, I merge it and submit a new PR soon.

PR #916, opened 2020-06-08
stock_packaging_calculator improvements

simahawk:
Atomic commits explain changes.

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -13,7 +13,7 @@ Packaging = namedtuple("Packaging", "id name qty")
 class Product(models.Model):
     _inherit = "product.product"
 
-    def product_qty_by_packaging(self, prod_qty):
+    def product_qty_by_packaging(self, prod_qty, with_contained=False):
         """Calculate quantity by packaging.
 
         The minimal quantity is always represented by the UoM of the product.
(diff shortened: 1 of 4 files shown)


simahawk:
CC @rousseldenis

stock_packaging_calculator/models/product.py
```diff
@@ -1,51 +1,76 @@
...
-            single_unit = self.uom_id
-            min_unit = (single_unit.factor, single_unit.name)
-        if min_unit:
-            packagings.append(min_unit)
+        packagings = [Packaging(x.id, x.name, x.qty) for x in self.packaging_ids]
+        # Add minimal unit
+        packagings.append(
+            # NOTE: the ID here could clash w/ one of the packaging's.
+            # If you create a mapping based on IDs, keep this in mind.
+            Packaging(self.uom_id.id, self.uom_id.name, self.uom_id.factor)
```
guewen, line 42: should we add a "model" key in the dictionary so we know to what model the id refers?

simahawk: I thought about a flag instead (eg: `is_unit=True`) but it's quite easy to get the unit elem: it's always the last one.

rousseldenis (approved):
LGTM

guewen (approved):

PR #919, opened 2020-06-09
[12.0][IMP] stock_move_location: Allow to select any location.

sergio-teruel:
cc @Tecnativa
With this PR the behavior is the same that a picking. If I can do a internal transfer and I can select any location, Why in this module can not do it??
https://github.com/odoo/odoo/blob/a9a7834c8766246d3308eb09893f06a5b3639447/addons/stock/views/stock_picking_views.xml#L263-L264


HaraldPanten (approved):
Small change. LGTM 👍

pedrobaeza (approved):

carlosdauden (approved):

sergio-teruel:
Can we merge it??

PR #923, opened 2020-06-12
[13.0] Add stock_move_auto_assign

guewen:
Automatically check availability of stock moves when a move is set to "done".

It uses queue jobs to verify the availability in order to have a minimal impact
on the user operations.

The conditions to trigger the check are:

* A move is marked as done
* The destination locations of the move lines are internal
* The move doesn't have successors in a chain of moves

At this point, jobs are generated:

* One job per product
* Any move waiting for stock in a parent (or same) location of the internal
  destination locations from the done move has its availability checked

Only one job is generated for an identical set of (product, locations).

* [ ] needs https://github.com/OCA/queue/pull/237

### DIFF
--- a/stock_move_auto_assign/models/product_product.py
+++ b/stock_move_auto_assign/models/product_product.py
@@ -1,0 +1,41 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+from odoo.addons.queue_job.job import job
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _moves_auto_assign_domain(self, locations):
+        return [
+            ("product_id", "=", self.id),
+            ("location_id", "parent_of", locations.ids),
+            ("state", "in", ("confirmed", "partially_available")),
+            # useless to try reserving a move that waits on another move
+            # or is MTO
+            ("move_orig_ids", "=", False),
+            ("procure_method", "=", "make_to_stock"),
+            # Do not filter on product_id.type by default because it uses an
+            # additional query on product_product and product_template.
+            # StockMove._prepare_auto_assign() already filtered out
+            # non-stockable products and # `_action_assign()` would filter them
+            # out anyway.
+        ]
+
+    @job(default_channel="root.stock_auto_assign")
+    def moves_auto_assign(self, locations):
+        """Try to reserve moves based on product and locations
+
+        When a product has been added to a location, it searches all*
+        the moves with a source equal or above this location and try
+        to reserve them.
+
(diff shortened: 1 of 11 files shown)


jgrandguillaume (approved):
Thanks for the job, LGTM

stock_move_auto_assign/models/product_product.py
```diff
@@ -0,0 +1,42 @@
...
+            # useless to try reserving a move that waits on another move
+            # or is MTO
+            ("move_orig_ids", "=", False),
+            ("procure_method", "=", "make_to_stock"),
+            # Do not filter on product_id.type by default because it uses an
+            # additional query on product_product and product_template.
+            # StockMove._prepare_auto_assign() already filtered out
+            # non-stockable products and # `_action_assign()` would filter them
+            # out anyway.
+        ]
```
LoisRForgeFlow, line 26: Wouldn't be good to add a date (date_expected) in this domain? This is easily extensible, but anyway you wouldn't want to reserve inventory for moves that are planned to be done next month for instance.

jgrandguillaume: @LoisRForgeFlow Odoo standard behavior is actually to reserve goods at each scheduler run whatever their dates. This module doesn't intend to change this principle, only avoid to launch the scheduler every 10 minutes.

If you want to avoid move reserving goods a month in advance, you shall use our excellent module :) https://github.com/OCA/wms/tree/13.0/sale_stock_available_to_promise_release 

This will allow you to chose what you release and intend the system to reserve or not.

With this module, every move within the warehouse is supposed to be treated and therefor reserved.

LoisRForgeFlow: Fair enough, thanks for the insights. Is your recommendation to reduce the frequency of the scheduler when using this module?

rousseldenis (approved):
LGTM.

Just one question, I don't know too much v13, but before, when trying to assign moves that are currently modified by user (treating move lines by modifying quantities), the assignation will erase moves lines and recreate them. Is this process safe?

guewen:
@rousseldenis Moves already available are ignored by `_action_assign()`, but possibly on moves partially available? (https://github.com/odoo/odoo/blob/03fb98f876ea03deef05acb74144d8e979a61f54/addons/stock/models/stock_move.py#L1170)

I didn't know this issue. But then it happens when the scheduler runs too?

HaraldPanten (approved):

marcelsavegnago (approved):

stock_move_auto_assign/models/product_product.py
jgrandguillaume: Yes, of course, as then you probably don't need it anymore for short term, only once a day shall be enough IMO

guewen:
Rebased and re-run pre-commit

PR #927, opened 2020-06-23
[12.0][IMP] stock_warehouse_calendar: robustness

LoisRForgeFlow:
when delta argument is a float `plan_days` method can return
unexpected False result. This can lead to silent errors.
Therefore we ensure that delta is a integer in the helper.

@ForgeFlow


AdriaGForgeFlow (approved):

AaronHForgeFlow (approved):
Code :+1:

JordiBForgeFlow (approved):
:+1: code review

HviorForgeFlow (approved):

PR #928, opened 2020-06-23
[13.0][IMP] stock_warehouse_calendar: robustness

LoisRForgeFlow:
when delta argument is a float `plan_days` method can return
unexpected False result. This can lead to silent errors.
Therefore we ensure that delta is a integer in the helper.

Forward port of https://github.com/OCA/stock-logistics-warehouse/pull/927.

@ForgeFlow


HaraldPanten (approved):
LGTM like https://github.com/OCA/stock-logistics-warehouse/pull/927

rousseldenis (approved):

PR #937, opened 2020-07-03
[12.0][UPT]stock_inventory_revaluation add init_hook.py to speed up installation

AaronHForgeFlow:
In companies with millions account move lines it takes ages to install this module

@ForgeFlow

### DIFF
--- a/stock_inventory_revaluation/__init__.py
+++ b/stock_inventory_revaluation/__init__.py
@@ -1,2 +1,3 @@
 from . import models
 from . import wizards
+from .init_hook import pre_init_hook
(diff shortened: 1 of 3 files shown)


rousseldenis:
@AaronHForgeFlow I'm maybe wrong but I think this is not sufficent as compute function will be bypassed as column is already created by your script.

AaronHForgeFlow:
In this case, the column is not computed, but for some reason it stays for long time storing null in the column. I will keep investigating. You can add the Work in Progress Label too.

AaronHForgeFlow:
wrong

PR #939, opened 2020-07-06
[IMP][stock_packaging_calculator] Add key is_unit for in the packaging calculator return

Tonow-c2c:
Add the is_unit key to easily check the type of the line

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -100,7 +100,12 @@ class Product(models.Model):
         for pkg in pkg_by_qty:
             qty_per_pkg, qty = self._qty_by_pkg(pkg.qty, qty)
             if qty_per_pkg:
-                value = {"id": pkg.id, "qty": qty_per_pkg, "name": pkg.name}
+                value = {
+                    "id": pkg.id,
+                    "qty": qty_per_pkg,
+                    "name": pkg.name,
+                    "is_unit": pkg.is_unit,
+                }
                 if with_contained:
                     contained = None
                     if not pkg.is_unit:
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -63,6 +63,7 @@ class Product(models.Model):
 
             {contained: [{id: 1, qty: 4, name: "Big box"}]}
         """
+        self.ensure_one()
         return self._product_qty_by_packaging(
             self._ordered_packaging(), prod_qty, with_contained=with_contained,
         )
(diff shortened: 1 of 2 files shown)


stock_packaging_calculator/models/product.py
```diff
@@ -100,7 +100,12 @@ def _product_qty_by_packaging(self, pkg_by_qty, qty, with_contained=False):
         for pkg in pkg_by_qty:
             qty_per_pkg, qty = self._qty_by_pkg(pkg.qty, qty)
             if qty_per_pkg:
-                value = {"id": pkg.id, "qty": qty_per_pkg, "name": pkg.name}
+                value = {
```
LoisRForgeFlow: Would you mind adding a hook here? I think it can be helpful to easily map the information you need without needing to browse the record using the ID downstream.

grindtildeath: Done here as I also needed it: https://github.com/Tonow-c2c/stock-logistics-warehouse/pull/1 :wink:

grindtildeath (approved):
:+1:

LoisRForgeFlow:
Travis is unhappy :crying_cat_face:

stock_packaging_calculator/models/product.py
LoisRForgeFlow: Thanks!

guewen:
@Tonow-c2c could you adapt the tests? 


```
First differing element 0:
{'id': 3, 'qty': 1, 'name': 'FOO Pallet', 'is_unit': False}
{'id': 3, 'qty': 1, 'name': 'FOO Pallet'}
```

guewen:
What about passing the record instead of `is_unit`? We'll have the information in `dict['record']._name`

guewen (approved):
Would be worth a squash IMO, but LGTM

simahawk:
@Tonow-c2c can you squash?

simahawk:
ops, I hit the btn by error :sob:

jbaudoux:
A small remark, when ordering the packaging by quantity, can you ensure that if we have a package with qty=1, it comes before the single unit?

simahawk:
> A small remark, when ordering the packaging by quantity, can you ensure that if we have a package with qty=1, it comes before the single unit?

it should be always the case because they are sorted before adding the unit. If you get it wrong, it does not depend on this module very likely.

jbaudoux:
> > A small remark, when ordering the packaging by quantity, can you ensure that if we have a package with qty=1, it comes before the single unit?
> 
> it should be always the case because they are sorted before adding the unit. If you get it wrong, it does not depend on this module very likely.

@simahawk They are sorted after. See:
https://github.com/OCA/stock-logistics-warehouse/blob/13.0/stock_packaging_calculator/models/product.py#L95

simahawk:
tackled here https://github.com/OCA/stock-logistics-warehouse/pull/949

PR #941, opened 2020-07-08
[13.0] Add stock_picking_consolidation_priority

guewen:
As the number of packing space is usually limited, if someone starts to prepare
an order by picking it in a zone, all other related picking in the other zones
should be of a higher priority. This should make a "started" order a priority
for all.

This module adds an option "Raise priority when partially available" on
operation types.

When the option is set on an operation type, the first time a packing operation
is made partially available (goods have been moved for instance from a picking),
all the other moves that will bring goods for the same packing transfer have
their priority raised to "Very Urgent". If the moves that bring goods to the
packing transfer involves a chain of several moves, all the moves of the chain
have their priority raised.

### DIFF
--- a/stock_picking_consolidation_priority/models/stock_move.py
+++ b/stock_picking_consolidation_priority/models/stock_move.py
@@ -21,80 +21,81 @@ class StockMove(models.Model):
 
         Consider this chain of moves::
 
-            PICK/001 ━►  PACK/001  ┓
-                                   ┃
-            PICK/002 ┓             ┣━► OUT/001
-                     ┣━► PACK/002  ┛
-            PICK/003 ┛
+                      PICK/001 ━►  PACK/001  ┓
+                                             ┃
+                      PICK/002 ┓             ┣━► OUT/001
+                               ┣━► PACK/002  ┛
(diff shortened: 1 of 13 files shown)


stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _consolidate_priority_domain(self):
+        return [("state", "not in", ("cancel", "done"))]
+
+    def _consolidate_priority_values(self):
+        return {"priority": self._consolidate_priority_value}
+
+    def _consolidate_priority(self):
+        query, params = self._query_get_consolidate_moves()
+        self.env.cr.execute(query, params)
```
grindtildeath, line 113: Shouldn't the models be flushed before executing the SQL?

grindtildeath (approved):
Just a question, otherwise LGTM

stock_picking_consolidation_priority/models/stock_move.py
guewen: yes, thanks! pushed a fixup

simahawk (approved):

PR #942, opened 2020-07-15
Re-license stock_packaging_calculator w/ LGPL

simahawk:


jgrandguillaume (approved):
LGTM, thx Simone we need this LGPL for DDMRP stack for info

marcelsavegnago (approved):

grindtildeath:
While there's a PR for it, wouldn't it make sense to move the module to product-attribute repo, where all the other modules related to product and packaging are? 
Also IMO renaming the module to product_packaging_calculator would make sense as we don't really depend on stock modules.

My 0.02 $ :wink:

HaraldPanten (approved):
minor changes

simahawk:
@grindtildeath you might be right but we don't have enough time to move it especially because it adds little value AFAIS.
You can add it to the ROADMAP if want.

PR #946, opened 2020-07-23
[IMP] New filters under minumum and over maximum

manuelcalerosolis:
Two new filters have been included, for low and above maximum stocks, which I think will be very useful.

image


rousseldenis:
@manuelcalerosolis Thank you for your contribution. Can you check flake8 ?

PR #951, opened 2020-08-17
[12.0][FIX] stock_available_unreserved

AdriaGForgeFlow:
This PR includes the fix proposed here #511 and also adds a pre_init_hook to ease the installation of this module in huge databases.

cc ~ @ForgeFlow

### DIFF
--- a/stock_available_unreserved/hooks.py
+++ b/stock_available_unreserved/hooks.py
@@ -1,0 +1,66 @@
+# Copyright 2020 ForgeFlow S.L. (https://www.forgeflow.com)
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    """
+    The objective of this hook is to speed up the installation
+    of the module on an existing Odoo instance.
+
+    Without this script, if a database has a few hundred thousand
+    partners, which is not unlikely, the update will take
+    at least a few hours.
+
+    The pre init script pre-computes the field using SQL.
+    """
+    add_field_unreserved_quantity(cr)
+    add_field_contains_unreserved(cr)
+
+
+def add_field_unreserved_quantity(cr):
+    """Add field unreserved_quantity and compute its value"""
+
+    cr.execute("""
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name='stock_quant' AND
+        column_name='unreserved_quantity'
+    """)
+    if not cr.fetchone():
+        cr.execute(
+            """
+            ALTER TABLE stock_quant ADD COLUMN unreserved_quantity numeric
+            DEFAULT 0.0;
+            """)
+        cr.execute(
+            """
+            UPDATE stock_quant
+            SET unreserved_quantity = quantity - reserved_quantity
+            WHERE reserved_quantity > 0.01
+            """
+        )
+
+
+def add_field_contains_unreserved(cr):
+    """Add field contains_unerserved and compute its value"""
+
+    cr.execute("""
+            SELECT column_name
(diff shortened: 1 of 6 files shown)


pedrobaeza:
As this is done in the middle of a version, you must keep the previous computed field, although providing the new one for optimized computations.

AdriaGForgeFlow:
> As this is done in the middle of a version, you must keep the previous computed field, although providing the new one for optimized computations.

I've brought it back. I know the computation of the field on the pre_init_hook is not exactly the one in python code but to translate the `_get_available_quantity` method to SQL language would take too long. Do you think like this is alright or I should just create the column and leave it uncomputed?

stock_available_unreserved/hooks.py
```diff
@@ -0,0 +1,66 @@
...
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name='stock_quant' AND
+        column_name='unreserved_quantity'
+    """)
+    if not cr.fetchone():
+        cr.execute(
+            """
+            ALTER TABLE stock_quant ADD COLUMN unreserved_quantity numeric
+            DEFAULT 0.0;
```
pedrobaeza: Don't add default, or it will be forever the default.

stock_available_unreserved/hooks.py
```diff
@@ -0,0 +1,66 @@
...
+            SELECT column_name
+            FROM information_schema.columns
+            WHERE table_name='stock_quant' AND
+            column_name='contains_unreserved'
+        """)
+    if not cr.fetchone():
+        cr.execute(
+            """
+            ALTER TABLE stock_quant ADD COLUMN contains_unreserved boolean
+            DEFAULT FALSE;
```
pedrobaeza: Don't add default

pedrobaeza (changes requested):

stock_available_unreserved/hooks.py
```diff
@@ -0,0 +1,64 @@
...
+logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    """
+    The objective of this hook is to speed up the installation
+    of the module on an existing Odoo instance.
+
+    Without this script, if a database has a few hundred thousand
+    partners, which is not unlikely, the update will take
```
pedrobaeza:
```suggestion
    quants, which is not unlikely, the update will take
```

stock_available_unreserved/tests/test_stock_available_unreserved.py
```diff
@@ -214,6 +214,7 @@ def test_stock_search(self):
         no_template = self.env['product.template']
         # Start: one template with three variants.
         # All variants have zero unreserved stock
+
```
pedrobaeza: Why adding an extra line here? PEP8 disregards empty lines in the middle of a method.

AdriaGForgeFlow: That was the cherry-pick from jordi. I have removed it.

stock_available_unreserved/hooks.py
AdriaGForgeFlow: :+1:

pedrobaeza (approved):
Please squash commits

marcelsavegnago (approved):

LoisRForgeFlow (approved):
LGTM

pedrobaeza:
@AdriaGForgeFlow @LoisRForgeFlow I have just seen that you need to add a migration script calling the pre-fill of the new field, or it will take a lot also for updating the module.

PR #959, opened 2020-09-02
[FIX][13.0][stock_inventory_preparation_filter] Remove migration scripts

ivantodorovich:
As explained here: https://github.com/OCA/stock-logistics-warehouse/pull/847#issuecomment-685783116

These migration scripts should implemented in OpenUpgrade, not in this module.

> This script (or at least most of it) belongs to OpenUpgrade. As it's actually trying to migrate the core stock module.
> 
> location_id was changed to location_ids in core
> https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L53
> https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L45
> 
> product_id was changed to product_ids in core
> https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_inventory.py#L58
> https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L50
> 
> lot_id and category_id where removed by odoo, AFAIK. it's nice that this module restores them, but the migration script should fail silently at least, or even not exist here (could be covered by openupgrade, in case this module is installed.. )
>
> It's not within the scope of this module to migrate these fields.

@jjscarafia @pedrobaeza @sergio-teruel @HaraldPanten


pedrobaeza:
OK in principle, but there may be some parts of code that should be kept here if belongs to this module. Can you confirm?

ivantodorovich:
Actually no, the 4 fields belong to core: `location_id`, `product_id`, `category_id`, `lot_id`. So all of this should be managed in OpenUpgrade.

`location_id` and `product_id` are straightforward, it can be implemented the same way in OpenUpgrade `stock`.
`lot_id` and `category_id` where removed in core, without a m2m replacement, but still they need to be managed in OpenUpgrade scope. Maybe by migrating them only if `stock_inventory_preparation_filter` is installed.. Or simply migrating them anyways, in any case if the module is not installed a few extra db columns won't hurt.

pedrobaeza (approved):

kos94ok-3D (approved):
I agree with @ivantodorovich 
We need to check that this module is installed and convert fields in the stock module

sergio-teruel (approved):

PR #961, opened 2020-09-03
[13.0][IMP] stock_location_position: Add barcode field to location view in same group like as v12.0

sergio-teruel:
cc @Tecnativa TT24428
please @carlosdauden @pedrobaeza review....

### DIFF
--- a/stock_location_position/views/stock_location.xml
+++ b/stock_location_position/views/stock_location.xml
@@ -14,6 +14,7 @@
                     <field name="posx" />
                     <field name="posy" />
                     <field name="posz" />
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But this module is not adding such `barcode` field. Why adding it here?

sergio-teruel:
Bacode only is removed in the stock location view...., Yhea!! Mybe I can add directly in stock_barcodes module...

pedrobaeza:
Yes, I think it could be a better place for that.

sergio-teruel:
OK... so close this PR---- :sob:

PR #963, opened 2020-09-05
[12.0] [IMP] stock_request - get the list of picking thru procurements

dominiquekonsuntack:
get the list of pickings from procurement
advantage: back orders will be fetched

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -117,8 +117,18 @@ class StockRequestOrder(models.Model):
     @api.depends('stock_request_ids.allocation_ids')
     def _compute_picking_ids(self):
         for record in self:
-            record.picking_ids = record.stock_request_ids.mapped('picking_ids')
-            record.picking_count = len(record.picking_ids)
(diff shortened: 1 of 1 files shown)


stock_request/models/stock_request_order.py
```diff
@@ -117,8 +117,18 @@ def _get_default_requested_by(self):
...
         for record in self:
-            record.picking_ids = record.stock_request_ids.mapped('picking_ids')
-            record.picking_count = len(record.picking_ids)
+            # The first query is kept for compatibility with the original method
+            # the second query is added for the pickings which were created as backorder / split
+
+            direct_pick_ids = record.stock_request_ids.mapped('picking_ids')
+            procurement_id = record.procurement_group_id
+            if procurement_id:
+                to_add_pick_ids = self.env['stock.picking'].search([('group_id', '=', procurement_id.id)])
```
rousseldenis, line 126: For performance reasons, you should do the search outside of the loop.

rousseldenis:
@dominiquekonsuntack Could you add a test for that ?

PR #968, opened 2020-09-16
[13.0]Multi company support for stock change qty reason

bizzappdev:
added support for multi-company for  stock change qty reason.

### DIFF
--- a/stock_change_qty_reason/views/stock_inventory_line_view.xml
+++ b/stock_change_qty_reason/views/stock_inventory_line_view.xml
@@ -19,8 +19,14 @@
             eval="[(4, ref('stock_change_qty_reason.group_qty_reason_preset'))]"
         />
         <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="before">
+                <field name="company_id" invisible="1" />
+            </xpath>
             <field name="reason" position="replace">
-                <field name="preset_reason_id" />
+                <field
+                    name="preset_reason_id"
+                    domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]"
+                />
             </field>
         </field>
     </record>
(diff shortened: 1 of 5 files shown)


stock_change_qty_reason/views/stock_inventory_line_view.xml
```diff
@@ -19,8 +19,14 @@
...
         />
         <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="before">
+                <field name="company_id" invisible="1" />
+            </xpath>
             <field name="reason" position="replace">
-                <field name="preset_reason_id" />
+                <field
+                    name="preset_reason_id"
+                    domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]"
```
LoisRForgeFlow, line 28: Instead of doing this, I think it is better if you add multi company security rules for `stock.inventory.line.reason`

bizzappdev: correct me if I am wrong. In V13 you can have multiple companies active at the same time. so reason can not be from the current company of user but it has to be the company from the stock scrap/Stock picking is in

LoisRForgeFlow: I see your point now, thanks for sharing. In any case, adding default security rules is still a good practice when you add the company_id field to some model so it can coexist with this view teak, would you mind adding them?

bizzappdev: I agree with you adding additional security rule based on allowed company ids will be a good thing.

bizzappdev: Done

bizzappdev:
@LoisRForgeFlow can you review again?

LoisRForgeFlow (approved):
LGTM

rousseldenis (approved):

PR #970, opened 2020-09-16
[IMP] stock_secondary_unit: Display info only for inventory users

sergio-teruel:
cc @Tecnativa TT25561
A sales user without inventory security group can not see the product form view.

ping @chienandalu @CarlosRoca13

### DIFF
--- a/stock_secondary_unit/migrations/13.0.1.0.1/post-migration.py
+++ b/stock_secondary_unit/migrations/13.0.1.0.1/post-migration.py
@@ -1,0 +1,22 @@
+# Copyright 2020 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Delete old security group for views.
+    views = [
+        "stock_secondary_unit.view_template_property_form",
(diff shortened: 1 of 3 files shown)


stock_secondary_unit/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,22 @@
...
+        "stock_secondary_unit.product_form_view_procurement_button",
+        "stock_secondary_unit.product_template_tree_view",
+        "stock_secondary_unit.product_product_tree_view",
+    ]
+    IrUiView = env["ir.ui.view"]
+    user_group = env.ref("uom.group_uom")
+    views_to_update = IrUiView.browse()
+    for view in views:
+        views_to_update |= env.ref(view)
+    views_to_update.write({"groups_id": [(2, user_group.id)]})
```
chienandalu: Is this the right operation for this case? Won't it delete the group itself?? From the docstring:

```
          ``(2, id, _)``
              removes the record of id ``id`` from the set, then deletes it
              (from the database). Can not be used in :meth:`~.create`.
          ``(3, id, _)``
              removes the record of id ``id`` from the set, but does not
              delete it. Can not be used in
              :meth:`~.create`.
```

sergio-teruel:
Ups!!! done

chienandalu (approved):

pedrobaeza (approved):

PR #974, opened 2020-09-22
[FIX] stock_secondary_unit: Compute secondary_unit_qty_available field with sudo

sergio-teruel:
cc @Tecnativa TT25561
In kanban view also the user has not inventory user group the available quantities has been computed.
I have tried adding compute_sudo=True to field but this is not working

ping @carlosdauden ping  @victoralmau

### DIFF
--- a/stock_secondary_unit/models/product.py
+++ b/stock_secondary_unit/models/product.py
@@ -15,7 +15,7 @@ class StockProductSecondaryUnit(models.AbstractModel):
     )
 
     def _compute_secondary_unit_qty_available(self):
-        for product in self:
+        for product in self.with_context(company_owned=True).sudo():
             if not product.stock_secondary_uom_id:
                 product.secondary_unit_qty_available = 0.0
             else:


pedrobaeza:
The potential problem I see is that in a multi-company environment, if the product is used by both companies, qty_available will be correct?

sergio-teruel:
Ok... And with a force_company??

sergio-teruel:
Maybe I can use company_owned context..

https://github.com/odoo/odoo/blob/dd160aa0ac4512a498a002e21c74e9546de371f8/addons/stock/models/product.py#L234

pedrobaeza:
Yeah, maybe, although I don't get the problem if all views already have inventory user group for avoiding salesman to see them.

sergio-teruel:
Please read the initial comment

sergio-teruel:
Closed by https://github.com/OCA/stock-logistics-warehouse/pull/1005

PR #998, opened 2020-10-27
[13.0][MIG] stock_orderpoint_generator: Migration to 13.0

victoralmau:
Please @joao-p-marques review it.

@Tecnativa TT25953


victoralmau:
Appear some errors in Travis related to addon `stock_vertical_lift_storage_type` but i don't understand that happened, somebody have any idea about it?

pedrobaeza:
Fix for the branch in #1000

stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -0,0 +1,193 @@
...
+                for discard_field in self._template_fields_to_discard():
+                    data.pop(discard_field)
+                for product_id in product_ids:
+                    vals = data.copy()
+                    vals["product_id"] = product_id.id
+                    if record.auto_min_qty:
+                        vals["product_min_qty"] = stock_min_qty.get(product_id.id, 0)
+                    if record.auto_max_qty:
+                        vals["product_max_qty"] = stock_max_qty.get(product_id.id, 0)
+                    orderpoint_model.create(vals)
```
rousseldenis: I would have used the multi create feature here.

stock_orderpoint_generator/models/product.py
```diff
@@ -0,0 +1,123 @@
...
+        """Returns a dict of products with a dict of historic moves as for
+           a list of historic stock values resulting from those moves. If
+           a location_id is passed, we can restrict it to such location"""
+        location = location_id and location_id.id
+        domain_quant_loc, domain_move_in_loc, domain_move_out_loc = self.with_context(
+            location=location
+        )._get_domain_locations()
+        if not to_date:
+            to_date = fields.Datetime.now()
+        domain_move_in = domain_move_out = [
```
rousseldenis: IMHO this function is too long. Please separate some logical parts (domain creation, product_move_dict creation).

rousseldenis (changes requested):

stock_orderpoint_generator/models/product.py
victoralmau: Ok, thanks, separated it now in different functions.

stock_orderpoint_generator/models/product.py
```diff
@@ -26,53 +27,23 @@ def create(self, vals):
...
-            to_date = fields.Datetime.now()
-        domain_move_in = domain_move_out = [
-            ("product_id", "in", self.ids),
-            ("state", "=", "done"),
-        ] + domain_move_in_loc
-        domain_move_out = [
-            ("product_id", "in", self.ids),
-            ("state", "=", "done"),
-        ] + domain_move_out_loc
+    def _get_stock_move_domain(self, domain_move=False, from_date=False, to_date=False):
```
joao-p-marques, line 37: IMHO it would be clearer to use `None` as default here
```suggestion
    def _get_stock_move_domain(self, domain_move=None, from_date=None, to_date=None):
```

stock_orderpoint_generator/models/product.py
```diff
@@ -26,53 +27,23 @@ def create(self, vals):
...
+        domain = [("product_id", "in", self.ids), ("state", "=", "done")] + domain_move
```
joao-p-marques, line 38: It would also be better to check if `domain_move` is not `False` (or `None`) before appending it, as those are the default values. It would increase resilience.

stock_orderpoint_generator/models/product.py
```diff
@@ -26,53 +27,23 @@ def create(self, vals):
...
-        # to obtain the stock snapshot in every moment
-        moves_out = move_obj.search_read(
-            domain_move_out, ["product_id", "product_qty", "date"], order="date asc"
-        )
-        for move in moves_out:
-            move["product_qty"] *= -1
-        # Merge both results and group them by product id as key
-        moves = moves_in + moves_out
+            domain += [("date", ">=", from_date)]
+        domain += [("date", "<=", to_date)]
```
joao-p-marques, line 41: Same here as above. If this will always be used, why is the default value `False` and never checked?

stock_orderpoint_generator/models/product.py
```diff
@@ -26,53 +27,23 @@ def create(self, vals):
...
+        return domain
+
+    def _set_product_moves_dict(
+        self, moves=False, location=False, from_date=False, to_date=False
```
joao-p-marques, line 45: Same suggestion as above here. 

Can you filter by date with said date value set to `False`?

I think `None` is a better default value here, at least for arguments that can't be used with the `False` value (like `moves`), but I don't know if this an Odoo/OCA standard.

However, if this value is not passed and the default is assumed, you will get an error bellow when running the loop for `moves`. You should make them obligatory args, or test if they are set before using them.

stock_orderpoint_generator/models/product.py
```diff
@@ -121,3 +92,37 @@ def _compute_historic_quantities_dict(
                 v["stock"] for k, v in product_moves.items()
             ]
         return product_moves_dict
+
+    def _compute_historic_quantities_dict(
+        self, location_id=False, from_date=False, to_date=False
```
joao-p-marques, line 97: Same suggestions/doubts as above.

joao-p-marques (commented):
Looks good overall. Just some questions and suggestions.

stock_orderpoint_generator/models/product.py
victoralmau: Everything is related to orderpoint_template model. These functions are called related to some False fields (that it's possible to contain False in some cases).
With these considerations IMO everithing it's correct and False default in functions it's OK.

joao-p-marques: Yes, I understand you can filter in the ORM with some fields set to `False` (like `location`, right?)
However, in this particular case, if you leave `moves` with `False` as default value and never check if is defined on the function call, you might get an error bellow in the loop:
```python
for move in moves:
```
as you might be trying to loop over a `False` value.

If `moves` is always needed, you should remove the default and leave it as an obligatory argument. Otherwise, you must check if it is still `False` before trying to iterate it (or `None` as default as, if this is a Python object and not a Odoo ORM object, it would make more sense to use that as default)

The same can happen with the dates. Does the ORM allow you to filter for `from_date=False` or `to_date=False`? 

> Everything is related to orderpoint_template model. 

I didn't really understand this part. Are you inheriting and overriding this method in particular?

joao-p-marques (approved):

pedrobaeza:
@rousseldenis is it OK for you?

rousseldenis (approved):
Code review

pedrobaeza:
I'm afraid we have exhausted Travis free build rate, so this won't work. Let's see how do we handle at OCA level.

PR #1000, opened 2020-10-28
[13.0] Remove _order override in stock.location

guewen:
The default order of stock.location is 'complete_name'.
Changing it to "name" can have side effects, and shouldn't be required
for stock_inventory_lockdown.

See https://github.com/OCA/stock-logistics-warehouse/pull/811/files#r512797834

### DIFF
--- a/stock_inventory_lockdown/models/stock_location.py
+++ b/stock_inventory_lockdown/models/stock_location.py
@@ -9,7 +9,6 @@ class StockLocation(models.Model):
     """Refuse changes during exhaustive Inventories"""
 
     _inherit = "stock.location"
-    _order = "name"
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
@LoisRForgeFlow do you agree?

HaraldPanten (approved):

rousseldenis (approved):

LoisRForgeFlow:
Sorry for being late, but I wanted to express my agreement with it anyway. It seemed a legacy issue from the initial version of the module. Thanks for the catch!

PR #1002, opened 2020-10-30
[10.0][IMP] stock_available_unreserved: Add hook method

rousseldenis:
Sometimes, it is needed to manipulate the available result without
requerying the quants. This hook allows it.


Cedric-Pigeon (approved):
Code review

PR #1004, opened 2020-11-02
[IMP + FIX] stock_move_location: Display reserved quantity in wizard lines. Reassign reserved quantities for quants moved.

sergio-teruel:
**Behavior before this PR:**

- Create a outgoing picking with any product with available stock
- Check availability
- Move all stock to other location with the wizard
- Run sheduler

The detailed operaciont have two records, one with 0 units and other (new) with the reserved quatities

**Behavior before this PR:**
- Create a outgoing picking with any product with available stock
- Check availability
- Move all stock to other location with the wizard
- Now the reserved quantities get from new location
- Run sheduler

All Ok.

cc @Tecnativa TT26510

### DIFF
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -90,6 +90,7 @@ class StockMoveLocationWizard(models.TransientModel):
                     "product_id": quant.product_id.id,
                     "move_quantity": quant.quantity,
                     "max_quantity": quant.quantity,
+                    "reserved_quantity": quant.reserved_quantity,
                     "origin_location_id": quant.location_id.id,
                     "lot_id": quant.lot_id.id,
                     "product_uom_id": quant.product_uom_id.id,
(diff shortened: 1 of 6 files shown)


stock_move_location/wizard/stock_move_location.py
```diff
@@ -199,7 +234,8 @@ def _get_group_quants(self):
         # Using sql as search_group doesn't support aggregation functions
```
carlosdauden, line 234: I know it is not a modification of this PR, but I don't quite understand what cannot be done with read_group

carlosdauden (approved):

pedrobaeza (approved):

stock_move_location/wizard/stock_move_location.py
rousseldenis: Indeed, you're right. Something to be done in another PR ? As this is bypassing security.

PR #1008, opened 2020-11-05
[13.0] Add picking_type_id in stock_reserve_rule list/search views

guewen:
It's an important information for the reservation rules,
and currently we have to records one-by-one to check the configuration.


rousseldenis (approved):
LGTM

bodedra (approved):
LG :+1:

guewen:
Moved in #1013 as the field is renamed

PR #1012, opened 2020-11-16
[FIX] stock_vertical_lift_kardex proxy

gurneyalex:
The proxy would not exit properly in case of a lost connection with the JMIF server.
The change allows the proxy to exit, giving a chance to an external monitoring system
to restart the service.

### DIFF
--- a/stock_vertical_lift_kardex/proxy/kardex-proxy.py
+++ b/stock_vertical_lift_kardex/proxy/kardex-proxy.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import os
 import ssl
+import sys
 import time
 
 import aiohttp  # pylint: disable=missing-manifest-dependency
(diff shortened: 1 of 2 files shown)


guewen (approved):

guewen:
pre-commit is not happy though

stock_vertical_lift_kardex/proxy/kardex-proxy.py
```diff
@@ -103,7 +105,9 @@ async def notify_odoo(self, msg):
             params = {"answer": msg, "secret": self.args.secret}
             async with session.post(url, data=params) as resp:
                 resp_text = await resp.text()
-                _logger.info("Reponse from Odoo: %s %s", resp.status, resp_text)
+                _logger.info(
+                    "Reponse from Odoo: %s %s", resp.status, resp_text
```
sebalix:
```suggestion
                    "Response from Odoo: %s %s", resp.status, resp_text
```

PR #1013, opened 2020-11-16
[13.0] stock_reserve_rule: Change picking type to many2many in reserve rules

guewen:
The picking type field allows to restrict the reservation rules to a picking type. There is no reason to restrict to only one picking type, this PR replaces the many2one per a many2many.

### DIFF
--- a/stock_reserve_rule/migrations/13.0.1.1.0/post-migration.py
+++ b/stock_reserve_rule/migrations/13.0.1.1.0/post-migration.py
@@ -1,0 +1,18 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    cr.execute(
+        """
+    INSERT INTO stock_picking_type_stock_reserve_rule_rel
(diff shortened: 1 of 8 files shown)


stock_reserve_rule/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    cr.execute(
+        """
+    INSERT INTO stock_picking_type_stock_reserve_rule_rel
+      (stock_reserve_rule_id, stock_picking_type_id)
+    SELECT id, picking_type_id
```
sebalix, line 13: I'm not sure we still have the `picking_type_id` field on `stock_reserve_rule` once the module has been updated.

guewen: We do. I checked this and run the upgrade on the project's DB. In fact, the column is removed after the migration script is called.

sebalix: Good to know, thanks!

sebalix (approved):

simahawk (approved):

PR #1021, opened 2020-12-01
[13.0][FIX] stock_vertical_lift: handles multiple move lines on pick

sebalix:
If there is two move lines for the same product in the vertical lift (stored in2 differents trays for instance), the pick scenario was failing when the user was processing the first line.

To circumvent this, instead of validating directly the move, we put the line in its own stock move, then we put the stock move in its own transfer and validate this one.

Methods used to do that have been copied from the `shopfloor` module, they probably deserves their own module as they are quite generic.

Ref. 1866

### DIFF
--- a/stock_vertical_lift/models/vertical_lift_operation_base.py
+++ b/stock_vertical_lift/models/vertical_lift_operation_base.py
@@ -11,6 +11,79 @@ from odoo.addons.base_sparse_field.models.fields import Serialized
 _logger = logging.getLogger(__name__)
 
 
+# The following methods have been copied from 'shopfloor' module (OCA/wms)
+# TODO: we should move them in a generic module
+
+
+def split_other_move_lines(move, move_lines):
+    """Substract `move_lines` from `move.move_line_ids`, put the result
+    in a new move and returns it.
+    """
+    other_move_lines = move.move_line_ids - move_lines
+    if other_move_lines or move.state == "partially_available":
+        qty_to_split = move.product_uom_qty - sum(move_lines.mapped("product_uom_qty"))
+        backorder_move_id = move._split(qty_to_split)
+        backorder_move = move.browse(backorder_move_id)
+        backorder_move.move_line_ids = other_move_lines
+        backorder_move._recompute_state()
+        backorder_move._action_assign()
+        move._recompute_state()
+        return backorder_move
+    return False
+
+
+def extract_and_action_done(move):
+    """Extract the moves in a separate transfer and validate them.
+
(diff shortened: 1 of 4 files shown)


stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -11,6 +11,90 @@
 _logger = logging.getLogger(__name__)
 
 
+# The following methods have been copied from 'shopfloor' module (OCA/wms)
+# TODO: we should move them in a generic module
+
+
+def split_other_move_lines(move, move_lines, intersection=False):
```
guewen: As we discussed on the other one, `intersection` is an issue, it doesn't compute the correct `qty_to_split` and should be removed (in favor of filtering the lines before the call to `split_other_move_lines`. Shouldn't you remove this parameter here?

sebalix: I could remove it here yes as it is unused.
But regarding the comment of @simahawk regarding this parameter, I checked and - I could be wrong - but it was difficult to drop it...

guewen: IIRC we could make the intersection before and pass these move lines

sebalix: Yep but I was struggling with the computation of the qty to split which differs with/without intersection (and I could not understand anymore why there is this difference). Need to take time to analyse again this stuff.

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -11,6 +11,79 @@
 _logger = logging.getLogger(__name__)
 
 
+# The following methods have been copied from 'shopfloor' module (OCA/wms)
```
simahawk, line 14: reported on https://github.com/OCA/wms/issues/76

simahawk (approved):
@sebalix please rebase n squash

sebalix:
@simahawk rebased & squashed

simahawk:
@sebalix 

File "/home/travis/build/OCA/stock-logistics-warehouse/stock_vertical_lift/models/stock_location.py", line 118, in _hardware_vertical_lift_fetch_tray_payload

raise NotImplementedError()

NotImplementedError

guewen:
The error should be fixed by https://github.com/odoo/odoo/pull/63979 but in the meanwhile is fixed by #1035.
@sebalix could you rebase?

sebalix:
rebased

guewen (approved):
Thanks!

PR #1024, opened 2020-12-03
[stock_quant_manual_assign]: clear the useless wizard data after confirm

fanha99:
after confirm, the data is useless, so clear the useless data

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -56,6 +56,9 @@ class AssignManualQuants(models.TransientModel):
             ml.qty_done = ml.product_qty
         move._recompute_state()
         move.mapped('picking_id')._compute_state()
+        # after confirm, the data is useless, so clear the useless data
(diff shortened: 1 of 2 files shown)


rousseldenis:
@fanha99 Could you put the module name in PR description with a better explanation ?

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -56,6 +56,9 @@ def assign_quants(self):
             ml.qty_done = ml.product_qty
         move._recompute_state()
         move.mapped('picking_id')._compute_state()
+        # after confirm, the data is useless, so clear the useless data
+        self.quants_lines.unlink()
+        self.unlink()
```
yostashiro, line 61: Is there an example of a situation where these are needed on transient models?

fanha99: @yostashiro 
yeah, my false, I turn off the auto vacuum by mistake. And I did not know how the transient models works.
Thanks
I think this pull request must be close.

PR #1030, opened 2020-12-16
[FIX]10.0 mrp stock qty modules compatibility

flotho:
Using virtual_available is a bad choice.
As we can see here : https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_available_immediately/models/product_product.py#L17 immediate_usable_qty is used in specific modules as the stock_available one is already defining immediate_usable_qty = virtual_available
This patch allow to make both modules "immediate" and "mrp" work together and also mrp stil work alone with this.

@Cedric-Pigeon , @Yajo  any review would be appreciate on this .

Regards

### DIFF
--- a/stock_available_mrp/models/__init__.py
+++ b/stock_available_mrp/models/__init__.py
@@ -3,4 +3,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import product_product
-from . import product_template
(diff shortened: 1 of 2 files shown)


yajo:
I'm no expert here, sorry.

rousseldenis:
@flotho Is this still applicable ?

PR #907, opened 2020-05-19
[13.0] Add stock_checkout_sync

guewen:
Replaces #862 

Regroup moves at the same location when they reach an operation type.

Example of use case:

* We use several operation types for Picking Transfers (Pick Shelves, Pick Reserve)
* We use several operation types for Packing (Pack, Pack Post)
* The "Checkout Synchronization" flag is activated on the 2 Packing operation types
* The Packing can occur in several areas (Packing/PACK1, Packing/PACK2, Packing/PACK3)

Based on these assumptions, we have these pickings, all for the same order:

* SHELF001, pick 2 goods in the shelves
* RESER001, pick 2 goods in the reserve
* PACK001, pack 3 of the goods (2 from SHELF001, 1 from RESER001)
* PPOST001, pack the remaining good of the reserve

The operator uses the button "Checkout Sync" on a pick transfer,
for instance SHELF001, and select Packing/PACK2 as destination.
We want all the lines that reach PACK001 to be updated with:

* The move lines (except the one that goes to PPOST001) must now reach
  Packing/PACK2
* The same moves' destination must reach Packing/PACK2 as well:
  we constrain the user to move the goods there as we want to move
  all the goods at the same place for the packing.

Selection_848

Works best when used with ``stock_available_to_promise_release``:

* When using the mentioned module, we assume that we release moves (which
  creates the whole chain of moves) only when we know that we have the
  quantities in stock (otherwise the module splits the delivery). Any exception
  (which is now rare when you release this ways) that would lead to a backorder
  is handled by canceling the remaining line.
* When the module is not used, then the destination of the backorders may be
  changed at the same time

Compatible with ``stock_dynamic_routing``.

PR #1035, opened 2020-12-22
[13.0][FIX] stock_vertical_lift: Error in tests

victoralmau:
Actually (13.0) `stock_vertical_lift` show errors in Travis CI test 
https://travis-ci.com/github/OCA/stock-logistics-warehouse/builds/209069134

`_compute_vertical_lift_kind` is not called automatically.
Probably exists a better solution but i don't understand a lot these addons and all dependencies.

Related to: https://github.com/OCA/stock-logistics-warehouse/pull/1021#issuecomment-746464712

ping @guewen

### DIFF
--- a/stock_vertical_lift/models/vertical_lift_shuttle.py
+++ b/stock_vertical_lift/models/vertical_lift_shuttle.py
@@ -145,6 +145,11 @@ class VerticalLiftShuttle(models.Model):
         record = self.env[model].search([("shuttle_id", "=", self.id)])
         if not record:
             record = self.env[model].create({"shuttle_id": self.id})
+        # Since https://github.com/odoo/odoo/commit/3a6ac95
+        # _compute_vertical_lift_shuttle_id() don't work fine
+        # These change fix it temporally
+        if "location_id" in record:
+            record.location_id._compute_vertical_lift_shuttle_id()
         return record
 
     def action_open_screen(self):


guewen:
I saw this one here: https://github.com/OCA/stock-logistics-warehouse/pull/907#issuecomment-749558769

It looks like a regression in Odoo.
(It happens since https://github.com/odoo/odoo/commit/3a6ac95ddf6a8719aa714a83fc339ba838c3811d)

victoralmau:
OK thanks. I will review the solution you propose. This PR was only intended to resolve the error that currently appears in Travis even though it is not the final solution (because it is not the correct one).

guewen:
Yes and thanks for this! Unfortunately I have no solution, all I know is that it didn't happen before the mentioned commit. Although I think we should add this call to the compute method once in the setUpClass of the test as a temporary measure rather than in the code which will do additional writes in the location.

victoralmau:
I have changed the "solution" to define the field. Do you think this is a better solution?

victoralmau:
Change apply, what do you think about it @guewen

guewen (approved):
Thank you very much!
Nice work-around

pedrobaeza (approved):

guewen:
Should be fixed by https://github.com/odoo/odoo/pull/63979 (can be reverted as soon as it is - hopefully - merged).

PR #953, opened 2020-08-19
[IMP] reason code location set as a scrap location if set

bodedra:
Issue Ref: https://github.com/OCA/stock-logistics-warehouse/issues/914

PR #1036, opened 2020-12-22
[FIX] Empty 'Reason Code' nulls the mandatory 'Scrap Location'

BT-cserra:
The 'Scrap Location' (field ```scrap_location_id```) is mandatory for the model ```stock.scrap```, while the 'Reason Code' (field ```reason_code_id```) is not. If the 'Reason Code' is set, then the location indicated there is taken as the new scrap location (this is done in the code, the user never sees anything in the user-interface).

The problem was: Because of how the check of the field ```scrap_location_id``` was done, it could be that we received a falsy value when creating/updating the object, which would result in an empty record for the Reason Code, and thus an empty new scrap location taken from there. The scrap location is a mandatory field, thus the user got the error of a mandatory field not being set.

### DIFF
--- a/scrap_reason_code/models/stock_scrap.py
+++ b/scrap_reason_code/models/stock_scrap.py
@@ -24,7 +24,7 @@ class StockScrap(models.Model):
             self.scrap_location_id = self.reason_code_id.location_id
 
     def write(self, vals):
-        if "reason_code_id" in vals:
+        if vals.get("reason_code_id"):
             vals.update(
                 {
                     "scrap_location_id": self.env["scrap.reason.code"]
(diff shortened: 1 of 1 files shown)


bodedra:
@BT-cserra Can you try with this patch? https://github.com/OCA/stock-logistics-warehouse/pull/953

BT-cserra:
> @BT-cserra Can you try with this patch? #953

Hi @bodedra I hadn't seen it, sorry. I was about to point out that your solution is not as efficient but then I realised that the ```location_id``` is not mandatory in ```scrap.reason.code```, thus my fix wouldn't solve the problem for the case of having a 'Reason Code' with no location set. Thus I close this PR and comment on yours. **Thanks**.

PR #1040, opened 2020-12-28
[12.0][FIX] stock_request: fix archived search view

jarroyomorales:
`stock.request` and `stock.request.order` dont have an `active` field, so these filters are causing an error.

@etobella


rousseldenis (approved):

etobella (approved):
I wonder if it is better to add the field or remove the filters... 🤔 
In any case, it solves an issue so LGTM and if someone needs it, they can add it later

LoisRForgeFlow (approved):

marcelsavegnago (approved):

HaraldPanten (approved):
LGTM

PR #1042, opened 2020-12-28
[14.0][ADD] stock_free_quantity

florian-dacosta:
This module should probably replaced `stock_available_unreserved` as Odoo have added a `free_qty` field on product.product since v13. (check https://github.com/OCA/stock-logistics-warehouse/pull/810)
This module add the native field in `product.product` form view and add the field on `product.template` + in `product.template` form view

### DIFF
--- a/stock_free_quantity/models/product_template.py
+++ b/stock_free_quantity/models/product_template.py
@@ -1,0 +1,33 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def _compute_quantities_dict(self):
+        prod_available = super()._compute_quantities_dict()
+        for template in self:
+            free_qty = 0
+            for p in template.with_context(active_test=False).product_variant_ids:
+                free_qty += p.free_qty
+            prod_available[template.id]["free_qty"] = free_qty
+        return prod_available
+
+    def _compute_quantities(self):
+        super()._compute_quantities()
+        res = self._compute_quantities_dict()
+        for template in self:
+            template.free_qty = res[template.id]["free_qty"]
+
+    def _search_free_qty(self, operator, value):
+        return [("product_variant_ids.free_qty", operator, value)]
+
+    free_qty = fields.Float(
+        "Free Quantity",
+        compute="_compute_quantities",
+        search="_search_free_qty",
+        compute_sudo=False,
+        digits="Product Unit of Measure",
+    )
--- a/setup/stock_free_quantity/odoo/addons/stock_free_quantity
+++ b/setup/stock_free_quantity/odoo/addons/stock_free_quantity
@@ -1,0 +1,1 @@
+../../../../stock_free_quantity
--- a/setup/stock_free_quantity/setup.py
+++ b/setup/stock_free_quantity/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/stock_free_quantity/__init__.py
+++ b/stock_free_quantity/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
--- a/stock_free_quantity/__manifest__.py
+++ b/stock_free_quantity/__manifest__.py
@@ -1,0 +1,17 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Free Quantity",
+    "version": "14.0.1.0.0",
+    "author": "akretion,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "development_status": "Production/Stable",
+    "category": "Warehouse",
+    "depends": ["stock"],
+    "license": "AGPL-3",
+    "data": [
+        "views/product_template_view.xml",
+        "views/product_product_view.xml",
+    ],
+    "installable": True,
+}
--- a/stock_free_quantity/models/__init__.py
+++ b/stock_free_quantity/models/__init__.py
@@ -1,0 +1,1 @@
+from . import product_template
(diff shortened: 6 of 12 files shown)


florian-dacosta:
@LoisRForgeFlow @AdriaGForgeFlow @AaronHForgeFlow @kos94ok-3D @jjscarafia

stock_free_quantity/README.rst
```diff
@@ -0,0 +1,117 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/stock-logistics-warehouse-13-0/stock-logistics-warehouse-13-0-stock_available
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/153/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module proposes several options to compute the quantity available to
```
jjscarafia: this readme content is overwritten later by the description file?

jjscarafia (commented):
Thanks for the PR.
Maybe I would add the free qty field on kanban and tree views as done with [stock_available_unreserved](https://github.com/OCA/stock-logistics-warehouse/blob/13.0/stock_available_unreserved/views/product_view.xml) module.
On tree views we could use the `optional="show"`

Besides that, I've tested and is working properly.
Thanks!

stock_free_quantity/models/product_template.py
```diff
@@ -0,0 +1,35 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def _compute_quantities_dict(self):
+        prod_available = super()._compute_quantities_dict()
+        for template in self:
+            free_qty = 0
+            for p in template.with_context(active_test=False).product_variant_ids:
+                free_qty += p.free_qty
```
kos94ok-3D, line 14: @pedrobaeza Can we replace to read_group? I don`t know how read_group works with computed fields.

pedrobaeza: I don't get what do you mean. Are you talking about using `read_group` inside a compute method? That's possible.

kos94ok-3D: <pre>
data = self.env['product.product'].with_context(active_test=False).read_group([
  ('product_tmpl_id', 'in', self.ids),
], ['product_tmpl_id', 'free_qty'], ['product_tmpl_id']):
res = {
  rec['product_tmpl_id'][0]: rec['free_qty'] for rec in data
}
for template in self:
  prod_available[template.id]["free_qty"] = res.get(template.id, 0.0)
</pre>

kos94ok-3D: Is it improve perfomance or not?

pedrobaeza: Yeah, indeed you get a performance improvement. It's the usual technique for doing compute fields efficient.

florian-dacosta:
> Maybe I would add the free qty field on kanban and tree views as done with [stock_available_unreserved](https://github.com/OCA/stock-logistics-warehouse/blob/13.0/stock_available_unreserved/views/product_view.xml) module.
> On tree views we could use the `optional="show"`

Done, I've added it in kanban and tree template views and in product tree view (optional).

stock_free_quantity/README.rst
florian-dacosta: It should be, I removed it anyway

stock_free_quantity/models/product_template.py
florian-dacosta: Well, I gave it a try but it dit not work.
I don't really have time to investigate this but I guess it should not be blocking.
Won't have time to really check if there is a significant performance boost.

kos94ok-3D: I checked and saw that read_group doesn`t allow add non stored field in the field_list section.

kos94ok-3D (approved):

jjscarafia (approved):
+1 functional test + code review

jjscarafia:
once merged, can we backport it to v13? what do you think?

florian-dacosta:
> once merged, can we backport it to v13? what do you think?

I don't think it is really in line with OCA policy, since there are already a module doing the same.
We would have 2 modules doing the same, in a different way... And we can't really remove the other one since it is already used.

stock_free_quantity/models/product_template.py
```diff
@@ -0,0 +1,35 @@
...
+    def _compute_quantities(self):
+        super()._compute_quantities()
+        res = self._compute_quantities_dict()
+        for template in self:
+            template.free_qty = res[template.id]["free_qty"]
+
+    def _search_free_qty(self, operator, value):
+        domain = [("free_qty", operator, value)]
+        product_variant_ids = self.env["product.product"].search(domain)
+        return [("product_variant_ids", "in", product_variant_ids.ids)]
```
rousseldenis: You can avoid a search returning directly the domain [("product_variant_ids.free_qty", operator, value)].

But I think this assertion is False as it does not aggregate product variants free qty for the template but search on each one.

florian-dacosta: Returning  `[("product_variant_ids.free_qty", operator, value)]`, the search will then be done after anyway right ? (in variant search method)
Anyway, I confess I just copied/pasted the method defined in stock module for the other template quantity fields.
I think we have to keep the same logic (not aggregating the variant free qty) for consistency, what do you think ?

rousseldenis: That depends on what this free_qty means on product template.

If you want to search templates that have variants' free qty ... values, then it's ok. But if you want to search templates that have free qty according to aggregated values, then it should be changed.

In the case of first solution, change _search method as I mentionned. It will avoid a search (your solution do a first search then  a second one after having returned the domain).

florian-dacosta: Ok, I changed the domain, and went for the first solution, as it is the Odoo native solution for all quantity fields on templates.

jjscarafia:
Just one comment, I believe this module should replace ["stock_available_immediately"](https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_available_immediately) instead of ["stock_available_unreserved"](https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_available_unreserved). Because, if I'm not wrong, free_qty is the same concept than the one on stock_available_immediately, but not exactly as in the unreserved one.

ventor-dev:
@florian-dacosta , what is status of this pull request? When it can be merged?

ventor-dev:
@pedrobaeza , seems it met all conditions and can be merged now

PR #1047, opened 2021-01-05
[MIG] 14.0 stock_generate_putaway_from_inventory

PierrickBrun:
Big refactoring since Odoo made a lot of changes to the putaway logic.

The code is simpler after the refactor


stock_generate_putaway_from_inventory/models/stock_location.py
```diff
@@ -0,0 +1,18 @@
...
+# @author Pierrick Brun <pierrick.brun@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockLocation(models.Model):
+    _inherit = "stock.location"
+
+    def _is_child(self, location):
```
hparfr: Can we use this https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_location_children instead of recursive call ?

hparfr (changes requested):

stock_generate_putaway_from_inventory/models/stock_inventory.py
```diff
@@ -8,52 +8,51 @@
...
+        for record in self:
+            if record.state != "done":
+                raise ValidationError(
+                    _(
+                        "Please validate the inventory before generating "
+                        "the putaway strategy."
+                    )
+                )
+            for location in record.location_ids:
+                record.line_ids._generate_putaway_rules(location)
```
kos94ok-3D: Why not iterate inside a `_generate_putaway_rules` function?
```suggestion
            record.line_ids._generate_putaway_rules(record.location_ids)
```

stock_generate_putaway_from_inventory/models/stock_inventory.py
```diff
@@ -8,52 +8,51 @@
...
+
+class StockInventoryLine(models.Model):
+    _inherit = "stock.inventory.line"
+
+    def _generate_putaway_rules(self, inventory_location):
+        # Eliminate lines for other IN locations
+        # and eliminate lines where quantity is null
+        self.filtered(
+            lambda x: x.location_id._is_child(inventory_location) and x.product_qty > 0
+        )._update_product_putaway_rule(inventory_location)
```
kos94ok-3D: I think it looks better and maybe faster:
<pre>
def _generate_putaway_rules(self, inventory_locations):
    for location_in in inventory_locations:
        self.filtered(
            lambda x: x.product_qty &gt; 0.0
                and x.parent_path.startswith((location_in.parent_path or '') + '/')
        )._update_product_putaway_rule(location_in)
</pre>

kos94ok-3D (commented):
Can avoid use a `_is_child` function

kos94ok-3D (approved):

hparfr (approved):
LGTM

PR #1048, opened 2021-01-06
14.0 mig stock move common dest

sebastienbeau:


sebastienbeau:
@grindtildeath

hparfr (approved):

stock_move_common_dest/models/stock_move.py
```diff
@@ -0,0 +1,67 @@
...
+            common_move_ids = res.get(move.id)
+            if common_move_ids:
+                move.common_dest_move_ids = [(6, 0, common_move_ids)]
+            else:
+                move.common_dest_move_ids = [(5, 0, 0)]
+
+    def _search_compute_dest_move_ids(self, operator, value):
+        moves = self.search([("id", operator, value)])
+        if not moves:
+            return [("id", "=", 0)]
```
rousseldenis: I'd prefer FALSE_DOMAIN

```suggestion
            from odoo.osv.expression import FALSE_DOMAIN
            return FALSE_DOMAIN
```

rousseldenis (changes requested):

simahawk (approved):
LG apart for @rousseldenis comment

hparfr:
@rousseldenis can you update your review ?

stock_move_common_dest/models/stock_move.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @api.depends(
+        "move_dest_ids",
+        "move_dest_ids.picking_id",
+        "move_dest_ids.picking_id.move_lines",
+        "move_dest_ids.picking_id.move_lines.move_orig_ids",
+    )
+    def _compute_common_dest_move_ids(self):
+        sql = self._common_dest_move_query()
+        self.env.cr.execute(sql, (tuple(self.ids),))
```
grindtildeath, line 50: Don't we need a `flush()` here? Probably a bug in 13.0 though...

grindtildeath (commented):
LG, but watch out

stock_move_common_dest/models/stock_move.py
gurneyalex: @grindtildeath It's a SELECT query, why would we need a flush?

grindtildeath: @gurneyalex I just looked a bit more into Odoo. documentation :sweat_smile: 

On one hand, we have the `flush` function description that is: "Process all the pending computations (on all models), and flush all the pending updates to the database." (https://www.odoo.com/documentation/14.0/reference/orm.html#odoo.models.Model.flush)
So even if we have a `SELECT` statement, I tend to think we need to flush the models/fields that are used in the query to make sure we'll have the same result than we'd have using ORM methods.

On the other hand, we have "SQL Execution" (https://www.odoo.com/documentation/14.0/reference/orm.html#sql-execution) that states the following: "Because models use the same cursor and the Environment holds various caches, these caches must be invalidated when altering the database in raw SQL, or further uses of models may become incoherent. It is necessary to clear caches when using CREATE, UPDATE or DELETE in SQL, but not SELECT (which simply reads the database)."
We should also remember this was written before v13.0 and maybe wasn't updated properly (https://www.odoo.com/documentation/12.0/reference/orm.html#low-level-sql) 

Then if we look into the code, we can see a few examples of code in Odoo where `flush` is used before `SELECT` statements:
https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_account.py#L104-L105
https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_account.py#L125-L132
https://github.com/odoo/odoo/blob/14.0/addons/product_margin/models/product_product.py#L106-L108

We can probably also see others where there's no flush though...
https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/models/pos_order.py#L895

Finally, as I understand it, v13.0 brought some improvements in ORM, where transactions are not necessarily written directly in the DB but left pending in memory to reduce the amount of interactions with the DB by regrouping updates.
Therefore, anytime we _bypass_ the ORM (i.e. executing SQL through DB cursor), we should make sure that the values still in the ORM cache are reflected in the DB or we risk data inconsistencies, and that's why Odoo provides `flush` method.

The explanations from Odoo experience seem to confirm my assumption: https://youtu.be/EBvuvyIJ1d0 (around 24-25 mins)
Too bad Odoo doesn't update its documentation properly, and this wasn't described in OCA migration guidelines neither....

gurneyalex: so yes, you are right, I read too fast (flush **before** makng the query)

hparfr:
@grindtildeath @rousseldenis @gurneyalex  ok for the merge ?

rousseldenis (approved):

PR #1052, opened 2021-01-11
[14.0][MIG][WIP] stock_picking_completion_info

hparfr:
Seems functional, tests need to be fixed


guewen:
Hi @hparfr, @jgrandguillaume would like to move this module to OCA/wms when migrating to 14.0, could you propose it there please? 
I added a remark on this issue on modules which have to be moved to OCA/wms: https://github.com/OCA/stock-logistics-warehouse/issues/981

Thanks!

PR #1053, opened 2021-01-12
[13.0][IMP] stock_secondary_unit: Use secondary units directly in pickings

sergio-teruel:
[IMP] stock_secondary_unit: Use product_secondary_unit mixin model from product-attribute repository.

cc @Tecnativa TT27363

### DIFF
--- a/stock_secondary_unit/models/stock_move.py
+++ b/stock_secondary_unit/models/stock_move.py
@@ -4,31 +4,45 @@ from odoo import api, fields, models
 from odoo.tools.float_utils import float_round
 
 
-class StockSecondaryUnitMixin(models.AbstractModel):
-    _name = "stock.secondary.unit.mixin"
-    _description = "Stock Secondary Unit Mixin"
-
-    secondary_uom_id = fields.Many2one(
-        comodel_name="product.secondary.unit", string="Second unit"
-    )
-    secondary_uom_qty = fields.Float(
-        string="Secondary Qty", digits="Product Unit of Measure"
-    )
-
-
 class StockMove(models.Model):
-    _inherit = ["stock.move", "stock.secondary.unit.mixin"]
+    _inherit = ["stock.move", "product.secondary.unit.mixin"]
     _name = "stock.move"
+    _secondary_unit_fields = {
+        "qty_field": "product_uom_qty",
+        "uom_field": "product_uom",
+    }
+
+    product_uom_qty = fields.Float(
+        store=True, readonly=False, compute="_compute_product_uom_qty", copy=True
+    )
+
+    @api.depends("secondary_uom_qty", "secondary_uom_id")
+    def _compute_product_uom_qty(self):
+        self._compute_helper_target_field_qty()
+
+    @api.onchange("product_uom")
+    def onchange_product_uom_for_secondary(self):
+        self._onchange_helper_product_uom_for_secondary()
 
     def _merge_moves_fields(self):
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
LGTM :+1:

stock_secondary_unit/models/stock_move.py
```diff
@@ -4,31 +4,49 @@
...
 
     def _merge_moves_fields(self):
         res = super()._merge_moves_fields()
-        res["secondary_uom_qty"] = self[-1:].secondary_uom_qty
+        res["secondary_uom_qty"] = sum(self.mapped("secondary_uom_qty"))
         return res
 
+    @api.model
+    def _prepare_merge_moves_distinct_fields(self):
+        # Not merge moves with distinct secondary units
```
pedrobaeza: You should put this as a docstring:
```suggestion
        """Don't merge moves with distinct secondary units."""
```

stock_secondary_unit/models/stock_move.py
```diff
@@ -4,31 +4,49 @@
...
+        return distinct_fields
+
 
 class StockMoveLine(models.Model):
-    _inherit = ["stock.move.line", "stock.secondary.unit.mixin"]
+    _inherit = ["stock.move.line", "product.secondary.unit.mixin"]
     _name = "stock.move.line"
+    _secondary_unit_fields = {"qty_field": "qty_done", "uom_field": "product_uom_id"}
+
+    secondary_uom_qty = fields.Float()
```
pedrobaeza: Why do you need to declare these fields?

stock_secondary_unit/models/stock_move.py
```diff
@@ -4,31 +4,49 @@
...
+    def _prepare_merge_moves_distinct_fields(self):
+        # Not merge moves with distinct secondary units
+        distinct_fields = super()._prepare_merge_moves_distinct_fields()
+        distinct_fields += ["secondary_uom_id"]
+        return distinct_fields
+
 
 class StockMoveLine(models.Model):
-    _inherit = ["stock.move.line", "stock.secondary.unit.mixin"]
+    _inherit = ["stock.move.line", "product.secondary.unit.mixin"]
```
pedrobaeza, line 41: UoM is not editable in these lines so no need of onchange?

stock_secondary_unit/models/stock_move.py
sergio-teruel: Good question.... yes... yes....
Due to compute order fields....

stock_secondary_unit/models/stock_move.py
sergio-teruel: Yes... but you can create a new stock move lines directly...

sergio-teruel:
Changes done

stock_secondary_unit/models/stock_move.py
pedrobaeza: I'm not understanding you, but any explanation should be put here as comment for later references.

stock_secondary_unit/models/stock_move.py
pedrobaeza: I also don't understand you. Is it needed or not? If so, I don't see it in your current code.

stock_secondary_unit/models/stock_move.py
sergio-teruel: Solved with this PR https://github.com/OCA/product-attribute/pull/773

PR #1055, opened 2021-01-13
[12.0][IMP] stock_quant_manual_assign: make qty_done fill optional

yostashiro:
There are cases where auto-filling of qty_done of stock move line is not desirable.
e.g. you assign quants manually for some of the moves in a picking and not the others,
in such case you need to go over all the moves in the picking to either remove qty_done
or fill it in to proceed with the validation of the entire moves. Auto-fill behavior is
also troublesome when this function is used in a manufacturing order. i.e. having
qty_done of the component move live messes up the outcome of the production.

This PR makes the auto-fill of qty_done optional per picking type.

Related: https://github.com/OCA/manufacture/pull/586

### DIFF
--- a/stock_quant_manual_assign/models/stock_picking_type.py
+++ b/stock_quant_manual_assign/models/stock_picking_type.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    auto_fill_qty_done = fields.Boolean(
+        "Auto-fill Quantity Done",
+        help="Select this in case done quantity of the stock move line should "
+             "be auto-filled when quants are manually assigned."
+    )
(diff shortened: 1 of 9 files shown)


yostashiro:
Will follow up and update the tests.

stock_quant_manual_assign/models/stock_picking_type.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+class StockPickingType(models.Model):
```
TimLai125, line 7: Flake8 is complaining there should be one more blank line here.
`stock_quant_manual_assign/models/stock_picking_type.py:6:1: E302 expected 2 blank lines, found 1`
```suggestion

class StockPickingType(models.Model):
```

TimLai125 (changes requested):

rousseldenis (approved):
Code review. LGTM

stock_quant_manual_assign/models/stock_picking_type.py
yostashiro: @TimLai125 Fixed it!

marcelsavegnago (approved):
LGTM.

TimLai125 (approved):
LGTM!

PR #1056, opened 2021-01-19
[FIX] stock_mts_mto_rule: Installation across companies

PabloPaezSheridan:
This change make the module to install  across all companies
 not only company we are on in the moment of the installation


rousseldenis (approved):
Makes sense

marcelsavegnago (approved):
LGTM

PR #1062, opened 2021-01-25
[13.0][IMP] stock_vertical_lift: Add skip operation to vertical lift shuttles

carlosserratoro:
In the screen for the vertical lift shuttles, accessible through
Inventory > Operations > Vertical Lift Shuttles, a new button has been
added to allow to skip an operation. This button can also be triggered
by scanning the barcode O-BTN.skip.svg that is inside the folder
'images'.

When a skip is done, the next stock.move.line to pick is chosen and
shown to the operator. A skipped move line is added to the end of the
list of pending move lines to be picked, so they will be shown again
in the future as soon as the other move lines have been successfully
processed.

This option was added because, sometimes, the operator can not process
a move line for whatever reason. Right now, the only way of proceeding
is to wait until he/she can effectively process it, which involves a
delay in the operations. With this new skip operation, the operator
can continue processing the rest of move lines.

### DIFF
--- a/stock_vertical_lift/models/stock_move_line.py
+++ b/stock_vertical_lift/models/stock_move_line.py
@@ -1,11 +1,19 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import fields, models
 
 
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
+
+    vertical_lift_skipped = fields.Boolean(
+        "Skipped in Vertical Lift?",
+        default=False,
+        help="If this flag is set, it means that when the move "
+        "was being processed in the Vertical Lift, the operator decided to "
+        "skip its processing.",
+    )
 
     def fetch_vertical_lift_tray_source(self):
         self.ensure_one()
--- a/stock_vertical_lift/models/vertical_lift_operation_pick.py
+++ b/stock_vertical_lift/models/vertical_lift_operation_pick.py
@@ -17,6 +17,7 @@ class VerticalLiftOperationPick(models.Model):
             ("scan_destination", "Scan New Destination Location"),
             ("save", "Pick goods and save"),
             ("release", "Release"),
+            ("skip", "Skip"),  # Virtual state.
         ]
 
     def _transitions(self):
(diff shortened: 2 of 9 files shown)


stock_vertical_lift/models/vertical_lift_operation_pick.py
```diff
@@ -70,10 +85,15 @@ def fetch_tray(self):
     def select_next_move_line(self):
         self.ensure_one()
         next_move_line = self.env["stock.move.line"].search(
-            self._domain_move_lines_to_do(), limit=1
+            self._domain_move_lines_to_do(), limit=1, order="vertical_lift_pick_skipped"
```
guewen: I think we still have to ensure the original order is kept, maybe by concatenating `vertical_lift_pick_skipped` and `self._order`?

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -287,6 +287,12 @@ def process_current(self):
         """
         raise NotImplementedError
 
+    def button_skip(self):
```
guewen: Is the skip button going to be implemented for Put/Inventory? If not, the `button_skip` button better has to be in vertical_lift_operation_pick.py only

stock_vertical_lift/models/stock_move_line.py
```diff
@@ -1,12 +1,21 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import fields, models
 
 
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
+    vertical_lift_pick_skipped = fields.Boolean(
```
guewen: Is it going to be implemented for the Put screen? I think `vertical_lift_skipped` is enough, because in any case the move line can't be both put and pick, so the field can be reused if ever we add skip on putaway lines

stock_vertical_lift/models/vertical_lift_operation_pick.py
```diff
@@ -17,22 +17,37 @@ def _selection_states(self):
...
             self.Transition(
                 "release", "scan_destination", lambda self: self.select_next_move_line()
             ),
             self.Transition("release", "noop"),
+            self.Transition(
+                "skip",
+                "scan_destination",
+                lambda self: self.select_next_move_line(),
+                direct_eval=True,
+            ),
```
guewen, line 43: Shouldn't you add `self.Transition("skip", "noop", direct_eval=True),` in case there is no more move line? When you skip, it should in normal case not happen as if you skip the last line, the same line will be re-opened, but if for any reason the line was the last, you skip it and for some reason it's no longer to be picked (deleted? done by someone else?), it returns to `noop`.

stock_vertical_lift/models/stock_move_line.py
carlosserratoro: It's only for the Pick.

stock_vertical_lift/models/vertical_lift_operation_base.py
carlosserratoro: OK, it's now a method of the child class, ```vertical.lift.operation.pick```.

stock_vertical_lift/models/vertical_lift_operation_pick.py
carlosserratoro: Done.

guewen (approved):
Thanks!

simahawk (approved):

PR #1063, opened 2021-01-25
[13.0][ADD] stock_location_last_inventory_date

carlosserratoro:
This new module adds a stored computed field on locations that stores
the last inventory date for the location, for validated inventories.
This is only computed for leaf locations, not parent ones.


stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+        """Store date of the last inventory for each leaf location"""
+        for loc in self:
+            if (
+                loc.usage != "view"
+                and not loc.child_ids
+                and loc.validated_inventory_ids
+            ):
+                loc.last_inventory_date = loc.validated_inventory_ids.sorted(
+                    lambda inventory: inventory.write_date
+                )[-1].write_date
```
guewen: Is there a reason to use `write_date` and not `date`? `write_date` could hypothetically change even on a very old inventory if we call `write()` on it

carlosserratoro: This was a mistake. There is no need to not use ```date```. I'll force-push a correction on this. Thanks!

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+        "stock.inventory",
+        relation="stock_inventory_stock_location_rel",
+        column1="stock_location_id",
+        column2="stock_inventory_id",
+        string="Stock Inventories",
+        help="Stock inventories in state validated for this location.",
+        domain="[('location_ids', 'in', id), ('state', '=', 'done')]",
+    )
+
+    @api.depends("validated_inventory_ids", "validated_inventory_ids.write_date")
```
guewen:
```suggestion
    @api.depends("validated_inventory_ids", "validated_inventory_ids.date")
```

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+    def _compute_last_inventory_date(self):
+        """Store date of the last inventory for each leaf location"""
+        for loc in self:
+            if (
+                loc.usage != "view"
+                and not loc.child_ids
+                and loc.validated_inventory_ids
+            ):
+                loc.last_inventory_date = loc.validated_inventory_ids.sorted(
+                    lambda inventory: inventory.write_date
```
guewen:
```suggestion
                    lambda inventory: inventory.date
```

guewen (changes requested):
Leftovers of `write_date` to replace by `date`

ajaniszewska-dev (approved):

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,44 @@
...
+        "stock.inventory",
+        relation="stock_inventory_stock_location_rel",
+        column1="stock_location_id",
+        column2="stock_inventory_id",
+        string="Stock Inventories",
+        help="Stock inventories in state validated for this location.",
+        domain="[('location_ids', 'in', id), ('state', '=', 'done')]",
+    )
+
+    @api.depends("validated_inventory_ids", "validated_inventory_ids.date")
```
sebalix: I would add `usage` and `child_ids` to the dependencies

carlosserratoro: You are right. Thanks! ─ done.

stock_location_last_inventory_date/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Location Last Inventory Date",
+    "summary": "Show the last inventory date for a leaf location",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Warehouse Management",
```
rousseldenis: Please set to 'Warehouse' to align https://odoo-community.org/shop categories (Odoo standard ones)

rousseldenis (approved):
LGTM. Minor comment

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,49 @@
...
+        column2="stock_inventory_id",
+        string="Stock Inventories",
+        help="Stock inventories in state validated for this location.",
+        domain="[('location_ids', 'in', id), ('state', '=', 'done')]",
+    )
+
+    @api.depends(
+        "validated_inventory_ids",
+        "validated_inventory_ids.date",
+        "validated_inventory_ids.usage",
```
rousseldenis:
```suggestion
        "validated_inventory_ids.location_id.usage",
```

rousseldenis (changes requested):
LGTM. Minor comment

guewen:
@carlosserratoro did you try with a normal stock user? I wonder if you won't have an access right issue without `compute_sudo` on the field.

stock_location_last_inventory_date/models/stock_location.py
carlosserratoro: Wait, I'm realising this depends is wrong also because of the child_ids, this should be added also: ```validated_inventory_ids.location_id.child_ids```.

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -0,0 +1,49 @@
...
+        "validated_inventory_ids.child_ids",
```
carlosserratoro:
```suggestion
        "validated_inventory_ids.location_id.child_ids",
```

carlosserratoro:
> @carlosserratoro did you try with a normal stock user? I wonder if you won't have an access right issue without `compute_sudo` on the field.

@guewen I tested using unit tests. For indirect recomputations (for instance those that happen when one of the fields in the ```depends``` is changed) it works fine without ```compute_sudo``` because a special user *Sytem* does it. However a call to the compute method with a hardcoded ```.with_user()``` has precedence over the ```compute_sudo```, so it fails, but I think this applies to any computed field in Odoo. To be extra-sure, I can add the ```compute_sudo``` **and** use ```sudo()``` inside the method.

For the moment I've added a new test having to do with permissions (without using ```sudo()```) Can you have a look again, please?

stock_location_last_inventory_date/__manifest__.py
carlosserratoro: Done.

stock_location_last_inventory_date/models/stock_location.py
carlosserratoro: I re-wrote the list of depends, there were a couple missing.

guewen:
@carlosserratoro That's good without adding `compute_sudo` or `sudo()` then. I don't think we should expect users able to directly call the compute method. Thanks for checking :)

guewen (approved):

rousseldenis (approved):

PR #1067, opened 2021-01-28
[13.0] stock_move_auto_assign: Fix concurrency race condition on picking state

guewen:
When the 2 last moves of a stock.picking are assigned at the same
time by 2 jobs (different products both available), none of the
transaction will see that it is the last move to be assigned. As a
result, the picking will stay in state "confirmed" even if all its
moves are assigned.

Lock the stock.picking records when we call auto_assign.
If we have many pickings touched for the same product, the lock
can be quite large, so we may have to find a better option.

### DIFF
--- a/stock_move_auto_assign/models/product_product.py
+++ b/stock_move_auto_assign/models/product_product.py
@@ -8,7 +8,6 @@ import psycopg2
 from odoo import models
 
 from odoo.addons.queue_job.exception import RetryableJobError
-from odoo.addons.queue_job.job import job
 
 _logger = logging.getLogger(__name__)
 
@@ -32,8 +31,6 @@ class ProductProduct(models.Model):
             # out anyway.
         ]
 
-    # TODO remove decorator
-    @job(default_channel="root.stock_auto_assign")
     def moves_auto_assign(self, locations):
         """Job trying to reserve moves based on product and locations
 
--- a/stock_move_auto_assign/models/product_product.py
+++ b/stock_move_auto_assign/models/product_product.py
@@ -1,9 +1,16 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+import logging
+
+import psycopg2
+
 from odoo import models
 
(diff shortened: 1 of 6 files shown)


stock_move_auto_assign/models/product_product.py
```diff
@@ -38,4 +43,23 @@ def moves_auto_assign(self, locations):
         """
         self.ensure_one()
         moves = self.env["stock.move"].search(self._moves_auto_assign_domain(locations))
+        pickings = moves.picking_id
+        if not pickings:
+            return
+        try:
+            self.env.cr.execute(
```
rousseldenis, line 50: Is it really wanted ??? Shouldn't be a retryable job ?

guewen: This will retry the job. The problem here is there's no conflict as if 2 jobs change the state of 2 different moves, none of them will update their common related picking, no concurrent write.

Each of the job will update their move, at the same time, so in both transactions, they'll have their respective move as "assigned" and the other one as "confirmed". So none will update the picking's state to "assigned".

I'm far to be convinced this is a good solution, but I have no other for now.

guewen: Another option could be to use `move.picking_id.write({})` on the stock.picking records were at least one move became assigned, that would update `write_date` on the stock.picking, so in case of concurrent move, one would retry, and we don't lock the stock.picking. But it means we have more code executed before the rollback happens.

guewen: @jbaudoux do you have an opinion on the way to solve this issue?

gurneyalex (approved):

jgrandguillaume (approved):
For my opinion having a large lock is less harmful than a wrong state of the picking. Plus, the lock may concerns many picking in such case I agree, but should not last for long right ? I mean the lock will be a very short duration. 

So +1 for me

jbaudoux (approved):

stock_move_auto_assign/models/product_product.py
jbaudoux: @guewen I like the explicit lock as you did.

PR #1068, opened 2021-01-29
[12.0] [FIX] stock_archive_constraint: Test bug product_uom_qty is a read-only field

CarlosRoca13:
cc @Tecnativa TT27886

ping @victoralmau @carlosdauden

### DIFF
--- a/stock_archive_constraint/tests/test_location_archive_constraint.py
+++ b/stock_archive_constraint/tests/test_location_archive_constraint.py
@@ -60,7 +60,6 @@ class TestLocationArchiveConstraint(SavepointCase):
         stock_move_line_form.location_id = location_id
         stock_move_line_form.location_dest_id = location_dest_id
         stock_move_line_form.product_id = product_id
-        stock_move_line_form.product_uom_qty = qty
         stock_move_line_form.qty_done = qty
         stock_move_line_form.state = 'done'
         stock_move_line_form.save()


CarlosRoca13:
Apply this change in a stock.move.line have no sense because the field is readonly 
image

carlosdauden (approved):
https://github.com/OCA/OCB/blob/52901f6aa97a47e8cef89695fcbab1fbaee1081d/addons/stock/views/stock_picking_views.xml#L296

stock_archive_constraint/tests/test_location_archive_constraint.py
```diff
@@ -60,7 +60,6 @@ def _create_stock_move_line(self, location_id, location_dest_id, product_id, qty
         stock_move_line_form.location_id = location_id
         stock_move_line_form.location_dest_id = location_dest_id
         stock_move_line_form.product_id = product_id
-        stock_move_line_form.product_uom_qty = qty
```
pedrobaeza, line 63: I think it should be:
```python
stock_move_line_form.product_qty = qty
```
although it can be the same without assigning such value.

carlosdauden: https://github.com/OCA/OCB/blob/52901f6aa97a47e8cef89695fcbab1fbaee1081d/addons/stock/models/stock_move_line.py#L77-L82

and this field is not defined in the picking view:
https://github.com/OCA/OCB/blob/52901f6aa97a47e8cef89695fcbab1fbaee1081d/addons/stock/views/stock_picking_views.xml#L282-L299

pedrobaeza: OK then

pedrobaeza (approved):

HaraldPanten (approved):

PR #1073, opened 2021-02-09
[13.0] Add stock_helper: share common code used by stock modules

guewen:
The first method is: StockLocation.is_sublocation_of()

This method is currently used in:

* wms/stock_dynamic_routing
* wms/shopfloor
* wms/stock_move_source_relocate
* stock-logistics-warehouse/stock_reserve_rule
* ddmrp/ddmrp

The goal will be to use this module as dependency instead of
reimplementing the method in each. stock_reserve_rule is done here.

Other methods should follow in "stock_helper".

Note: I opened odoo/odoo#53866 to propose a
generic version of this method, expecting odoo's opinion, but got no
answer.


guewen:
wms repo updates: https://github.com/OCA/wms/pull/165

etobella (approved):
LGTM Code review

hparfr:
I don't like the name "stock_helper". Too generic.
What do you think to rename it stock_location_tree_helper, stock_location_navigation, stock_location_namingishard? 

It can be merged with the module stock_location_children  https://github.com/OCA/stock-logistics-warehouse/pull/855 ?

Having more common function to traverse graph of objects like locations, or categories, BOM, ... is beneficial and can limit slow implementations.

guewen:
Stock location children adds a field, this should be a module that only adds methods / doesn't change the data model.

I don't like the too generic name either, but it is meant to be generic at the same time.

@sebalix has some methods to add related to stock moves and maybe others models to add.

And then, if the module is too specific to a model or method, someone will complain it is too specific... 😉

stock_helper/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020-2021 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Stock Move Helpers",
```
simahawk:
```suggestion
    "name": "Stock Helpers",
```

stock_helper/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "summary": "Add methods shared between various stock modules",
+    "version": "13.0.1.1.0",
```
simahawk:
```suggestion
    "version": "13.0.1.0.0",
```

simahawk (approved):

stock_helper/__manifest__.py
guewen: Why did I wrote "Move" here? :shrug:

PR #1075, opened 2021-02-10
[13.0][FIX] stock_cubiscan: Default packaging quantity

rousseldenis:


rousseldenis:
@jgrandguillaume @grindtildeath It seems weird to create default packaging with qty = 0.

I've problem with packaging_uom module with that. It seems your tests are running fine with this change.

grindtildeath:
@rousseldenis I don't have the context here, but I poked @jgrandguillaume on our internal chat :wink:

jgrandguillaume (approved):
@rousseldenis I think it's ok. We did that to easily spot the lines that needs info (0 means not fulfilled). But If you prefer to set to 1 let's do this.

grindtildeath (approved):

PR #1079, opened 2021-02-10
[MIG] stock_location_tray: Migration to 14.0

phuctranfxvn:
Standard migration, no new feature added


rousseldenis:
@phuctranerp Could you add tests to cover missing parts? Thanks

hailangvn:
Thanks for your comment. I will do it.

hailangvn:
Good day @rousseldenis, the PR has been updated. Thanks.

stock_location_tray/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Location Trays",
     "summary": "Organize a location as a matrix of cells",
-    "version": "13.0.1.0.2",
+    "version": "14.0.1.0.2",
```
sebalix: Version number to reset to `14.0.1.0.0`

stock_location_tray/models/stock_location.py
```diff
@@ -128,34 +128,6 @@ def write(self, vals):
                 location.child_ids.write({"posz": vals["posz"]})
         return True
 
-    @api.constrains("active")
-    def _tray_check_active(self):
```
sebalix: Why this is considered dead code? As we still have this on 13.0, I try to understand the reason to remove it on 14.0, and maybe backporting it on 13.0 if it is relevant?
The commit message mentions https://github.com/OCA/stock-logistics-warehouse/pull/987 but I'm not sure to understand the link.

hailangvn: There is change in odoo/odoo for both 13.0 and 14.0 at https://github.com/odoo/odoo/commit/b6ba6038ad8f419fb5ba805c10c987eb3ff93530#diff-0b83f13eeb39606e0cc2ebe1443b40c01c19bf49ec16ab100dbcb0d0f4c9e4b6R99-R106. With the change, `UserError(_('You still have some product in locations %s')` is always risen before the two `ValidationError(_("Tray locations cannot be archived when they contain products.")` and `UserError(_("Trays cannot be modified when they contain products.")` of this module. Therefor, the #987 changes tests to meet that flow change.

To prove the dead code, we have two hacks: change context to bypass odoo/odoo checking as https://github.com/OCA/stock-logistics-warehouse/pull/1079/commits/79fa70556a01f644986d92067a46315278830b8b#diff-ce46c42b625b31fbfc9cd2646ccfe9d647a894001a331cd54d3ce0cf04b466a3R161-R163 and  https://github.com/OCA/stock-logistics-warehouse/pull/1079/commits/79fa70556a01f644986d92067a46315278830b8b#diff-ce46c42b625b31fbfc9cd2646ccfe9d647a894001a331cd54d3ce0cf04b466a3R183-R187. We believe that in daily use, there is no hack like that.

sebalix (approved):

stock_location_tray/models/stock_location.py
sebalix: Thank you for the explanation (and for adding it to the commit message!)

TDu (approved):

PR #1090, opened 2021-02-23
[MIG] stock_vertical_lift: Migration to 14.0

dzungtran89:
Dependencies:
- stock_location_tray: https://github.com/OCA/stock-logistics-warehouse/pull/1079
- web_ir_actions_act_view_reload: https://github.com/OCA/web/pull/1816
- web_notify: https://github.com/OCA/web/pull/1824


stock_vertical_lift/models/vertical_lift_shuttle.py
```diff
@@ -0,0 +1,276 @@
...
+        event occurred to inform the user on their screen.
+
+        The method is private only to prevent xml/rpc calls to
+        interact with the screen.
+        """
+        # XXX do we want to do something special in the notification?
+        self._operation_for_mode()._send_notification_refresh()
+
+
+class VerticalLiftShuttleManualBarcode(models.TransientModel):
```
rousseldenis: Could you maybe split one model per file ?

mmequignon: ping @dzungtran89

hailangvn: Good day @rousseldenis, @mmequignon, the model has been split to https://github.com/OCA/stock-logistics-warehouse/blob/8ba7d977a5f513f934c6d23ce47f59a304972a5c/stock_vertical_lift/wizards/vertical_lift_shuttle_manual_barcode.py. This PR has been rebased and updated with latest test requirement. Thanks for your comments.

hailangvn: I am checking the travis...

hailangvn:
Good day @rousseldenis, @mmequignon, the travis and runbot are green now.

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -0,0 +1,502 @@
...
+
+def split_other_move_lines(move, move_lines):
+    """Substract `move_lines` from `move.move_line_ids`, put the result
+    in a new move and returns it.
+    """
+    other_move_lines = move.move_line_ids - move_lines
+    if other_move_lines or move.state == "partially_available":
+        qty_to_split = move.product_uom_qty - sum(move_lines.mapped("product_uom_qty"))
+        backorder_move_vals = move._split(qty_to_split)
+        backorder_move = move.create(backorder_move_vals)
```
sebalix, line 27: You have to confirm the new move:
```suggestion
        backorder_move = move.create(backorder_move_vals)
        backorder_move._action_confirm(merge=False)
```
See https://github.com/OCA/wms/blob/b0ceac26303ef8278d641fb050399d3c965ecc6c/shopfloor/models/stock_move.py#L41-L43

hailangvn:
Good day @sebalix, thanks for your help. The PR has been updated.

simahawk:
@hailangvn can you check the build?

hailangvn:
> @hailangvn can you check the build?

@simahawk, thanks for your approval. The PR has been rebased to latest 14.0 and updated to remove test dependency.

mmequignon (approved):

stock_vertical_lift/models/vertical_lift_operation_base.py
simahawk: @hailangvn what about this?  Also, are we sure we always get values? We might have an empty dict AFAIR. Can you check?

stock_vertical_lift/models/vertical_lift_command.py
```diff
@@ -0,0 +1,52 @@
...
+    command = fields.Char(required=True)
+    answer = fields.Char()
+    error = fields.Char()
+    shuttle_id = fields.Many2one("vertical.lift.shuttle", required=True)
+
+    def record_answer(self, answer):
+        name = self._get_key(answer)
+        record = self.search([("name", "=", name)], limit=1)
+        if not record:
+            if not getattr(threading.currentThread(), "testing", False):
```
simahawk: do we really need this?

hailangvn: Thanks @simahawk for your review. I will work on this.
> @hailangvn what about this? Also, are we sure we always get values? We might have an empty dict AFAIR. Can you check?

hailangvn:
Good day @simahawk, this is the investigation.
> @hailangvn what about this? Also, are we sure we always get values? We might have an empty dict AFAIR. Can you check?
* https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L1599 shows that empty response only happens when (a) split qty is 0, (b) split qty >= stock quantity
(a) tested with split qty = 0: no operation is created for vertical lift shuttle so cannot run to the code.
(b.1) testing with split qty = stock quantity = 30: 3 operations. Scanning and saving each operation had no issue. Last one has no other_move_lines so the _split was not run.
(b.2) testing with split qty = 35 > stock quantity = 30: 3 operations. Last one had two lines with qty 10 and 5 correspondingly. The _split result was not empty.
* To be in safe side, a check is put in place to process when backorder is not empty.

hailangvn:
We found that tests of `stock_location_tray` were failed because of `stock_vertical_lift`. So we need to add a fix.

sebalix (approved):
LG, need to squash the last commits into the `[MIG] stock_vertical_lift: Migration to 14.0` one, I pre-approve

hailangvn:
Thanks @sebalix for the comment. The PR has been updated.

simahawk (approved):

PR #1094, opened 2021-02-24
13.0. stock_warehouse_orderpoint_stock_info: New filters under minumum and …

manuelcalerosolis:
Add two new filters in stock warehouse stock, products over maximum and under minimum.

### DIFF
--- a/stock_warehouse_orderpoint_stock_info/models/stock_warehouse_orderpoint.py
+++ b/stock_warehouse_orderpoint_stock_info/models/stock_warehouse_orderpoint.py
@@ -29,16 +29,26 @@ class StockWarehouseOrderpoint(models.Model):
     product_category = fields.Many2one(
         string="Product Category", related="product_id.categ_id", store=True
     )
+    product_under_minimum = fields.Boolean(
+        string="Product Under Minimum",
+        compute="_compute_product_available_qty",
+        search="_search_product_under_minimum",
+    )
+    product_over_maximum = fields.Boolean(
+        string="Product Over Maximum",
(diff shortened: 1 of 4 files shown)


rousseldenis:
@manuelcalerosolis Could you add few tests for searches ?

manuelcalerosolis:
> Could you add few tests for searches ?

@rousseldenis Can you tell me where I can find tests for the filters in other oca modules? To inspire me, I have never done them.

stock_warehouse_orderpoint_stock_info/models/stock_warehouse_orderpoint.py
```diff
@@ -29,30 +29,60 @@ class StockWarehouseOrderpoint(models.Model):
...
+                        "product_under_minimum": product["qty_available"]
+                        < order.product_min_qty,
+                        "product_over_maximum": product["qty_available"]
+                        > order.product_max_qty,
                     }
                 )
+
+    def _search_product_under_minimum(self, operator, value):
+        if operator == "=":
+            recs = self.search([]).filtered(lambda x: x.product_under_minimum is True)
```
rousseldenis, line 76: @manuelcalerosolis Write a test that searches for product with the field ('product_under_minimum', '=', True).

Moreover, IMHO this is a little bulky as it result of two searches (one with all products, then a search with ids).

manuelcalerosolis: @rousseldenis certainly the search is very bad, currently the module is not being used in version 13.0 I will prepare it for version 15.0

PR #1095, opened 2021-02-24
[13.0][IMP] stock_move_location: Take into account owner for move quants

sergio-teruel:
cc @Tecnativa TT28438

### DIFF
--- a/stock_move_location/i18n/es.po
+++ b/stock_move_location/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 11.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-11-02 20:30+0000\n"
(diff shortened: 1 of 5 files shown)


rousseldenis:
Hi @sergio-teruel Thanks for this. Could you add a test and remove .po modifications ?

chienandalu (approved):
Ok, although tests would be welcome indeed

Translation changes maybe would fit better in its own commit

pedrobaeza:
@carlosdauden can you confirm this?

carlosdauden (approved):
Please review @rousseldenis comments

pedrobaeza:
The changes on the language files are because there are new terms that are translated in this PR. I have checked that there's no pending translating to push and no conflict in https://translation.odoo-community.org/projects/stock-logistics-warehouse-13-0/stock-logistics-warehouse-13-0-stock_move_location/es/#repository

About the test, it's always desirable, but this is more a fix than a imp, and we need it on the repo. Later the regression test can be added:

/ocabot merge patch

PR #1101, opened 2021-03-04
[13.0] Fix stock_vertical_lift skip screen

santostelmo:
In vertical lift we have a button to skip screens. When we reach the last screen, we must be able to move to the first one.

### DIFF
--- a/stock_vertical_lift/models/vertical_lift_operation_pick.py
+++ b/stock_vertical_lift/models/vertical_lift_operation_pick.py
@@ -1,6 +1,8 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from itertools import cycle
+
 from odoo import _, models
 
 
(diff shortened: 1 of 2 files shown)


stock_vertical_lift/models/vertical_lift_operation_pick.py
```diff
@@ -83,15 +85,30 @@ def _domain_move_lines_to_do_all(self):
     def fetch_tray(self):
         self.current_move_line_id.fetch_vertical_lift_tray_source()
 
+    def get_next_move_line(self, order):
```
sebalix:
```suggestion
    def _get_next_move_line(self, order):
```
As this method can't be used by the RPC layer (object returned can't be serialized).

stock_vertical_lift/models/vertical_lift_operation_pick.py
```diff
@@ -83,15 +85,30 @@ def _domain_move_lines_to_do_all(self):
...
+        def get_next(move_lines, current_move):
+            if not move_lines:
+                return False
+            move_lines_cycle = cycle(move_lines)
+            if not current_move:
+                return next(move_lines_cycle)
+            # Point to current_move and then return the next
+            while next(move_lines_cycle) != current_move:
+                continue
+            return next(move_lines_cycle)
```
sebalix: Nitpick, to avoid confusion between `stock.move` and `stock.move.line`
```suggestion
        def get_next(move_lines, current_move_line):
            if not move_lines:
                return False
            move_lines_cycle = cycle(move_lines)
            if not current_move_line:
                return next(move_lines_cycle)
            # Point to current_move_line and then return the next
            while next(move_lines_cycle) != current_move_line:
                continue
            return next(move_lines_cycle)
```

stock_vertical_lift/models/vertical_lift_operation_pick.py
santostelmo: Done

stock_vertical_lift/models/vertical_lift_operation_pick.py
santostelmo: Done

sebalix (approved):

gurneyalex (approved):

simahawk:
@santostelmo please squash :pray:

santostelmo:
@simahawk squashed

PR #1106, opened 2021-03-11
[13.0][ADD] stock_vertical_lift_empty_tray_check: is the tray empty?

carlosserratoro:
A vertical lift retrieves a tray and places it in front of the
user, and depending on the quantity the user takes from it,
it adapts the pending quantity in the tray. However, because of
errors, it could be that the system thinks the tray is empty
while it is not. With this module, when the system things the
tray is empty, while in the step for the release of the tray,
the operator is asked explicitly to check if the tray is
actually empty. Depending on his/her answer (yes/no) an inventory
adjustment is created stating the situation.

To activate this optional feature, a new configuration setting
has been added to Inventory > Configuration > Settings, named
'Check Empty Tray'. It is deactivated by default.

Developing decisions:
- The screens shown to the operator are actually wizards, but
  since in the original module, `stock_vertical_lift`, they
  were considered on the source tree as views, this has been
  continued here.
- It has been decided, to not change the current workflow of
  the operators, to embed the new check inside the step for
  the 'release'. So, a new screen appears to ask for the
  visual inspection of whether the tray is empty or not. In
  order to test this easily, the method `button_release` of
  the module `stock_vertical_lift` has been slighly modified
  so that it always returns. This is so that we can check
  in the unit-tests for the outcome of the intermediate
  screen (i.e. wizard) in the unit-tests ─ similarly to how
  it is done when validating a picking that can result in
  a backorder.

### DIFF
--- a/stock_vertical_lift_empty_tray_check/models/res_config_settings.py
+++ b/stock_vertical_lift_empty_tray_check/models/res_config_settings.py
@@ -1,0 +1,11 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    vertical_lift_empty_tray_check = fields.Boolean(
+        "Vertical lift: Check Empty Tray",
+        default=False,
+        config_parameter="vertical_lift_empty_tray_check",
+    )
--- a/stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick.py
+++ b/stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick.py
@@ -1,0 +1,46 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, models
+from odoo.tools import float_is_zero
+
+
+class VerticalLiftOperationPick(models.Model):
+    _inherit = "vertical.lift.operation.pick"
+
+    def button_release(self):
+        """Release the operation, go to the next
+
+        By default it asks the user to inspect visually if the tray is empty.
+        """
+        icp = self.env["ir.config_parameter"].sudo()
+        tray_check = icp.get_param("vertical_lift_empty_tray_check")
+        skip_zero_quantity_check = self.env.context.get("skip_zero_quantity_check")
+        if not skip_zero_quantity_check and tray_check:
+            uom_rounding = self.product_id.uom_id.rounding
+            if float_is_zero(self.tray_qty, precision_rounding=uom_rounding):
+                return self._check_zero_quantity()
+
+        return super().button_release()
+
+    def _check_zero_quantity(self):
+        """Show the wizard to check for real-zero quantity."""
+        view = self.env.ref(
+            "stock_vertical_lift_empty_tray_check."
+            "vertical_lift_operation_pick_zero_check_view_form"
+        )
+        wizard_model = "vertical.lift.operation.pick.zero.check"
+        wizard = self.env[wizard_model].create(
+            {"vertical_lift_operation_pick_id": self.id}
+        )
+        return {
+            "name": _("Is the tray empty?"),
+            "type": "ir.actions.act_window",
+            "view_mode": "form",
+            "target": "new",
+            "views": [(view.id, "form")],
+            "view_id": view.id,
+            "res_model": wizard_model,
+            "res_id": wizard.id,
+            "context": self.env.context,
+        }
--- a/stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick_zero_check.py
+++ b/stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick_zero_check.py
@@ -1,0 +1,87 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import _, fields, models
+
+
+class VerticalLiftOperationPickZeroCheck(models.TransientModel):
+    _name = "vertical.lift.operation.pick.zero.check"
+    _description = "Make sure the tray location is empty"
+
+    vertical_lift_operation_pick_id = fields.Many2one("vertical.lift.operation.pick")
+
+    def _get_data_from_operation(self):
+        """Return picking, location and product from the operation shuttle"""
+        operation = self.vertical_lift_operation_pick_id
+
+        # If the move is split into several move lines, it is
+        # moved to another picking, being a backorder of the
+        # original one. We are always interested in the original
+        # picking that was processed at first, so if the picking
+        # is a backorder of another picking, we take that other one.
+        picking = operation.picking_id.backorder_id or operation.picking_id
+        location = operation.current_move_line_id.location_id
+        product = operation.product_id
+        return operation, picking, location, product
+
+    def button_confirm_empty(self):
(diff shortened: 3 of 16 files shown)


stock_vertical_lift_empty_tray_check/models/res_config_settings.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    vertical_lift_empty_tray_check = fields.Boolean("Check Empty Tray", default=False)
```
simahawk:
```suggestion
    vertical_lift_empty_tray_check = fields.Boolean("Vertical lift: Check Empty Tray", default=False,  config_parameter="vertical_lift_empty_tray_check")
```

stock_vertical_lift_empty_tray_check/models/res_config_settings.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def set_values(self):
```
simahawk: trash these overrides

stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick.py
```diff
@@ -0,0 +1,44 @@
...
+
+class VerticalLiftOperationPick(models.Model):
+    _inherit = "vertical.lift.operation.pick"
+
+    def button_release(self):
+        """Release the operation, go to the next
+
+        By default it asks the user to inspect visually if the tray is empty.
+        """
+        icp = self.env["ir.config_parameter"]
```
simahawk:
```suggestion
        icp = self.env["ir.config_parameter"].sudo()
```
otherwise is broken 99%

stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick_zero_check.py
```diff
@@ -0,0 +1,87 @@
...
+from odoo import _, models
+
+
+class VerticalLiftOperationPickZeroCheck(models.TransientModel):
+    _name = "vertical.lift.operation.pick.zero.check"
+    _description = "Make sure the tray location is empty"
+
+    def _get_data_from_operation(self):
+        """Return picking, location and product from the operation shuttle"""
+        active_id = self.env.context.get("active_id")
```
simahawk: maybe a computed field like `current_move_line` (or `current_move`, not clear what you get here)

stock_vertical_lift_empty_tray_check/models/vertical_lift_operation_pick_zero_check.py
```diff
@@ -0,0 +1,87 @@
...
+        # moved to another picking, being a backorder of the
+        # original one. We are always interested in the original
+        # picking that was processed at first, so if the picking
+        # is a backorder of another picking, we take that other one.
+        picking = operation.picking_id.backorder_id or operation.picking_id
+        location = operation.current_move_line_id.location_id
+        product = operation.product_id
+        return operation, picking, location, product
+
+    def confirm_empty(self):
```
simahawk:
```suggestion
    def [action_|button_]confirm_empty(self):
```

stock_vertical_lift_empty_tray_check/tests/test_pick.py
```diff
@@ -0,0 +1,82 @@
...
+
+class TestPick(VerticalLiftCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.picking_out = cls.env.ref(
+            "stock_vertical_lift.stock_picking_out_demo_vertical_lift_1"
+        )
+        cls.env["ir.config_parameter"].set_param("vertical_lift_empty_tray_check", True)
```
simahawk: Use `sudo` in case the user is changed in base class (normally this tests should run w/ stock user)

simahawk (changes requested):

sebalix (approved):

stock_vertical_lift_empty_tray_check/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_vertical_lift_operation_pick_zero_check_stock_user,access_vertical_lift_operation_pick_zero_check stock user,model_vertical_lift_operation_pick_zero_check,stock.group_stock_user,1,1,1,1
```
sebalix: Runbot is complaining about this, you should remove the file:

    2021-03-11 17:53:03,262 183 WARNING openerp_test odoo.modules.loading: The transient model vertical.lift.operation.pick.zero.check (Make sure the tray location is empty) should not have explicit access rules!

carlosserratoro:
@simahawk thanks for your review. The changes are in https://github.com/OCA/stock-logistics-warehouse/pull/1106/commits/a2c23708e04e4a7f522dd526ecf07f00bd6bf8d6.

simahawk (approved):
pls remove the ACL and squash

stock_vertical_lift_empty_tray_check/security/ir.model.access.csv
carlosserratoro: @sebalix it was removed in the last commit.

stock_vertical_lift_empty_tray_check/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+When a tray is released, and the system things it is empty,
```
TDu:
```suggestion
When a tray is released, and the system thinks it is empty,
```

TDu (approved):

stock_vertical_lift_empty_tray_check/views/res_config_setting_views.xml
```diff
@@ -0,0 +1,31 @@
...
+<odoo>
+    <data>
+        <record id="res_config_settings_view_form" model="ir.ui.view">
+            <field name="name">res.config.settings.view.form.inherit.stock</field>
+            <field name="model">res.config.settings</field>
+            <field name="priority" eval="30" />
+            <field name="inherit_id" ref="stock.res_config_settings_view_form" />
+            <field name="arch" type="xml">
+                <xpath
+                    expr="//div[descendant-or-self::field[@name='group_stock_adv_location'] and @class='col-12 col-lg-6 o_setting_box']"
```
simahawk: I missed this... this select seems weird. What are you trying to select? Also, it causes
`021-03-12 10:15:17,541 185 WARNING openerp_test odoo.addons.base.models.ir_ui_view: Error-prone use of @class in view res.config.settings.view.form.inherit.stock (): use the hasclass(*classes) function to filter elements by their classes`

simahawk: as discussed: use `/ancestor::` to get to the closest setting_box

carlosserratoro: Changed by ```//field[@name='group_stock_adv_location']/ancestor::div[hasclass('o_setting_box')]``` as you suggested. Thanks!

jbaudoux:
Naming is confusing because you are checking if the cell is empty, not the tray. The names should be fixed to remove the confusion.

PR #1105, opened 2021-03-10
[12.0][IMP] stock_request_analytic + stock_request_purchase: Add Analytic Tags and propagate to PO

joao-p-marques:
Needs https://github.com/OCA/account-analytic/pull/366 to add `analytic_tag_ids` field to `stock.move`.

@Tecnativa
TT28488

ping @pedrobaeza @chienandalu

PR #1107, opened 2021-03-12
[12.0][FIX] stock_request_purchase: Check if fields exists previously and prevent error about it

victoralmau:
Check if fields (`analytic_account_id` + `analytic_tag_ids`) exists previously and prevent error about it.

Related to https://github.com/OCA/stock-logistics-warehouse/pull/1105

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa

### DIFF
--- a/stock_request_purchase/models/stock_rule.py
+++ b/stock_request_purchase/models/stock_rule.py
@@ -11,14 +11,17 @@ class StockRule(models.Model):
                                      product_uom, values, po, supplier):
         vals = super(StockRule, self)._prepare_purchase_order_line(
             product_id, product_qty, product_uom, values, po, supplier)
-        if 'stock_request_id' in values:
-            stock_request = self.env["stock.request"].browse(values['stock_request_id'])
-            vals['stock_request_ids'] = [(4, stock_request.id)]
-            if stock_request.analytic_account_id:
-                vals['account_analytic_id'] = stock_request.analytic_account_id.id
-            if stock_request.analytic_tag_ids:
-                vals['analytic_tag_ids'] = [
-                    (4, tag.id) for tag in stock_request.analytic_tag_ids
+        if "stock_request_id" in values:
+            request = self.env["stock.request"].browse(values["stock_request_id"])
+            vals["stock_request_ids"] = [(4, request.id)]
+            if (
+                request._fields.get("analytic_account_id")
+                and request.analytic_account_id
+            ):
+                vals["account_analytic_id"] = request.analytic_account_id.id
+            if request._fields.get("analytic_tag_ids") and request.analytic_tag_ids:
+                vals["analytic_tag_ids"] = [
+                    (4, tag.id) for tag in request.analytic_tag_ids
                 ]
         return vals
 
--- a/stock_request_purchase/tests/test_stock_request_purchase.py
+++ b/stock_request_purchase/tests/test_stock_request_purchase.py
@@ -233,6 +233,8 @@ class TestStockRequestPurchase(common.TransactionCase):
 
     def test_create_request_analytic_data(self):
         """Single Stock request with analytic data"""
+        if not self.stock_request._fields.get("analytic_account_id"):
+            self.skipTest('No analytic addon installed')
         expected_date = fields.Datetime.now()
         analytic_accont = self.env["account.analytic.account"].create({
             "name": "Test Analytic Account",


stock_request_purchase/tests/test_stock_request_purchase.py
```diff
@@ -233,6 +233,8 @@ def test_view_actions(self):
 
     def test_create_request_analytic_data(self):
         """Single Stock request with analytic data"""
+        if 'analytic_account_id' not in self.env["stock.request"]:
+            raise unittest.SkipTest('No analytic addon installed')
```
joao-p-marques:
```suggestion
            self.SkipTest('No analytic addon installed')
```

joao-p-marques:
@pedrobaeza this is a side effect of https://github.com/OCA/stock-logistics-warehouse/pull/1105
As `stock_request_analytic` adds the analytic fields and `stock_request_purchase` is the one that propagates it to the PO, I missed the situation where `stock_request_purchase` might be running without `stock_request_analytic`, so we need to account for that.

Thanks @victoralmau

stock_request_purchase/tests/test_stock_request_purchase.py
```diff
@@ -233,6 +233,8 @@ def test_view_actions(self):
 
     def test_create_request_analytic_data(self):
         """Single Stock request with analytic data"""
+        if 'analytic_account_id' not in self.env["stock.request"]:
```
joao-p-marques: This shouldn't be necessary, but just in case:
```suggestion
        if 'analytic_account_id' not in self.env["stock.request"] or `analytic_tag_ids` not in self.env["stock.request"]:
```

joao-p-marques (approved):

stock_request_purchase/models/stock_rule.py
```diff
@@ -12,13 +12,13 @@ def _prepare_purchase_order_line(self, product_id, product_qty,
...
             product_id, product_qty, product_uom, values, po, supplier)
         if 'stock_request_id' in values:
-            stock_request = self.env["stock.request"].browse(values['stock_request_id'])
-            vals['stock_request_ids'] = [(4, stock_request.id)]
-            if stock_request.analytic_account_id:
-                vals['account_analytic_id'] = stock_request.analytic_account_id.id
-            if stock_request.analytic_tag_ids:
+            request = self.env["stock.request"].browse(values['stock_request_id'])
+            vals['stock_request_ids'] = [(4, request.id)]
+            if 'analytic_account_id' in request and request.analytic_account_id:
```
pedrobaeza: Shouldn't it be `if request._fields.get("analytic_account_id' in request")`?

And please use double quotes for avoiding pre-commit changes in 13.0.

stock_request_purchase/models/stock_rule.py
```diff
@@ -11,14 +11,14 @@ def _prepare_purchase_order_line(self, product_id, product_qty,
...
-            vals['stock_request_ids'] = [(4, stock_request.id)]
-            if stock_request.analytic_account_id:
-                vals['account_analytic_id'] = stock_request.analytic_account_id.id
-            if stock_request.analytic_tag_ids:
-                vals['analytic_tag_ids'] = [
-                    (4, tag.id) for tag in stock_request.analytic_tag_ids
+        if "stock_request_id" in values:
+            request = self.env["stock.request"].browse(values["stock_request_id"])
+            vals["stock_request_ids"] = [(4, request.id)]
+            if request._fields.get("analytic_account_id" in request):
```
pedrobaeza:
```suggestion
            if request._fields.get("analytic_account_id"):
```
and the same for the other

stock_request_purchase/tests/test_stock_request_purchase.py
```diff
@@ -233,6 +233,11 @@ def test_view_actions(self):
 
     def test_create_request_analytic_data(self):
         """Single Stock request with analytic data"""
+        if (
+            'analytic_account_id' not in self.env["stock.request"]
```
pedrobaeza: Switch this as well and no need to test both conditions.

stock_request_purchase/tests/test_stock_request_purchase.py
```diff
@@ -233,6 +233,8 @@ def test_view_actions(self):
 
     def test_create_request_analytic_data(self):
         """Single Stock request with analytic data"""
+        if self.stock_request._fields.get("analytic_account_id"):
```
pedrobaeza:
```suggestion
        if not self.stock_request._fields.get("analytic_account_id"):
```

stock_request_purchase/models/stock_rule.py
```diff
@@ -11,14 +11,14 @@ def _prepare_purchase_order_line(self, product_id, product_qty,
...
-                vals['analytic_tag_ids'] = [
-                    (4, tag.id) for tag in stock_request.analytic_tag_ids
+        if "stock_request_id" in values:
+            request = self.env["stock.request"].browse(values["stock_request_id"])
+            vals["stock_request_ids"] = [(4, request.id)]
+            if request._fields.get("analytic_account_id"):
+                vals["account_analytic_id"] = request.analytic_account_id.id
+            if request._fields.get("analytic_tag_ids"):
+                vals["analytic_tag_ids"] = [
+                    (4, tag.id) for tag in request.analytic_tag_ids
```
joao-p-marques: I think we also need to check if the value is defined before using it:
```suggestion
            if request._fields.get("analytic_account_id") and request.analytic_account_id:
                vals["account_analytic_id"] = request.analytic_account_id.id
            if request._fields.get("analytic_tag_ids") and request.analytic_tag_ids:
                vals["analytic_tag_ids"] = [
                    (4, tag.id) for tag in request.analytic_tag_ids
```
Because of Python's lazy comparation, it shouldn't introduce any errors.

It is needed because we might have the field present in the model, but not defined, thus giving an error when trying to retrieve its id.

joao-p-marques (approved):

joao-p-marques (approved):

etobella (approved):

PR #1111, opened 2021-03-24
[13.0][FIX] stock_request_purchase: Fix user permissions in tests

joao-p-marques:
Some other modules might add more constraints and ACLs to the models needed to manage stock requests. We can make it more resilient by using a more general group in the tests
As the general "Inventory / User" group inherits the "Stock Request User" group, we can use that instead (see stock_request module)
We don't need that in the manager as it inherits the "Inventory / User" group, thus ensuring the correct user permissions.

@Tecnativa
TT28895

ping @sergio-teruel @victoralmau


JordiBForgeFlow (approved):

rousseldenis (approved):
Makes sense

etobella (approved):

PR #1114, opened 2021-03-25
[13.0][ADD] New module stock_quant_expiration_date_tree

manuelcalerosolis:
New module stock_quant_expiration_date_tree.

Allows to display expirations dates on stock quant tree view.


rousseldenis (approved):
LGTM

stock_quant_expiration_date_tree/models/stock_quant.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    life_date = fields.Datetime(string="End of Life Date", related="lot_id.life_date",)
+    use_date = fields.Datetime(string="Best before Date", related="lot_id.use_date",)
+    removal_date = fields.Datetime(
+        string="Removal Date", related="lot_id.removal_date",
+    )
```
yostashiro, line 14: @manuelcalerosolis Sorry to comment late but I think these lines should be removed as the standard code already has it. https://github.com/odoo/odoo/blob/8f42ad0334e316b3aa94b70dd0a4aca4403bb0e5/addons/product_expiry/models/stock_quant.py#L10

PR #1115, opened 2021-03-29
[13.0][ADD] New module 'stock_helper_delivery'

sebalix:
Include a new helper method to get the first picking with a carrier from an operation in the middle of a chain.
Used for instance to get the carrier from any non-delivery operation, it will returns the related delivery which has a carrier defined.


stock_helper_delivery/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Stock Helpers - Delivery",
+    "summary": "Add methods shared between various stock+delivery modules",
+    "version": "13.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "category": "Hidden",
+    "depends": ["stock", "delivery", "stock_helper"],
```
simahawk: should it really depend on stock_helper?

TDu (approved):

stock_helper_delivery/__manifest__.py
sebalix: Not really, I was re-using the tests class from `stock_helper` but we could remove this dependency if it's really needed

sebalix: Updated to get rid of this dependency + increase test coverage

stock_helper_delivery/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def get_carrier_from_chained_picking(self):
```
simahawk: as reported on the other PR: make it return the picking instead. To be renamed as `get_picking_with_carrier_from_chain` or something like that.

stock_helper_delivery/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+        """Returns the proper carrier for the current transfer.
+
+        For non-delivery transfers this method will return the carrier used
+        on the related delivery transfer if there is one.
+        """
+        self.ensure_one()
+        if self.carrier_id:
+            return self.carrier_id
+
+        def get_moves_dest(moves):
```
simahawk: as reported on the other PR: loop and stop at the 1st one that matches.

simahawk (changes requested):

stock_helper_delivery/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
...
+
+        def get_first_move_dest(moves):
+            for move in moves.move_dest_ids:
+                if move.state not in ("cancel", "done"):
+                    return move
+
+        move_dest = get_first_move_dest(self.move_lines)
+        while move_dest:
+            if move_dest.picking_id.carrier_id:
+                return move_dest.picking_id
```
mmequignon, line 30:
```suggestion
        if move_dest:
            if move_dest.picking_id.carrier_id:
                return move_dest.picking_id
```
At most you'll have 1 record. No need to use "while" here, or am I missing something?

mmequignon (approved):
technical review : LG

stock_helper_delivery/models/stock_picking.py
simahawk: it means it jumps on the next move till you get a dest move. It stops either if you find the carrier or if you don't have any dest move

stock_helper_delivery/models/stock_picking.py
```diff
@@ -17,15 +24,21 @@ def get_picking_with_carrier_from_chain(self):
...
+    def _get_ship_from_chain(self, done=False):
+        """Returns the shipment related to the current operation."""
+        move_dest = get_first_move_dest(self.move_lines, done=done)
+        while move_dest:
+            picking = move_dest.picking_id
+            if picking.picking_type_id.code == "outgoing":
+                return picking
+            move_dest = get_first_move_dest(move_dest, done=done)
+        # Should return an empty record if we reach this line
+        return self.browse()
```
TDu, line 44: Should it not return an empty move instead of stock.picking ?

sebalix: No this method returns a `stock.picking` record (the delivery one if we found one), it never returns a move.

TDu: :+1:

TDu (approved):

sebalix:
I want to replace this module by a more functional one (not a "helper" one, difficult to maintain depending on the dependencies, and a bit hackish), the new module will implement the same kind of methods usable by other modules.

stock_helper_delivery/models/stock_picking.py
mmequignon: Well, if I read get_first_move_dest, it'll return a single record.
What I understand from your explaination is something like this
```python
        def get_first_move_dest(moves):
            for move in moves.move_dest_ids:
                if move.state not in ("cancel", "done"):
                    yield move
                    
        while get_first_move_dest(self.move_lines):
            if move_dest.picking_id.carrier_id:
                return move_dest.picking_id
```
Anyway, I'm probably missing something.
Forget about this

sebalix:
Closing, replaced by https://github.com/OCA/delivery-carrier/pull/354

PR #1120, opened 2021-04-07
[FIX] stock_vertical_lift: remove dependency

gurneyalex:
remove dependency on web_notify, as we suspect it is causing
issues on the production instance of our customer by creating
a huge load of queries on /longpolling/poll


sebalix:
LG, we'll try.

PR #1121, opened 2021-04-07
[13.0] Add stock_measuring_device

gurneyalex:
This is a rework of the stock_cubiscan module to allow supporting
different kind of measuring devices, with different communication
protocols. We use components from OCA/connector to abstract the
implementation of the backend (some devices are synchronous, others are
asynchronous...).

This module by itself does nothing, it needs one of the support module providing a driver to be present (see below).
    
Related pull requests:
stock_measuring_device_cubiscan #1129 
stock_measuring_device_zippcube #1128


stock_measuring_device/components/measuring_device_component.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_device(self):
+        success = self._do_test_device()
+        device = self.collection
+        if success and device.state == "not_ready":
+            device.state = "ready"
+        elif not success and device.state == "ready":
+            device.state = "not_ready"
+
+    def _do_test_device(self):
```
simahawk: I guess this has to be overridden by specific implementation?
A docstring would help :wink:

stock_measuring_device/components/measuring_device_component.py
```diff
@@ -0,0 +1,35 @@
...
+    def preprocess_measures(self, measures):
+        """perform required parsing, key mapping and possible unit conversion"""
+        return measures
+
+    def post_update_packaging_measures(self, measures, packaging, wizard_line):
+        """hook called after the update of the packaging or wizard line update"""
+        pass
+
+    def get_measure(self):
+        pass
```
simahawk: `raise NotImplementedError` maybe?

stock_measuring_device/models/measuring_device.py
```diff
@@ -0,0 +1,105 @@
...
+            kwargs,
+        )
+        self.ensure_one()
+        measuring_device = self._get_measuring_device()
+        return getattr(measuring_device, method)(*args, **kwargs)
+
+    def _get_measuring_device(self):
+        with self.work_on(self._name) as _forward:
+            return _forward.component(usage=self.device_type)
+
```
simahawk:
```suggestion
        with self.work_on(self._name) as work:
            return work.component(usage=self.device_type)

```

stock_measuring_device/wizard/measuring_wizard.py
```diff
@@ -0,0 +1,178 @@
...
+                        self.product_id.product_height, mm_uom
+                    ),
+                }
+            )
+        return [vals]
+
+    def _prepare_packaging_lines(self):
+        vals_list = []
+        product_packaging = self.env["product.packaging"]
+        packaging_types = self.env["product.packaging.type"].search([])
```
simahawk, line 60: maybe a silly question as I miss a bit of context, but... this search and the following loop could generate tons of wiz lines, no? Is there any way to limit this?

stock_measuring_device/wizard/measuring_wizard.py
```diff
@@ -0,0 +1,178 @@
...
+        self.ensure_one()
+        res = self.device_id.open_wizard()
+        res["res_id"] = self.id
+        return res
+
+    def on_barcode_scanned(self, barcode):
+        self.ensure_one()
+        prod = self.env["product.product"].search([("barcode", "=", barcode)], limit=1)
+        self.product_id = prod
+        self.onchange_product_id()
```
simahawk: the call to onchange here seems superfluous as you are setting the value right above, no?

simahawk (changes requested):

stock_measuring_device/models/measuring_device.py
gurneyalex: you had requested me to use _forward in your previous review...

stock_measuring_device/wizard/measuring_wizard.py
gurneyalex: you typically don't have that many packaging types: unit, box, retail box, pallet. Give or take a few. So no.

stock_measuring_device/models/measuring_device.py
simahawk: no, I asked you to use call `_forward` to forward calls to the device, not to call `_forward` the `work` context of the component :wink:

simahawk (approved):
LG, pleas squash

mmequignon (approved):

PR #1128, opened 2021-04-15
[ADD] stock_measuring_device_zippcube

gurneyalex:
support for Bosche Zippcube measuring devices

includes #1121

PR #1129, opened 2021-04-15
[ADD] stock_measuring_device cubiscan

gurneyalex:
add support for cubiscan devices

includes #1121


gurneyalex:
I'm not sure how to handle the transition from stock_cubiscan to this one? 

options:
* Remove stock_cubiscan from the repo
* Replace it with an empty module depending on this one (possibly with an upgrade script to create the measuring device from the record?) 
* Leave stock_cubiscan, with a note saying it is deprecated and recommending to use this module? And not migrating it to 14.0

mmequignon (approved):

simahawk:
done https://github.com/OCA/stock-logistics-warehouse/pull/1121 and https://github.com/OCA/stock-logistics-warehouse/pull/1128 for zipcube. Cubiscan stays as it was since we are not using it anymore

PR #1132, opened 2021-04-17
[13.0][IMP] stock_request, consolidate imp/fix from v12

kittiu:
- [X] Improve cancellation detection https://github.com/OCA/stock-logistics-warehouse/pull/895
- [x] fix archived search view https://github.com/OCA/stock-logistics-warehouse/pull/1040
- [x] stock_request:- Improved _compute_qty code https://github.com/OCA/stock-logistics-warehouse/pull/960

PR #1134, opened 2021-04-19
[14.0][MIG] stock_request

kittiu:
This one is the standard migration form v13, but also include fixed in this PR https://github.com/OCA/stock-logistics-warehouse/pull/1132


rvalyi (approved):
LGTM

PR #1135, opened 2021-04-19
[13.0][FIX] stock_available_unreserved, include scrap location on the compute.

guimarc-br:
Pull request created to include the scrap location on the compute value for the available_unreserved.

So when we have a scrap location with a specific product it will be computed on the available_not_res field.


guimarc-br:
Can someone take a look please ?

kos94ok-3D:
@guimarc-br I think it is wrong.
Qty available exclude a scrap location so why unreserve have to include it?

rousseldenis:
@guimarc-br Agree with @kos94ok-3D 

Makes no sense.

PR #1141, opened 2021-04-23
Zippcube measure update

mmequignon:
When the measure update is called from somewhere else than the
device wizard, an exception is raised.


sebalix (commented):
Not needed if https://github.com/OCA/stock-logistics-warehouse/pull/1144 is merged instead?

mmequignon:
I close this PR, as its only commit is included in https://github.com/OCA/stock-logistics-warehouse/pull/1144

PR #1154, opened 2021-05-18
[13.0][FIX] stock_request_picking_type - fixing stock.picking.type.kanban view

baimont:
Reproduce : 

Go to Inventory/Overview.
Click on Stock Requests.
image
Leads to an error.
image
Furthermore, buttons on this kanban card use filters that does not exist, so they don't work.


sanrav (approved):
Functional testing OK.
No error message when I click on Stock requests in the Inventory dashboard and filters are properly applied.

rousseldenis (approved):
Code and functional review

PR #1054, opened 2021-01-12
[14.0][MIG] stock_orderpoint_manual_procurement

dzungtran89:

PR #1163, opened 2021-05-28
14.0 mig stock orderpoint manual procurement

DavidBForgeFlow:
Migration to v14.
Adding definition of "_get_date_planned" in the file "stock_warehouse_orderpoint.py" because in v13 this function was defined in 'odoo>addons>stock>models>stock_warehouse.py' but it does not exist anymore neither in this file nor in similar ones.

cc @LoisRForgeFlow


rousseldenis:
@DavidBForgeFlow This has already done there : #1054 

Could you review that one instead ?

DavidBForgeFlow:
@rousseldenis Thank you for saying!! I'm closing this one and going to review the other.

PR #1167, opened 2021-06-03
[13.0] s_packaging_calculator: misc fix/imp

simahawk:
See atomic commits for details.

ref: 2535

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -24,6 +24,7 @@ class Product(models.Model):
         help="Technical field to store contained packaging. ",
     )
 
+    @api.depends_context("lang")
     @api.depends("packaging_ids.qty")
     def _compute_packaging_contained_mapping(self):
         for rec in self:
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -1,6 +1,8 @@
 # Copyright 2020 Camptocamp SA
+# @author: Simone Orsi <simone.orsi@camptocamp.com>
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
 
+import unicodedata
 from collections import namedtuple
 
 from odoo import api, models
@@ -10,6 +12,8 @@ from odoo.addons.base_sparse_field.models.fields import Serialized
 
 # Unify records as we mix up w/ UoM
 Packaging = namedtuple("Packaging", "id name qty barcode is_unit")
+
+NO_BREAK_SPACE_CHAR = unicodedata.lookup("NO-BREAK SPACE")
 
 
 class Product(models.Model):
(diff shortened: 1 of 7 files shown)


stock_packaging_calculator/models/product.py
```diff
@@ -145,3 +149,57 @@ def _prepare_qty_by_packaging_values(self, packaging, qty_per_pkg):
...
+            return ""
+
+        # Exclude unit qty and reuse it later
+        unit_qty = None
+        has_only_units = True
+        _qty_by_packaging = []
+        for pkg_qty in qty_by_packaging:
+            if pkg_qty["is_unit"]:
+                unit_qty = pkg_qty["qty"]
+                continue
```
mmequignon, line 175: what if you have more than 1 packaging with units? Is it even possible?
Shouldn't we set "unit_qty" as 0 and sum the qties instead?

mmequignon (approved):

stock_packaging_calculator/models/product.py
simahawk: impossible. The unit is added explicitly at the end of the pkg list and it's always the product uom.

mmequignon: Okay then

stock_packaging_calculator/models/product_qty_by_packaging_mixin.py
```diff
@@ -0,0 +1,48 @@
...
+# @author: Sébastien Alix <sebastien.alix@camptocamp.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+from odoo import api, fields, models
+
+
+class ProductQtyByPackagingMixin(models.AbstractModel):
+    """Allow displaying product qty by packaging.
+    """
+
+    _name = "product.qty_by_packaging.mixin"
```
TDu, line 12: It does not look to me that the new mixin is covered by test, maybe I am mistaken ?

simahawk: no, I wanted to but I had no time to add fake models. BUT it's tested in #1168 on real models :wink:

simahawk: @TDu good for you? Then I can merge and rebase #1168

stock_packaging_calculator/models/product.py
```diff
@@ -140,3 +150,57 @@ def _prepare_qty_by_packaging_values(self, packaging, qty_per_pkg):
...
+            has_only_units = False
+            _qty_by_packaging.append(pkg_qty)
+        # Browse them all at once
+        records = self.env["product.packaging"].browse(
+            [x["id"] for x in _qty_by_packaging]
+        )
+        _qty_by_packaging_as_str = self.env.context.get(
+            "_qty_by_packaging_as_str", self._qty_by_packaging_as_str
+        )
```
sebalix, line 184: Not related to this PR, I'm wondering if it is a good practice or advisable to pass non-serializable objects into the context?

sebalix (approved):

PR #1168, opened 2021-06-03
[13.0] Add glue modules for displaying qty by packaging

simahawk:
- [x] Depends on / includes https://github.com/OCA/stock-logistics-warehouse/pull/1167

Consider only last 4 commits.

### DIFF
--- a/stock_vertical_lift_kardex/proxy/kardex-proxy.py
+++ b/stock_vertical_lift_kardex/proxy/kardex-proxy.py
@@ -6,7 +6,7 @@ import os
 import ssl
 import time
 
-import aiohttp
+import aiohttp  # pylint: disable=missing-manifest-dependency
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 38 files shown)


sebalix (approved):

stock_vertical_lift_kardex/proxy/kardex-proxy.py
```diff
@@ -6,7 +6,7 @@
 import ssl
 import time
 
-import aiohttp
+import aiohttp  # pylint: disable=missing-manifest-dependency
```
mmequignon, line 9: why not adding this dependency in the manifest instead of hiding this?

mmequignon (approved):

stock_vertical_lift_kardex/proxy/kardex-proxy.py
simahawk: because is not required. This is not for the module to work, this is need only if you run this proxy script (and there's a specific `requirements.txt` in this folder ;)

simahawk:
merge branch was green, merging was still hanging... I merged manually

PR #1173, opened 2021-06-09
14.0 mig stock change qty reason

MeetKD:


rousseldenis:
This has been migrated through https://github.com/OCA/stock-logistics-warehouse/pull/1320

PR #1174, opened 2021-06-09
[FIX] - stock_available_mrp: local variable referenced before assignment

sbejaoui:


stock_available_mrp/models/product_product.py
```diff
@@ -83,6 +83,7 @@ def _compute_available_quantities_dict(self):
             # Need by product (same product can be in many BOM lines/levels)
             exploded_components = exploded_boms[product.id]
             component_needs = product._get_components_needs(exploded_components)
+            bom_id = product.bom_id
```
pedrobaeza: You can take the occasion to rename to a more proper name `bom` instead of `bom_id`.

pedrobaeza (approved):
This option also serves

pedrobaeza:
Fast tracking this obvious patch:

/ocabot merge patch

PR #1185, opened 2021-06-24
[13.0] stock_cycle_count error on location_id

celm1990:
Hi, when i try open Accuracy Stats get following error
image

image

field in` stock.inventory` is `location_ids(many2many)` https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_inventory.py#L45-L49

### DIFF
--- a/stock_cycle_count/models/stock_location.py
+++ b/stock_cycle_count/models/stock_location.py
@@ -115,9 +115,6 @@ class StockLocation(models.Model):
         self.ensure_one()
         action = self.env.ref("stock_cycle_count.act_accuracy_stats")
         result = action.read()[0]
-        result["context"] = {"search_default_location_id": self.id}
-        new_domain = (
-            result["domain"][:-1] + ", ('location_id', 'child_of', active_ids)]"
-        )
+        new_domain = result["domain"][:-1] + ", ('location_ids', 'in', active_ids)]"
         result["domain"] = new_domain
         return result


celm1990:
someone help me with this PR please?

CC: @LoisRForgeFlow @jimhoefnagels

stock_cycle_count/models/stock_location.py
```diff
@@ -115,9 +115,9 @@ def action_accuracy_stats(self):
         self.ensure_one()
         action = self.env.ref("stock_cycle_count.act_accuracy_stats")
         result = action.read()[0]
-        result["context"] = {"search_default_location_id": self.id}
+        result["context"] = {"search_default_location_ids": self.name}
```
pedrobaeza: I doubt this is correct, as you are mixing text with IDs. In any case, it should be `self.ids`.

stock_cycle_count/models/stock_location.py
```diff
@@ -115,9 +115,9 @@ def action_accuracy_stats(self):
...
         new_domain = (
-            result["domain"][:-1] + ", ('location_id', 'child_of', active_ids)]"
+            result["domain"][:-1] + ", ('location_ids', 'child_of', active_ids)]"
```
pedrobaeza: Not sure if operator `child_of` works with many2many fields.

stock_cycle_count/models/stock_location.py
celm1990: The reason: this text is pass to search view as input, `ids` not working.

stock_cycle_count/models/stock_location.py
celm1990: I tested changes before create PR and working fine
See 
https://github.com/odoo/odoo/blob/13.0/odoo/osv/expression.py#L1009

stock_cycle_count/models/stock_location.py
pedrobaeza: But that's wrong. Having other locations with similar name, they will be filtered in here.

celm1990: @pedrobaeza  If pass `ids ` is wrong on `searchview`
image

I update code for pass complete_name, is OK now??

celm1990: After change
image

pedrobaeza: The problem is that you can't search by IDs on a m2m, but I think you should convert the domain to `id in ...` instead.

celm1990: I Update my code, is OK now?

stock_cycle_count/models/stock_location.py
```diff
@@ -115,9 +115,9 @@ def action_accuracy_stats(self):
         self.ensure_one()
         action = self.env.ref("stock_cycle_count.act_accuracy_stats")
         result = action.read()[0]
-        result["context"] = {"search_default_location_id": self.id}
+        result["context"] = {"search_default_location_ids": self.complete_name}
```
pedrobaeza: You should remove this.

celm1990:
@pedrobaeza changes are OK? or I need adapt something

pedrobaeza (approved):

LoisRForgeFlow (approved):
Thanks

LoisRForgeFlow:
wrong command, not sure if editing it was enough so I restart the merge:

/ocabot merge patch

PR #1190, opened 2021-07-02
[13.0] [FIX] stock_inventory_include_exhausted: Filtered with product…

nicomacr:
… who has already created inventory line to avoid duplicated line.

Is is similar to odoo did for v14 https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_inventory.py#L287 to consider if the inventory line has already created to avoid create again for the same product and location.

### DIFF
--- a/stock_inventory_include_exhausted/models/stock_inventory.py
+++ b/stock_inventory_include_exhausted/models/stock_inventory.py
@@ -18,6 +18,8 @@ class StockInventory(models.Model):
 
         if self.include_exhausted:
             domain = [("qty_available", "=", 0), ("type", "=", "product")]
+            product_ids = [data.get("product_id") for data in vals]
+            domain.append(("id", "not in", product_ids))
             if self.product_ids:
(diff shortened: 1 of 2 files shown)


dreispt (approved):

stock_inventory_include_exhausted/models/stock_inventory.py
```diff
@@ -18,6 +18,8 @@ def _get_inventory_lines_values(self):
 
         if self.include_exhausted:
             domain = [("qty_available", "=", 0), ("type", "=", "product")]
+            product_ids = [data.get("product_id") for data in vals]
```
rousseldenis, line 21: @nicomacr Thanks for this. Could you add a test that highlight this ?

nicomacr: Ready!!

nicomacr:
@rousseldenis done! ready for merge?

PR #1202, opened 2021-07-23
[13.0] [MIG] stock_reserve: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT30806

please @joao-p-marques @victoralmau review this


victoralmau (approved):
Code review OK.

stock_reserve/model/stock_reserve.py
```diff
@@ -0,0 +1,181 @@
...
+        self.release_reserve()
+        return super().unlink()
+
+    @api.onchange("product_id")
+    def _onchange_product_id(self):
+        """ set product_uom and name from product onchange """
+        # save value before reading of self.move_id as this last one erase
+        # product_id value
+        self.move_id.write({"product_id": self.product_id.id})
+        self.move_id.onchange_product_id()
```
joao-p-marques: I have some doubts about the way this is done:

1. Is it "safe" to do a `write` in an `onchange`? Than can end up triggering too many writes, or even cause some weird behavior. Could it maybe be converted to a `compute` and apply a similar behavior?
2. Is it still necessary to manually trigger the `onchange` in the `stock.move`? 

As I said, they are doubts, maybe this is the best solution, just wanted to confirm 🙂

joao-p-marques (commented):
Code review. Just a couple of questions...

stock_reserve/model/stock_reserve.py
pedrobaeza: No, you can't do a write on an onchange. That's totally forbidden.

CarlosRoca13: On the second point, I think that's necessary, because the change is launched by code. If we just assign the `product_id` on the `stock.move`, we will not take the value of `name` and `product_uom`

stock_reserve/model/stock_reserve.py
```diff
@@ -200,3 +179,27 @@ def open_move(self):
             views=[(view_id, "form")], res_id=self.move_id.id,
         )
         return action_dict
+
+    def write(self, vals):
+        res = super().write(vals)
+        if (
+            "product_uom_qty" in vals
+            and self.state == "partially_available"
+            and self.product_id.virtual_available >= 0
```
rousseldenis: IMHO you should use float_compare here

stock_reserve/model/stock_reserve.py
```diff
@@ -200,3 +179,27 @@ def open_move(self):
...
+        ):
+            self.reserve()
+        return res
+
+    @api.model
+    def assign_waiting_confirmed_reserve_moves(self):
+        reservations_to_assign = self.search(
+            [
+                ("state", "in", ["confirmed", "waiting", "partially_available"]),
```
rousseldenis: Maybe externalize domain retrieval in a function ?

rousseldenis (changes requested):
Seems great. Code review

CarlosRoca13:
@rousseldenis changes done

> Maybe externalize domain retrieval in a function ?

I don't know if you mean what I have done 😅😅

stock_reserve/model/stock_reserve.py
rousseldenis: > @rousseldenis changes done
> 
> > Maybe externalize domain retrieval in a function ?
> 
> I don't know if you mean what I have done

Nearly, in order to allow inheritance, domain retrieval is better in a separate function, but for the whole one.

Like:

self.search(self._get_reservations_to_assign_domain())

rousseldenis (approved):
LGTM. Code review

CarlosRoca13:
@rousseldenis thanks for the clarification, changes done :)

PR #1204, opened 2021-07-27
[13.0][IMP] stock_quant_manual_assign: make qty_done fill optional

LoisRForgeFlow:
There are cases where auto-filling of qty_done of stock move line is not desirable.
e.g. you assign quants manually for some of the moves in a picking and not the others,
in such case you need to go over all the moves in the picking to either remove qty_done
or fill it in to proceed with the validation of the entire moves. Auto-fill behavior is
also troublesome when this function is used in a manufacturing order. i.e. having
qty_done of the component move live messes up the outcome of the production.

Forward port of https://github.com/OCA/stock-logistics-warehouse/pull/1055.

@ForgeFlow

### DIFF
--- a/stock_quant_manual_assign/__init__.py
+++ b/stock_quant_manual_assign/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizard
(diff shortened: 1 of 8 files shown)


LoisRForgeFlow:
@yostashiro could you review?

yostashiro (approved):
Functional review. 👍

MateuGForgeFlow (approved):

PR #1210, opened 2021-08-09
Deleting fr_FR.po files

pasgou:
It seems that fr_FR.po files are useless, cause Odoo load fr.po files for french_FRENCH.
fr_XX is used only for XX in (BE, CA, ... and so on).


pedrobaeza (approved):
Let's fast-track it.

PR #1211, opened 2021-08-10
12.0 [IMP] simplify_stock_move_location

tack-yabeQuartile:
make code clear and readable.
Deleted method was created in the commit https://github.com/OCA/stock-logistics-warehouse/commit/3a9fb80e8224e106c30f4b796d6aea6ea183bb91 on 29 Jun 2020.
I suggest displaying not only max_qty and move_qty but non-reserved_qty, which is more smart and user-friendly, I guess.


tack-yabeQuartile:
This PR will be closed because I have misunderstood the role of the topic method.

PR #1217, opened 2021-08-16
[12.0][IMP] stock_move_location: add option to propose available quantities

yostashiro:
This PR adds the available quantity column in the move location wizard line,
and allows users to configure in the operation type as necessary to make the available quantity
(non-reserved quantity) the default move quantity instead of the max quantity (quantity on hand).

With this, the risk of unexpectedly moving some reserved quants can be mitigated.

image


TomohiroAbe-Quartile (approved):
Functional : OK
Code : LGTM.

tack-yabeQuartile (approved):
I felt this is a nice function. LGTM from viewpoints of both functional and code.

PR #879, opened 2020-04-07
[12.0][FIX] user can delete lines

jarroyomorales:
Unlink permision of stock_requests was missing for stock request users.

@HviorForgeFlow @etobella

PR #645, opened 2019-07-08
[FIX] stock_request: Allow users to delete stock requests

etobella:
Without this, users cannot delete lines of stock request orders. The unlink cannot be used unless the state is `draft`

@jbeficent

PR #1219, opened 2021-08-22
[14.0][FIX] stock_request: Stock Request User can delete Stock Request

newtratip:
1. Setting user role to `Stock Request User`
2. Create Stock Request Order
3. Add Item Lines
4. Can not delete item lines
Selection_145

This PR is fix item lines should be delete by stock request user.

cc: @kittiu @ps-tubtim

### DIFF
--- a/stock_request/security/ir.model.access.csv
+++ b/stock_request/security/ir.model.access.csv
@@ -1,5 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_stock_request_user,stock request user,model_stock_request,group_stock_request_user,1,1,1,
+access_stock_request_user,stock request user,model_stock_request,group_stock_request_user,1,1,1,1
 access_stock_request_manager,stock request manager,model_stock_request,group_stock_request_manager,1,1,1,1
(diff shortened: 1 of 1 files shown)


kittiu:
@newtratip The module intention is for stock manager to delete only, and not allow for stock user. I think this is by design since version 11.

I also don't know why it happen this way. But I think you should consult with the original author of this module, if you still want to proceed.

newtratip:
@kittiu, There is PR for `Allow users to delete stock requests` https://github.com/OCA/stock-logistics-warehouse/pull/645 in V.11 and `Fix user can delete lines` https://github.com/OCA/stock-logistics-warehouse/pull/879 in V.12. we can cherry pick this one.

In V.11, V.12 we can delete stock request
In V.13, V.14 we can not delete stock request

I think we should fix it.

kittiu:
> @kittiu, There is PR for `Allow users to delete stock requests` #645 in V.11 and `Fix user can delete lines` #879 in V.12. we can cherry pick this one.
> 
> In V.11, V.12 we can delete stock request
> In V.13, V.14 we can not delete stock request
> 
> I think we should fix it.

OK. Thanks for explaination.

kittiu (approved):

rousseldenis (approved):

HaraldPanten (approved):
LGTM

PR #1220, opened 2021-08-22
[12.0] stock_request: allowed translating status values, model stock.…

alexeirivera87:
small adjustment to allow translating the status values of a stock request

### DIFF
--- a/stock_request/__manifest__.py
+++ b/stock_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock Request",
     "summary": "Internal request for stock",
-    "version": "12.0.1.5.0",
+    "version": "12.0.1.5.1",
     "license": "LGPL-3",
     "website": "https://github.com/stock-logistics-warehouse",
     "author": "Eficent, "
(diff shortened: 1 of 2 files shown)


etobella:
This is not  necessary, odoo automatically detects this. On weblate, you can see that this values can be translated, as you can see in the screenshot for spanish
imagen

rousseldenis:
> This is not necessary, odoo automatically detects this. On weblate, you can see that this values can be translated, as you can see in the screenshot for spanish
> imagen

Agree. This is not needed.

alexeirivera87:
@rousseldenis 

Hello
I believe that what is being translated are the filters in the search views of stock.request and stock.request.order, not the values of the field state.

alexeirivera87:
Hello @rousseldenis @pedrobaeza 

The values of the field state, are taken from a function which takes the values from a variable defined outside the class

`    state = fields.Selection(
        selection=_get_request_states, string='Status',
        copy=False, default='draft', index=True,
        readonly=True, track_visibility='onchange',
    )`

In the .po file, these are the filters in the search views of stock.request and stock.request.order, not the values of the field state:

#. module: stock_request
#: model_terms:ir.ui.view,arch_db:stock_request.stock_request_order_search
#: model_terms:ir.ui.view,arch_db:stock_request.stock_request_search
msgid "Cancelled"
msgstr "Annulé"

PR #1221, opened 2021-08-23
[14.0] [MIG] stock_inventory_chatter

mariadforgeflow:
Standard migration to v14


bizzappdev:
This module migration to v14 not needed as it is available in [standard odoo](https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_inventory.py#L15).

PR #1226, opened 2021-08-25
[14.0][ADD][WIP] stock_inventory_revaluation_adjustment

patrickrwilson:
This module add a workflow to update costs of products in bulk, similar to the inventory adjustment workflow. It allows the user to track see the impact before posting the changes.

Any product type can be updated and tracked.

It also allows the option to specify a different account for the changes if the storable product generates journal entries. This is done via the adjustment type.


patrickrwilson:
@Chandresh-OSI During testing this modules without the mrp one I get the following error because without the manufacturing module installed, BoM's don't exist. It would be best if these BoM calculations were put into the mrp module so if someone wants to use this module by itself they don't need the MRP module.

 File "C:\Odoo\Odoo 14.0e\server\odoo\addons\stock_inventory_revaluation\models\cost_adjustment.py", line 162, in action_post
    for bom in line.product_id.bom_line_ids.mapped("bom_id"):
Exception

The above exception was the direct cause of the following exception:

  File "C:\Odoo\Odoo 14.0e\server\odoo\http.py", line 640, in _handle_exception
  File "C:\Odoo\Odoo 14.0e\server\odoo\http.py", line 316, in _handle_exception
    raise exception.with_traceback(None) from new_cause
AttributeError: 'product.product' object has no attribute 'bom_line_ids'

patrickrwilson:
@Chandresh-OSI Small change I noticed. When the proposed cost on a product is $0, we don't want to set the future cost to that when it's added as this might cause users to mistakenly set costs to zero. Please modify this so if the proposed cost = 0.0 then it will use the current cost in the future cost field, otherwise if proposed cost is not 0.0 then it will use that value.
image

rousseldenis:
@patrickrwilson This is not a new module. It exists there : https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_inventory_revaluation

Please feel free to migrate it.

dreispt:
Thank you for pointing that out @rousseldenis - we missed that. 
We discussed about this, and at this point we can't afford to change these set of modules to fit the original `stock_inventory_revaluation` one.

So what we'll do for now is to rename this module, and avoid the name collision.
When we have budget from another project, we will reconsider migrating the original revaluation module and refactoring the additional features.
We'll make sure this is described in the module's README.

dreispt:
@patrickrwilson @Chandresh-OSI I suggest we rename this module to `stock_inventory_revaluation_adjustment`, and the description should have a paragraph about the overlap with `stock_inventory_revaluation`.
Thanks

stock_inventory_revaluation/models/cost_adjustment.py
```diff
@@ -0,0 +1,243 @@
...
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+from odoo.tools.float_utils import float_compare
+
+from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
+
+
+class CostAdjustment(models.Model):
+    _name = "cost.adjustment"
```
rousseldenis: IMHO, I would have either split this model in a base module, either you need to rename to something like 'stock.cost.adjustment'.

dreispt: I agree that namespaces are not properly taken case here.
@Chandresh-OSI @patrickrwilson I  hope this has limited impact on the code.

stock_inventory_revaluation_adjustment/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+
+{
+    "name": "Stock Inventory Revaluation",
+    "summary": "Adds a workflow to adjusting costs on products",
+    "version": "14.0.1.0.0",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "category": "Stock",
+    "license": "AGPL-3",
+    "depends": ["stock_account", "mrp"],
```
dreispt, line 12:
```suggestion
    "depends": ["stock_account"],
```

CasVissers-360ERP: I agree this module should not be depending on MRP here. 

There are more changes necessary though:

1. remove bom_count on product.template
2. remove bom_ids

stock_inventory_revaluation_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,164 @@
...
+        string="Product Type", related="product_id.type", readonly=True
+    )
+    product_original_cost = fields.Float(
+        string="Current Cost",
+        readonly=True,
+        default=0,
+        copy=False,
+    )
+    product_cost = fields.Float(
+        string="Future Cost",
```
CasVissers-360ERP, line 40: Add digits="Product Price",

stock_inventory_revaluation_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,164 @@
...
+        check_company=True,
+        index=True,
+        required=True,
+    )
+    product_type = fields.Selection(
+        string="Product Type", related="product_id.type", readonly=True
+    )
+    product_original_cost = fields.Float(
+        string="Current Cost",
+        readonly=True,
```
CasVissers-360ERP, line 35: Add digits="Product Price",

stock_inventory_revaluation_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+    product_cost = fields.Float(
+        string="Future Cost",
+        readonly=True,
+        states={"confirm": [("readonly", False)]},
+        default=0,
+        copy=False,
+    )
+    difference_cost = fields.Float(
+        string="Difference",
```
CasVissers-360ERP, line 47: Add digits="Product Price",

stock_inventory_revaluation_adjustment/views/cost_adjustment_line.xml
```diff
@@ -0,0 +1,144 @@
...
+                    name="product_id"
+                    readonly="context.get('default_product_id', False)"
+                    attrs="{'readonly': ['|', ('is_editable', '=', False), ('state', '!=', 'confirm')]}"
+                    options="{'no_create': True}"
+                />
+                <field name="product_type" optional="hide" />
+                <field name="categ_id" optional="hide" />
+                <field
+                    name="product_original_cost"
+                    widget="monetary"
```
CasVissers-360ERP, line 31: Only allows for 2 digits

stock_inventory_revaluation_adjustment/views/cost_adjustment_line.xml
```diff
@@ -0,0 +1,144 @@
...
+                    name="action_get_origin_cost"
+                    title="Refresh quantity"
+                    icon="fa-refresh"
+                    type="object"
+                    attrs="{'invisible': ['|', ('price_outdated', '=', False), ('state', '=', 'posted')]}"
+                />
+                <field
+                    name="product_cost"
+                    optional="show"
+                    widget="monetary"
```
CasVissers-360ERP, line 48: Only allows for 2 digits

stock_inventory_revaluation_adjustment/views/cost_adjustment_line.xml
```diff
@@ -0,0 +1,144 @@
...
+                    options="{'currency_field': 'currency_id'}"
+                    sum="Sum"
+                />
+                <field
+                    name="difference_cost"
+                    optional="show"
+                    widget="monetary"
```
CasVissers-360ERP, line 55: Only allows for 2 digits

stock_inventory_revaluation_adjustment/models/cost_adjustment.py
```diff
@@ -0,0 +1,241 @@
...
+                    negative.product_id.display_name,
+                    negative.product_cost,
+                )
+            )
+        return True
+
+    def _remove_unchanged_lines(self):
+        for line in self.line_ids:
+            precision = self.env["decimal.precision"].precision_get(
+                "Product Unit of Measure"
```
CasVissers-360ERP, line 109: "Product Price"

CasVissers-360ERP (changes requested):
Multiple changes necessary to allow more than 2 digits

CasVissers-360ERP:
Functional review, nice module! Looks very lean, but dependency on MRP should definitely be removed to allow more flexibility.

stock_inventory_revaluation_adjustment/models/product.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @api.model
+    def _change_standard_price(self, new_price, account=False):
+        """Override the helper method in stock_account.product.py to create
+         the stock valuation layers and the account moves.
+        :param new_price: new standard price, account: cost adjustment type account
+        """
+        # Check Access Rights for Journal Entry creation.
+
+        new_price_round = self.env.company.currency_id.round(new_price)
```
CasVissers-360ERP, line 20: Why do we round the price here?

rousseldenis:
@patrickrwilson @dreispt @LoisRForgeFlow This seems to have duplicate code with this : https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_inventory_revaluation

with same naming but without dependency on it. So, something has to be clarified...

rousseldenis:
> @patrickrwilson @Chandresh-OSI I suggest we rename this module to `stock_inventory_revaluation_adjustment`, and the description should have a paragraph about the overlap with `stock_inventory_revaluation`. Thanks

@dreispt This could lead to confusion. stock_cost_adjustement should be better.

But I continue to think two modules in OCA should not duplicate a behaviour. @LoisRForgeFlow what do you think ?

stock_inventory_revaluation_adjustment/models/product.py
CasVissers-360ERP: I think this nis actually a base issue. (reported to Odoo)

https://github.com/odoo/odoo/blob/1d73bc7d5458b3bb3a566aa4b4a4b101a5329b9a/addons/stock_account/models/product.py#L230

If you using pricing while > 2 digits price adjustments are incorrectly calculated. Actually the whole COGS mechanism is broken for > 2 digits: https://github.com/odoo/odoo/commit/306f3be81192eef504927e30483856669b36b047

rousseldenis:
@patrickrwilson @dreispt What's the status of this?

dreispt:
@rousseldenis Will https://github.com/OCA/stock-logistics-warehouse/pull/1226#issuecomment-977619298 be blocking for this PR?

rousseldenis:
@dreispt @patrickrwilson Comments of @CasVissers-360ERP have not been attended.

For mine, please add a ROADMAP readme fragment in order not forgetting to adapt code to existing stock_inventory_revaluation module.

PR #1227, opened 2021-08-25
[14.0][ADD][WIP] stock_inventory_revaluation_mrp

patrickrwilson:
This module extends the stock_inventory_valuation module to include MRP/BoM components. It adds a 'details' model to pull in BoM products to better see the impact for when a components price changes. It also will process the cost rollup process on the BoM's once the cost adjustment is posted.

Depends on https://github.com/OCA/stock-logistics-warehouse/pull/1226


stock_inventory_revaluation_mrp/models/cost_adjustment.py
```diff
@@ -0,0 +1,123 @@
...
+            for bom in boms:
+                if bom.product_id:
+                    self.get_product_bom_component(bom.product_id)
+                else:
+                    product = self.env["product.product"].search(
+                        [("product_tmpl_id", "=", bom.product_tmpl_id.id)], limit=1
+                    )
+                    self.get_product_bom_component(product)
+
+    def _create_cost_details(self, boms, product):
```
rousseldenis: You should reimplement this loop in order to :

- Do the searches before the loop (and then filter on criteria inside).
- Create a list of dict that will help you to do a create multi at the end.

stock_inventory_revaluation_mrp/models/cost_adjustment.py
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                )
+        return True
+
+    def action_post(self):
+        res = super().action_post()
+        for line in self.line_ids:
+            for bom in line.product_id.bom_line_ids.mapped("bom_id"):
+                if bom.product_id:
+                    bom.product_id.sudo().button_bom_cost()
```
rousseldenis, line 230: You should build a recordset of products and then call button_bom_cost once to get multi implementation:

https://github.com/odoo/odoo/blob/14.0/addons/mrp_account/models/product.py#L20

stock_inventory_revaluation_mrp/models/cost_adjustment_detail.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2021 - Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class CostAdjustmentDetail(models.Model):
+    _name = "cost.adjustment.detail"
```
rousseldenis: If it's linked strongly to mrp, I suggest to write somewhere 'mrp'.

Some parts shouldn't be in stock_inventory_revaluation ?

stock_inventory_revaluation_mrp/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,63 @@
...
+        "mrp.bom", string="BOMs", compute="_compute_set_productions_boms"
+    )
+    bom_count = fields.Integer(
+        string="BOM's", compute="_compute_set_productions_boms", readonly=False
+    )
+
+    @api.depends("product_id", "state")
+    def _compute_set_productions_boms(self):
+        for line in self:
+            if line.state not in ("posted"):
```
rousseldenis: You should set a value for all entries of recordset. So an else is needed.

stock_inventory_revaluation_mrp/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,63 @@
...
+                # Set MO's
+                if line.mrp_production_ids:
+                    line.mrp_production_ids = [(5,)]
+                productions = self.env["mrp.production"].search(
```
rousseldenis: Do the search before the loop and add:

`('move_raw_ids.product_id', 'in', self.mapped("product_id.id"))`

to get only the concerned ones.

stock_inventory_revaluation_mrp/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,63 @@
...
+                    [("state", "in", ["draft", "confirmed", "progress"])]
+                )
+                for production in productions:
```
rousseldenis: Due to the above comment, this can be simplified.

stock_inventory_revaluation_mrp/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,63 @@
...
+                    components = production.move_raw_ids.mapped("product_id")
+                    for product in components:
+                        if line.product_id.id == product.id:
+                            line.mrp_production_ids = [(4, production.id)]
+                # Set BOMs
+                if line.bom_ids:
+                    line.bom_ids = [(5,)]
+                bom_lines = self.env["mrp.bom.line"].search(
```
rousseldenis: Same remark about searches in loops

rousseldenis (changes requested):
IMHO you should reimplement the logic to avoid huge performances problems.

Issue #1139, opened 2021-04-21
[13.0] stock_mts_mto_rule: Error when migrating database


juhk:
OpenUpgrade migration from version 12.0 to 13.0 fails with following error:
```
2021-04-21 14:48:04,603 1 INFO mig_12_13 odoo.modules.loading: loading stock_mts_mto_rule/data/stock_data.xml 
2021-04-21 14:48:04,627 1 ERROR mig_12_13 odoo.modules.registry: Failed to load registry 
Traceback (most recent call last):
  File "/var/lib/openupgrade/odoo/tools/convert.py", line 717, in parse
    self._tag_root(de)
  File "/var/lib/openupgrade/odoo/tools/convert.py", line 679, in _tag_root
    f(rec)
  File "/var/lib/openupgrade/odoo/tools/convert.py", line 582, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/var/lib/openupgrade/odoo/models.py", line 4144, in _load_records
    data['record']._load_records_write(data['values'])
  File "/var/lib/openupgrade/odoo/models.py", line 4082, in _load_records_write
    self.write(values)
  File "/var/lib/openupgrade/odoo/models.py", line 3657, in write
    self._check_company()
  File "/var/lib/openupgrade/odoo/models.py", line 3260, in _check_company
    raise UserError("\n".join(lines))
odoo.exceptions.UserError: ("Incompatible companies on records:\n- 'Make To Order + Make To Stock' belongs to company False and 'Products' (product_ids: ...) belongs to another company.", '')
```
Migration completed successfully after adding `<data noupdate="1">` tag to stock_mts_mto_rule/data/stock_data.xml file.

PR #1231, opened 2021-09-02
[FIX] stock_mts_mto_rule: don't update route

chienandalu:
Odoo declares routes as noupdate records so the user has the control
over them. Another side effect of updatable route records is on
initializing the record when OpenUpgrade is used to upgrade, which in a
multicompany environment with mixed companies in the products could lead
to a false company restriction error (maybe due to a bad declaration of
restrictions in the product_ids field).

cc @Tecnativa TT31601

Related issue: https://github.com/OCA/stock-logistics-warehouse/issues/1139

### DIFF
--- a/stock_mts_mto_rule/__manifest__.py
+++ b/stock_mts_mto_rule/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Stock MTS+MTO Rule",
     "summary": "Add a MTS+MTO route",
-    "version": "13.0.1.2.0",
+    "version": "13.0.1.3.0",
     "development_status": "Mature",
     "category": "Warehouse",
     "website": "https://github.com/OCA/stock-logistics-warehouse",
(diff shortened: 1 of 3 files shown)


stock_mts_mto_rule/migrations/13.0.1.3.0/pre_migration.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Tecnativa David Vidal
```
pedrobaeza: It should be with slash, not underscore: `pre-migration.py`

pedrobaeza (approved):

pedrobaeza:
Let's fast-track this for solving the problem in migrations:

/ocabot merge nobump

HaraldPanten (approved):

chienandalu:
Looks like there's some issue with `stock_request` tests in the main branch

chienandalu:
Looks like a random issue. Maybe you can try and merge again?

PR #1153, opened 2021-05-13
[14.0][MIG] stock_request_submit (Version 13.0 to 14.0)

patrickrwilson:

PR #1232, opened 2021-09-02
[14.0][IMP]stock_request: Migrated a "Submitted" state flow #595

Vimal-OSI:
#595


patrickrwilson:
@Vimal-OSI There was a version 12 module for the submit state, it's found here: https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_request_submit
and there is an open PR for the migration here: #1153 

Since that module has been migrated can you revert these submitted changes?

CC: @osi-scampbell

Vimal-OSI:
I'm closing this PR,because of the similar changes are available in this PR #1153

PR #1237, opened 2021-09-07
[14.0][FIX] stock_move_location :  onchange picking_type_id

### LINKED ISSUE #1233: [14.0] stock_move_location needs system rights
Today, after installing stock_move_location on v14.0,
I realized that when clicking on Immediate transfer or Planned transfer on the wizard, an authorization error was raised for a non admin user, stating that it needs access to action_window.

These accesses have been removed in v14 (these were allowed on previous versions) with https://github.com/odoo/odoo/commit/de4213b771f641cbc218ebb1f759b09e064100bf#diff-e275aaa5de11da2815c2f0cc18bc046f967df5e4eba94999ee55ffe8926c87bd

Therefore the module should be reworked to create directly the action to be executed when clicking the buttons rather than trying to read an existing one for modifying it afterwards here : https://github.com/OCA/stock-logistics-warehouse/blob/183fb6443bdbfbe6a98415b4a356b1d376ec6117/stock_move_location/wizard/stock_move_location.py#L225

Also, I have noticed some other issues, which are not so critical. or possible improvements, I list them here since these might be tackled at the same time : 

- [x] IMP : origin and / or destination location could be filled from operation type if specified (#1237)
- [ ] IMP : add product in move line tree is not needed and confusing since all products in source locations are already filled, therefore adding a product from another location makes no sense since it will cause the picking to fail, also the tree view shown when clicking on add product only list IDs without other columns, which is not useable --> add product may need to be removed

jordancoest:
fix right issues reported in #1233 + IMP : origin and / or destination location filled from operation type if specified

### DIFF
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -138,6 +138,13 @@ class StockMoveLocationWizard(models.TransientModel):
                     )
         return res
 
+    @api.onchange("picking_type_id")
+    def _onchange_picking_type_id(self):
+        if self.picking_type_id.default_location_src_id:
+            self.origin_location_id = self.picking_type_id.default_location_src_id
+        if self.picking_type_id.default_location_dest_id:
+            self.destination_location_id = self.picking_type_id.default_location_dest_id
+
     @api.onchange("origin_location_id")
     def _onchange_origin_location_id(self):
         if not self.env.context.get("origin_location_disable", False):


remi-filament (approved):
Tested OK, thanks @jordancoest !

stock_move_location/wizard/stock_move_location.py
```diff
@@ -222,11 +229,16 @@ def action_move_location(self):
         return self._get_picking_action(picking.id)
 
     def _get_picking_action(self, pickinig_id):
-        action = self.env.ref("stock.action_picking_tree_all").read()[0]
```
rousseldenis: This is correct, but should be done with sudo

rousseldenis (changes requested):

stock_move_location/wizard/stock_move_location.py
remi-filament: Do you think it should be preferred to recreating an action as done below ?

rousseldenis: Nope, in fact, the good way is:

action = self.env['ir.actions.act_window']._for_xml_id("stock.action_picking_tree_all")

remi-filament: Thanks, just did, it works !

rousseldenis:
@remi-filament Could you rebase too ?

remi-filament:
Actually the fix was already made with #1326 here : https://github.com/OCA/stock-logistics-warehouse/commit/1a446ffbcfdfc60da7544b3750072fe4c4f5c227
Therefore I only kept the onchange on picking_type_id

rousseldenis (approved):

rousseldenis:
@jordancoest Could you rebase ?

remi-filament:
@rousseldenis done !

rousseldenis:
@remi-filament Thanks. Do we agree there is no change in access rights ?

remi-filament:
Yes, already covered in https://github.com/OCA/stock-logistics-warehouse/pull/1326
Jordan left the team and I am not allowed to change the title of the PR though !

PR #1244, opened 2021-09-13
[IMP] stock_cycle: save field readonly(warehouse_ids)

celm1990:
OCA conteo cliclico


rousseldenis (approved):
Seems legit indeed

HaraldPanten (approved):

celm1990:
@pedrobaeza I see the problem is another module, you can retry please?
image

pedrobaeza:
If the problem is still there, the retry will fail again.

PR #1246, opened 2021-09-15
[13.0][IMP] stock_lot_filter_available: Add more conditions to avoid errors

CarlosRoca13:
cc @Tecnativa TT31888

ping @sergio-teruel @carlosdauden

### DIFF
--- a/stock_lot_filter_available/models/stock_production_lot.py
+++ b/stock_lot_filter_available/models/stock_production_lot.py
@@ -1,6 +1,8 @@
 # Copyright 2021 Tecnativa - Carlos Roca
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import _, api, fields, models
+import operator as operator_lib
+
+from odoo import _, fields, models
 
 
 class StockProductionLot(models.Model):
@@ -8,12 +10,16 @@ class StockProductionLot(models.Model):
 
     product_qty = fields.Float(search="_search_product_qty")
 
-    @api.model
-    def _is_unsupported_search_operator(self, operator):
-        return operator != ">"
-
     def _search_product_qty(self, operator, value):
-        if self._is_unsupported_search_operator(operator) or value:
+        operator_mapping = {
+            ">": operator_lib.gt,
+            ">=": operator_lib.ge,
+            "<": operator_lib.lt,
+            "<=": operator_lib.le,
+            "=": operator_lib.eq,
+            "!=": operator_lib.ne,
+        }
+        if operator not in operator_mapping:
             raise ValueError(_("Unsupported search"))
         ids = []
         domain = [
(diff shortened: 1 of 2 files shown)


stock_lot_filter_available/models/stock_production_lot.py
```diff
@@ -23,7 +24,18 @@ def _search_product_qty(self, operator, value):
...
+            or (operator == "<=" and value >= 0)
+        ):
+            ids.extend(
+                self.env["stock.production.lot"]
+                .search([("id", "not in", involved_lot_ids)])
+                .ids
+            )
         for group in groups:
-            if group["quantity"] > value:
+            if safe_eval("{} {} {}".format(group["quantity"], operator, value)):
```
pedrobaeza: Please don't use `safe_eval`, but use the operators functions and call them:

https://docs.python.org/3/library/operator.html

```python
mapping = {
    ">": operator.gt,
    "<": operator.lt,
    ...
}
mapping[operator](group["quantity"], value)
```

CarlosRoca13:
@pedrobaeza changed

stock_lot_filter_available/models/stock_production_lot.py
```diff
@@ -23,7 +23,26 @@ def _search_product_qty(self, operator, value):
         groups = self.env["stock.quant"].read_group(
             domain, ["lot_id", "quantity"], ["lot_id"]
         )
+        involved_lot_ids = [group["lot_id"][0] for group in groups]
+        operator_mapping = {
+            ">": operator.gt,
```
pedrobaeza: You need to do `import operator` at the top.

pedrobaeza: And there will be a conflict on the name, so you should do `import operator as operator_lib` or similar, and use it with `operator_lib.gt`, etc.

CarlosRoca13:
Yes, sorry, changed

stock_lot_filter_available/models/stock_production_lot.py
```diff
@@ -10,9 +12,9 @@ class StockProductionLot(models.Model):
 
     @api.model
     def _is_unsupported_search_operator(self, operator):
-        return operator != ">"
+        return operator not in ["=", "!=", ">", ">=", "<", "<="]
 
-    def _search_product_qty(self, operator, value):
+    def _search_product_qty(self, op, value):
```
pedrobaeza: Is the initial signature with `operator` arg? If so, we shouldn't change it, but do the other thing I explained.

stock_lot_filter_available/models/stock_production_lot.py
```diff
@@ -10,9 +12,9 @@ class StockProductionLot(models.Model):
...
         if self._is_unsupported_search_operator(operator) or value:
```
carlosdauden: You can define operator_mapping at this point and check if operator is in dictionary to avoid keep _is_unsupported_search_operator method

CarlosRoca13:
@pedrobaeza @carlosdauden changed

CarlosRoca13:
please, can you review it again?

pedrobaeza (approved):

PR #1249, opened 2021-09-16
[14.0][MIG] stock_orderpoint_route: Migration to 14.0

dzungtran89:


stock_orderpoint_route/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "website": "https://github.com/OCA/stock-logistics-warehouse",
     "author": "ForgeFlow, Camptocamp, Odoo Community Association (OCA)",
     "category": "Warehouse",
-    "depends": ["stock"],
+    "depends": ["sale", "stock"],
```
grindtildeath: Any reason to add this dependency?

grindtildeath (changes requested):
LG, out of the extra dependency.

stock_orderpoint_route/__manifest__.py
dzungtran89: hello @grindtildeath , thx for your review. It seems to be a mistake, I had tested on a db with pre-installed `sale` module, and if I didn't put `sale` as dependency, the test case would have been failed due to a constraint notnull of `sale_line_warn`. I should have tested on a fresh db instead, I'll update the PR

dzungtran89: @grindtildeath I pushed the update

simahawk (approved):

mmequignon (approved):

dzungtran89:
we're checking, the test cases of module `stock_request` failed

nilshamerlinck:
Hi @simahawk we suspect a non-deterministic issue in that case, but we have run a few builds on our side and the failure rate is low; while we investigate further, would you mind trying to merge it again?

nilshamerlinck:
Hi @simahawk this PR will unblock the failing merge: https://github.com/OCA/stock-logistics-warehouse/pull/1262

PR #1252, opened 2021-09-17
[12.0][FIX] stock_available_info_popup: warehouse context

chienandalu:
It's necessary to provide the proper warehouse and scheduled date context to keep the coherence with the main computations that filter quantities depending on them.

TT31716

please review @ernestotejeda @pedrobaeza


pedrobaeza (approved):
Odoo does this correctly in v13?

chienandalu:
Well, in v13 this extension still exists but extending the core, so the patch is needed as well.

carlosdauden (approved):

pedrobaeza:
/ocabot merge patch

Please move it to v13 then.

PR #1258, opened 2021-09-21
[14.0][ADD] stock_location_warehouse

BernatPForgeFlow:
This module adds the associated warehouse to every location. If we have a tree of locations and sublocations, this module will update the associated warehouse every time we modify a location parent.
If a location changes its father location, it will change its associated warehouse and all its childs too.

@forgeflow


BernatPForgeFlow:
@Forgeflow

stock_location_warehouse/models/stock_location.py
```diff
@@ -0,0 +1,25 @@
+from collections import OrderedDict
+
+from odoo import api, fields, models
+
+
+class StockLocation(models.Model):
+
+    _inherit = "stock.location"
+    warehouse_id = fields.Many2one(
+        comodel_name="stock.warehouse", compute="_compute_warehouse_id"
```
CasVissers-360ERP: @BernatPForgeFlow Nice module!!

Any reason why you didn't make the field stored? Think it would be easy to have the field available in the database and UI thing like grouping?

stock_location_warehouse/models/stock_location.py
```diff
@@ -0,0 +1,25 @@
+from collections import OrderedDict
+
+from odoo import api, fields, models
+
+
+class StockLocation(models.Model):
+
+    _inherit = "stock.location"
```
MiquelRForgeFlow:
```suggestion
    _inherit = "stock.location"

```
aesthetics

stock_location_warehouse/models/stock_location.py
```diff
@@ -0,0 +1,25 @@
...
+    warehouse_id = fields.Many2one(
+        comodel_name="stock.warehouse", compute="_compute_warehouse_id", store=True
+    )
+
+    @api.depends("parent_path")
+    def _compute_warehouse_id(self):
+        warehouses = self.env["stock.warehouse"].search(
```
MiquelRForgeFlow:
```suggestion
        warehouses = self.env["stock.warehouse"].with_context(active_test=False).search(
```

stock_location_warehouse/models/stock_location.py
BernatPForgeFlow: @CasVissers-360ERP Yes, nice observation. I deleted the stored value because the previous function had nested loops with locations. Now we have exactly backported just what odoo15 does and it works fine. If I put `store=True`, then code stops working and it would need to be refactored again, and it doesn't seem trivial to me.

JordiBForgeFlow (approved):

CasVissers-360ERP (approved):
Useful addition.

stock_location_warehouse/models/stock_location.py
BernatPForgeFlow: @CasVissers-360ERP Finally the code is refactored and the field `warehouse_id` is stored. How do you see it?

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
Functional review, LGTM

JordiBForgeFlow:
@LoisRForgeFlow can you merge?

LoisRForgeFlow:
> @LoisRForgeFlow can you merge?

I'm waiting travis to see that tests go green.

PR #1266, opened 2021-10-01
[FIX] stock_orderpoint_route: avoid missing route_ids field issue

hailangvn:
This is to avoid issue below for any module that inherits tree view of
the model `stock.warehouse.orderpoint`. Example: `stock_orderpoint_manual_procurement`

Field route_ids used in field route_id default domain
([('id', 'in', route_ids)]) must be present in view but is missing.

The tree view defines `route_id` at https://github.com/odoo/odoo/blob/038816725a988a0ada3d6f8518a436ea46a81eba/addons/stock/views/stock_orderpoint_views.xml#L52

### DIFF
--- a/stock_orderpoint_route/views/stock_warehouse_orderpoint_views.xml
+++ b/stock_orderpoint_route/views/stock_warehouse_orderpoint_views.xml
@@ -17,4 +17,21 @@
             </field>
         </field>
     </record>
+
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
Fastracking as it is breaking other module migration:

/ocabot merge minor

PR #1268, opened 2021-10-04
[FIX] sale_stock_available_info_popup: cache miss

chienandalu:
cc @Tecnativa TT29615

please review @CarlosRoca13 @ernestotejeda

### DIFF
--- a/sale_stock_available_info_popup/models/sale_order.py
+++ b/sale_stock_available_info_popup/models/sale_order.py
@@ -16,10 +16,9 @@ class SaleOrderLine(models.Model):
     @api.depends("product_id", "product_uom_qty")
     def _compute_immediately_usable_qty_today(self):
         qty_processed_per_product = defaultdict(lambda: 0)
-        remaining = self.env["sale.order.line"]
+        self.immediately_usable_qty_today = False
         for line in self.sorted(key=lambda r: r.sequence):
             if not line.display_qty_widget:
-                remaining |= line
                 continue
             product = line.product_id
             qty_processed = qty_processed_per_product[product.id]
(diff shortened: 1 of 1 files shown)


sale_stock_available_info_popup/models/sale_order.py
```diff
@@ -16,15 +16,13 @@ class SaleOrderLine(models.Model):
     @api.depends("product_id", "product_uom_qty")
     def _compute_immediately_usable_qty_today(self):
         qty_processed_per_product = defaultdict(lambda: 0)
-        remaining = self.env["sale.order.line"]
+        self.immediately_usable_qty_today = False
```
pedrobaeza, line 19: False or 0?

chienandalu: Works either way. Odoo sets similar fields to `False` by default so I stuck with that choice

sale_stock_available_info_popup/models/sale_order.py
```diff
@@ -16,15 +16,13 @@ class SaleOrderLine(models.Model):
...
             if not line.display_qty_widget:
-                remaining |= line
                 continue
             product = line.product_id
             qty_processed = qty_processed_per_product[product.id]
             line.immediately_usable_qty_today = (
                 product.immediately_usable_qty - qty_processed
             )
             qty_processed_per_product[product.id] += line.product_uom_qty
-        remaining.write({"immediately_usable_qty_today": False})
```
pedrobaeza, line 30: @ernestotejeda a write should never be done in a `compute`, but `update`. I think this is also happening in v12.

CarlosRoca13 (approved):

pedrobaeza (approved):

pedrobaeza:
Please remove the write thing in 12.0 as well

/ocabot merge patch

PR #1269, opened 2021-10-04
[13.0] procurement_auto_create_group: Set partner on created procurement group

jbaudoux:
When a SO is confirmed, the created procurement group contains the shipping address.
On a stock rule, you can define a shipping address. So if the stock rule is configured to create a new procurement group, that one should also contain the shipping address.
This allows compatibility with the module https://github.com/OCA/stock-logistics-workflow/tree/13.0/stock_picking_group_by_partner_by_carrier
that will regroup delivery moves for a same partner in a same delivery picking.

cc @JordiBForgeFlow @LoisRForgeFlow  

Note for merging: use nobump option (the version has been manually updated)

### DIFF
--- a/procurement_auto_create_group/models/procurement_group.py
+++ b/procurement_auto_create_group/models/procurement_group.py
@@ -1,8 +1,12 @@
 # Copyright 2017-2020 ForgeFlow, S.L.
+# Copyright 2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, models
-from odoo.exceptions import UserError
+import logging
+
+from odoo import api, models
+
+_logger = logging.getLogger(__name__)
 
 
 class ProcurementGroup(models.Model):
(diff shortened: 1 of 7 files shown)


procurement_auto_create_group/models/stock_rule.py
```diff
@@ -13,3 +15,13 @@ class StockRule(models.Model):
...
             self.auto_create_group = False
+
+    @api.model
+    def _prepare_auto_procurement_group_data(self):
+        name = self.env["ir.sequence"].next_by_code("procurement.group") or False
+        if not name:
+            raise UserError(_("No sequence defined for procurement group."))
+        return {
+            "name": name,
+            "partner_id": self.partner_address_id.id,
```
TDu, line 25: As this function is declared as `api.model` this line will not work, I think.

TDu (commented):
Would be good also to update the tests accordingly.

procurement_auto_create_group/models/stock_rule.py
jbaudoux: Indeed. Fixed.

jbaudoux:
Test updated

procurement_auto_create_group/models/stock_rule.py
```diff
@@ -13,3 +15,12 @@ class StockRule(models.Model):
...
         if self.group_propagation_option != "propagate":
             self.auto_create_group = False
+
+    def _prepare_auto_procurement_group_data(self):
+        name = self.env["ir.sequence"].next_by_code("procurement.group") or False
+        if not name:
+            raise UserError(_("No sequence defined for procurement group."))
+        return {
+            "name": name,
+            "partner_id": self.partner_address_id.id,
```
LoisRForgeFlow, line 25: What if you don't have partner_address_id in this rule?

jbaudoux: it will evaluate to False
```
>>> env['res.partner'].id
False
```

LoisRForgeFlow: That's good then :+1:

LoisRForgeFlow (approved):
:+1: 

Note for merging: use nobump option (the version has been manually updated)

TDu (approved):

procurement_auto_create_group/models/procurement_group.py
```diff
@@ -1,34 +1,25 @@
...
             and values.get("date_planned")
         ):
-            group_data = self._prepare_auto_procurement_group_data()
+            group_data = rule._prepare_auto_procurement_group_data()
             group = self.env["procurement.group"].create(group_data)
             values["group_id"] = group
-        return result
-
-    @api.model
-    def _prepare_auto_procurement_group_data(self):
```
simahawk, line 28: IMO you should not remove this but:

* log a deprecation msg
* update the group_data with it till it's removed

jbaudoux: done

jbaudoux:
squashed

simahawk (approved):

simahawk:
bad command, build canceled.

/ocabot merge minor

PR #1270, opened 2021-10-05
[FIX] stock_request: add delete permission to group_stock_request_user

albariera:
I add delete permissions to the group_stock_request_user, because when creating a stock.request.order, the user could not delete items.


etobella (approved):
It was present on 12.0

:+1:

LoisRForgeFlow (approved):

PR #1273, opened 2021-10-06
[13.0][stock_fifo_update_kit_cost][new]

JordiBForgeFlow:
This module allows the cost of a product that is sold as a kit to be updated automatically when the components are delivered.

@ForgeFlow

Todo: 
* [ ] Add tests

### DIFF
--- a/stock_fifo_update_kit_cost/models/product.py
+++ b/stock_fifo_update_kit_cost/models/product.py
@@ -1,0 +1,30 @@
+# Copyright 20121 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+class Product(models.Model):
+    _inherit = "product.product"
+
+    def _run_fifo(self, quantity, company):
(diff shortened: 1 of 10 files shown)


stock_fifo_update_kit_cost/models/product.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import models
+
+
+class Product(models.Model):
+    _inherit = "product.product"
+
+    def _run_fifo(self, quantity, company):
+        vals = super(Product, self)._run_fifo(quantity, company)
+        phantom_boms = (
```
AaronHForgeFlow, line 12: Isn't it possible to use de _bom_find method? https://github.com/odoo/odoo/blob/13.0/addons/mrp/models/mrp_bom.py#L186

JordiBForgeFlow: @AaronHForgeFlow we checked, and it could find the wrong bom (not the phantom one)

AaronHForgeFlow (commented):
Functional test :+1:  pending to fix pre-commit

rousseldenis:
@JordiBForgeFlow

AaronHForgeFlow:
Just rebased the branch

PR #1282, opened 2021-10-23
[14.0][ADD] stock_production_lot_country_of_origin

rousseldenis:


JordiBForgeFlow:
@rousseldenis Finally how did you integrate with Intrastat?

Have you considered simply using the country of the supplier identified in the invoice during the intrastat declaration?

rousseldenis:
> @rousseldenis Finally how did you integrate with Intrastat?
> 
> Have you considered simply using the country of the supplier identified in the invoice during the intrastat declaration?

@JordiBForgeFlow I did this : https://github.com/OCA/intrastat-extrastat/pull/150

This is a corner case as our customer uses one product for several possible origins (we manage them through lots).

As the country of origin should be the one where products were manufactured, this is the only mean we found to do it (quite easily).

https://www.taxdoo.com/en/blog/intrastat-changes-vat-id-and-country-of-origin-10447/

rousseldenis:
@apineux @baimont @sanrav A little review here ?

PR #1284, opened 2021-10-25
[12.0][ADD] stock_realign_move_and_quants

robinkeunen:
A bug in Odoo 12 leads to the desynchronisation of the stock
quants and the stock move lines. The root cause has yet to be found.
This modules fixes the symptoms and not the cause. It aims to
1. Realign the stock quants and the stock move lines
2. Correct on the fly future desynchronisation

To do that, we assume the source of truth are the move lines rather than
the quants.

CAREFUL : this module has not been tested on a warehouse with multi-location or using packages.

More information on this issue: https://github.com/odoo/odoo/issues/62139


robinkeunen:
Compared to @gurneyalex 's gists, I changed

- the comparison in the wizard uses float_compare
- removed `def fix_reserved_quantity_in_quants` in stock.quant (unused)
- adding inventory line, fixing the quants and computing inventory line's theoretical quantity is done in one pass on the problematic quantities.

rousseldenis:
@robinkeunen Could you rebase ?

robinkeunen:
@rousseldenis here you go

rousseldenis (approved):
Code review. LGTM

rousseldenis:
@robinkeunen Don't forget to install pre-commit locally

PR #1279, opened 2021-10-18
[14.0] [ADD] stock_inventory_revaluation_wip: Added new module in the repo.

Chandresh-OSI:
Depends on #1227

PR #1289, opened 2021-10-28
[14.0][ADD] stock_inventory_revaluation_wip

patrickrwilson:
Depends on #1227


patrickrwilson:
@Chandresh-SerpentCS here is the new PR you requested since the other one was close (PR #1279 )

dreispt:
@patrickrwilson @Chandresh-SerpentCS Pelse fix warning: `odoo.models: The model add.multi.product has no _description`

rousseldenis:
@patrickrwilson This is not an addition. This module already exists. Please feel free to migrate it instead.

https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_inventory_revaluation

dreispt (changes requested):
@Chandresh-SerpentCS 
This is missing a dependency on `mrp_account_analytic_wip`.
That is the module that assigns "Cost Type" Products to workcenter operations.

Also, I see references to `analytic_cost_ids` in `stock_inventory_revaluation_mrp` that should be not there, as that module does not have and should not have those dependencies.
Those references should be added as extension in this module.

rousseldenis:
@patrickrwilson @dreispt What's the status of this ?

dreispt:
We are doing QA on this and has active WIP.
I was considering moving it away from here, as I see little probability to see it merged.

PR #1293, opened 2021-10-29
[11.0][FIX] stock_request: tests

AaronHForgeFlow:
Basically an access error was raising before the validation error. In this case the validation error has nothing to do with the user that triggers it. The admin user will get the error so the test will still be valid as it is now.

@LoisRForgeFlow


LoisRForgeFlow (approved):

LoisRForgeFlow:
The problem is only present in v11. Since it is an old verison, let's accept this quick patch for the sake of a healthy travis.

/ocabot merge nobump  

(only changing tests, no bump needed)

PR #1299, opened 2021-11-03
[14.0][ADD] stock_orderpoint_origin: add source documents to origin field

dreispt:


stock_orderpoint_origin/models/procurement_group.py
```diff
@@ -0,0 +1,34 @@
...
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def run(self, procurements, raise_user_error=True):
+        # Store the reorder source Procurement Groups
+        # To be used for reference and link navigation
+        # Also updates the Origin field of the PO
+        Forecast = self.env["report.stock.report_product_product_replenishment"]
```
rousseldenis, line 15: That sounds strange to add business data from reporting data ...

rousseldenis (changes requested):
This sounds strange as mixing different concepts.

Could you please do a PR per module?

Could you please rename module as there is no reference to orderpoints in these ?

stock_orderpoint_origin/models/procurement_group.py
dreispt: How else can you know what originated the demand?

dreispt:
@rousseldenis 
Notice that the two modules are added in separate commits. But submitting them together makes sense, since the second one is a simple module link and does not stand on itself.

This feature if for "orderpoints" aka Replenishments.
It adds information to the procurements generated by it, even if its code is not touched directly. So I feel the name is appropriate.

rousseldenis:
> This feature if for "orderpoints" aka Replenishments.
> It adds information to the procurements generated by it, even if its code is not touched directly. So I feel the name is appropriate.

Nope. You mixed two different concepts.

Replenishments represent a 'stock demand' for a product on a location.

Orderpoints represent minimum stock rules for a product on a location.

You can trigger replenishment from different manners. Manually on the product, by the replenish wizard or by an orderpoint trigger when the product is lower the minimum level.

That's why my comment has sense. Please change this.

dreispt:
@rousseldenis
This features enriches the "Orign" data for POs generated from the "Replenishment" menu.
The "Replenishment" menu open a list for `stock.warehouse.orderpoint` records.
It uses a different filter and records than the Reordering rules, but they both share the same underlying model.

So this feature enhances what the Orderpoint model does, specifically when running from the Replenishment menu. This is why that name was picked.

What would you suggest as an alternative name?

rousseldenis:
Ok, I see.

No it's ok. Odoo put those data in a report, but can be useful to have those computations in model directly.

Indeed, the origin comes from move_out._get_source_document() but the way to get to that is quite long.

rousseldenis (approved):

dreispt:
@rousseldenis Thank you.

> No it's ok. Odoo put those data in a report, but can be useful to have those computations in model directly.

That was exactly my thought when this need came to my hands.
Unfortunately I did not find a direct way to find the demand origins, other than the Forecast report data.
And there was little point in re-implementing that report's logic.

This is a bit a "heuristic" approach, but the final result was useful to get a good idea of the answer to "why are we purchasing this?", even if it might not be 100% exact.

dreispt:
image

dreispt:
FYI added a fix for duplication issue found with multi-step routes.

Before:
image

After: 
image

PR #1301, opened 2021-11-08
[14.0][FIX] stock_mts_mto_rule: check mts+mto in mrp

alan196:
This PR add the functionality to use mts+mto in procurements created from MRP module.

### DIFF
--- a/stock_mts_mto_rule/models/__init__.py
+++ b/stock_mts_mto_rule/models/__init__.py
@@ -1,2 +1,3 @@
 from . import stock_rule
 from . import stock_warehouse
+from . import stock_move
--- a/stock_mts_mto_rule/models/stock_move.py
+++ b/stock_mts_mto_rule/models/stock_move.py
@@ -1,0 +1,47 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+from odoo.tools import float_compare, float_is_zero
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _adjust_procure_method(self):
+        precision = self.env["decimal.precision"].precision_get(
+            "Product Unit of Measure"
+        )
+        for move in self:
+            product_id = move.product_id
+            domain = [
+                ("location_src_id", "=", move.location_id.id),
+                ("location_id", "=", move.location_dest_id.id),
+                ("action", "!=", "push"),
+            ]
+            rules = self.env["procurement.group"]._search_rule(
+                False, product_id, move.warehouse_id, domain
+            )
+            if not rules or rules and rules.action != "split_procurement":
+                return super()._adjust_procure_method()
+            needed_qty = rules.get_mto_qty_to_order(
+                move.product_id, move.product_qty, move.product_uom, {}
+            )
+            if float_is_zero(needed_qty, precision_digits=precision):
+                move.procure_method = "make_to_stock"
+            elif (
+                float_compare(needed_qty, move.product_qty, precision_digits=precision)
+                == 0.0
+            ):
+                move.procure_method = "make_to_order"
+            else:
+                mts_qty = move.product_uom_qty - needed_qty
+                move.copy(
+                    {
+                        "product_uom_qty": needed_qty,
+                        "procure_method": "make_to_order",
+                    }
+                )
+                move.write(
+                    {"product_uom_qty": mts_qty, "procure_method": "make_to_stock"}
+                )
+                move._action_assign()


jonerikceberio:
Been working in a very similar solution, my two cents:

- We thought that it should be a separate module: mrp_mts_mto_rule, as the method is called in mrp components (not sure why the original method is in stock)
- The solution does not work without a rule with action split from stock to production. In our code we inherited the _get_global_route_rules_values (as the original stock_mts_mto_rule) to create such a rule.

So I understand that the afore code does not work without creating that rule manually, is that so?

jonerikceberio:
Indeed there is another possible problem: In the case of a double dependency of a component, e.g. A has a BoM with B and C, and B has a BoM with C, the method is not able to recognize that the stock of C has been used for producing A in the production of B. 
In the same example, with BoMs usin 1 component each, if you have 5 units of C and you sell 4 units of A, the method finds no needed_qty (zero) of C. This is because the make_to_stock moves are not confirmed and therefore they do not reserve stock. And so it uses the same 4 units twice.
If you don't mind I can PR your branch so you can see the differences with your code.

elvise:
@alan196 really cool! 
Any feedbacks about comments of @jonerikceberio ?

elvise:
> Indeed there is another possible problem: In the case of a double dependency of a component, e.g. A has a BoM with B and C, and B has a BoM with C, the method is not able to recognize that the stock of C has been used for producing A in the production of B.
> In the same example, with BoMs usin 1 component each, if you have 5 units of C and you sell 4 units of A, the method finds no needed_qty (zero) of C. This is because the make_to_stock moves are not confirmed and therefore they do not reserve stock. And so it uses the same 4 units twice.
> If you don't mind I can PR your branch so you can see the differences with your code.

@jonerikceberio good point! 
any progress in this direction?

rousseldenis:
@alan196 Could you rebase?

rousseldenis:
@alan196 Could you add a test that highlight this change ?

elvise:
@alan196 just a gentle reminder :)

elvise:
> > Indeed there is another possible problem: In the case of a double dependency of a component, e.g. A has a BoM with B and C, and B has a BoM with C, the method is not able to recognize that the stock of C has been used for producing A in the production of B.
> > In the same example, with BoMs usin 1 component each, if you have 5 units of C and you sell 4 units of A, the method finds no needed_qty (zero) of C. This is because the make_to_stock moves are not confirmed and therefore they do not reserve stock. And so it uses the same 4 units twice.
> > If you don't mind I can PR your branch so you can see the differences with your code.
> 
> @jonerikceberio good point! any progress in this direction?

@jonerikceberio just a gentle reminder :)

elvise:
> Indeed there is another possible problem: In the case of a double dependency of a component, e.g. A has a BoM with B and C, and B has a BoM with C, the method is not able to recognize that the stock of C has been used for producing A in the production of B. 
> 
> In the same example, with BoMs usin 1 component each, if you have 5 units of C and you sell 4 units of A, the method finds no needed_qty (zero) of C. This is because the make_to_stock moves are not confirmed and therefore they do not reserve stock. And so it uses the same 4 units twice.
> 
> If you don't mind I can PR your branch so you can see the differences with your code.


@alan196 thanks for your commit!
Do you have considered @jonerikceberio scenario ?

alan196:
@elvise no I don't 😞 currently I'm a bit busy, but If you or @jonerikceberio propose the change in a PR I'll take the time to review and make the proper tests.

Sadly, at this moment, I cannot take the time to develop that feature.

elvise:
@pedrobaeza have you time for this PR ?

pedrobaeza:
I'm not using this module anymore, sorry.

elvise:
@rousseldenis merge ? :)

rousseldenis:
> @rousseldenis merge ? :)

As said above, tests should be added.

Moreover, I would say that code that depends on mrp should be added in a dedicated module.

Somehow `stock_mts_mto_mrp_rule`.

elvise:
> > @rousseldenis merge ? :)
> 
> As said above, tests should be added.
> 
> Moreover, I would say that code that depends on mrp should be added in a dedicated module.
> 
> Somehow `stock_mts_mto_mrp_rule`.

Sorry! Got it

elvise:
> > @rousseldenis merge ? :)
> 
> As said above, tests should be added.
> 
> Moreover, I would say that code that depends on mrp should be added in a dedicated module.
> 
> Somehow `stock_mts_mto_mrp_rule`.

@alan196 have you time for follow these points ?

francesco-ooops:
Hi @alan196, do you think you can add tests soon? 

or is it ok if we complete your work in a new PR in order to merge it?

Thanks!

francesco-ooops:
> Hi @alan196, do you think you can add tests soon?
> 
> or is it ok if we complete your work in a new PR in order to merge it?
> 
> Thanks!

@alan196 gentle reminder :)

alan196:
@francesco-ooops sorry for the late response, is ok for me.

Sadly at this moment I don't have the time to write tests, but as soon as you make the PR please tag me and I will review

francesco-ooops:
@alan196 sure thanks!

PR #1303, opened 2021-11-09
[14.0][IMP] stock_available_unreserved: Add hooks

rousseldenis:


sebalix (approved):

rousseldenis:
/ocabot merge patch Fast tracking this as no breaking change

pedrobaeza:
Why 2 methods and one empty? If you want to hook into it, you can always inherit and call super before and after, so I think it's enough with one, isn't it?

About the ocabot command, you should put it in a separate line, as all the line is parsed.

rousseldenis:
> Why 2 methods and one empty? I

The first one was there in v10. I've let it in order to keep compatibility for people doing migrations (and we all know that implementation changes are costly). https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_available_unreserved/models/product.py#L83

PR #1304, opened 2021-11-10
[14.0][IMP] stock_available_unreserved make field optional in tree view

remi-filament:
Add optional attribute for tree view so that field can be hidden (like for the other ones in tree view)

### DIFF
--- a/stock_available_unreserved/views/product_view.xml
+++ b/stock_available_unreserved/views/product_view.xml
@@ -6,7 +6,12 @@
         <field name="inherit_id" ref="stock.view_stock_product_template_tree" />
         <field name="arch" type="xml">
             <field name="qty_available" position="after">
-                <field name="qty_available_not_res" />
+                <field
+                    name="qty_available_not_res"
+                    attrs="{'invisible':[('type', '!=', 'product')]}"
+                    optional="show"
+                    decoration-danger="qty_available_not_res &lt; 0"
+                />
             </field>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


rousseldenis:
@remi-filament Could you check pre-commit ? You should install it prior committing locally.

remi-filament:
Hi, I had no issue reported locally, the issue reported with pre-commit seems to come from JS file from stock_inventory_discrepancy...
I am not sure why this was not reported before that module was merged ?

rousseldenis:
@remi-filament This is the real error : Can you fix it ?

image

remi-filament:
Sure, sorry I missed it, the change was pending in my unstaged files !

rousseldenis (approved):

jordancoest (approved):

pedrobaeza:
This should be reverted ASAP, as adding a computed field in the tree makes the view terribly slow, having to do the computation for the number of elements in the tree (usually 80), and it doesn't matter you have it hidden, it will always be computed.

rousseldenis:
> This should be reverted ASAP, as adding a computed field in the tree makes the view terribly slow, having to do the computation for the number of elements in the tree (usually 80), and it doesn't matter you have it hidden, it will always be computed.

What are you talking about ? No field was added

pedrobaeza:
Indeed, this PR is not adding it (only making it optional), sorry, but the rationale is the same, so I think the field should be removed from the view.

rousseldenis:
> Indeed, this PR is not adding it (only making it optional), sorry, but the rationale is the same, so I think the field should be removed from the view.

There from all versions, so your ask is unrelevant. That should not be changed.

pedrobaeza:
Then at least a note on the README can be added warning about that IMO.

PR #1305, opened 2021-11-11
[14.0][ADD] stock_exception

newtratip:


stock_exception/models/stock.py
```diff
@@ -0,0 +1,49 @@
...
+            self.ignore_exception = False
+
+    def action_confirm(self):
+        if self.detect_exceptions() and not self.ignore_exception:
+            return self._popup_exceptions()
+        return super().action_confirm()
+
+    @api.model
+    def _get_popup_action(self):
+        action = self.env.ref("stock_exception.action_stock_exception_confirm")
```
rousseldenis, line 48:
```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id("stock_exception.action_stock_exception_confirm")
```

rousseldenis (changes requested):
Code review. Looks great. One change

rousseldenis:
@newtratip Could you attend comments ?

stock_exception/models/stock.py
newtratip: Fixed. may i ask, why change `self.env["ir.actions.act_window"]._for_xml_id(...)` is better `self.env.ref(...)` ?

newtratip: @rousseldenis, I used this code `self.env["ir.actions.act_window"]._for_xml_id(...)` is error, I rollback the code.
Selection_735

rousseldenis: Ok, I didn't know the base_exception behaviour.

It does the sudo().read. But that's that part which is not correct (in base exception).

Of course you can keep your code.

Saran440 (approved):
LGTM, Functional Test on Production :+1:

rousseldenis (approved):

PR #1306, opened 2021-11-11
[14.0][MIG] stock_inventory_cost_info: Migration to 14.0

victoralmau:
Migration to 14.0

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT31837


stock_inventory_cost_info/hooks.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def pre_init_hook(cr):
+    cr.execute(
+        """ALTER TABLE stock_inventory_line
+    ADD COLUMN adjustment_cost numeric
+    DEFAULT 0"""
+    )
+
+    cr.execute(
```
rousseldenis, line 12: Maybe an explanation on this as setting a default and then removing it ???

stock_inventory_cost_info/models/stock_inventory.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class InventoryLine(models.Model):
+    _inherit = "stock.inventory.line"
+
+    currency_id = fields.Many2one(
+        string="Currency", related="inventory_id.company_id.currency_id", readonly=True
```
rousseldenis:
```suggestion
        string="Currency", related="inventory_id.company_id.currency_id"
```

rousseldenis (changes requested):
Code review. Little change and one question

stock_inventory_cost_info/hooks.py
victoralmau: Please @ernestotejeda, can you clarify if a default is not really necessary and reduce code?

pedrobaeza: This is needed for speeding up installation and have consistent info. If not executed, all existing inventory lines will be filled with current cost price, which is not correct and slow.

sergio-teruel: It would not be enough create a column without  the default, I do not see where the field is filled..

pedrobaeza: Creating it without default would lead to null values. We should think if we want such null values and how they are seen in pivot table

sergio-teruel: Oks...

sergio-teruel (approved):

pedrobaeza:
@rousseldenis is it ok for you?

rousseldenis (approved):

PR #1307, opened 2021-11-11
[14.0][MIG] stock_putaway_product_template: Migration to 14.0

victoralmau:
Migration to 14.0

Please, @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT31847


rousseldenis (approved):
Code review

stock_putaway_product_template/models/stock_putaway_rule.py
```diff
@@ -19,16 +19,15 @@ class StockPutawayRule(models.Model):
     @api.depends("product_id")
     def _compute_product_tmpl_id(self):
         for rec in self:
-            rec.product_tmpl_id = False
```
pedrobaeza: There's a path where you don't assign any value, so I still see this as needed.

victoralmau: If I'm not mistaken v14 already defines a default value in the registers with no value in a compute method and therefore takes care of this. In views it works fine if that's what you mean.

pedrobaeza: OK, I see that is a computed writeable, so it's OK then.

pedrobaeza (approved):

PR #1308, opened 2021-11-15
[14.0][FIX] stock_order_point_move_link get move values

TDu:
The `order_point_id` can be set as an empty recordset by the core.
Which breaks later in a write.

### DIFF
--- a/stock_orderpoint_move_link/models/stock.py
+++ b/stock_orderpoint_move_link/models/stock.py
@@ -27,8 +27,8 @@ class StockRule(models.Model):
             company_id,
             values,
         )
-        if "orderpoint_id" in values:
+        if "orderpoint_id" in values and values["orderpoint_id"]:
             vals["orderpoint_ids"] = [(4, values["orderpoint_id"].id)]
-        elif "orderpoint_ids" in values:
+        elif "orderpoint_ids" in values and values["orderpoint_ids"]:
             vals["orderpoint_ids"] = [(4, o.id) for o in values["orderpoint_ids"]]
         return vals


bodedra (commented):
LG: +1

Can you do same thing for `orderpoint_ids`

TDu:
> LG: +1
> 
> Can you do same thing for `orderpoint_ids`

I did, although not as critical because of the loop on `values["orderpoint_ids"]` values.

stock_orderpoint_move_link/models/stock.py
```diff
@@ -27,8 +27,8 @@ def _get_stock_move_values(
             company_id,
             values,
         )
-        if "orderpoint_id" in values:
+        if "orderpoint_id" in values and values["orderpoint_id"]:
```
mmequignon, line 30: I would have written something like that
```suggestion
        orderpoints = values.get("orderpoint_id") or values.get("orderpoint_ids")
        if orderpoints:
            vals["orderpoint_ids"] = [(4, orderpoint.id) for orderpoint in orderpoints]
```

simahawk (approved):

stock_orderpoint_move_link/models/stock.py
rousseldenis: @TDu Do you want to apply this ?

TDu: @rousseldenis The change makes sense, but it is still a cosmetic one IMO. So I will not

rousseldenis (approved):

PR #1311, opened 2021-11-16
[14.0][FIX] stock_location_tray: crash when creating a stock.location

florentx:
Crash: Got a MemoryError when creating a new `stock.location` with tray_type_id set.

How to reproduce:
 - create a Tray Type on a new database
 - Click the "Locations" action on the top right
 - Hit "Create" and enter a new `stock.location` with this tray type

And boom!

### DIFF
--- a/stock_location_tray/models/stock_location.py
+++ b/stock_location_tray/models/stock_location.py
@@ -188,6 +188,7 @@ class StockLocation(models.Model):
                         "posz": posz,
                         "location_id": location.id,
                         "company_id": location.company_id.id,
+                        "tray_type_id": False,
                     }
                     values.append(subloc_values)
         if values:


florentx:
Some technical explanations of the issue and the fix.

 -> button box "Locations" on the tray type view is linked to method `open_locations`:
 ```
    def open_locations(self):
        action = self.env.ref("stock.action_location_form").read()[0]
        action["domain"] = [("tray_type_id", "in", self.ids)]
        if len(self.ids) == 1:
            action["context"] = {"default_tray_type_id": self.id}
        return action
```

 -> and due to this `default_tray_type_id`, the sub-locations of the Tray were created with same `tray_type_id` as their parent

 -> so there was an infinite recursion, because for each cell of the Tray new sub-cells were created with same `tray_type_id`, etc ...

 -> `MemoryError`

----
The fix is to ensure that `tray_type_id` is empty for cells which are created automatically in a Tray

StephaneMangin:
Normally, this kind of issue should be solved at the relational model level, this is clearly a bad relation management between parent and children, this should be avoided by adding a relational constraint on the model.

But if it works in all known cases, that's fine.

StephaneMangin (approved):
LG

sebalix (approved):

PR #1318, opened 2021-11-25
[13.0][IMP] stock_request: fixes the company selection on the procurement

etobella:
This PR fixes the company selection on the procurement

@LoisRForgeFlow


JordiBForgeFlow (approved):

JordiBForgeFlow:
The hook is not needed anymore, then right?

etobella:
The fix is needed

LoisRForgeFlow (approved):

PR #1325, opened 2021-12-09
[14.0][FIX] stock_location_tray: fix access on act_window in v14

florentx:
Tested locally, both changes behave as expected. Thank you @sebalix


sebalix (approved):

sebalix:
Fast tracking this one as the addon doesn't work properly if we are not admin.
/ocabot merge patch

sebalix:
@florentx I didn't see your commit message, the next time better to not include the version number in messages (excepting for migration commits), as commits could be ported from one branch to another, this makes the message irrelevant afterwards.

PR #1184, opened 2021-06-23
[13.0][MIG] stock_picking_type_user_restriction

dani-xtendoo:
stock_picking_type_restriction migrate to 13.0

PR #1329, opened 2021-12-15
[14.0][MIG] stock_picking_type_user_restriction

rousseldenis:
Based on :

https://github.com/OCA/stock-logistics-warehouse/pull/1184

with https://github.com/xtendoo-corporation/stock-logistics-warehouse/pull/1

Changes: 

- [x] More restrictive rule: people that are in 'User: Assigned Operation Types Only' group will really see only its picking and picking types


rousseldenis:
@eLBati I want to have your advice on this.

The rules are done with a OR on assigned_user_ids = False.
This implies all picking types that do not have a user is visible to everyone even with the group.
Shouldn't we add a group 'See all own picking operations and assigned to nobody' (with the rule assigned_user_ids = False with that group) ? Or does it add complexity ?

eLBati:
> This implies all picking types that do not have a user is visible to everyone even with the group

This follows the same principle followed by multi company rules: if no restriction on the record > the record is accessible.
Do you see problems with this?

dreispt (approved):
Code review

rousseldenis:
@eLBati @FrancoMaxime @SMaciasOSI  Could you review ?

marcelsavegnago (approved):
LGTM

PR #1333, opened 2021-12-17
[14.0][UPD] stock_warehouse_calendar: development status to stable.

LoisRForgeFlow:
It has been a while since this is realeased and it can be considered stable.


LoisRForgeFlow:
Fasttracking.

/ocabot merge nobump

PR #1334, opened 2021-12-17
[14.0][IMP] stock_cycle_count: check modify

MateuGForgeFlow:
Create method not_modify to enable inheritance and customization.
cc @LoisRForgeFlow

### DIFF
--- a/stock_cycle_count/models/stock_inventory.py
+++ b/stock_cycle_count/models/stock_inventory.py
@@ -90,11 +90,7 @@ class StockInventory(models.Model):
 
     def write(self, vals):
         for inventory in self:
-            if (
-                inventory.cycle_count_id
-                and "state" not in vals.keys()
-                and inventory.state == "draft"
-            ):
+            if inventory._allow_write(vals):
                 raise UserError(
                     _(
                         "You cannot modify the configuration of an Inventory "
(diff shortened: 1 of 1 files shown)


stock_cycle_count/models/stock_inventory.py
```diff
@@ -90,15 +90,16 @@ def create(self, vals):
...
+            if inventory.not_modify(vals):
                 raise UserError(
                     _(
                         "You cannot modify the configuration of an Inventory "
                         "Adjustment related to a Cycle Count."
                     )
                 )
         return super(StockInventory, self).write(vals)
+
+    def not_modify(self, vals):
```
LoisRForgeFlow: can you make this internal an put a more meaningful name? like `_allow_write`

stock_cycle_count/models/stock_inventory.py
```diff
@@ -90,15 +90,16 @@ def create(self, vals):
...
+        if self.cycle_count_id and "state" not in vals.keys() and self.state == "draft":
+            return True
+        return False
```
LoisRForgeFlow: simply: `return self.cycle_count_id and "state" not in vals.keys() and self.state == "draft"`

LoisRForgeFlow (approved):

PR #1341, opened 2021-12-27
[14.0][FIX/IMP] stock_mts_mto_rule: Allow split with no group / Improve tests

rousseldenis:


rousseldenis:
@PierrickBrun @florian-dacosta

elvise:
@rousseldenis can you rebase ?
(For runboat)

rousseldenis:
@florian-dacosta @PierrickBrun What do you think of this ?

rousseldenis:
@florian-dacosta @elvise

elvise (approved):
LGTM! (Functional)

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -93,7 +93,7 @@ def _run_split_procurement(self, procurements):
                 # to adjust the product's free qty
                 group_id = mts_procurement.values.get("group_id")
                 group_domain = expression.AND(
-                    [domain, [("group_id", "=", group_id.id)]]
+                    [domain, [("group_id", "=", group_id and group_id.id or False)]]
```
florian-dacosta, line 96: Actually, I have a hard time understanding why this block of code with the `moves_to_assign` exists...
The comment is about adapting free qty, which is not used by the module AFAIK.
I don't really like it because I don't think it is the purpose of this module to assign stuff...One may not want to assign stock moves automatically this way. But this matter is not related to the PR...

Anyway, if this part of code is really useful I guess the change make sense. But, without the group it could trigger the assignment of a huge number of moves and it could become a problem (from a performance point of view for instance).

Hard to approve the change since I don't understand this part...

stock_mts_mto_rule/tests/test_mto_mts_route.py
```diff
@@ -94,6 +129,52 @@ def test_mts_mto_route_split(self):
...
+                    "test",
+                    self.warehouse.company_id,
+                    vals,
+                )
+            ]
+        )
+        moves = self.env["stock.move"].search(
+            [("product_id", "=", self.product.id), ("group_id", "=", False)]
+        )
+        self.assertEqual(3, len(moves))
```
florian-dacosta, line 155: I wonder if the test would not be clearer if we add a `('state', '!=', 'done')` in the domain and then test that we have 2 moves. Because knowing the feature of this module, I know I run a procurement and expect 2 moves instead of 1, I though it was a bit confusing the have a test of a number of moves that is not really related to what we test, just because we had to make an inventory !
Well, it is really a minor comment, it is not blocking at all.

stock_mts_mto_rule/models/stock_rule.py
rousseldenis: Mmmh, long time. I need to check in which case this is needed.

I think this is in case of inter warehouse operations with stock requests (and no procurement group).

elvise: looks reasonable

PR #1345, opened 2022-01-07
[14.0][FIX] stock_request: use sudo to run procurements to prevent permission issues

alan196:
This fixes the following issue:
If a user makes a request for a product that you need to manufacture, but the user doesn't have permission to
MRP module and you don't want to give them access, it will raise an error that you don't have permission to
see the mrp.bom.


LoisRForgeFlow (approved):

LoisRForgeFlow:
Thanks, fast-tracking this fix:

/ocabot merge patch

PR #1346, opened 2022-01-10
14.0 mig stock reserve

mt-software-de:
This is the migration for stock_reserve
There is an other pull request for stock_reserve_sale https://github.com/OCA/stock-logistics-warehouse/pull/1347


rousseldenis (approved):

CasVissers-360ERP:
@rousseldenis 
Looks like your merge failed?

rousseldenis:
> @rousseldenis Looks like your merge failed?

It wasn't a merge but 'migration' command is to set the good message in Migration issue.

CasVissers-360ERP (approved):
Functional review.

CasVissers-360ERP:
Have been using the module for some time now, forgot to give feedback.

stock_reserve/models/stock_reserve.py
```diff
@@ -0,0 +1,218 @@
...
+        return res
+
+    @api.model
+    def get_location_from_ref(self, ref):
+        """Get a location from a xmlid if allowed
+        :param ref: tuple (module, xmlid)
+        """
+        try:
+            location = self.env.ref(ref, raise_if_not_found=True)
+            location.check_access_rule("read")
```
gdgellatly, line 102: I am not sure this is enough. In multicompany, you can pass access rule check but fail record rule I think. At least in multicompany test this is what I'm getting.

EDIT: Sorry this is wrong. It does apply the exception to Record Rules, just it needs to catch AccessError not except_orm

gdgellatly (commented):
I've given a thorough functional review of #1346 and #1347 and there is quite a bit I didn't like.

Firstly the generation of the Reservation location and the defaulting everywhere. It is 100% not multicompany compatible and just leads to constant errors. We shouldn't have a module that requires development out of the box to work MC.

Now in my case I solved it by creating an init hook that creates a Reservation location for each warehouse, added a warehouse_id column with onchanges, and set the picking type from that warehouse. Otherwise in multicompany you get 2 location errors and a picking type error by default. But really I feel like these defaults should just not exist at all.

We can default company and use an onchange there (and if adding a warehouse field, onchanges there for locations and company). In sale_stock_reserve, we can also get the warehouse from the sale.

I also feel like at the point of reserving stock, that perhaps it should try to reserve then, not when the scheduler runs, but not so important.

Next (in #1347) the sale views should allow draft,sent states not just draft. If anything, it is more likely you would reserve stock, after emailing a customer a quote.

Finally in sale for calculating is_stock_reservable of the line. That boolean is ultra confusing and doesn't work. Everything just ends up True that I tested, so I unrolled and just if/elif the whole thing and it works as expected. I think the issue probably comes from the procure method check, because in my customisation we unpack that to reserve the BoM raw materials. - sorry it is both together I reviewed as one will copy to other.

stock_reserve/models/stock_reserve.py
```diff
@@ -0,0 +1,218 @@
+# Copyright 2013 Camptocamp SA - Guewen Baconnier
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+from odoo.exceptions import except_orm
```
gdgellatly, line 4:
```suggestion
from odoo.exceptions import AccessError
```

stock_reserve/models/stock_reserve.py
```diff
@@ -0,0 +1,218 @@
...
+    @api.model
+    def get_location_from_ref(self, ref):
+        """Get a location from a xmlid if allowed
+        :param ref: tuple (module, xmlid)
+        """
+        try:
+            location = self.env.ref(ref, raise_if_not_found=True)
+            location.check_access_rule("read")
+            location_id = location.id
+        except (except_orm, ValueError):
```
gdgellatly, line 104:
```suggestion
        except (AccessError, ValueError):
```

rousseldenis:
> Firstly the generation of the Reservation location and the defaulting everywhere. It is 100% not multicompany compatible and just leads to constant errors. We shouldn't have a module that requires development out of the box to work MC.
> 
> Now in my case I solved it by creating an init hook that creates a Reservation location for each warehouse, added a warehouse_id column with onchanges, and set the picking type from that warehouse. Otherwise in multicompany you get 2 location errors and a picking type error by default. But really I feel like these defaults should just not exist at all.

@mt-software-de Could you either solve this missing features, either (at least) put this in readme/ROADMAP.rst ?

Could you also fix suggested changes ?

rousseldenis:
Or, @gdgellatly could you maybe provide the init hooks in a PR on @mt-software-de branch ?

mt-software-de:
sorry for not answering here. 
The only thing i have done was basically copying the code from version 14.0 to 15.0
I basically don't need the module anymore, because we achieved our goals differently. 
I will do the few changes here.

tarbib:
> sorry for not answering here. The only thing i have done was basically copying the code from version 14.0 to 15.0 I basically don't need the module anymore, because we achieved our goals differently. I will do the few changes here.

Hello @mt-software-de 
I'm curious, how did you manage to achieve your goals differently? We have a quite a reservation process to establish here and would be happy to hear about good practices

mt-software-de:
> > sorry for not answering here. The only thing i have done was basically copying the code from version 14.0 to 15.0 I basically don't need the module anymore, because we achieved our goals differently. I will do the few changes here.
> 
> Hello @mt-software-de I'm curious, how did you manage to achieve your goals differently? We have a quite a reservation process to establish here and would be happy to hear about good practices

Hey tarbib,
yes of course
we created two modules
- one module we called it stock_halt_rule_run_pull
hooks into _run_pull of stock.rule 
it is filtering the procurement.groups by a domain
with this we are able to stop a procurement.group's pull

- the second module is is able to make the reservation by
calling the _action_launch_stock_rule of the sale.order.line's
with the interaction of the other module we are able to stop the pull
and this leads to only a outgoing picking for the sale.order
the created moves are flaged with a new field so that we can disable the action_assign for those

I hope i could help you with it?

tarbib:
> > > sorry for not answering here. The only thing i have done was basically copying the code from version 14.0 to 15.0 I basically don't need the module anymore, because we achieved our goals differently. I will do the few changes here.
> > 
> > 
> > Hello @mt-software-de I'm curious, how did you manage to achieve your goals differently? We have a quite a reservation process to establish here and would be happy to hear about good practices
> 
> Hey tarbib, yes of course we created two modules
> 
> * one module we called it stock_halt_rule_run_pull
>   hooks into _run_pull of stock.rule
>   it is filtering the procurement.groups by a domain
>   with this we are able to stop a procurement.group's pull
> * the second module is is able to make the reservation by
>   calling the _action_launch_stock_rule of the sale.order.line's
>   with the interaction of the other module we are able to stop the pull
>   and this leads to only a outgoing picking for the sale.order
>   the created moves are flaged with a new field so that we can disable the action_assign for those
> 
> I hope i could help you with it?

Sorry I did not get the notification of your reply
Thanks a lot for it. We don't use the sale module, so we decided to go forward with stock reserve module.
Happy new year ;)

stock_reserve/models/product.py
```diff
@@ -0,0 +1,59 @@
...
+        for product in self:
+            product.reservation_count = sum(
+                product.product_variant_ids.mapped("reservation_count")
+            )
+
+    def action_view_reservations(self):
+        self.ensure_one()
+        ref = "stock_reserve.action_stock_reservation_tree"
+        product_ids = self.mapped("product_variant_ids.id")
+        action_dict = self.env.ref(ref).read()[0]
```
victoralmau, line 23:
```suggestion
        product_ids = self.mapped("product_variant_ids.id")
        action_dict = self.env["ir.actions.act_window"]._for_xml_id(
            "stock_reserve.action_stock_reservation_tree"
        )
```

stock_reserve/models/product.py
```diff
@@ -0,0 +1,59 @@
...
+                ("product_id", "=", product.id),
+                ("state", "in", ["draft", "assigned"]),
+            ]
+            reservations = self.env["stock.reservation"].search(domain)
+            product.reservation_count = sum(reservations.mapped("product_qty"))
+
+    def action_view_reservations(self):
+        self.ensure_one()
+        ref = "stock_reserve.action_stock_reservation_tree"
+        action_dict = self.env.ref(ref).read()[0]
```
victoralmau, line 52:
```suggestion
        action_dict = self.env["ir.actions.act_window"]._for_xml_id(
            "stock_reserve.action_stock_reservation_tree"
        )
```

stock_reserve/models/stock_reserve.py
```diff
@@ -0,0 +1,218 @@
...
+    @api.onchange("product_uom_qty")
+    def _onchange_quantity(self):
+        """ On change of product quantity avoid negative quantities """
+        if not self.product_id or self.product_uom_qty <= 0.0:
+            self.product_uom_qty = 0.0
+
+    def open_move(self):
+        self.ensure_one()
+        action = self.env.ref("stock.stock_move_action")
+        action_dict = action.read()[0]
```
victoralmau, line 181:
```suggestion
        action_dict = self.env["ir.actions.act_window"]._for_xml_id(
            "stock.stock_move_action"
        )
```

victoralmau (changes requested):
A few minor changes and please, fix pre-commit errors.

victoralmau:
Superseed by https://github.com/OCA/stock-logistics-warehouse/pull/1604

PR #1357, opened 2022-01-21
[15.0][MIG] stock_request_kanban

BernatPForgeFlow:


rousseldenis (commented):
@BernatPForgeFlow Thanks for this. Could you check tests?

Do you think js changes can be backported ? If yes, could you isolate those changes in a separate commit ?

Thanks

BernatPForgeFlow:
> @BernatPForgeFlow Thanks for this. Could you check tests?
> 
> Do you think js changes can be backported ? If yes, could you isolate those changes in a separate commit ?
> 
> Thanks

@rousseldenis No, I tried this changes in v14 and is not working. I think the migration to Owl in v15 is enough. Thank you for the review.

rousseldenis (approved):

LoisRForgeFlow (approved):

PR #1358, opened 2022-01-24
[14.0][ADD] stock_inventory_line_custom_cost

newtratip:


stock_inventory_line_custom_cost/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Inventory Adjustment - Custom Cost",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Ecosoft,Odoo Community Association (OCA)",
+    "maintainer": "newtratip",
```
rousseldenis:
```suggestion
    "maintainers": ["newtratip"],
```

rousseldenis (changes requested):
Code review.

Maybe 'custom' is a little bit too generic. I would have said 'Manual cost' or something like that. No ?

Saran440:
@newtratip @rousseldenis it's make sense to change module and name to `manual_cost`

Case: **Increase** product -> LGTM :+1: 
Case: **Reduce** product -> it not working. BUT I think it's correct 

**Ex (with FIFO)**
Product A has 10 units and cost = 100
1. we use inventory adjustment it to 15 units and manual cost = 70. it will use **manual cost** * diff qty = 70 * 5
2. we use inventory adjustment it to 5 units and manual cost = 70. it will use **cost** * diff qty = 100 * 5

I'm not sure for case reduct product by adjustment (2). Should it compute by `manual_cost` or `cost` (standard)
For me it should be compute by cost because product loss but we know product cost.

WDYT?

PR #1359, opened 2022-01-26
[12.0][FIX] stock_available_unreserved: pre_init_hook

yostashiro:
Before this commit, unreserved_quantity and contains_unreserved
were not updated at install when there was quantity and not
reserved_quantity.


rousseldenis (approved):
Code review. Seems legit

kos94ok-3D (approved):

PR #1365, opened 2022-01-31
[14.0][FIX] Fix access right issues when using actions

sebalix:


florian-dacosta (approved):

stock_available_unreserved/models/product_template.py
```diff
@@ -32,13 +32,14 @@ def action_open_quants_unreserved(self):
         quant_ids = quants.filtered(
             lambda x: x.product_id.qty_available_not_res > 0
         ).ids
-        result = self.env.ref("stock.group_stock_multi_locations").read()[0]
-        result["domain"] = [("id", "in", quant_ids)]
-        result["context"] = {
+        xmlid = "stock.group_stock_multi_locations"
+        action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
+        action["domain"] = [("id", "in", quant_ids)]
+        action["context"] = {
```
sebalix: Travis is failing on this, but it seems than even the previous code was wrong: it makes no sense to read a `res.groups` record and then add some values such as `domain` in it and return it as an action window for `stock.quant`... it is the `_for_xml_id` method which does some checks against the data model that raises.
I'm fixing it.

sebalix: Fixed, there was an action defined in the current addon for this use-case.

i-vyshnevska (approved):

PR #1368, opened 2022-02-09
[14.0][FIX] stock_orderpoint_manual_procurement: wrong parameter in _prepare_procurement_values

mariadforgeflow:
This remained from older odoo versions, but in v14  _prepare_procurement_values doesn't have item quantity as a parameter. Giving  this parameter to the function was setting date_deadline  = item.qty.

@ForgeFlow


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

PR #1372, opened 2022-02-15
[10.0] stock_quant_merge: add company_id and owner_id in merge domain

alexis-via:


alexis-via:
company_id is present in merge domain in v8:
https://github.com/OCA/stock-logistics-warehouse/blob/8.0/stock_quant_merge/models/stock.py#L22
owner_id is an addition, but it's absolutely needed when that feature is used !

rousseldenis (approved):
Code review. A test can help

PR #1375, opened 2022-02-18
[13.0][FIX] stock_inventory_preparation_filter: Filtering by lots is adding all lots of products

CarlosRoca13:
cc @Tecnativa TT34672

Screeenshot inventory:
imagen

Before this patch, if we create a stock inventory for some lots, we get a line for each lot of the products difined on lots.
imagen

After this patch, we get the lines for the lots defined.
imagen

Please @carlosdauden @chienandalu review it


carlosdauden (approved):

pedrobaeza (approved):

pedrobaeza:
Please fw-port it to v14

PR #1376, opened 2022-02-21
[IMP] stock_inventory_preparation_filter: Filtering by lots is adding all lots of products

CarlosRoca13:
Before this patch, if we create a stock inventory for some lots, we get a line for each lot of the products difined on lots.

After this patch, we get the lines for the lots defined.

cc @Tecnativa TT34672

fw-por of https://github.com/OCA/stock-logistics-warehouse/pull/1375

### DIFF
--- a/stock_inventory_preparation_filter/README.rst
+++ b/stock_inventory_preparation_filter/README.rst
@@ -64,6 +64,7 @@ Contributors
   * Pedro M. Baeza
   * David Vidal
   * Sergio Teruel
+  * Carlos Roca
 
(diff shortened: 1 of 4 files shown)


CarlosRoca13:
ping @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this, as already merged in previous version:

/ocabot merge patch

pedrobaeza:
Next time, this is a fix, as it's solving a bug.

PR #1377, opened 2022-02-22
[14.0][FIX] stock_orderpoint_origin: support multiwarehouse case

dreispt:
.

### DIFF
--- a/stock_orderpoint_origin/models/procurement_group.py
+++ b/stock_orderpoint_origin/models/procurement_group.py
@@ -15,9 +15,12 @@ class ProcurementGroup(models.Model):
         Forecast = self.env["report.stock.report_product_product_replenishment"]
         new_procurements = []
         for procurement in procurements:
+            ForecastWH = Forecast
+            warehouse = procurement.values.get("warehouse_id")
+            if warehouse:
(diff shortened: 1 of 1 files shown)


patrickrwilson (approved):
LGTM

stock_orderpoint_origin/models/procurement_group.py
```diff
@@ -15,9 +15,10 @@ def run(self, procurements, raise_user_error=True):
         Forecast = self.env["report.stock.report_product_product_replenishment"]
         new_procurements = []
         for procurement in procurements:
+            warehouse = procurement.values["warehouse_id"]
+            ForecastWH = Forecast.with_context(warehouse=warehouse.id)
```
rousseldenis: I think warehouse can be void too. That has to be taken into account.

rousseldenis (changes requested):
@dreispt @patrickrwilson I think it's time to add tests on this in order to avoid regressions. Thanks

stock_orderpoint_origin/models/procurement_group.py
dreispt: @rousseldenis Done

rousseldenis (approved):

PR #1152, opened 2021-05-13
[13.0][MIG] stock_available_mrp: Migration to 13.0

victoralmau:
Migration to 13.0.

Please @chienandalu can you review it?

@Tecnativa TT27873

PR #1378, opened 2022-02-22
[13.0][FIX] stock_available_mrp: Set correctlly immediately_usable_qty value.

victoralmau:
Related to: https://github.com/OCA/stock-logistics-warehouse/pull/1152#commitcomment-67223998

Set correctlly `immediately_usable_qty` value (related to https://github.com/OCA/stock-logistics-warehouse/pull/1379).

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT27873

### DIFF
--- a/stock_available_mrp/models/product_product.py
+++ b/stock_available_mrp/models/product_product.py
@@ -86,7 +86,7 @@ class ProductProduct(models.Model):
             bom_id = product.bom_id
             if not component_needs:
                 # The BoM has no line we can use
-                potential_qty = immediately_usable_qty = 0.0
+                potential_qty = 0.0
             else:
                 # Find the lowest quantity we can make with the stock at hand
                 components_potential_qty = min(
@@ -109,8 +109,7 @@ class ProductProduct(models.Model):
                 )
 
             res[product.id]["potential_qty"] = potential_qty
-            immediately_usable_qty = potential_qty if bom_id.type != "phantom" else 0
-            res[product.id]["immediately_usable_qty"] += immediately_usable_qty
+            res[product.id]["immediately_usable_qty"] = potential_qty
 
         return res, stock_dict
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

chienandalu (approved):

Cedric-Pigeon:
@victoralmau would be great to adapt tests, isn't it ?

pedrobaeza:
Which kind of adaptation, Cédric? They are green. If you have steps to reproduce the previous problem, a regression test can be done then to avoid such regression.

Cedric-Pigeon:
Well just wondering why test were not failing before the fix. :thinking:

pedrobaeza:
@victoralmau if you think in any case where the previous code could fail, please express it in a test. If not, please tell and at least let's merge this for avoiding further problems.

rousseldenis:
Such changes that have impact on returned values can easily expressed through a test. That has to be done

victoralmau:
A bit of context on this might help.
The change that was made in the migration is related to this comment https://github.com/OCA/stock-logistics-warehouse/pull/1152#pullrequestreview-663199244 by my colleague @chienandalu .

Reviewing again the change that I have suggested and the tests I think that it is really not necessary. The variables `potential_qty` and `bom_id` are always defined. I have created a local test of a product with a bom without components and no error appears.

@Cedric-Pigeon, can you explain a use case to reproduce the error and see if any changes are really necessary?

Cedric-Pigeon:
Well I have a bom of type "Phantom". None of the components have availabilities -> so I expected to get immediately_available = 0
But due to Odoo compute, the value is 1 because the min (virtual_available) of components is 1

victoralmau:
I've added several tests to check some use cases, but I don't think it's necessary to apply any changes to the current code.

Cedric-Pigeon:
@rousseldenis reproduced the bug here: https://github.com/OCA/stock-logistics-warehouse/pull/1379

victoralmau:
> @rousseldenis reproduced the bug here: #1379

Ok thanks, I have applied the change that I THINK should be applied. It is important that all reviewers have an opinion about it.

pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
Travis seems broken for now: https://app.travis-ci.com/github/OCA/stock-logistics-warehouse/jobs/561591463

I will retry later

PR #1381, opened 2022-02-24
[13.0][ADD]: account_move_line_stock_info script post_install for update references on new column

celm1990:
When install module on database with data, new column is NULL, purpose is FILL column with data

Inspired by https://github.com/OCA/account-financial-tools/blob/13.0/account_move_line_sale_info/hooks.py

### DIFF
--- a/account_move_line_stock_info/hooks.py
+++ b/account_move_line_stock_info/hooks.py
@@ -1,0 +1,17 @@
+# Copyright 2022 Carlos Lopez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def post_init_hook(cr, registry):
+
+    """ INIT stock_move_id references in acount move line from account_move"""
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml
+            SET stock_move_id = am.stock_move_id
+        FROM account_move am
+        WHERE am.id = aml.move_id
+            AND am.stock_move_id IS NOT NULL AND aml.stock_move_id IS NULL;
+    """
+    )
--- a/account_move_line_stock_info/__init__.py
+++ b/account_move_line_stock_info/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import models
+from .hooks import post_init_hook
--- a/account_move_line_stock_info/__manifest__.py
+++ b/account_move_line_stock_info/__manifest__.py
@@ -14,4 +14,5 @@
         "views/account_move_line_view.xml",
         "views/stock_move_view.xml",
     ],
+    "post_init_hook": "post_init_hook",
 }


celm1990:
ping @OCA/logistics-maintainers @AaronHForgeFlow

account_move_line_stock_info/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+def post_init_hook(cr, registry):
+
+    """ INIT stock_move_id references in acount move line from account_move"""
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml
+            SET stock_move_id = am.stock_move_id
+        FROM account_move am
+        WHERE am.id = aml.move_id AND am.stock_move_id IS NOT NULL;
```
AaronHForgeFlow:
```suggestion
        WHERE am.id = aml.move_id AND am.stock_move_id IS NOT NULL AND aml.stock_move_id IS NULL;
```

AaronHForgeFlow (commented):
I did not know the stock move reference is in the account move reference. When is that field filled in the account move? 
When receiving a picking there may be are several stock moves involved but only one journal entry. So I guess for those cases the stock move information is missing in the account move too, isn't it?

account_move_line_stock_info/hooks.py
celm1990: Hi @AaronHForgeFlow, relations are:

**Odoo:**
`stock.move > account.move`(not with picking) see 
https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/account_move.py#L9
with this module **account_move_line_stock_info**
`stock.move > account.move.line`

for each `stock.move` odoo creat a `account.move`, see
https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/stock_move.py#L461
https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/stock_move.py#L279



with PR i try fill field into `account.move.line` from `account.move`

Respect to your suggestion, `AND aml.stock_move_id IS NULL` i not take into account because this script is executed when module is installed and column is new and is empty

AaronHForgeFlow: Ok. Thanks for the explanation.

Regarding my suggestion, `AND aml.stock_move_id IS NULL` it is just a suggestion. I understand the column will be null for most cases. But imagine someone has filled the column using a custom script/module. You will not want the column overwritten in that case. I think it is not harmful to add the condition.

celm1990: Thank you so much, I applied your suggestion  ;)

AaronHForgeFlow (approved):
Code review LGTM!

celm1990:
ping https://github.com/orgs/OCA/teams/logistics-maintainers

account_move_line_stock_info/hooks.py
```diff
@@ -0,0 +1,18 @@
...
+    """ INIT stock_move_id references in acount move line from account_move"""
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml
+            SET stock_move_id = am.stock_move_id
+        FROM account_move am
+        WHERE am.id = aml.move_id
+            AND am.stock_move_id IS NOT NULL
+            AND aml.stock_move_id IS NULL;
```
AaronHForgeFlow: perhaps you can remove this line, to make this even to https://github.com/OCA/stock-logistics-warehouse/pull/1385

For me both solutions are fine

celm1990: @AaronHForgeFlow  done

celm1990:
@AaronHForgeFlow  i see errors on test but are on other modules, any idea or can continue with merge this PR?

account_move_line_stock_info/hooks.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def post_init_hook(cr, registry):
+
+    """ INIT stock_move_id references in acount move line from account_move"""
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml
```
rousseldenis, line 11: @AaronHForgeFlow @celm1990 My question looking at this query: Why the stock_move_id field is not a related on account_move one ?

AaronHForgeFlow: This module was created in v8 and Odoo in that version Odoo was not adding the stock_move_id in the account move I believe

When Odoo added that field I think that this module could be adapted to that (making it related stored), but in any case I think it is still consistent. It uses the prepare_account_move_line method:https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/stock_move.py#L375 which it is always called form the same stock move that is used to fill the information in the account_move. 
Perhaps in newer versions this could be adapted.

PR #1385, opened 2022-03-08
[14.0][IMP] account_move_line_stock_info: Add init hook

JordiBForgeFlow:
We want to make sure that we move the stock_move_id from the account.move
to the account.move.line

@NuriaXForgeFlow @AaronHForgeFlow

### DIFF
--- a/account_move_line_stock_info/hooks.py
+++ b/account_move_line_stock_info/hooks.py
@@ -1,0 +1,16 @@
+# Copyright 2022 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def post_init_hook(cr, registry):
+
+    """INIT stock references in acount move line"""
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml SET stock_move_id = am.stock_move_id
+        FROM account_move am
+        WHERE am.id = aml.move_id
+        AND am.stock_move_id IS NOT NULL;
+    """
+    )
--- a/account_move_line_stock_info/__init__.py
+++ b/account_move_line_stock_info/__init__.py
@@ -1,3 +1,2 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
 from . import models
+from .hooks import post_init_hook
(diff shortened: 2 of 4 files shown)


account_move_line_stock_info/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+def post_init_hook(cr, registry):
+
+    """ INIT stock references in acount move line """
+    # FOR stock moves
+    cr.execute(
+        """
+        UPDATE account_move_line aml SET stock_move_id = am.stock_move_id
+        FROM account_move am
+        WHERE am.id = aml.move_id
+        AND am.stock_move_id IS NOT NULL;
```
AaronHForgeFlow, line 14: Maybe also:
```suggestion
        AND am.stock_move_id IS NOT NULL AND aml.stock_move_id IS NULL;
```

MiquelRForgeFlow:
```suggestion
        AND am.stock_move_id IS NOT NULL AND aml.stock_move_id IS NULL;
```
hehe

AaronHForgeFlow: Opps yes!

flotho:
Hi @JordiBForgeFlow , easy comments to fix from your colleagues.

account_move_line_stock_info/hooks.py
JordiBForgeFlow: I don't agree with this. We want to make sure that we overwrite any previous value.

flotho: > I don't agree with this. We want to make sure that we overwrite any previous value.

thanks for this explanation.
I'm not experienced enough on this part to approve or request a change, I let you as expert to "fight"

AaronHForgeFlow:
Rebased and fixed conflicts.

> We want to make sure that we overwrite any previous value.

I don't like to overwrite previous values, but in the real world that column is empty when installing the module, so it is ok to me.

AaronHForgeFlow (approved):

flotho (approved):

rousseldenis:
@JordiBForgeFlow Ready ?

PR #1395, opened 2022-03-24
[13.0][FIX] sale_stock_available_info_popup: right context

chienandalu:
We must ensure the right context is provided to the product in order to compute the right quantities in relation to the other values of the popup.

cc @Tecnativa TT35179

please review @victoralmau @pedrobaeza

### DIFF
--- a/sale_stock_available_info_popup/models/sale_order.py
+++ b/sale_stock_available_info_popup/models/sale_order.py
@@ -18,9 +18,15 @@ class SaleOrderLine(models.Model):
         qty_processed_per_product = defaultdict(lambda: 0)
         self.immediately_usable_qty_today = False
         for line in self.sorted(key=lambda r: r.sequence):
+            if line.order_id.commitment_date:
+                date = line.order_id.commitment_date
+            else:
+                date = line._expected_date()
             if not line.display_qty_widget:
                 continue
-            product = line.product_id
+            product = line.product_id.with_context(
+                to_date=date, warehouse=line.warehouse_id.id
+            )
             qty_processed = qty_processed_per_product[product.id]
             line.immediately_usable_qty_today = (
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You always forget to put the version in the PR title (not in the commit - there it shouldn't be put -).

pedrobaeza (approved):

chienandalu:
> You always forget to put the version in the PR title (not in the commit - there it shouldn't be put -).

I know, sorry :face_with_spiral_eyes:

victoralmau (approved):

sale_stock_available_info_popup/models/sale_order.py
```diff
@@ -18,9 +18,15 @@ def _compute_immediately_usable_qty_today(self):
         qty_processed_per_product = defaultdict(lambda: 0)
         self.immediately_usable_qty_today = False
         for line in self.sorted(key=lambda r: r.sequence):
+            if line.order_id.commitment_date:
+                date = line.order_id.commitment_date
```
rousseldenis, line 22: What is making the decision to take the commitment date and not the order date?

Shouldn't both computations in different fields ?

chienandalu: Hi, @rousseldenis It comes from core: https://github.com/odoo/odoo/blob/13.0/addons/sale_stock/models/sale_order.py#L243-L246

rousseldenis: Could you add a test with commitment_date ?

chienandalu: Added test for that condition and simplified test declaration using `Form`

pedrobaeza:
Merging as the petition has been attended:

/ocabot merge patch

Can you fw-port it to 14?

PR #1399, opened 2022-03-29
[FIX] pre-commit black: fixed _unicodefun patch code for Click 8.1.0

kos94ok-3D:
See: https://github.com/psf/black/pull/2966


rousseldenis:
This is solved there : https://github.com/OCA/stock-logistics-warehouse/pull/1400

PR #1400, opened 2022-03-30
Updated from template - Switch to Github actions

rousseldenis:

PR #1401, opened 2022-03-31
[ADD] Stock Location Content Template

Hardik-OSI:


stock_location_content_template/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "Stock Location Content Template",
+    "summary": "Define templates of location content and check the report for discrepancies",
+    "version": "14.0.1.0.0",
+    "category": "stock",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "maintainers": ["max3903"],
+    "license": "AGPL-3",
+    "depends": ["stock", "sale_management"],
```
max3903, line 12:
```suggestion
    "depends": ["stock"],
```

.pre-commit-config.yaml
```diff
@@ -104,7 +104,7 @@ repos:
           - --settings=.
         exclude: /__init__\.py$
   - repo: https://github.com/acsone/setuptools-odoo
-    rev: 2.6.0
+    rev: 3.0.3
```
rousseldenis: Please don't modify here. The good method is to apply copier update in another PR

rousseldenis: @max3903 Click version should not be pinned. Black has to be updated. 

Please review this instead https://github.com/OCA/stock-logistics-warehouse/pull/1400/files

rousseldenis: @max3903 I will merge an update there that is less intrusive: #1418

rousseldenis (commented):
@Hardik-OSI @max3903 This is quite strange. Could you describe why you don't use reordering rules in this case?

max3903:
> @Hardik-OSI @max3903 This is quite strange. Could you describe why you don't use reordering rules in this case?

Reordering rules need to be created for each location. Here we want to define few templates but use them on hundreds of locations. It is easier to update one template than hundreds of rules.

rousseldenis:
> > @Hardik-OSI @max3903 This is quite strange. Could you describe why you don't use reordering rules in this case?
> 
> Reordering rules need to be created for each location. Here we want to define few templates but use them on hundreds of locations. It is easier to update one template than hundreds of rules.

This has to be clearly explained in README. Moreover, the module name does not reflect what it does really.
Could you take time to clarify ?

rousseldenis:
> 

Functionally I wanted to say

rousseldenis:
> > > @Hardik-OSI @max3903 This is quite strange. Could you describe why you don't use reordering rules in this case?
> > 
> > 
> > Reordering rules need to be created for each location. Here we want to define few templates but use them on hundreds of locations. It is easier to update one template than hundreds of rules.
> 
> This has to be clearly explained in README. Moreover, the module name does not reflect what it does really. Could you take time to clarify ?

And why not creating ordering rules templates ?

 - You define the rule lines like here:
image

- And you find a way to create ordering rules for a list of locations. By that way, you reuse the standard stock Odoo behaviour and avoid stuff like this : https://github.com/OCA/stock-logistics-warehouse/pull/1401/files#diff-ff28961d99f577c8872449c9f0eead27e17f4ac1f463792e3d302632273b0a2cR95
That could lead to extra function to `create_<whatever>_pick()` and quite a mess.

- When you create the ordering rules, you make a link to the template, so, if you want to recreate them, it's easy to identify them from the normal ones.

stock_location_content_template/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,5 @@
-* This module allows you to define template of location with list of products and quantities supposed to be there.
+This module allows you to define template of location with list of products and quantities supposed to be there.
```
rousseldenis, line 1: It does more than that.

It does not use the stock rules at all. User should be warned of that.

In a more global view, I think this will lead to extra code that will 'pollute' this repo because the behaviour is aside of Odoo.

IMHO it deserves its own repo for a 'Simple Warehouse Management'.

@OCA/logistics-maintainers What do you think ?

@max3903

stock_location_content_template/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import test_apartment_template
```
rousseldenis, line 3: I think your particular requirements should not appear in code.

stock_location_content_template/tests/test_apartment_template.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import datetime
+
+from odoo.tests import common
+
+
+class TestApartmentTemplate(common.TransactionCase):
+    def setUp(self):
+        super(TestApartmentTemplate, self).setUp()
```
rousseldenis, line 10: Use setUpClass instead in conjunciton with SavepointCase

stock_location_content_template/readme/DESCRIPTION.rst
hparfr: +1 on a big warning that it's quite different than one may expect. 

> IMHO it deserves its own repo for a 'Simple Warehouse Management'.

I think a different repo may be needed if there is more modules like this one.

rousseldenis: @hparfr https://github.com/OCA/stock-logistics-warehouse/pull/1419

hparfr: I suggest "Different wh mgnt" then.

rousseldenis: > I suggest "Different wh mgnt" then.

Mmmmh. 'different' in which manner?

hparfr: We can't say it's a "simple" case when the need is to manage lot of rules en masse.

rousseldenis: Yes, I wanted to say 'from Odoo way of doing things'.

But apart naming, to summarize, I think we agree to say this is not an extension of Odoo standard (for stock flows). With that assertion, we can assume a different repository is quite a sane way of doing it if we want to welcome it under OCA umbrella in order to:

- not introducing confusion for users
- gather all extension modules in the same playground
- minimize risk for testing

rousseldenis:
@Hardik-OSI @max3903 Could you rebase?

stock_location_content_template/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Location Content Template",
```
rousseldenis, line 4: That's better !

stock_location_content_template/models/stock_location_content_check_line.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class StocklocationContentCheckLine(models.Model):
+    _name = "stock.location.content.check.line"
+    _description = "Stock Location Content Check Lines"
+    _rec_name = "product_id"
+
+    product_id = fields.Many2one("product.product", string="Product")
+    check_id = fields.Many2one("stock.location.content.check", string="Check")
+    expected_qty = fields.Float(
```
rousseldenis, line 14: You should use digit precision for all stock flows.

stock_location_content_template/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
...
+
+We didn't use reordering rules for maintenance purposes: Reordering rules need to be defined and maintained for each location. If you have hundreds of locations, updating one template is more convenient than updating hundreds of rules.
```
rousseldenis, line 3: A question comes to me: Why not applying a template on reordering rules instead ? That will avoid all the stuff about moves creation etc... @max3903 @dreispt

stock_location_content_template/models/stock_location_content_check.py
```diff
@@ -0,0 +1,242 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class StocklocationContentCheck(models.Model):
+    _name = "stock.location.content.check"
```
rousseldenis, line 8: Maybe splitting this in another module should be great:
 - Location content template management
 - Location content check management

stock_location_content_template/models/stock_location_content_check.py
```diff
@@ -0,0 +1,242 @@
...
+        return super(StocklocationContentCheck, self).create(vals)
+
+    def action_confirm(self):
+        for rec in self:
+            rec.state = "confirmed"
+
+    def action_complete(self):
+        for rec in self:
+            rec._create_out_pick()
+            rec._create_internal_pick()
```
rousseldenis, line 54: You assume there won't exist any complex stock rules

stock_location_content_template/__manifest__.py
rousseldenis, line 12:
```suggestion
    "depends": ["stock"],
```

I don't see any sale dependency here

stock_location_content_template/models/stock_location_content_check.py
```diff
@@ -0,0 +1,242 @@
...
+        picking_obj = self.env["stock.picking"]
+        for rec in self:
+            move_lines_self_parent = []
+            move_lines_parent_self = []
+            if rec.location_id.location_id.usage == "view":
+                raise UserError(_("You can't use view type location in transfer."))
+            wh_id = rec.location_id.get_warehouse()
+            pick_type_id = wh_id.int_type_id
+            for line in rec.line_ids:
+                if not line.replenished_qty == 0.0:
```
rousseldenis, line 105: As said for fields definitions, you need to use digit precision and float_compare here.

stock_location_content_template/models/stock_location_content_check.py
```diff
@@ -0,0 +1,242 @@
...
+                picking.action_confirm()
+                picking.action_assign()
+                for movel in picking.move_lines:
+                    movel.quantity_done = movel.product_uom_qty
+                picking.button_validate()
+            rec.state = "closed"
+
+    @api.onchange("location_id")
+    def _onchange_location_id(self):
+        check_lines = []
```
rousseldenis, line 195: Isn't this onchange very slow ?

stock_location_content_template/models/stock_location_content_check_line.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @api.onchange("expected_qty", "counted_qty")
+    def _onchange_replenished_qty(self):
+        for rec in self:
+            rec.replenished_qty = rec.expected_qty - rec.counted_qty
+
+    @api.constrains("expected_qty", "counted_qty", "replenished_qty")
+    def _check_quantities(self):
+        for rec in self:
+            if rec.counted_qty + rec.replenished_qty > rec.expected_qty:
```
rousseldenis, line 51: Use float_compare and digit precision

stock_location_content_template/models/stock_location_content_check_line.py
```diff
@@ -0,0 +1,57 @@
...
+                raise UserError(
```
rousseldenis, line 52: In constraints you must use ValidationError instead.

stock_location_content_template/models/stock_location_content_template.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockLocationContentTemplate(models.Model):
+    _name = "stock.location.content.template"
+    _description = "Stcok Location Content Template"
```
rousseldenis, line 8:
```suggestion
    _description = "Stock Location Content Template"
```

stock_location_content_template/models/stock_location_content_template.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockLocationContentTemplate(models.Model):
```
rousseldenis, line 6: I think this deserves a multi-company support. But can be put in ROADMAP

stock_location_content_template/report/stock_location_content_report.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (C) 2022 Open Source Integrators (https://www.opensourceintegrators.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class LocationContentReport(models.TransientModel):
+    _name = "stock.location.content.report"
```
rousseldenis, line 8:
```suggestion
    _name = "stock.location.content.template.report"
```
To avoid possible naming collision

rousseldenis (changes requested):
Code review

rousseldenis:
@Hardik-OSI @max3903 What's the status on this ?

rousseldenis:
@Hardik-OSI @max3903 Gentle reminder

max3903:
@SMaciasOSI Can you take over and get this merged? Reach out to me if you have any question.

Thanks @rousseldenis and sorry for the delay.

rousseldenis:
> Thanks @rousseldenis and sorry for the delay.

No problem. Spring household :-)

PR #1403, opened 2022-04-05
[FIX] stock_available_mrp: normal BoM available to promise

chienandalu:
Normal BoM products (those which are manufactured) are regular stored products and their immediately_usable_qty will be summed with potential stock. This was the expected behavior of the module and it was lost at some point.

Phantom BoM products (kits) don't have real stock so their available to promise quantity will be the same as the potential.

As an improvement, we've added the possibility to override the sum of potential and available to promise. In some cases such addition doesn't make sense as we don't know how long can take to manufacture those potential units.

cc @Tecnativa TT35589

please review @victoralmau @pedrobaeza

### DIFF
--- a/stock_available_mrp/models/product_product.py
+++ b/stock_available_mrp/models/product_product.py
@@ -109,8 +109,19 @@ class ProductProduct(models.Model):
                 )
 
             res[product.id]["potential_qty"] = potential_qty
-            res[product.id]["immediately_usable_qty"] = potential_qty
-
+            # Normal BoM products (those which are manufactured) are regular stored
+            # products and their immediately_usable_qty will be the sum of its regular
+            # available to promise quanity as storable product and its potential qty
+            # Phantom BoM products (kits) don't have real stock so their available
+            # to promise quantity will be the same as the potential.
+            if bom_id.type == "phantom":
+                res[product.id]["immediately_usable_qty"] = potential_qty
+            # We can override at BoM level to add the potential quantity to the
+            # manufactured product. This could be the case when the manufacturing
+            # proccess is uncertain and the promising such quantities could be
+            # missleading.
+            elif not bom_id.add_potential_exception:
+                res[product.id]["immediately_usable_qty"] += potential_qty
         return res, stock_dict
 
     def _explode_boms(self):
--- a/stock_available_mrp/README.rst
+++ b/stock_available_mrp/README.rst
@@ -37,14 +37,24 @@ with the components available to promise.
 .. contents::
    :local:
 
+Usage
+=====
+
+It's possible to avoid adding potential stock to available to promise stock at Bill of
+Materials level. This is interesting in case that sum is missleading due to uncertain
+manufacturing processes. To do so:
+
+- Go to that specific Bill of Materials and in the Miscellaneous tab set the option
+  **Avoid adding potential to available to promise** on.
+
 Known issues / Roadmap
 ======================
 
 Known issues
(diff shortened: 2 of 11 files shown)


victoralmau (approved):

pedrobaeza (approved):

stock_available_mrp/models/product_product.py
```diff
@@ -109,8 +109,12 @@ def _compute_available_quantities_dict(self):
...
 
             res[product.id]["potential_qty"] = potential_qty
-            res[product.id]["immediately_usable_qty"] = potential_qty
-
+            # Normal BoM products (those which are manufactured) are regular stored
+            # products and their immediately_usable_qty will follow general rules.
+            # Phantom BoM products (kits) don't have real stock so their available
+            # to promise quantity will be the same as the potential.
+            if bom_id.type == "phantom":
+                res[product.id]["immediately_usable_qty"] = potential_qty
```
pedrobaeza, line 118:
```suggestion
            res[product.id]["immediately_usable_qty"] += potential_qty
```

pedrobaeza (changes requested):
Aborted merge and discussing this alternative

chienandalu:
@pedrobaeza I added the config options we commented aside. Now the module works as expected by the tests but the behavior can be changed by company and overrided by BoM.

I have to add the config views and update the readme and comments, but take a look (maybe the config fields names can be improved...)

@victoralmau take a look as well for this more flexible approach

victoralmau:
> @pedrobaeza I added the config options we commented aside. Now the module works as expected by the tests but the behavior can be changed by company and overrided by BoM.
> 
> I have to add the config views and update the readme and comments, but take a look (maybe the config fields names can be improved...)
> 
> @victoralmau take a look as well for this more flexible approach

I think this approach is better at allowing the behavior to be configured.

stock_available_mrp/models/product_product.py
```diff
@@ -109,8 +109,17 @@ def _compute_available_quantities_dict(self):
...
+            elif (
+                self.env.company.stock_available_mrp_add_potential
```
pedrobaeza: Is this company always the more suitable one for the computation?

pedrobaeza (commented):
Please expand tests with this new options for avoiding regressions

stock_available_mrp/models/product_product.py
chienandalu: BoMs `company_id` field isn't required, this one is the most neutral I came up with. Another option is to set it globally as config parameter.

chienandalu: I'm thinking that maybe we could simply have the default behavior of adding potential to available to promise stock with option of overriding it at BoM level. This way, it could be still possible to have a default value at instance level and we don't have to deal with possible company quirks

pedrobaeza: OK for me

pedrobaeza: Default value in BoM should preserve the old behavior

chienandalu: Yes. The old behavior was to add them: https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_available_mrp/models/product_product.py#L123

chienandalu:
- [x] Recoverd former behavior
- [x] Added behavior overriding options at BoM
- [x] Added tests

pedrobaeza (approved):
Thanks for extra testing and the feature

PR #1407, opened 2022-04-20
[14.0][IMP] stock_packaging_calculator: improve performance

sebalix:
This change avoids a long running process (while loop) when dealing with high volume of quantities (several thousands) when the packaging qty is 1.

In our case it saves about 10s when updating quantities on a PO.

Ref. 3522

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -118,7 +118,12 @@ class Product(models.Model):
             "_packaging_values_handler", self._prepare_qty_by_packaging_values
         )
         for pkg in pkg_by_qty:
-            qty_per_pkg, qty = self._qty_by_pkg(pkg.qty, qty)
+            # Boost perf: no need to deduce the qty_per_pkg if the pkg_qty is 1
+            if float_compare(pkg.qty, 1, precision_digits=self.uom_id.rounding) == 0:
+                qty_per_pkg = int(qty)
+                qty = 0.0
+            else:
+                qty_per_pkg, qty = self._qty_by_pkg(pkg.qty, qty)
             if qty_per_pkg:
                 value = prepare_values(pkg, qty_per_pkg)
                 if with_contained:


stock_packaging_calculator/models/product.py
```diff
@@ -118,7 +118,12 @@ def _product_qty_by_packaging(self, pkg_by_qty, qty, with_contained=False):
             "_packaging_values_handler", self._prepare_qty_by_packaging_values
         )
         for pkg in pkg_by_qty:
-            qty_per_pkg, qty = self._qty_by_pkg(pkg.qty, qty)
+            # Boost perf: no need to deduce the qty_per_pkg if the pkg_qty is 1
+            if float_compare(pkg.qty, 1, precision_digits=self.uom_id.rounding) == 0:
```
i-vyshnevska, line 122:
```suggestion
            if float_compare(qty, 1, precision_digits=self.uom_id.rounding) == 0:
```
I don't know for sure just base on reading seems another qty needs to be used

sebalix: The code was correct (I guess), `qty` is the whole qty (e.g. 150 000) while `pkg.qty` is the packaging qty (e.g. 1), and the `_qty_by_pkg()` method subtract `pkg.qty` from `qty`:
```python
    def _qty_by_pkg(self, pkg_qty, qty):
        """Calculate qty needed for given package qty."""
        qty_per_pkg = 0
        while (
            float_compare(qty - pkg_qty, 0.0, precision_digits=self.uom_id.rounding)
            >= 0.0
        ):
            qty -= pkg_qty
            qty_per_pkg += 1
        return qty_per_pkg, qty
```

mmequignon: can we have multiple pkg_qties in pkg_by_qty?
If not, why not using divmod?

TDu (approved):

simahawk (approved):
seems legit.

FTR what's the point of duplicating the "improve(ment)" word in this commit?

`[IMP] s_p_calculator: improve performance` -> `stock_packaging_calculator: improve perf`

Rule of thumb: if the commit msg is already explaining what you did, IMO there's no need for the [TAG] (even if is in the Odoo guidelines)"

stock_packaging_calculator/models/product.py
sebalix: I didn't understood your questions?

sebalix:
@simahawk commit message reworded

stock_packaging_calculator/models/product.py
mmequignon: I don't know what represents pkg_by_qty, if qty is 1000, and pkg_qty is 20, then the result is `qty / pkg_qty`.
Might save some time for other cases too, no?

sebalix: Maybe, I didn't analyzed the code too much so that's a question for @simahawk.
I didn't wanted to refactor the whole thing and wanted to avoid the simplest case where the pkg.qty is 1, making the while loop useless.

PR #1409, opened 2022-04-27
[14.0][ADD] module: product_route_profile

Kev-Roche:
This module replaces the initial concept of route_ids with a new concept of "route profile", coming with a company-specific and priority route profile.

### DIFF
--- a/product_route_profile/models/product_template.py
+++ b/product_route_profile/models/product_template.py
@@ -1,0 +1,63 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    route_profile_id = fields.Many2one("route.profile", string="Route Profile")
+    force_route_profile_id = fields.Many2one(
+        "route.profile",
+        string="Priority Route Profile",
+        company_dependent=True,
+        help="If defined, the "
+        "priority route profile will be used and will replace the "
+        "route profile, only for this company.",
+    )
+
+    route_ids = fields.Many2many(
+        compute="_compute_route_ids",
+        inverse="_inverse_route_ids",
+        search="_search_route_ids",
+        store=False,
+    )
+
+    @api.depends("route_profile_id", "force_route_profile_id")
+    @api.depends_context("company")
+    def _compute_route_ids(self):
+        for rec in self:
+            if rec.force_route_profile_id:
+                rec.route_ids = [(6, 0, rec.force_route_profile_id.route_ids.ids)]
+            elif rec.route_profile_id:
+                rec.route_ids = [(6, 0, rec.route_profile_id.route_ids.ids)]
+            else:
+                rec.route_ids = False
+
+    def _search_route_ids(self, operator, value):
+        return [
+            "|",
+            ("force_route_profile_id.route_ids", operator, value),
+            "&",
+            ("force_route_profile_id", "=", False),
+            ("route_profile_id.route_ids", operator, value),
+        ]
+
+    def _inverse_route_ids(self):
+        profiles = self.env["route.profile"].search([])
+        for rec in self:
+            for profile in profiles:
+                if rec.route_ids == profile.route_ids:
+                    rec.route_profile_id = profile
+                    break
+            else:
+                vals = rec._prepare_profile()
+                rec.route_profile_id = self.env["route.profile"].create(vals)
+
+    def _prepare_profile(self):
+        return {
+            "name": " / ".join(self.route_ids.mapped("name")),
+            "route_ids": [(6, 0, self.route_ids.ids)],
+        }
--- a/product_route_profile/models/route_profile.py
+++ b/product_route_profile/models/route_profile.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class RouteProfile(models.Model):
+    _name = "route.profile"
+    _description = "Route Profile"
+
+    name = fields.Char("Name")
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        default=lambda self: self.env.company.id,
+        required=False,
+        string="Company",
+    )
+    route_ids = fields.Many2many(
+        "stock.location.route",
+        string="Routes",
+        domain=[("product_selectable", "=", True)],
+    )
--- a/product_route_profile/readme/USAGE.rst
+++ b/product_route_profile/readme/USAGE.rst
@@ -1,0 +1,9 @@
+**Route profile**
+In Inventory > Configuration > Settings > Routes Profiles
+-  Create some Route profile depending on your needs
+
+
+**On product**
+On each template product, in inventory page, we can select:
+- **Route Profile**: a default profile, common to all companies
+- **Priority Route Profile**: a profile specific to each company and priority if existing.
--- a/product_route_profile/__init__.py
+++ b/product_route_profile/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from .hooks import post_init_hook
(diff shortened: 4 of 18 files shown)


product_route_profile/models/route_profile.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class RouteProfile(models.Model):
+    _name = "route.profile"
+
```
CasVissers-360ERP, line 11: Missing description

CasVissers-360ERP (changes requested):
One minor feedback point, we will start our functional tests now. Nice module!

product_route_profile/models/route_profile.py
```diff
@@ -0,0 +1,21 @@
...
+    name = fields.Char("Name")
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        default=lambda self: self.env.company.id,
+        required=False,
+        string="Company",
+    )
+    route_ids = fields.Many2many(
+        "stock.location.route",
+        string="Routes",
```
florian-dacosta, line 21: I think, for consistency, you should add the domain : `domain=[('product_selectable', '=', True)]` like it is on the original `route_ids` field on template

CasVissers-360ERP (approved):
Functional review, some code feedback. But module works great. Nice one!

product_route_profile/models/route_profile.py
Kev-Roche: Good point ! I add it.

product_route_profile/models/route_profile.py
Kev-Roche: Done, thank you !

product_route_profile/models/product_template.py
```diff
@@ -0,0 +1,35 @@
...
+        "route profile, only for this company.",
+    )
+
+    route_ids = fields.Many2many(
+        compute="_compute_route_ids",
+        store=False,
+    )
+
+    @api.depends("route_profile_id", "force_route_profile_id")
+    @api.depends_context("company")
```
rousseldenis, line 29: Where does it depend on this ?

product_route_profile/models/product_template.py
```diff
@@ -0,0 +1,35 @@
...
+        "route.profile",
+        string="Priority Route Profile",
+        company_dependent=True,
+        help="If defined, the "
+        "priority route profile will be used and will replace the "
+        "route profile, only for this company.",
+    )
+
+    route_ids = fields.Many2many(
+        compute="_compute_route_ids",
```
rousseldenis, line 22: Maybe a search method should be a great improvement

product_route_profile/models/product_template.py
florian-dacosta: `force_route_profile_id` is company dependent

rousseldenis: Yes, saw it. But doesn't Odoo ORM aware of that through that field ?

Your new depends IMHO should not add that entry.

rousseldenis:
@Kev-Roche @florian-dacosta IMHO, we should avoid breaking standard behaviour without an option to activate the feature.

At least you should extract this module from tests to do it separately

product_route_profile/models/product_template.py
Kev-Roche: After some checks, it appears to be unnecessary, I remove it.

product_route_profile/models/product_template.py
Kev-Roche: True, I add it to the roadmap

Kev-Roche:
> @Kev-Roche @florian-dacosta IMHO, we should avoid breaking standard behaviour without an option to activate the feature.
> 
> At least you should extract this module from tests to do it separately

I extract the tests from the others.

From my point of view, if we install this module, it is for use it. We prefer to avoid the activation / setting step if you don't mind.

product_route_profile/models/product_template.py
Kev-Roche: After deeper investigation, @api.depends_context("company") is relevant here, as we can see with the fail test (previous commit) and the fix in the last commit.

rousseldenis:
> > @Kev-Roche @florian-dacosta IMHO, we should avoid breaking standard behaviour without an option to activate the feature.
> > At least you should extract this module from tests to do it separately
> 
> I extract the tests from the others.
> 
> From my point of view, if we install this module, it is for use it. We prefer to avoid the activation / setting step if you don't mind.

IMHO, the best is to control features through parameters. Moreover, if it breaks standard behaviour.

Ok, to put in roadmap, but keep it in mind, you'll see it could help you especially if you want to deploy it without activating the feature yet.

product_route_profile/models/product_template.py
rousseldenis: Ok, I think this is a problem in ORM as company_dependant fields that are in depends should imply that behaviour. But ok

sebastienbeau:
@rousseldenis I think adding an option to activate the feature will complexify the code, for something you do not want as real feature (only need it for solving testing compatibility).
So it better/simplier to test the module separately

rousseldenis:
> @rousseldenis I think adding an option to activate the feature will complexify the code, for something you do not want as real feature (only need it for solving testing compatibility). So it better/simplier to test the module separately

In this case, if you don't want an option, shouldn't route_ids field on product have an inverse method in order to set the correct profile (or to create one ?) ?

Kev-Roche:
@rousseldenis I add the search method as previously discussed, you are right, it is very useful. 
About your last comment, we can set and create a route profile (in inventory>settings>Route profiles) and then set the profile in route_profile_id and/or force_route_profile_id in the product template. In that way, route_ids will be equal to the (force)route_profile_id.
We also add a hook to make compatible previous route_ids with the new concept of route profile.

product_route_profile/models/product_template.py
```diff
@@ -32,3 +34,13 @@ def _compute_route_ids(self):
                 rec.route_ids = [(6, 0, self.route_profile_id.route_ids.ids)]
             else:
                 rec.route_ids = False
+
+    def _search_route_ids(self, operator, value):
+        profiles = self.env["route.profile"].search([("route_ids", operator, value)])
```
rousseldenis: You can avoid a big search doing this directly in the return:

`
return [
"|",
            ("force_route_profile_id.route_ids", operator, value),
            "&",
            ("force_route_profile_id", "=", False),
            ("route_profile_id.route_ids", operator, value),
]
`

Kev-Roche: done, I also squashed the commits.

rousseldenis:
@Kev-Roche Could you fix this ?

Kev-Roche:
@rousseldenis , @sebastienbeau, @CasVissers-360ERP, @florian-dacosta 
I fixed the tests, can we add this module ? 
fixes : 
- add user in addition to administrator access right
- delete first test since it was not really relevelant here and most importantly, the test was inconsistant with some modules (add "default route" on product.template and consequently create a route_profile_id).

If ok, i'll squash the commits.

CasVissers-360ERP (approved):
LGTM!

florian-dacosta (approved):
LGTM

rousseldenis:
@Kev-Roche Could you remove travis file as we use Github actions?

Kev-Roche:
> 
@rousseldenis, if I remove the travis file, the tests will not work anymore, are they ?

rousseldenis:
> > 
> 
> @rousseldenis, if I remove the travis file, the tests will not work anymore, are they ?

Of course they do. If you look at checks, both are running :smiley:

Kev-Roche:
> > > 
> > 
> > 
> > @rousseldenis, if I remove the travis file, the tests will not work anymore, are they ?
> 
> Of course they do. If you look at checks, both are running smiley

good point,  I also squash the commits

rousseldenis (approved):

product_route_profile/readme/USAGE.rst
```diff
@@ -0,0 +1,15 @@
+
+Settings
+=====================
```
rousseldenis:
```suggestion
Settings
========
```

rousseldenis: @Kev-Roche Could you check titles?

Kev-Roche: Done

elvise:
@rousseldenis is it ready for merge ?

rousseldenis:
@Kev-Roche you should run oca-genaddon-readme locally to solve errors.

PR #1411, opened 2022-05-09
[14.0][ADD] stock_warehouse_calendar_orderpoint

sebalix:
<s>Depends on https://github.com/OCA/stock-logistics-warehouse/pull/1412 to fix the build.</s>

image


stock_warehouse_calendar_orderpoint/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Warehouse Calendar (reordering rules)",
+    "summary": "Adds a calendar to the Warehouse for reordering rules",
+    "version": "14.0.1.0.0",
+    "license": "LGPL-3",
```
rousseldenis, line 7: I think it's AGPL seeing files headers

stock_warehouse_calendar_orderpoint/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        "views/res_config_settings.xml",
+        "views/stock_warehouse.xml",
+        "views/stock_warehouse_orderpoint.xml",
+    ],
+    "demo": [
+        "demo/resource_calendar.xml",
+    ],
+    "installable": True,
+    "development_status": "Production/Stable",
+    "maintainers": [],
```
rousseldenis: Wouldn't you adopt it ?

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,50 @@
...
+        for orderpoint in self.with_context(bypass_delay_description=True):
+            op_calendar = orderpoint.warehouse_id.orderpoint_calendar_id
+            if not op_calendar:
+                continue
+            if not orderpoint.product_id or not orderpoint.location_id:
+                orderpoint.lead_days_date = False
+                continue
+            lead_days = orderpoint._get_lead_days()
+            # Get the next planned date to execute this orderpoint
+            start_date = op_calendar.plan_days(1, fields.Datetime.now())
```
rousseldenis: Maybe call now() once before the loop ?

stock_warehouse_calendar_orderpoint/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module adds an Reordering Calendar to the Warehouse. This calendar
+can then used by reordering rules to compute the forecasted date when goods
```
rousseldenis, line 2: @sebalix You can maybe remind where it can be found ?

https://github.com/odoo/odoo/blob/14.0/addons/stock/static/src/xml/stock_orderpoint.xml#L27

rousseldenis (changes requested):
At least harmonize licensing

stock_warehouse_calendar_orderpoint/readme/DESCRIPTION.rst
sebalix: A screenshot is maybe welcome indeed. This PR was also adding this field in the tree view.

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
sebalix: Fixed

stock_warehouse_calendar_orderpoint/__manifest__.py
sebalix: Bad code snippet, all have been fixed now I guess.

sebalix: And it has to be LGPL as the dependency `stock_warehouse_calendar` is LGPL.

ivantodorovich (approved):

sebalix:
The popover forecast widget is now also updated to display the base reordering date so it's easier for the user to understand the computation:
image

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,51 @@
...
+            lead_days = orderpoint._get_lead_days()
+            # Get the next planned date to execute this orderpoint
+            start_date = op_calendar.plan_days(1, now)
+            lead_days_date = start_date + relativedelta.relativedelta(days=lead_days)
+            # Postpone to the next available workday if needed
+            if (
+                orderpoint.warehouse_id.orderpoint_on_workday
+                and orderpoint.warehouse_id.calendar_id
+            ):
+                lead_days_date = datetime.combine(lead_days_date.date(), time.min)
```
mmequignon: If lead_days_date is `2022-06-09 23:00:00` and wh TZ is `Europe/Paris`, then the datetime should be `2022-06-10 01:00:00`.
Also, `combine(datetime, time.min)` is enough.

sebalix: For the second remark you mean this?
```python
lead_days_date = datetime.combine(lead_days_date, time.min)
```
:+1: 
And I guess there is no need to play with the TZ if I do this.

mmequignon: No, because lead_days_date isn't TZ aware. If you apply time.min on it, you might get the wrong day:
- E.G -> 2022-06-02 00:30:00 would become 2022-06-02 00:00:00
But if it is supposed to be UTC+2, we should have 2022-06-01 22:30:00 which would become 2022-06-01 00:00:00

mmequignon: I would say that the rule of thumb is to make a datetime TZ aware before:
- `datetime.date()`
- `combine(datetime, time.min/max)`
- edit: everything that removed the `time` information from a datetime

And to convert it to a naive datetime just before storing the datetime in the DB.

sebalix: Everything has been updated and also tested with a delay as float (with https://github.com/OCA/purchase-workflow/pull/1484) to ensure upcoming compatibility. Thanks for your help.

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,62 @@
...
+        for orderpoint in self.with_context(bypass_delay_description=True):
+            op_calendar = orderpoint.warehouse_id.orderpoint_calendar_id
+            if not op_calendar:
+                continue
+            if not orderpoint.product_id or not orderpoint.location_id:
+                orderpoint.lead_days_date = False
+                continue
+            lead_days = orderpoint._get_lead_days()
+            # Get the next planned date to execute this orderpoint
+            start_date = orderpoint._get_next_reordering_date()
```
sebalix, line 31: TODO: `_get_next_reordering_date` should return Today (like in std) instead of `False`
Goal is being able to use the `orderpoint_on_workday` feature without an orderpoint calendar.

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,62 @@
...
+        return lead_days
+
+    def _get_next_reordering_date(self):
+        self.ensure_one()
+        calendar = self.warehouse_id.orderpoint_calendar_id
+        if calendar:
+            # TODO: should we take into account days off or the reordering
+            # calendar with 'compute_leaves=True' here?
+            return calendar.plan_hours(0, fields.Datetime.now())
+        return False
```
sebalix, line 53: TODO:
```suggestion
        return fields.Datetime.now()
```

stock_warehouse_calendar_orderpoint/models/stock_rule.py
```diff
@@ -0,0 +1,26 @@
...
+        if self.env.context.get("orderpoint_id"):
+            op_id = self.env.context["orderpoint_id"]
+            op = self.env["stock.warehouse.orderpoint"].browse(op_id)
+            reordering_date = op._get_next_reordering_date()
+            # Display the reordering date first so the user will know from which
+            # date the computation is done.
+            if reordering_date:
+                reordering_date = misc.format_date(self.env, reordering_date)
+                title = _("Reordering Date")
+                delay_description = (
```
sebalix, line 22: TODO: As `_get_next_reordering_date` will always return a date, we need to test on the OP calendar here
```suggestion
            op_calendar = op.warehouse_id.orderpoint_calendar_id
            # Display the reordering date first so the user will know from which
            # date the computation is done.
            if op_calendar:
                reordering_date = op._get_next_reordering_date()
                reordering_date = misc.format_date(self.env, reordering_date)
                title = _("Reordering Date")
                delay_description = (
```

stock_warehouse_calendar_orderpoint/models/res_company.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    orderpoint_calendar_id = fields.Many2one(
```
ivantodorovich, line 10: It is only used as default value in `stock.warehouse`, so it's a bit confusing to have it in general settings.

Imagine you have already some warehouses configured (which you do in all databases when you install this module), and then you change this setting.

The warehouses' calendars wouldn't be changed
see (`StockWarehouse._default_orderpoint_calendar_id`)


IMO you should drop this field on the company / general settings. Or at least rename it to `default_orderpoint_*` and clearly state that's going to be used as a default for new warehouses... but TBH having a setting to set a default value on a warehouse is overkill. You only create warehouses once..

stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+
+from odoo import fields, models
+
+
+class StockWarehouse(models.Model):
+    _inherit = "stock.warehouse"
+
+    orderpoint_calendar_id = fields.Many2one(
```
ivantodorovich, line 10: Could you explain the difference between the `orderpoint_calendar_id` and the [`calendar_id` defined in `stock_warehouse_calendar`](https://github.com/OCA/stock-logistics-warehouse/blob/f13ff658c2e8625ad264701108083610aed6fab8/stock_warehouse_calendar/models/stock_warehouse.py#L12)?

It kind of seems to me that, conceptually, they both refer to the same thing, no?
In which case would you use one but not the other?

stock_warehouse_calendar_orderpoint/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,62 @@
...
+class StockWarehouseOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    @api.depends(
+        "rule_ids",
+        "product_id.seller_ids",
+        "product_id.seller_ids.delay",
+        "warehouse_id.orderpoint_calendar_id",
+    )
+    def _compute_lead_days(self):
```
ivantodorovich, line 19: It seems to me this code should be implemented in `stock_warehouse_calendar`, as it would be the same scope only the original implementation maybe forgot about the lead_days computation and its impact in the created purchase orders from orderpoints.

I would love to see a hook method to retrieve the calendar_id here.
(thinking of an usecase we have, to have calendars set on suppliers)

stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
sebalix: The `calendar_id` field is used to define the working hours of your warehouse, while the `orderpoint_calendar_id` is used to define the time when your orderpoints will be run.

stock_warehouse_calendar_orderpoint/models/res_company.py
sebalix: Yes it was in discussion, we are waiting feedback after some usage by the customer so we'll probably rework this wobbly part.

rousseldenis:
@sebalix Do you have any customer feedback on this ?

PR #1413, opened 2022-05-09
[14.0][FIX] stock_available_mrp: normal BoM available to promise

CarlosRoca13:
Normal BoM products (those which are manufactured) are regular stored
products and their immediately_usable_qty will be summed with potential
stock. This was the expected behavior of the module and it was lost at
some point.

Phantom BoM products (kits) don't have real stock so their available to
promise quantity will be the same as the potential.

As an improvement, we've added the possibility to override the sum of
potential and available to promise. In some cases such addition doesn't
make sense as we don't know how long can take to manufacture those potential
units.

cc @Tecnativa TT35841

fw-port of 

- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1403

please @chienandalu @victoralmau

### DIFF
--- a/stock_available_mrp/README.rst
+++ b/stock_available_mrp/README.rst
@@ -37,14 +37,24 @@ with the components available to promise.
 .. contents::
    :local:
 
+Usage
+=====
+
+It's possible to avoid adding potential stock to available to promise stock at Bill of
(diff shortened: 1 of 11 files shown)


chienandalu (approved):
Thanks ;)

victoralmau (approved):

PR #1414, opened 2022-05-09
[14.0][IMP] scrap_reason_code (Require Reason Code & Scrap Location)

patrickrwilson:
If you try to perform a scrap without a reason code or the reason code doesn't have a scrap location, it will give an error to the user that a mandatory field is needed. This is easily solved by requiring the reason and scrap location on the reason.

### DIFF
--- a/scrap_reason_code/views/reason_code_view.xml
+++ b/scrap_reason_code/views/reason_code_view.xml
@@ -14,7 +14,7 @@
                     <field name="description" />
                 </group>
                 <group>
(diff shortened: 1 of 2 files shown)


patrickrwilson:
FYI: After discussing this with others, we will change this so it won't require the fields however if the reason code is not set then it will use the scrap location on the scrap order.

patrickrwilson:
Closing to reduce confusion and will open a new PR.

PR #1415, opened 2022-05-10
[FIX] scrap_reason_code (Location Not Set)

patrickrwilson:
Currently the reason code and scrap location on the code has to be set otherwise an error is generated stating the location is required.
This fixes that so if the reason code is not used or there is not a location on the code then the location on the default scrap location on the order is used, otherwise it will use the code's location.

### DIFF
--- a/scrap_reason_code/models/stock_scrap.py
+++ b/scrap_reason_code/models/stock_scrap.py
@@ -25,21 +25,23 @@ class StockScrap(models.Model):
 
     def write(self, vals):
         if "reason_code_id" in vals:
-            vals.update(
-                {
-                    "scrap_location_id": self.env["scrap.reason.code"]
-                    .browse(vals.get("reason_code_id"))
-                    .location_id
-                }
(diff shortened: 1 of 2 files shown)


rousseldenis (commented):
@patrickrwilson Thanks for this.

Is this covered by a test ?

patrickrwilson:
@rousseldenis I will work on getting the test updated.

patrickrwilson:
@rousseldenis @dreispt this is ready for your review once it's done with checks. Thank you.

bodedra (changes requested):
cherry-pick this commit 3a4a74923900bf0293a45144bda8ccdafd6a226f and add a new commit for the unit test 

V13 Fix PR Ref: https://github.com/OCA/stock-logistics-warehouse/pull/953

patrickrwilson:
I managed to mess up these commits so I will close this and create a new clean PR instead of trying to continue to clean this one.

PR #1416, opened 2022-05-11
[14.0][FIX] scrap_reason_code (Location Set Fix)

patrickrwilson:
There is an error if no location is set on the scrap reason or no scrap reason is set on the order. Because the onchange sets the location on the order this removes the create/write where it's also setting the location.


dreispt:
This looks like it is doing the same as an onchange.
Why not implementing it as an onchange?

patrickrwilson:
@bodedra fixed this issue in v13, so this just cherry picks that same fix and adds test for coverage.
@rousseldenis this is available for your review also.

scrap_reason_code/models/stock_scrap.py
```diff
@@ -25,21 +25,23 @@ def _onchange_reason_code_id(self):
 
     def write(self, vals):
         if "reason_code_id" in vals:
```
rousseldenis: As the code is exactly the same in both functions:

```suggestion
        self._update_scrap_reason_code_location(vals)
        
        (...)
        def _update_scrap_reason_code_location(vals):
            if "reason_code_id" in vals:
            (...)
```

rousseldenis: So, if I understand well, if a reason is entered in scrap and that reason has a location, change the scrap location ?

patrickrwilson: Yes, that is how it works in my testing. I could try to change this to just function via the onchange as @dreispt suggested in another comment, but I'm not sure if there was another reason as to why the developers who created this decided to have the create/write methods, maybe in the event a scrap is called from another process maybe. @bodedra do you know the reasoning for the write/create methods?

bodedra: @patrickrwilson I don't have all the Customer scenarios for the scrapping but if I recall correctly, there are many cases onchange doesn't trigger, meaning the field is not visible in the views. Scrapping parts from MO, WO and Inventory level. I am open to replacing logic into an onchange method.

patrickrwilson: I've updated it removing the create/write methods and the new test. Testing works fine with just the onchange event. It won't let me squash the commits though, so if it's better to close this PR and create yet another so the commits are clean, please let me know if that is preferred.

rousseldenis: My question at 10 000 $ is: wasn't write/create code done to work with imports (don't want to remove functionnality for existing users)?

rousseldenis: @bodedra

bodedra: @rousseldenis yup, thank you. I recall it now. I developed it to support imports and external system interaction where onchange doesn't trigger. Let's final with write/create method. @patrickrwilson sorry for the back and forth.

patrickrwilson: No problem, because it won't let me squash commits and such I'm going to close this and reopen one more to have a clean PR.

patrickrwilson:
Closing to have a clean PR since it won't let me squash these commits.

PR #1418, opened 2022-05-16
[IMP] Update templates with copier

rousseldenis:


rousseldenis:
This one is less intrusive than #1400

PR #1427, opened 2022-06-02
[14.0][FIX] stock_picking_cancel_confirm: cancel pickings without lines

ps-tubtim:
This pull request fix is about canceled pickings without lines.

**Before:**  Pickings not cancelled
cancel_picking

**After:** Pickings cancelled
cancel_picking_2

### DIFF
--- a/stock_picking_cancel_confirm/models/stock_picking.py
+++ b/stock_picking_cancel_confirm/models/stock_picking.py
@@ -12,4 +12,9 @@ class StockPicking(models.Model):
     def action_cancel(self):
         if not self.filtered("cancel_confirm"):
             return self.open_cancel_confirm_wizard()
-        return super().action_cancel()
+        res = super().action_cancel()
+        # Ensure to cancel the picking w/o moves that wasn't cancelled by odoo core
+        self.filtered(lambda l: l.state != "cancel" and not l.move_lines).write(
+            {"state": "cancel"}
+        )
+        return res
--- a/stock_picking_cancel_confirm/README.rst
+++ b/stock_picking_cancel_confirm/README.rst
@@ -53,7 +53,10 @@ Authors
 Contributors
 ~~~~~~~~~~~~
 
-* Kitti U. <kittiu@ecosoft.co.th>
+* `Ecosoft <http://ecosoft.co.th>`__:
+
+  * Kitti Upariphutthiphong. <kittiu@ecosoft.co.th>
+  * Pimolnat Suntian <pimolnats@ecosoft.co.th>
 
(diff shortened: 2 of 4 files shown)


stock_picking_cancel_confirm/wizard/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import cancel_confirm
```
kittiu, line 3: **In my option, we don't need to patch any of this wizard.**
It was the logic of Odoo core in picking.action_cancel() that do not cancel picking (whether it is a bug or the intention)

BUT, if we need to fix it we should fix at the right place, which is in action_cancel() itself.

kittiu: **In my option, we don't need to patch any of this wizard.**
It was the logic of Odoo core in picking.action_cancel() that do not cancel picking (whether it is a bug or the intention)

BUT, if we need to fix it we should fix at the right place, which is in action_cancel() itself.

kittiu: I propose the following code for **def action_cancel()**

```
    def action_cancel(self):
        if not self.filtered("cancel_confirm"):
            return self.open_cancel_confirm_wizard()
        res = super().action_cancel()
        # Ensure to cancel the picking w/o moves that wasn't cancelled by odoo core
        self.filtered(lambda l: l.state != "cancel" and not l.move_lines).write({"state": "cancel"})
        return res
```

ps-tubtim: @kittiu Thank you for your suggestion. I will change it.

stock_picking_cancel_confirm/models/stock_picking.py
```diff
@@ -12,4 +12,9 @@ class StockPicking(models.Model):
     def action_cancel(self):
         if not self.filtered("cancel_confirm"):
             return self.open_cancel_confirm_wizard()
-        return super().action_cancel()
+        res = super().action_cancel()
+        # Ensure to cancel the picking w/o moves that wasn't cancelled by odoo core
+        self.filtered(lambda l: l.state != "cancel" and not l.move_lines).write(
```
rousseldenis, line 17: @ps-tubtim Could you add a test with no moves ?

But I think as state field is computed one and readonly, this should not work.

In core, picking remains in 'draft' if there is no move inside.

rousseldenis:
@ps-tubtim

PR #1432, opened 2022-06-07
[15.0][MIG] stock_secondary_unit: Migration to 15.0

CarlosRoca13:
cc @Tecnativa TT36576

please @sergio-teruel @carlosdauden review this

PR #1434, opened 2022-06-08
15.0 mig stock secondary unit

st-anagpal:


sergio-teruel:
Hi @st-anagpal There is already an open PR: https://github.com/OCA/stock-logistics-warehouse/pull/1432

PR #1437, opened 2022-06-10
[FIX] stock_request_picking_type: ValueError: malformed node or string

max3903:


stock_request_picking_type/__manifest__.py
```diff
@@ -15,6 +15,6 @@
         "views/stock_request_order_views.xml",
         "views/stock_picking_views.xml",
     ],
-    "development_status": "Beta",
+    "development_status": "Production/Stable",
```
max3903:
```suggestion
    "development_status": "Beta",
```

PR #1441, opened 2022-06-17
[14.0][IMP] stock_quant_manual_assign: make several columns sortable in wizard

MiquelRForgeFlow:
We want to be able to sort by some columns in the wizard. Only stored columns are sortable.

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -116,6 +116,7 @@ class AssignManualQuantsLines(models.TransientModel):
     _name = "assign.manual.quants.lines"
     _description = "Assign Manual Quants Lines"
     _rec_name = "quant_id"
+    _order = "location_id, lot_id, package_id"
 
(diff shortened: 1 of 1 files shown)


yostashiro (approved):
LGTM

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -116,6 +116,7 @@ class AssignManualQuantsLines(models.TransientModel):
     _name = "assign.manual.quants.lines"
     _description = "Assign Manual Quants Lines"
     _rec_name = "quant_id"
+    _order = "lot_id, location_id, package_id"
```
pedrobaeza: I think you should sort first by location

MiquelRForgeFlow: Done :)

pedrobaeza (approved):

MiquelRForgeFlow:
Forward ported in https://github.com/OCA/stock-logistics-warehouse/pull/1445.

PR #1443, opened 2022-06-20
[13.0][ADD] stock_picking_dest_moves

albariera:
This addon allows to easily access destination pickings (in the case of chained routes) through a button in the "parent" picking view.


stock_picking_dest_moves/views/stock_picking.xml
```diff
@@ -0,0 +1,36 @@
...
+                    <div class="o_field_widget o_stat_info">
+                       <span class="o_stat_value">
+                            <field
+                                name="dest_picking_count"
+                                widget="statinfo"
+                                nolabel="1"
+                                class="mr4"
+                            />
+                        </span>
+                        <span class="o_stat_text">Destination Pickings</span>
```
LoisRForgeFlow: "Destination Transfer" would be more aligned with the typical UI nomenclature for stock.picking.

LoisRForgeFlow (approved):
Tested on runboat :+1:

stock_picking_dest_moves/views/stock_picking.xml
albariera: You're right! Thanks :)

JordiBForgeFlow:
@albariera Would also be interesting to add preceding pickings. Then the module would named something like 'stock_picking_show_linked'.  @juanydforgeflow has been working on preparing something like this.

albariera:
@JordiBForgeFlow Thanks! @JuanyDForgeflow could you send the links of what you are working on? :blush:

JuanyDForgeflow:
@albariera this is the link https://github.com/OCA/stock-logistics-warehouse/pull/1444

rousseldenis:
@LoisRForgeFlow @JordiBForgeFlow @albariera @JuanyDForgeflow 
So, could we converge in order to not duplicate features in different modules ?

albariera:
If everyone agrees I'll close this PR, as this functionality is already covered and extended at #1444, so I'll just backport it to 13.0.

PR #1450, opened 2022-07-12
[14.0][FIX] stock_quant_manual_assign: fix access error issues

LoisRForgeFlow:
When not all features are activated in inventory (locations, lots,
packages and owners) selecting a quant would raise an access error.

It is not possible to directly read the related fields, e.g
`self.lot_id`, because they are stored fields and have a group
restriction, so an access error would raise. To work around it,
we should access these fields from the quant:
`self.quant_id.lot_id`.

@ForgeFlow


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
A forward port should be done.

pedrobaeza (approved):

PR #1455, opened 2022-07-18
[13.0] stock_request: issue with backorder.

adrienpeiffer:
When creating several stock requests on the same procurement group with the same product, only one stock move exists. When the stock move is partially delivered to complete the first stock request, it works. When the rest of the quantity is delivered with the backorder, the quantities are added to the stock request already done. 

https://youtu.be/eXmPlXeiADY 

https://github.com/OCA/stock-logistics-warehouse/runs/7391979937#step:8:667


adrienpeiffer:
https://github.com/OCA/stock-logistics-warehouse/runs/7391979937?check_suite_focus=true#step:8:670

stock_request/models/stock_move.py
```diff
@@ -55,24 +55,82 @@ def copy_data(self, default=None):
...
+            "request_name": request.name,
+            "picking_name": move.picking_id.name,
+            "product_name": move.product_id.name_get()[0][1],
+            "product_qty": allocated_qty,
+            "product_uom": move.product_uom.name,
+            "location_name": move.location_dest_id.name_get()[0][1],
+        }
+
     def _action_done(self, cancel_backorder=False):
+        for move in self.filtered(lambda m: m.exists() and m.allocation_ids):
```
rousseldenis, line 117: @adrienpeiffer Is this treatment could not be done in check_done() instead ?

adrienpeiffer: I'm not sure I understand. Do you mean in the check_done method of stock request? If so, then my answer is no. The check_done method is responsible for checking if we can change the status of the stock request not for changing the quantities on the allocations.

rousseldenis: Ok, so, moving that code in a separate function should be better.

adrienpeiffer: Which part of the code do you mean ?

rousseldenis (approved):
LGTM

sebalix:
@adrienpeiffer can you remove the terms `13.0` from the commit? They will obviously be ported to other branches afterwards.

adrienpeiffer:
@sebalix Done !

adrienpeiffer:
@sebalix Is this one ok for you ?

rousseldenis:
@sebalix Could you update your review ?

PR #1461, opened 2022-07-30
[14.0] add stock_inventory_location_state

sebastienbeau:
Add a module to track location done during an inventory (needed as dependency for shopfloor)


rousseldenis:
@sebastienbeau Could you fix tests and rebase ?

PR #1464, opened 2022-08-08
[15.0][MIG] stock_orderpoint_procure_location

AaronHForgeFlow:
This module allows to procure in a different location than the location of the orderpoint.

cc @ForgeFlow @JoanSForgeFlow


stock_orderpoint_procure_location/tests/test_stock_orderpoint_procure_location.py
```diff
@@ -0,0 +1,176 @@
...
+                            "delay": 5,
+                            "min_qty": 1,
+                            "price": 10,
+                        },
+                    )
+                ],
+            }
+        )
+
+    def _create_purchase(self, line_products):
```
rousseldenis, line 94: @AaronHForgeFlow It seems that this is not used.

PR #1471, opened 2022-08-25
[13.0] stock_available: Update models/product_template.py

vkanev:
Negative quantities on any variant should not have any impact on product.template availability at all.

Example:
We have a t-shirt in two colors with following availabilities:
red - 1 piece in stock
white - 0 pieces in stock.

Someone pre-order a white t-shirt and qty will be -1.

Result:
t-shirt (product.template) immediately_usable_qty = 0
red t-shirt (product variant of the same product.template) immediately_usable_qty = 1
white t-shirt (product variant of the same product.template) immediately_usable_qty = -1 - MUST be purchased from supplier OR cancel the sale order BUT should not affect product.template available qty.

### DIFF
--- a/stock_available/models/product_template.py
+++ b/stock_available/models/product_template.py
@@ -31,7 +31,7 @@ class ProductTemplate(models.Model):
                 [
                     variants_dict[p.id]["immediately_usable_qty"]
                     - variants_dict[p.id]["potential_qty"]
-                    for p in template.product_variant_ids
+                    for p in template.product_variant_ids if p.immediately_usable_qty > 0 #negative quantities should not have any impact on availability
                 ]
             )
             potential_qty = max(
--- a/stock_available/models/product_template.py
+++ b/stock_available/models/product_template.py
@@ -31,7 +31,8 @@ class ProductTemplate(models.Model):
                 [
                     variants_dict[p.id]["immediately_usable_qty"]
                     - variants_dict[p.id]["potential_qty"]
-                    for p in template.product_variant_ids if p.immediately_usable_qty > 0 #negative quantities should not have any impact on availability
+                    for p in template.product_variant_ids
+                    if p.immediately_usable_qty > 0 #negative quantities should not have any impact on availability
                 ]
             )
             potential_qty = max(


rousseldenis:
@vkanev This is not true.

You should fix the negative stock quantity instead.

rousseldenis:
To help you, there is this module: https://github.com/OCA/stock-logistics-workflow/tree/14.0/stock_no_negative

vkanev:
Temporary negative stock is possible to be present when internal stock checks are triggered by hand. This way using "procurement rules" Odoo creates POs to Suppliers and "fixes" the negative stock in natural way. BUT this can be done manually and meanwhile negative stock impacts product_template stock availability.

pedrobaeza:
I agree this module shouldn't filter out negatives quantities, and template availability is only informative. Everything is done at variant (product.product) level.

rousseldenis:
@vkanev Could you add a test to highlight this ?

stock_available/models/product_template.py
```diff
@@ -32,6 +32,7 @@ def _compute_available_quantities_dict(self):
                     variants_dict[p.id]["immediately_usable_qty"]
                     - variants_dict[p.id]["potential_qty"]
                     for p in template.product_variant_ids
+                    if p.immediately_usable_qty > 0 #negative quantities should not have any impact on availability
```
rousseldenis, line 35: You should use float_compare

rousseldenis:
> I agree this module shouldn't filter out negatives quantities, and template availability is only informative. Everything is done at variant (product.product) level.

I see

pedrobaeza:
This should be closed.

vkanev:
@rousseldenis Sorry but never done this, so don't know how to do it properly.

My use case is that I have same product with different variants. Some variants are sold only by request and as a result their quantities are most of the time zero, or whenever they are ordered - negative for a couple of days before scheduling procurement. During these couple of days or more (in rare cases) the negative stock is something that shouldn't impact other quantities.

As this is, as I see, my use case, it's not spread wide :)

pedrobaeza:
Check the availability at variant level, not at template level. When handling variants, such information at template level doesn't serve too much even if no negatives.

PR #1472, opened 2022-08-26
[13.0][IMP] account_move_line_stock_info: show stock move in list view

AaronHForgeFlow:
Put the stock move in the journal item tree view. I think it is very useful when analyzing the journal items of a stock transfer:
image

Also, some pre-commit fixes included
Also, restrict the stock move views to stock users

cc @ForgeFlow

### DIFF
--- a/account_move_line_stock_info/views/account_move_line_view.xml
+++ b/account_move_line_stock_info/views/account_move_line_view.xml
@@ -6,8 +6,23 @@
         <field name="inherit_id" ref="account.view_move_line_form" />
         <field name="arch" type="xml">
             <field name="quantity" position="before">
-                <field name="stock_move_id" />
+                <field name="stock_move_id" groups="stock.group_stock_user" />
             </field>
         </field>
     </record>
+    <record id="view_move_line_tree" model="ir.ui.view">
+        <field name="name">account.move.line.treee</field>
+        <field name="model">account.move.line</field>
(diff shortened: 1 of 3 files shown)


DavidJForgeFlow:
image
I get this error when installing, actually quantity field doesn't appear in account.view_move_line_tree, maybe some dependency is missing?

AaronHForgeFlow:
@DavidJForgeFlow you were right, the quantity field is in the account_move_line_product module. But the dependency is not needed, I just changed the view. Thanks!

account_move_line_stock_info/views/account_move_line_view.xml
```diff
@@ -6,8 +6,19 @@
...
             </field>
         </field>
     </record>
+    <record id="view_move_line_tree" model="ir.ui.view">
+        <field name="name">account.move.line.treee</field>
+        <field name="model">account.move.line</field>
+        <field name="inherit_id" ref="account.view_move_line_tree" />
+        <field name="arch" type="xml">
+            <field name="partner_id" position="before">
+                <field name="stock_move_id" groups="stock.group_stock_user" />
```
rousseldenis: @AaronHForgeFlow Maybe optional hide ?

AaronHForgeFlow: Yes, I am fine with that

rousseldenis (approved):

PR #1491, opened 2022-09-29
[13.0][IMP] stock_orderpoint_purchase_link: Add smart-button on orderpoints

pedrobaeza:
We provide a link from the orderpoint to the involved purchase orders.

@Tecnativa TT39644

PR #1492, opened 2022-09-30
[14.0][IMP] stock_orderpoint_purchase_link: Add smart-button on orderpoints

pedrobaeza:
We provide a link from the orderpoint to the involved purchase orders.

FW-port of #1491 

@Tecnativa TT39644

### DIFF
--- a/stock_orderpoint_purchase_link/models/stock_warehouse_orderpoint.py
+++ b/stock_orderpoint_purchase_link/models/stock_warehouse_orderpoint.py
@@ -1,7 +1,8 @@
 # Copyright 2018-20 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2022 Tecnativa - Pedro M. Baeza
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
 
 
 class StockWarehouseOrderpoint(models.Model):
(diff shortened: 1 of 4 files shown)


stock_orderpoint_purchase_link/models/stock_warehouse_orderpoint.py
```diff
@@ -13,3 +14,17 @@ class StockWarehouseOrderpoint(models.Model):
...
+
+    @api.depends("purchase_line_ids")
+    def _compute_purchase_count(self):
+        for record in self:
+            record.purchase_count = len(record.purchase_line_ids.order_id)
+
+    def button_view_purchases(self):
+        self.ensure_one()
+        orders = self.purchase_line_ids.order_id
+        action = self.env.ref("purchase.purchase_rfq").sudo().read()[0]
```
rousseldenis: Please use _for_xml_id

rousseldenis (changes requested):

stock_orderpoint_purchase_link/models/stock_warehouse_orderpoint.py
pedrobaeza: Well, in this case it's the same, isn't it?

rousseldenis: Odoo provides function to do that, so if the implementation changes, nothing to do here...

pedrobaeza: OK, changed.

victoralmau (approved):

rousseldenis (approved):

PR #1497, opened 2022-10-06
[14.0][FIX] stock_quant_reservation_info: only reservations from internal locations

DavidJForgeFlow:
Previously when seeing the reservations of the product, all reservations were being added, we just want the ones that come from internal locations (the ones that are being counted on product views).

@ForgeFlow


LoisRForgeFlow (approved):
Thanks!

PR #1502, opened 2022-10-10
[16.0] [MIG] stock_packaging_calculator

TDu:


lmignon (approved):

lmignon:
@rousseldenis Your review is welcome...

stock_packaging_calculator/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+{
+    "name": "Stock packaging calculator",
+    "summary": "Compute product quantity to pick by packaging",
+    "version": "16.0.1.0.0",
+    "development_status": "Alpha",
```
rousseldenis: @TDu @simahawk Still Alpha ?

TDu: It is installed on at least one production instance, so yes the status could be changed. Now for which one ? I am not sure ? Mature, Production/Stable or still Beta ?

rousseldenis: You can switch to Beta which is the default status

TDu: Thanks for your answer, done.

simahawk:
Wondering: shall we take the time finally to move it to product-attribute and rename it to `product_packaging_calculator`?
I named it like this because is mostly used for logistics but it has no dependency on stock modules.

rousseldenis:
> Wondering: shall we take the time finally to move it to product-attribute and rename it to `product_packaging_calculator`? I named it like this because is mostly used for logistics but it has no dependency on stock modules.

@simahawk I tend to say all stuffs related to packaging should go into stock-logistics-tracking repository (see my previous contributors mail about that).

Even if in this case, this depends on product module only

rousseldenis:
Let this as is. We can evaluate to migrate to product repo on further versions.

/ocabot merge patch

PR #1503, opened 2022-10-10
[16.0] [MIG] stock_mts_mto_rule

msteinfeld-sirum:


stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -43,16 +43,13 @@ def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
```
rousseldenis: @msteinfeld-sirum  Why changing this ?

msteinfeld-sirum: @rousseldenis cause we want the free quantity. Virtual available ist now incoming and outgoing only
So free quantity ist real available quantity

rousseldenis: @msteinfeld-sirum Yes, I know that, but that was done at purpose to use virtual available field to order.

If you want to use another, please maybe add a hook and make your own implementation.

I don't have the use case we want to use the free_qty.

florian-dacosta: I guess using the free_qty can make sense in some use cases, but virtual qty make sense too and it has been this way for years.
You can't just change the default behavior of the module, but I guess it could be a parameter on company or warehouse level for instance. (with default configuration beeing virtual qty)

stock_mts_mto_rule/view/pull_rule.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
     <record id="stock_location_route_form_view_mto_mto" model="ir.ui.view">
-        <field name="name">stock.location.route.form.mts.mto</field>
-        <field name="model">stock.location.route</field>
+        <field name="name">stock.location.form.mts.mto</field>
```
rousseldenis: `<field name="name">stock.route.form.mts.mto</field>`

rousseldenis (changes requested):

stock_mts_mto_rule/models/stock_rule.py
msteinfeld-sirum: @rousseldenis changed it back to original code

@florian-dacosta it was not my aim to change the behaviour. it was my aim to keep the right behaviour, as i was assuming, the field free_qty was new..but it isn't so i changed back to virtual_available.
The configuration at warehouse level is on my todo list for a later PR

stock_mts_mto_rule/view/pull_rule.xml
msteinfeld-sirum: @rousseldenis changed

rousseldenis (approved):

florian-dacosta (approved):
LGTM

PR #1512, opened 2022-10-21
[15.0][ADD] stock_inventory

DavidJForgeFlow:
This module allows to group Inventory Adjustments and have a group traceability (like before 15.0).

@ForgeFlow

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -1,0 +1,216 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.osv import expression
+
+
+class InventoryAdjustmentsGroup(models.Model):
+    _name = "stock.inventory"
+    _description = "Inventory Adjustment Group"
+    _order = "date desc, id desc"
+
+    name = fields.Char(required=True, default="Inventory", string="Inventory Reference")
+
+    date = fields.Datetime(default=lambda self: fields.Datetime.now())
+
+    state = fields.Selection(
+        [("draft", "Draft"), ("in_progress", "In Progress"), ("done", "Done")],
+        default="draft",
+    )
+
+    owner_id = fields.Many2one(
+        "res.partner", "Owner", help="This is the owner of the inventory adjustment"
+    )
+
+    location_ids = fields.Many2many(
+        "stock.location", string="Locations", domain="[('usage', '=', 'internal')]"
+    )
+
+    product_selection = fields.Selection(
+        [
+            ("all", "All Products"),
+            ("manual", "Manual Selection"),
+            ("category", "Product Category"),
+            ("one", "One Product"),
+            ("lot", "Lot/Serial Number"),
+        ],
+        default="all",
+        required=True,
+    )
+
+    product_ids = fields.Many2many("product.product", string="Products")
+
+    stock_quant_ids = fields.Many2many("stock.quant", string="Inventory Adjustment")
+
+    category_id = fields.Many2one("product.category", string="Product Category")
+
+    lot_ids = fields.Many2many(
+        "stock.production.lot",
+        string="Lot/Serial Numbers",
+    )
+
+    stock_move_ids = fields.One2many(
+        "stock.move.line",
+        "inventory_adjustment_id",
+        string="Inventory Adjustments Done",
+    )
+
+    count_stock_quants = fields.Integer(
+        compute="_compute_count_stock_quants", string="Adjustments"
+    )
+
+    count_stock_quants_string = fields.Char(
+        compute="_compute_count_stock_quants", string="Adjustments"
+    )
+
+    count_stock_moves = fields.Integer(
+        compute="_compute_count_stock_moves", string="Stock Moves Lines"
+    )
+
+    @api.depends("stock_quant_ids")
+    def _compute_count_stock_quants(self):
+        self.count_stock_quants = len(self.stock_quant_ids)
+        count_todo = len(
+            self.stock_quant_ids.search(
+                [("id", "in", self.stock_quant_ids.ids), ("to_do", "=", "True")]
+            )
+        )
+        self.count_stock_quants_string = "{} / {}".format(
+            count_todo, self.count_stock_quants
+        )
+
+    @api.depends("stock_move_ids")
+    def _compute_count_stock_moves(self):
+        sm_ids = self.mapped("stock_move_ids").ids
+        self.count_stock_moves = len(sm_ids)
+
+    def _get_quants(self, locations):
+        domain = []
+        base_domain = self._get_base_domain(locations)
+        if self.product_selection == "all":
+            domain = self._get_domain_all_quants(base_domain)
+        elif self.product_selection == "manual":
+            domain = self._get_domain_manual_quants(base_domain)
+        elif self.product_selection == "one":
+            domain = self._get_domain_one_quant(base_domain)
+        elif self.product_selection == "lot":
+            domain = self._get_domain_lot_quants(base_domain)
+        elif self.product_selection == "category":
+            domain = self._get_domain_category_quants(base_domain)
+        return self.env["stock.quant"].search(domain)
+
+    def _get_base_domain(self, locations):
+        return [
+            "|",
+            ("location_id", "in", locations.mapped("id")),
+            ("location_id", "in", locations.child_ids.ids),
+        ]
+
+    def _get_domain_all_quants(self, base_domain):
+        return base_domain
+
+    def _get_domain_manual_quants(self, base_domain):
+        return expression.AND(
+            [base_domain, [("product_id", "in", self.product_ids.ids)]]
+        )
+
+    def _get_domain_one_quant(self, base_domain):
+        return expression.AND(
+            [
+                base_domain,
+                [
+                    ("product_id", "in", self.product_ids.ids),
+                ],
+            ]
+        )
+
+    def _get_domain_lot_quants(self, base_domain):
+        return expression.AND(
+            [
+                base_domain,
+                [
+                    ("product_id", "in", self.product_ids.ids),
+                    ("lot_id", "in", self.lot_ids.ids),
+                ],
+            ]
+        )
+
+    def _get_domain_category_quants(self, base_domain):
+        return expression.AND(
+            [
+                base_domain,
+                [
+                    "|",
+                    ("product_id.categ_id", "=", self.category_id.id),
+                    ("product_id.categ_id", "in", self.category_id.child_id.ids),
+                ],
+            ]
+        )
+
+    def action_state_to_in_progress(self):
+        active_rec = self.env["stock.inventory"].search(
+            [
+                ("state", "=", "in_progress"),
+                "|",
+                ("location_ids", "in", self.location_ids.mapped("id")),
+                ("location_ids", "in", self.location_ids.child_ids.ids),
+            ],
+            limit=1,
+        )
+        if active_rec:
+            raise ValidationError(
+                _(
+                    "There's already an Adjustment in Process using one requested Location: %s"
+                )
+                % active_rec.name
+            )
+        self.state = "in_progress"
+        self.stock_quant_ids = self._get_quants(self.location_ids)
+        self.stock_quant_ids.update({"to_do": True})
+        return
+
+    def action_state_to_done(self):
+        self.state = "done"
+        self.stock_quant_ids.update({"to_do": True})
+        return
+
+    def action_state_to_draft(self):
+        self.state = "draft"
+        self.stock_quant_ids.update({"to_do": True})
+        self.stock_quant_ids = None
+        return
+
+    def action_view_inventory_adjustment(self):
+        result = self.env["stock.quant"].action_view_inventory()
+        ia_ids = self.mapped("stock_quant_ids").ids
+        result["domain"] = [("id", "in", ia_ids)]
+        result["search_view_id"] = self.env.ref("stock.quant_search_view").id
+        result["context"]["search_default_to_do"] = 1
+        return result
+
+    def action_view_stock_moves(self):
+        result = self.env["ir.actions.act_window"]._for_xml_id(
+            "stock_inventory.action_view_stock_move_line_inventory_tree"
+        )
+        sm_ids = self.mapped("stock_move_ids").ids
+        result["domain"] = [("id", "in", sm_ids)]
+        result["context"] = []
+        return result
+
+    @api.constrains("product_selection", "product_ids")
+    def _check_one_product_in_product_selection(self):
+        for rec in self:
+            if len(rec.product_ids) > 1:
+                if rec.product_selection == "one":
+                    raise ValidationError(
+                        _(
+                            "When 'Product Selection: One Product' is selected"
+                            " you are only able to add one product."
+                        )
+                    )
+                elif rec.product_selection == "lot":
+                    raise ValidationError(
+                        _(
+                            "When 'Product Selection: Lot Serial Number' is selected"
+                            " you are only able to add one product."
+                        )
+                    )
--- a/setup/stock_inventory/odoo/addons/stock_inventory
+++ b/setup/stock_inventory/odoo/addons/stock_inventory
@@ -1,0 +1,1 @@
+../../../../stock_inventory
--- a/setup/stock_inventory/setup.py
+++ b/setup/stock_inventory/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/stock_inventory/__init__.py
+++ b/stock_inventory/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/stock_inventory/__manifest__.py
+++ b/stock_inventory/__manifest__.py
@@ -1,0 +1,20 @@
+{
+    "name": "Stock Inventory Adjustment",
+    "version": "15.0.1.0.0",
+    "license": "LGPL-3",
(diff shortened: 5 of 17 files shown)


rousseldenis:
@DavidJForgeFlow Thanks for this. @pedrobaeza I think you could look at it following what we discussed on OCA days.

pedrobaeza:
Yeah, I see.

David, please name the module `stock_inventory_adjustment` (without `s` and I suppose you wanted to say "grouped inventory", but I think that having "inventory adjustment" in the name is enough.

stock_group_inventory_adjustments/models/stock_adjustment.py
```diff
@@ -0,0 +1,83 @@
+import datetime
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class InventoryAdjustmentsGroup(models.Model):
+    _name = "stock.adjustment"
```
pedrobaeza: Can't this still be called `stock.inventory` for avoiding migration scripts?

rousseldenis: Maybe we can have a feedback by someone having done Odoo migrations to 15.0 to see what they have done with those records.

DavidJForgeFlow:
> Yeah, I see.
> 
> David, please name the module `stock_inventory_adjustment` (without `s` and I suppose you wanted to say "grouped inventory", but I think that having "inventory adjustment" in the name is enough.

All references changed. Also changed the model name to stock.inventory, is it ok like this?

stock_inventory_adjustment/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Stock Inventory Adjustment",
+    "version": "15.0.1.1.0",
```
GuillemCForgeFlow:
```suggestion
    "version": "15.0.1.0.0",
```

stock_inventory_adjustment/views/stock_inventory.xml
```diff
@@ -0,0 +1,124 @@
...
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            type="object"
+                            name="action_view_inventory_adjustment"
+                            class="oe_stat_button"
+                            icon="fa-pencil-square-o"
+                            string="Adjustments"
+                            attrs="{'invisible':[('state', 'in', ['2', '3'])]}"
+                        >
+                            <field name="count_stock_quants" readonly="1" />
```
GuillemCForgeFlow:
```suggestion
                            attrs="{'invisible':['|', ('state', 'in', ['2', '3']), ('count_stock_quants', '=', 0)]}"
                        >
                            <field name="count_stock_quants" readonly="1" string="Adjustments" />
```
For all smart buttons:
- Move the string to the field
- Also, make it invisible if the count is 0.

DavidJForgeFlow: Done!

GuillemCForgeFlow: Great :+1: . I will wait for the other changes then.

stock_inventory_adjustment/models/stock_inventory.py
```diff
@@ -0,0 +1,87 @@
...
+    def _compute_count_stock_quants(self):
+        self.count_stock_quants = len(self.stock_quant_ids)
+
+    @api.depends("stock_move_ids")
+    def _compute_count_stock_moves(self):
+        sm_ids = self.mapped("stock_move_ids").ids
+        self.count_stock_moves = len(sm_ids)
+
+    def action_state_to_in_progress(self):
+        active_rec = self.env["stock.inventory"].search([("state", "=", "1")])
```
rousseldenis: I don't like the use of integer as state value as looking at this code, you cannot determine what state is behind

stock_inventory_adjustment/models/stock_inventory.py
```diff
@@ -0,0 +1,87 @@
...
+        if active_rec:
+            raise ValidationError(
+                _("There's already an Adjustment in Process: %s") % active_rec.name
+            )
+        self.state = "1"
+        if self.product_selection == "all":
+            for location_id in self._origin.location_ids:
```
rousseldenis: Please use:
```suggestion
            self.stock_quant_ids : self.env["stock.quant"].search([("location_id", "in", self._origin.location_ids.ids)])
            
```

DavidJForgeFlow: Done! thank you!

stock_inventory_adjustment/models/stock_inventory.py
DavidJForgeFlow: Also done, thank you again!

stock_inventory_adjustment/views/stock_inventory.xml
DavidJForgeFlow: You changes are done :smile:

GuillemCForgeFlow (approved):
LGTM! :top: :+1:

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+        "stock.location", string="Location", domain="[('usage', '=', 'internal')]"
+    )
+
+    product_selection = fields.Selection(
+        [
+            ("all", "All Products"),
+            ("manual", "Manual Selection"),
+            ("category", "Product Category"),
+            ("one", "One Product"),
+            ("lot", "Lot Serial Number"),
```
CasVissers-360ERP: String = "Lot" would be cleaner here?

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+
+    product_ids = fields.Many2many("product.product", string="Products")
+
+    stock_quant_ids = fields.Many2many("stock.quant", string="Inventory Adjustment")
+
+    category_id = fields.Many2one("product.category", string="Product Category")
+
+    lot_ids = fields.Many2many(
+        "stock.production.lot",
+        string="Lot/Serial Number",
```
CasVissers-360ERP: String = "Lots" would be cleaner here?

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+
+    @api.depends("stock_move_ids")
+    def _compute_count_stock_moves(self):
+        sm_ids = self.mapped("stock_move_ids").ids
+        self.count_stock_moves = len(sm_ids)
+
+    def action_state_to_in_progress(self):
+        active_rec = self.env["stock.inventory"].search([("state", "=", "in_progress")])
+        if active_rec:
```
CasVissers-360ERP, line 159: Should we extend this? Checking if there are counts with same product or same lots or same locations? Counting two locations at the same time should be possible?

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+                    ]
+                )
+        for quant in self.stock_quant_ids:
+            quant.to_do = True
+        return
+
+    def action_state_to_done(self):
+        self.state = "done"
+        for quant in self.stock_quant_ids:
+            quant.to_do = True
```
CasVissers-360ERP: It is still a manual action to confirm the quant count?

CasVissers-360ERP (changes requested):
Functional review

stock_inventory/models/stock_inventory.py
rousseldenis: I would say `Lot/Serial Number` to use Odoo naming

rousseldenis: @DavidJForgeFlow Maybe an 'owner' management too ?

CasVissers-360ERP: I agree with Lot/Serial Number

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+    def action_state_to_in_progress(self):
+        active_rec = self.env["stock.inventory"].search([("state", "=", "in_progress")])
+        if active_rec:
+            raise ValidationError(
+                _("There's already an Adjustment in Process: %s") % active_rec.name
+            )
+        self.state = "in_progress"
+        if self.product_selection == "all":
+            for location in self._origin.location_ids:
+                self.stock_quant_ids = self.env["stock.quant"].search(
```
rousseldenis: I would have put all domains in individual methods in order to allow override

stock_inventory/models/stock_inventory.py
CasVissers-360ERP: Edit: Lot/Serial Numbers

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,193 @@
...
+                        ("location_id", "in", location.child_ids.ids),
+                    ]
+                )
+        for quant in self.stock_quant_ids:
+            quant.to_do = True
+        return
+
+    def action_state_to_done(self):
+        self.state = "done"
+        for quant in self.stock_quant_ids:
```
rousseldenis: Better : 

```suggestion
        self.stock_quant_ids.update({"to_do": True})
```

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: @rousseldenis what do you mean with 'owner'? Add a responsible field?
The other changes are done, thanks!

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: Done thanks!

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: I don't think so because you can only do an Adjustment at a time, so repeated counts are "imposible" (unless something magic happens).

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: Yes, confirming a count is still manual.

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: Done, thanks!

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: Done!

stock_inventory/models/stock_inventory.py
CasVissers-360ERP: Ok, this is very confusing? Should we at least give a warning to the user that the count(s) still needs to be confirmed? Setting a group to done for me means that the count is confirmed.

CasVissers-360ERP: What if a quant is being mutated which is currently in a stock adjustment which is not done yet, should we not block this? Same like not allowing 2 counts? Overall I have a sense the module is confusing now. Why do we add a state to stock_inventory if nothing is being done / blocked or checked? If it is simply a group of quants, then we can create a group without a workflow?

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,200 @@
...
+from odoo.exceptions import ValidationError
+
+
+class InventoryAdjustmentsGroup(models.Model):
+    _name = "stock.inventory"
+    _description = "Inventory Adjustment Group"
+    _order = "date desc, id desc"
+
+    name = fields.Char(
+        required=True, default="Inventory ", string="Inventory Reference"
```
```suggestion
        required=True, default="Inventory", string="Inventory Reference"
```

stock_inventory/models/stock_inventory.py
rousseldenis: This one : https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_quant.py#L76

stock_inventory/models/stock_inventory.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+    @api.depends("stock_move_ids")
+    def _compute_count_stock_moves(self):
+        sm_ids = self.mapped("stock_move_ids").ids
+        self.count_stock_moves = len(sm_ids)
+
+    def _get_all_quants(self, locations):
+        return self.env["stock.quant"].search(
+            [
```
rousseldenis: Maybe you could put domains retrievals in separate methods to easy possible inheritance?

DavidJForgeFlow: All the different type of Inventory Adjustments are computed in different methods, what do you mean with putting the domains retrievals in separate methods?

stock_inventory/models/stock_inventory.py
DavidJForgeFlow: Done, thank you!

AaronHForgeFlow (commented):
Works Good!

Just a minor thing that needs to be fixed:

* Create an inventory adjustment for one product only.
* Start the adjustment
* Change the quantity and click on Apply
* The stock move line is successfully linked
* Go to the stock.inventory again and then to the adjustments, remove the filter "Todo".
* Change the quantity again to a new one and click on apply. 
* Issue: A new quant is successfully created but it is not linked to the inventory adjustment, that is, from the button Stock move lines I can only see one of the stock moves lines

DavidJForgeFlow:
> Works Good!
> 
> Just a minor thing that needs to be fixed:
> 
>     * Create an inventory adjustment for one product only.
> 
>     * Start the adjustment
> 
>     * Change the quantity and click on Apply
> 
>     * The stock move line is successfully linked
> 
>     * Go to the stock.inventory again and then to the adjustments, remove the filter "Todo".
> 
>     * Change the quantity again to a new one and click on apply.
> 
>     * Issue: A new quant is successfully created but it is not linked to the inventory adjustment, that is, from the button Stock move lines I can only see one of the stock moves lines

Thank you Aaron, this is already fixed!

The code is now better to inherit because of some change of how domains work, thanks to @GuillemCForgeFlow and @rousseldenis for the idea. 

Also, added the possiblilty of doing two Adjustments on two different Locations at the same time.

GuillemCForgeFlow:
Thanks for the work @DavidJForgeFlow :+1: !

AaronHForgeFlow (approved):
Functional review LGTM!

CasVissers-360ERP:
@AaronHForgeFlow If you did a functional approval I am really curious about feedback on my questions. For me this module currently makes no sense. It is just a group of quants and could better be called stock_quant_group.

AaronHForgeFlow:
Hi @CasVissers-360ERP  It is called stock_inventory in order to respect what it was in standard in previous versions, otherwise the migration will get more complex.

I guess the main discussion regards this comment:

> What if a quant is being mutated which is currently in a stock adjustment which is not done yet, should we not block this? Same like not allowing 2 counts? Overall I have a sense the module is confusing now. Why do we add a state to stock_inventory if nothing is being done / blocked or checked? If it is simply a group of quants, then we can create a group without a workflow?

I think that if the quant is altered while the inventory adjustment is done then there is no inconsistencies in the database in this case, because Odoo will internally handle this. It may be confusing for users that quants did a change while they were doing the adjustments.  In that case maybe it is necessary to avoid the modification of the quant until the inventory adjustment is completed. But that can be done on a separate module I think. I think there was also a module for that in this repo.

CasVissers-360ERP:
@AaronHForgeFlow 

You only answered one of my concerns? My main concern: Why do we add a state to stock_inventory if nothing is being done / blocked or checked? In addition it is not an inventory / inventory count. An inventory count start which blocking the location, counting the stock, and adjusting the stock level accordingly. I miss the whole idea of creating a count in Odoo (group of counts). But if there is no logic in the group/inventory I feel like it will be very confusing for the end-user. I have a hard time understanding the benefits as well. 

If the goal of this module is to simply create a group of quants, just remove the state and rename the module.

AaronHForgeFlow:
Hi @CasVissers-360ERP ,  I am just saying that the module works as my customer needs. They just need a document that records and keep a history of the changes on inventory. They are not worried about that restriction because the lifecycle of an inventory adjustment is too short for them and that kind of "issues" are not relevant. Nothing is blocked the quant in this module but it can be blocked on a module that depends on this one. Perhaps that feature can be included here, I am not against that.

CasVissers-360ERP:
@AaronHForgeFlow Well maybe the module is not generic/mature enough to be on OCA module?

AaronHForgeFlow:
@CasVissers-360ERP I think it is very generic module, it recovers the standard inventory adjustments feature form previous versions and that is proved to be generic enough. The only thing that may be missing is the locking on the quant that is being inspected. It is ok to me to implement that restriction on this module, but I will not block this PR because of that. Would you be fine with this PR if it implements that restriction?

Those discussions are good for the quality of the module so I am glad you commented and I hope more people joins to the discussion here.

ventor-dev (approved):
Great work! Customers still need this functionality to be able to group inventories for individual products into single big document. That is much better for Warehouse manager to review it later
 Not sure I understand why Odoo removed it. 

@DavidJForgeFlow thank you for the work! Hope it will be merged soon into OCA. Approved from my side

CasVissers-360ERP:
@AaronHForgeFlow I don't agree with this " it recovers the standard inventory adjustments "

A big thing missing is:
https://github.com/odoo/odoo/blob/e9a7283ea33a5e122afa53f73f717ed45ce9874d/addons/stock/models/stock_inventory.py#L131

But since everybody else seems to approve this PR I think it's best for me to exit the discussion.

LoisRForgeFlow:
Obviously this can receive improvements and changes, but since I think it is a common requirement for customers coming from previous Odoo versions, let's at least cover the gap with this (I guess better this than nothing).

Thanks to all participants! Do not hesitate to open PRs to include improvements and changes once this gets merged.

/ocabot merge nobump

PR #1511, opened 2022-10-21
[14.0][IMP] stock_request + stock_request_submit: Show error in stock.request.order in the "Confirm" or "Submit" buttons if there are no items.

victoralmau:
Changes done:
- [x] Show error in `stock.request.order` in the "_Confirm_" or "_Submit_" buttons if there are no items.
- [x] Show Confirm button from `stock.request.order` only to Stock Request Manager users.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT39761

PR #1514, opened 2022-10-24
[15.0][IMP] stock_request: Show error in stock.request.order in the "Confirm" or "Submit" buttons if there are no items.

victoralmau:
FWP from 14.0: https://github.com/OCA/stock-logistics-warehouse/pull/1511

Changes done:
- [x] Show error in `stock.request.order` in the "_Confirm_" or "_Submit_" buttons if there are no items.

@Tecnativa TT39761

### DIFF
--- a/stock_request/models/res_company.py
+++ b/stock_request/models/res_company.py
@@ -12,3 +12,11 @@ class ResCompany(models.Model):
     stock_request_allow_virtual_loc = fields.Boolean(
         string="Allow Virtual locations on Stock Requests"
     )
+    stock_request_confirm_user_permission = fields.Selection(
+        selection=[
+            ("user", "User"),
+            ("manager", "Manager"),
+        ],
+        default="user",
+        string="Confirm type button in Stock Requests",
+    )
--- a/stock_request/views/stock_request_order_views.xml
+++ b/stock_request/views/stock_request_order_views.xml
@@ -22,7 +22,15 @@
                         name="action_confirm"
                         string="Confirm"
                         type="object"
-                        attrs="{'invisible': [('state', 'not in', ['draft'])]}"
+                        attrs="{'invisible': ['|', ('state', 'not in', ['draft']), ('stock_request_confirm_user_permission', '!=', 'user')]}"
+                        groups="stock_request.group_stock_request_user"
+                    />
+                    <button
+                        name="action_confirm"
+                        string="Confirm"
+                        type="object"
+                        attrs="{'invisible': ['|', ('state', 'not in', ['draft']), ('stock_request_confirm_user_permission', '!=', 'manager')]}"
+                        groups="stock_request.group_stock_request_manager"
                     />
                     <button
                         name="action_cancel"
@@ -36,6 +44,7 @@
                         type="object"
                         string="Set to Draft"
                     />
+                    <field name="stock_request_confirm_user_permission" invisible="1" />
                     <field name="state" widget="statusbar" />
                 </header>
                 <sheet>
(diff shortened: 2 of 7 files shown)


stock_request/views/stock_request_order_views.xml
```diff
@@ -23,6 +23,7 @@
                         string="Confirm"
                         type="object"
                         attrs="{'invisible': [('state', 'not in', ['draft'])]}"
+                        groups="stock_request.group_stock_request_manager"
```
rousseldenis, line 33: Why changing this behaviour ?

rousseldenis: @victoralmau Stock request users should have access to confirmations. If you want to restrict, please either do it on your side or add an option to configure this behaviour.

victoralmau: Perhaps another possible approach would be to add a setting to define who can use confirm button (user or manager). What do you think about this?

rousseldenis: Yes and protect the action method through that

stock_request/models/res_company.py
```diff
@@ -12,3 +12,11 @@ class ResCompany(models.Model):
     stock_request_allow_virtual_loc = fields.Boolean(
         string="Allow Virtual locations on Stock Requests"
     )
+    stock_request_confirm_user_permission = fields.Selection(
+        selection=[
+            ("user", "User"),
+            ("manager", "Manager"),
+        ],
+        default="user",
```
pedrobaeza, line 20:
```suggestion
        default="manager",
```

rousseldenis: Nope as this changes the base behaviour from the beginning of this module. As already said, you can override that in your project.

pedrobaeza: Yes, but that's the intended behavior. It's a nonsense to have 2 steps (submit and confirm), when both profiles can make by default both. We'll bump major version.

rousseldenis:
@JordiBForgeFlow @kittiu I want your advice here

stock_request/models/res_company.py
rousseldenis: > We'll bump major version.

Yes, but it will be applied on further version. I think about migrations

pedrobaeza: We can put a migration script for those coming from previous versions that preserve the other, but I really think it's a mistake to let it as it was before. If Jordi/Kitti thinks the same, we can even remove this mechanism and let the new default only.

rousseldenis: @pedrobaeza I've reviewed 14.0 module and I don't find any submit state.

There are only 'draft' one (default) and open one. Could you review that ?

pedrobaeza: OK, I see the problem. The new module `stock_request_submit` is the one to fix:

- It should change the button "Confirm" to launch the method to put the request into submit.
- The button submit should call the method `confirm`.
- This button is the one to be protected by manager permission.

victoralmau:
I close this PR because finally it is not necessary to add any changes to v15 (changes are added in `stock_request_submit`).

pedrobaeza:
But what about the message when 0 stock request lines?

victoralmau:
> But what about the message when 0 stock request lines?

Ok sorry, can you reopen it?

pedrobaeza:
It's nos possible anymore it seems. Do another one, and the fw-port of the submit thing is also pending, no?

PR #1516, opened 2022-10-24
[13.0][MIG] stock_inventory_verification_request

dnplkndll:


rousseldenis:
@dnplkndll Have you ever looked at stock_cycle_count module as it seems there is an overlap here ?

https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_cycle_count

dnplkndll:
> @dnplkndll Have you ever looked at stock_cycle_count module as it seems there is an overlap here ?
> 
> https://github.com/OCA/stock-logistics-warehouse/tree/13.0/stock_cycle_count

I was actually migrating this due to the reference in the stock_cycle_count readme

PR #1520, opened 2022-10-26
[FIX] stock_request: add sudo to allow stock ops without stock request access

adrienpeiffer:
For the moment, if we confirm a stock operation without access to related stock request, we got an error.


pedrobaeza (approved):

ps-tubtim (approved):
LGTM

rousseldenis (approved):

PR #1521, opened 2022-10-27
[11.0][FIX] stock_move_location: _get_available_quantity has to return a tuple

wildi1:
**Problem**
_get_available_quantity() returns non tuple in some cases

**How to reproduce**
1. Run Move To Location
2. Add Product, wich is not on stock
3. Click Confirm
4. Error "int is not iterable" occurs

Was already fixed in 13.0, but not in  11.0, 12.0, 14.0, 15.0

### DIFF
--- a/stock_move_location/wizard/stock_move_location_line.py
+++ b/stock_move_location/wizard/stock_move_location_line.py
@@ -122,7 +122,7 @@ class StockMoveLocationWizardLine(models.TransientModel):
         more than exists."""
         self.ensure_one()
         if not self.product_id:
-            return 0
+            return 0, 0
         if self.env.context.get("planned"):
             # for planned transfer we don't care about the amounts at all
             return self.move_quantity, 0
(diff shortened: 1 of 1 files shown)


mathiasfrancke:
looks okay for me but the some CI check failed.

OSevangelist:
@wildi1 can you check and fix the CI stuff ?

rousseldenis:
@wildi1 @OSevangelist @mathiasfrancke I will look at open PR's in 11.0 if we need to switch to github actions (this will solve tests).

OSevangelist:
@rousseldenis super! As soon as this reaches green state i think we would be able to merge it. Cheers Frederik

rousseldenis:
@wildi1 Could you rebase ?

wildi1:
@rousseldenis can it be merged now? A test had not exist before, too.

rousseldenis:
> @rousseldenis can it be merged now? A test had not exist before, too.

Even if the fix is trivial, could you do a test to avoid regression?

rousseldenis:
@wildi1

wildi1:
Hi @rousseldenis,
I checked to make a unit test. I would have to build two complete new wizard scenarios.
In my mind it is too much efford, because the fix is obvious. 
This method is just called at one position and there, two return parameters are required.

For 100% covarage you also have to remove
```
if not self.product_id:
     return 0, 0
```
in Line 124,125 because product_id is required. 

Please consider my point and think about, to merge it without too much effort.

PR #1535, opened 2022-11-04
[14.0][IMP] stock_cycle_count: Implement constrains in inventory adjustment.

JordiBForgeFlow:
The previous method was too restrictive, and would cause conflicts in cases where another module would introduce changes to the inventory adjustment that were not really affecting to the cycle count.

We implement the minimum necessary constrains to ensure consistency between the cycle count and the related inventory adjusment.

cc @DavidJForgeFlow  @LoisRForgeFlow


DavidJForgeFlow (approved):

LoisRForgeFlow (approved):
Absolutely :+1:

PR #1541, opened 2022-11-09
[MIG][16.0] scrap_reason_code

PierrickBrun:


scrap_reason_code/models/stock_scrap.py
```diff
@@ -37,7 +37,8 @@ def write(self, vals):
...
 
-    @api.model
-    def create(self, vals):
-        self._update_scrap_reason_code_location(vals)
-        return super(StockScrap, self).create(vals)
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            self._update_scrap_reason_code_location(vals)
+        return super().create(vals_list)
```
dreispt, line 44: Is this really needed? @api.model works fine AFAIK.

dreispt (approved):

scrap_reason_code/models/stock_scrap.py
PierrickBrun: You're right, I thought this was an OCA recommendation to change it for @api.model_create_multi but it is not the case.

I guess we can keep it like this since this is equivalent

rousseldenis: @PierrickBrun @dreispt This is in fact better as if multi recordset is used, the create would be more performant (as each value would have been modified before in batch.

rousseldenis (approved):

PR #1543, opened 2022-11-10
[14.0] ADD stock_multi_warehouse_security

petrus-v:
PR as been re-open here: https://github.com/OCA/stock-logistics-warehouse/pull/1927
---

With this module you are able to set a restricted list
of allowed warehouses that user can see and operate with.

This module is inspired from the experiences of
`stock_multi_warehouse_security <https://github.com/akretion/stock-logistics-warehouse/tree/12-muli-wh-security/stock_multi_warehouse_security/>`_
on version 12.0 but has some key differences on user experience:

* In this module there is no new groups, user is able to see allowed warehouses
  only or all if not set.
* So in this module there is no "current warehouse" concept on user (
  in v12.0 that module was based on `base_multi_warehouse
  <https://github.com/akretion/stock-logistics-warehouse/tree/12-base-multi_warehouse/base_multi_warehouse>`_
  which allowed users to switch between warehouses).

As request by review, I've split this initial work in 2 modules:
* #1558 which add warehouse_id in various models
* This PR is adding stock_multi_warehouse_security (and based on #1558)


petrus-v:
@florian-dacosta I would be happy to get some review from you and also your agreement regarding the usage of your commit as beginning of this module. Also regarding the move of the license !

stock_multi_warehouse_security/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "Akretion, Pierre Verkest, Odoo Community Association (OCA)",
+    "maintainers": ["petrus-v"],
+    "license": "AGPL-3",
+    "installable": True,
+    "summary": "Restrict user access in multi-warehouse environment",
+    "depends": [
+        "stock",
+        "sale_stock",
```
florian-dacosta: Why the dependency on `sale` is needed ?

stock_multi_warehouse_security/models/stock_location.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class StockLocation(models.Model):
+    _inherit = "stock.location"
+
+    warehouse_id = fields.Many2one(
+        "stock.warehouse", compute="_compute_warehouse", store=True, index=True
+    )
+
+    @api.depends("location_id")
```
florian-dacosta: It should probably depend on `location_id.warehouse_id` in case a grand parent location is set afterward or change parent itself, etc

stock_multi_warehouse_security/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+        else:
+            locations = self.env["stock.location"].browse(
+                [vals["location_id"], vals["location_dest_id"]]
+            )
+            whs = locations.mapped("warehouse_id")
+            wh = len(whs) <= 1 and whs or self.env["stock.warehouse"]
+        return wh
+
+    @api.model
+    def create(self, vals):
```
florian-dacosta: Would be nice to transform it in `@api.model_create_multi` I guess

stock_multi_warehouse_security/models/stock_quant.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    warehouse_id = fields.Many2one(
+        "stock.warehouse", compute="_compute_warehouse_id", store=True, index=True
+    )
+
+    @api.depends("location_id")
```
florian-dacosta: I guess I would depend on `location_id.warehouse_id` here also.
Also, I am not sure why not to have a related field here, it is possible the comment is obsolete

stock_multi_warehouse_security/security/stock_security.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" ?>
+<odoo>
```
florian-dacosta: Deserves a noupdate="1"

florian-dacosta (commented):
Hello @petrus-v 
Thanks for the module proposition. Please keep in mind the module was quicky port to v12 but never used, the one on v8 is currently used on production
No problem for the change of license or use of commit of course.

Some remarks : 
I believe this module should be splitted, mainly the new `warehouse_id` fields could be useful and practical without having the security rules. And there is already a module adding the field on location for instance : `stock_location_warehouse`.
I think we can have a second module that add the field on all models but res.users
And then this module would add the field on res.users + the security rules, what do you think ?

Also, the goal with the current warehouse and the security rules depending on it was kind of have the same behaviour of the multi company (in v12 and less). As a target, I wonder if we should try to have something similar with v13-16 behaviour (beeing able to choose multiple warehouses among the allowed one in a list)
I guess we can have it this way and re-think it the day we have a module that implement this warehouse selection.

stock_multi_warehouse_security/__manifest__.py
petrus-v: Because I wanted to test, even user is not allowed to access to a given warehouse data, if employee is a saler he should be able to set a forbidden warehouse data access as default warehouse used while selling products cf https://github.com/OCA/stock-logistics-warehouse/pull/1543/files#diff-2fa3811d5404bb7e50aeb963eb796aeabafd818fb3f029d53c37468fbba3dae7R9. Probably I should only make sure read access on warehouse in my test and remove the dependency :thinking:

petrus-v: I've remove the dependency an skip test if module is not installed

stock_multi_warehouse_security/models/stock_location.py
petrus-v: This module is based on stock_location_warehouse now !

stock_multi_warehouse_security/models/stock_move.py
petrus-v: not sure to see the benefit here, if developer create multiple moves, according the code of  _get_move_default_warehouse each moves could concern different warehouse based on provided fields picking_type_id, location_id and location_dest_id. Do you want me to create a cache on those possible keys and reuse it to avoid few requests ?

We should probably carry on this thread on #1558

stock_multi_warehouse_security/security/stock_security.xml
petrus-v: done

stock_multi_warehouse_security/models/stock_quant.py
petrus-v: I've moved to a related field then !

damdam-s (approved):

petrus-v:
PR as been re-open here: https://github.com/OCA/stock-logistics-warehouse/pull/1927

PR #1547, opened 2022-11-11
[14.0][IMP] refactor product_route_profile

Kev-Roche:
Refactor product_route_profile to be easier to work with other modules. 
Also hide route profiles if no route is present on the product.

### DIFF
--- a/product_route_profile/models/product_template.py
+++ b/product_route_profile/models/product_template.py
@@ -25,16 +25,20 @@ class ProductTemplate(models.Model):
         store=False,
     )
 
+    def _get_routes(self):
+        self.ensure_one()
+        if self.force_route_profile_id:
+            return self.force_route_profile_id.route_ids
+        elif self.route_profile_id:
+            return self.route_profile_id.route_ids
+        else:
+            return self.env["stock.location.route"]
+
     @api.depends("route_profile_id", "force_route_profile_id")
     @api.depends_context("company")
     def _compute_route_ids(self):
         for rec in self:
-            if rec.force_route_profile_id:
-                rec.route_ids = [(6, 0, rec.force_route_profile_id.route_ids.ids)]
-            elif rec.route_profile_id:
-                rec.route_ids = [(6, 0, rec.route_profile_id.route_ids.ids)]
-            else:
-                rec.route_ids = False
+            rec.route_ids = rec._get_routes()
 
     def _search_route_ids(self, operator, value):
         return [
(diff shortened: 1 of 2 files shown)


product_route_profile/models/product_template.py
```diff
@@ -45,19 +49,23 @@ def _search_route_ids(self, operator, value):
             ("route_profile_id.route_ids", operator, value),
         ]
 
+    def _get_or_create_profile(self, profiles, routes):
+        self.ensure_one()
+        if not routes:
+            return self.env["route.profile"]
```
rousseldenis, line 55: Better : 
```suggestion
            return self.env["route.profile"].browse()
```

product_route_profile/models/product_template.py
```diff
@@ -25,16 +25,20 @@ class ProductTemplate(models.Model):
...
     )
 
+    def _get_routes(self):
+        self.ensure_one()
+        if self.force_route_profile_id:
+            return self.force_route_profile_id.route_ids
+        elif self.route_profile_id:
+            return self.route_profile_id.route_ids
+        else:
+            return self.env["stock.location.route"]
```
rousseldenis, line 35: Better :

```suggestion
            return self.env["stock.location.route"].browse()
```

product_route_profile/models/product_template.py
```diff
@@ -45,19 +49,23 @@ def _search_route_ids(self, operator, value):
...
+        if not routes:
+            return self.env["route.profile"]
+        for profile in profiles:
+            if routes == profile.route_ids:
+                return profile
+        vals = self._prepare_profile(routes)
+        return self.env["route.profile"].create(vals)
+
     def _inverse_route_ids(self):
         profiles = self.env["route.profile"].search([])
```
rousseldenis, line 63: This makes me think: to limit the search, this could be great:

```suggestion
        profiles = self.env["route.profile"].search([("route_ids", "in", self.mapped("route_ids").ids])
```

But never tested such query

rousseldenis (approved):
@Kev-Roche Minor changes

CasVissers-360ERP:
@Kev-Roche 
`            <xpath expr="//field[@name='route_ids']/parent::div" position="attributes">
                <attribute name="attrs">{'invisible': True}</attribute>
            </xpath>`

Could we update this to: 

`            <xpath expr="//field[@name='route_ids']" position="attributes">
                <attribute name="attrs">{'readonly': True}</attribute>
            </xpath>`

Hiding the field also hides the cool widget Odoo has to show the route in a visual way. 

EDIT: in addition it's nice to see what a profile changes. The same like product_profile work for example

bosd:
@Kev-Roche Kind reminder to attend to the comments/reviews

PR #1548, opened 2022-11-11
[14.0][ADD] product_route_profile_internal_resupply

Kev-Roche:
This module allows to add Internal Supplies Routes separately to other routes in route_profile.
depends of refactor here : https://github.com/OCA/stock-logistics-warehouse/pull/1547


bosd:
@Kev-Roche Can you re-trigger a build?

PR #1549, opened 2022-11-14
[16.0][MIG] stock_move_location

tupaq:


rousseldenis:
@tupaq Thanks for this. Could you fix tests ?

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,332 @@
...
+    )
+    picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type", default=_get_default_picking_type_id
+    )
+    picking_id = fields.Many2one(
+        string="Connected Picking", comodel_name="stock.picking"
+    )
+    edit_locations = fields.Boolean(default=True)
+    apply_putaway_strategy = fields.Boolean()
+
```
AungKoKoLin1997, line 64:
```suggestion
`clear_move_lines = fields.Boolean()
@api.onchange("clear_move_lines")
    def clear_move_line(self):
        if self.clear_move_lines == True:
            self._clear_lines()
`
```

AungKoKoLin1997: If the user want to move only some products from source to destination and there is a lot of products in source location, user need to remove products one by one. So, I think we should add the option for removing all line. That can save time.

stock_move_location/wizard/stock_move_location.py
```diff
@@ -0,0 +1,332 @@
...
+                if line_val.get("max_quantity") <= 0:
+                    continue
+                line = line_model.create(line_val)
+                line.max_quantity = line.get_max_quantity()
+                line.reserved_quantity = line.reserved_quantity
+                lines.append(line)
+            self.update(
+                {"stock_move_location_line_ids": [(6, 0, [line.id for line in lines])]}
+            )
+
```
AungKoKoLin1997, line 329:
```suggestion
          return {'domain': {'stock_move_location_line_ids': [('id', 'in', [line.id for line in lines])]}}
```

AungKoKoLin1997: I think we should add domain in many2many. If not, when we click add line, all line will be displayed. If user chooses the line that location is different with parent original location id, that will cause error.

stock_move_location/wizard/stock_move_location.xml
```diff
@@ -0,0 +1,154 @@
...
+                        <field name="origin_location_disable" invisible="True" />
+                        <field
+                            name="origin_location_id"
+                            attrs="{'readonly': [('origin_location_disable', '=', True)]}"
+                        />
+                        <field name="destination_location_disable" invisible="True" />
+                        <field
+                            name="destination_location_id"
+                            attrs="{'readonly': [('destination_location_disable', '=', True)]}"
+                        />
```
AungKoKoLin1997, line 38:
```suggestion
                        />
                        <field name="clear_move_lines" widget="boolean_toggle" />
```

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -0,0 +1,136 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools import float_compare
+
+
+class StockMoveLocationWizardLine(models.TransientModel):
+    _name = "wiz.stock.move.location.line"
+    _description = "Wizard move location line"
+
```
AungKoKoLin1997, line 13:
```suggestion
   _rec_name = "product_id
```

AungKoKoLin1997: This will help for choosing line when user chooses the product they want. Currently showing with ID.
Screenshot from 2022-12-01 12-29-04

AungKoKoLin1997 (commented):
Functional Review: LGTM
There are some points for enhancement but we can work after this PR is merged.

stock_move_location/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Move Stock Location",
-    "version": "15.0.1.0.2",
+    "version": "16.0.1.0.2",
```
rousseldenis: You need to reset to `16.0.1.0.0`

rousseldenis (approved):
Code review. Minor comment

PR #1550, opened 2022-11-14
[14.0][IMP] stock_quant_manual_assign: reserved field as readonly

mariadforgeflow:
This field shouldn't be editable since it doesn't change anything when editing it in the wizard and it's confusing


LoisRForgeFlow (approved):
:+1:

pedrobaeza (approved):

PR #1552, opened 2022-11-15
[14.0][FIX] stock_reserve_rule: fix _apply_strategy_empty_bin

mt-software-de:
Empty bin Rule
Now checks if there is an other product which is allocating the location

### DIFF
--- a/stock_reserve_rule/models/stock_reserve_rule.py
+++ b/stock_reserve_rule/models/stock_reserve_rule.py
@@ -219,11 +219,26 @@ class StockReserveRuleRemoval(models.Model):
         # Group by location (in this removal strategies, we want to consider
         # the total quantity held in a location).
         quants_per_bin = quants._group_by_location()
-
         # We take goods only if we empty the bin.
         # The original ordering (fefo, fifo, ...) must be kept.
-        rounding = fields.first(quants).product_id.uom_id.rounding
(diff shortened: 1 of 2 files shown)


mt-software-de:
@jbaudoux could please have a look?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -222,8 +222,18 @@ def _apply_strategy_empty_bin(self, quants):
 
         # We take goods only if we empty the bin.
         # The original ordering (fefo, fifo, ...) must be kept.
-        rounding = fields.first(quants).product_id.uom_id.rounding
+        product = fields.first(quants).product_id
+        rounding = product.uom_id.rounding
         for location, location_quants in quants_per_bin:
+            if self.env["stock.quant"].search(
```
jbaudoux: Can't we lower the amount of queries ?

jbaudoux:
cc @sebalix

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: use read_group instead outside of for loop

jbaudoux (approved):

sebalix (approved):
LG

PR #1563, opened 2022-11-29
[ADD] product_stock_coverage

robinkeunen:
Compute estimated stock coverage based on POS sales over a date range.

This module adds these fields on product template :

- Computation range in days
- Sales over computation range
- Average daily sales over computation range
- Stock coverage in days
- Effective sale price over computation range

The default date range is 14 days, it can be changed on each article.

---

Caveats: 

- This module computes coverage based on `pos.order` and not on `sale.order`
- This module also computes the effective sale price

For these reasons, I'm not sure it has a place in the OCA. Eager for feedbacks.

---

This module is similar to module [stock_demande_estimate](https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_demand_estimate) but rather computes the coverage on each product.


rousseldenis:
@robinkeunen Have you ever look at this :

https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_demand_estimate

I think you could use your sales estimation in order to fill in stock demand estimates.

@LoisRForgeFlow @JordiBForgeFlow I don't remember if it is used by sale flow apart ddmrp one

robinkeunen:
Hi @rousseldenis and thanks for the reply.

From what I understand, `stock_demand_estimate` does not take sale into account. It is computing `daily_sales` based on data you enter on the model. Are you suggesting to add the computations I'm doing here to `stock.demand.estimate` ?

I would then need to create a demand estimate for each product I want to track. Our customer wants those fields available on all products.

rousseldenis:
In fact, as I think we should avoid modules that do both quite the same thing:

Yes, IMHO, we should fill demand estimates, then, expand that module in order to display values on products.

That's my feeling seeing both modules functionalities.

As I see daily average usage computation is also done there:

https://github.com/OCA/ddmrp/blob/13.0/ddmrp/models/stock_buffer.py#L880

And there is already a method to compute past demands quantites by date range:

https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_demand_estimate/models/stock_demand_estimate.py#L151


My thoughts at this time.

If @LoisRForgeFlow or @JordiBForgeFlow can help to avoid having duplicate modules

rousseldenis:
And I would have extracted the adu calculation model from ddmrp in order to use it in other flows

robinkeunen:
I agree with you, this is a bit too close from existing code. Unfortunately, it needs a bit too much refactoring at this time. I'll put this back into our repo and add your comments to the roadmap.rst.

Thanks for your time and advice 👍

PR #1564, opened 2022-12-01
[15.0][IMP] stock_archive_constraint: Tests more resilient when other modules have been installed

sergio-teruel:
cc @Tecnativa 

ping @pilarvargas-tecnativa @stefan-tecnativa @chienandalu

### DIFF
--- a/stock_archive_constraint/models/product_product.py
+++ b/stock_archive_constraint/models/product_product.py
@@ -3,13 +3,21 @@
 
 from odoo import _, api, models
 from odoo.exceptions import ValidationError
+from odoo.tools import config
 
 
 class ProductProduct(models.Model):
     _inherit = "product.product"
 
+    def _skip_check_archive_constraint_condition(self):
+        return config["test_enable"] and not self.env.context.get(
+            "test_stock_archive_constraint"
+        )
+
     @api.constrains("active")
     def _check_active_stock_archive_constraint_stock_quant(self):
+        if self._skip_check_archive_constraint_condition():
+            return
         res = self.env["stock.quant"].search(
             [
                 ("location_id.usage", "in", ("internal", "transit")),
@@ -29,6 +37,8 @@ class ProductProduct(models.Model):
 
     @api.constrains("active")
     def _check_active_stock_archive_constraint_stock_move(self):
+        if self._skip_check_archive_constraint_condition():
+            return
         res = self.env["stock.move"].search(
             [
                 ("product_id", "in", self.filtered(lambda x: not x.active).ids),
@@ -47,6 +57,8 @@ class ProductProduct(models.Model):
 
     @api.constrains("active")
     def _check_active_stock_archive_constraint_stock_move_line(self):
+        if self._skip_check_archive_constraint_condition():
+            return
         res = self.env["stock.move.line"].search(
             [
                 ("product_id", "in", self.filtered(lambda x: not x.active).ids),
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

stock_archive_constraint/tests/test_location_archive_constraint.py
```diff
@@ -9,6 +9,9 @@ class TestLocationArchiveConstraint(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env = cls.env(
+            context=dict(cls.env.context, test_stock_archive_constraint=True)
```
chienandalu, line 13: <del>Wouldn't this go in the conflicting module test?</del>

sergio-teruel: This one is the module

pedrobaeza: No, it's for testing this functionality. The absence of this context is what makes that the constraints are not raised for the tests of the other modules.

chienandalu: Nevermind. I got it already

chienandalu (approved):

carlosdauden (approved):

stock_archive_constraint/models/product_product.py
```diff
@@ -3,13 +3,21 @@
...
 from odoo import _, api, models
 from odoo.exceptions import ValidationError
+from odoo.tools import config
 
 
 class ProductProduct(models.Model):
     _inherit = "product.product"
 
+    def _skip_check_archive_constraint_condition(self):
+        return config["test_enable"] and not self.env.context.get(
```
rousseldenis, line 13: @sergio-teruel 

As I suppose you do post-install tests, this could have been avoided through a configuration parameter that says 'Avoid product archiving when...'.

Such code should be avoided as many as possible

pedrobaeza: Adding a configuration that enables the only goal of the module is also not something optimal. Users expect to install the module to have the only feature it contains without doing anything more, and the problem is only on integration test. This one is a good compromise.

rousseldenis: I understand as this is a debate I already have many times.

I see two cons:

- Having in code something that depends on tests is not understandable from a pure programmatic point of view
- Activating or deactivating feature through interface is tedious for people that configures Odoo, but has the advantage of not depending on module state (and on your deployment times).

pedrobaeza: A comment can be added to clarify its usage, although the variable names are very straight-forward, but this is also a compromise solution for reducing the maintenance burden. Having the "interface" option means:

- Add such field in the proper place (company, res.config.settings, system parameter...).
- Add a view that works properly for doing the configuration (res.config.settings is specially ugly).
- Handle translations.
- etc.

It's too much trade-off for a module that does only this feature.

rousseldenis: > * Add such field in the proper place (company, res.config.settings, system parameter...).
> * Add a view that works properly for doing the configuration (res.config.settings is specially ugly).
> * Handle translations.

Isn't it our normal job ???

pedrobaeza: Not for something that is only a side effect.

rousseldenis: > Not for something that is only a side effect.

That's your point of view. As you are changing the default behaviour by adding a constraint, mine is it's a good habit to have a configuration for that.

Trying it, it's adopting it :smiley:

PR #1576, opened 2022-12-08
[13.0][FIX] Update from copier template

sebalix:


sebalix:
Fast tracking this one to fix the CI

PR #1577, opened 2022-12-08
[14.0][FIX] Stock Move Location: apply putaway rules

joshkreud:
The function called to get the putaway rule when creating the stock.move.lines does not exist in odoo 14.0.
https://github.com/OCA/stock-logistics-warehouse/blob/5830833b99df972329e173633dbec27ca002e05f/stock_move_location/wizard/stock_move_location_line.py#L106

- `get_putaway_strategy` should actuall be [_get_putaway_strategy](https://github.com/odoo/odoo//blob/076baeacbbdf92ae1b500e6716fd95d32acc4a53/addons/stock/models/stock_location.py#L135)
- Extend test to actually commit the wizard and check the resulting move.lines

P.s. This is my first PR to the OCA, so please let me know if I missed something.

### DIFF
--- a/stock_move_location/tests/test_common.py
+++ b/stock_move_location/tests/test_common.py
@@ -120,10 +120,21 @@ class TestsCommon(common.SavepointCase):
     def set_product_amount(
         self, product, location, amount, lot_id=None, package_id=None, owner_id=None
     ):
+        """Set available stock Quantity to 'amount'"""
+        current_qty = self.env["stock.quant"]._get_available_quantity(
+            product,
+            location,
+            lot_id=lot_id,
+            package_id=package_id,
+            owner_id=owner_id,
+        )
+        # Since _update_available_quantity decreases or increases,
+        # we need to first get the current amount.
+        change_amount = amount - current_qty
         self.env["stock.quant"]._update_available_quantity(
             product,
             location,
-            amount,
+            change_amount,
             lot_id=lot_id,
             package_id=package_id,
             owner_id=owner_id,
--- a/stock_move_location/tests/test_move_location.py
+++ b/stock_move_location/tests/test_move_location.py
@@ -156,6 +156,14 @@ class TestMoveLocation(TestsCommon):
             putaway_line.destination_location_id, self.internal_loc_2_shelf
         )
 
+        # Actually commit the wizard and check stock.move.line.location_dest_id
+        ret = wizard.action_move_location()
+        picking = self.env["stock.picking"].browse([ret["res_id"]])
+        putaway_move_line = picking.move_line_ids.filtered(
+            lambda l: l.product_id == self.product_no_lots
+        )[0]
+        self.assertEqual(putaway_move_line.location_dest_id, self.internal_loc_2_shelf)
+
     def test_delivery_order_assignation_after_transfer(self):
         """
         Make sure using the wizard doesn't break assignation on delivery orders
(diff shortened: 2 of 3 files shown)


etobella (approved):
Thanks @joshkreud technically it looks good

rousseldenis (approved):

rousseldenis:
@joshkreud Do you have the ability to add tests for putaway apply ?

joshkreud:
@rousseldenis I'm sorry but what do you mean by that? 

I extended the `test_wizard_with_putaway_strategy` test to commit the picking and check the `stock.move.line` destination location. 
https://github.com/OCA/stock-logistics-warehouse/blob/e9fb7bfe09af990d4f9d32f9ced86c103d1a9f54/stock_move_location/tests/test_move_location.py#L159-L165

rousseldenis:
> @rousseldenis I'm sorry but what do you mean by that?
> 
> I extended the `test_wizard_with_putaway_strategy` test to commit the picking and check the `stock.move.line` destination location.
> 
> https://github.com/OCA/stock-logistics-warehouse/blob/e9fb7bfe09af990d4f9d32f9ced86c103d1a9f54/stock_move_location/tests/test_move_location.py#L159-L165

Sorry, tired :sweat_smile:

joshkreud:
Heya, 

I discovered another bug a couple of lines down. 
This function should actually always return a tuple. 

Since I needed to set the product quantity to 0 for one test, I noticed that `_set_product_amount` actually didn't _set_ the amount but _increased/decreased_. 

Since it affects the same function, this PR seems the right place.

rousseldenis:
> Heya,
> 
> I discovered another bug a couple of lines down. This function should actually always return a tuple.
> 
> Since I needed to set the product quantity to 0 for one test, I noticed that `_set_product_amount` actually didn't _set_ the amount but _increased/decreased_.
> 
> Since it affects the same function, this PR seems the right place.

So, do you need to fix something more here ?

joshkreud:
@rousseldenis Nope all done now!

stock_move_location/tests/test_common.py
```diff
@@ -118,12 +118,20 @@ def setup_product_amounts(self):
         )
 
     def set_product_amount(
-        self, product, location, amount, lot_id=None, package_id=None, owner_id=None
+        self, product, location, target_qty, lot_id=None, package_id=None, owner_id=None
```
rousseldenis, line 121: @joshkreud Why changing function signature from the original ?

rousseldenis: @joshkreud In fact, I wouldn't have changed the signature here as some would have inherited from it in further modules. Can you change ?

joshkreud: Yeah I actually thought the same, but since the function previously did not and now does what you would expect from it's name, I thought it would be good to provoke an error in downstream tests. 
But I reset the signature and added a docstring clarifying what the function does now.

rousseldenis (approved):

joshkreud:
Hey, 
I just saw that pre-commit failed, because one comment line is to long. 
Should I fix that or ignore?

rousseldenis:
> Hey, I just saw that pre-commit failed, because one comment line is to long. Should I fix that or ignore?

Nope, you have to fix it.

I advise you to install pre-commit locally. It will fix all those stuffs when you are committing (or at least warn you)

joshkreud:
Alrighty, Done!

Apparently my linters are based on the OCA checks, but with 120 line width. 
88 seems super narrow but alright `pre-commit install` to the rescue ;) 

Now I just need an easy way to execute codecov locally.

rousseldenis:
> Now I just need an easy way to execute codecov locally.

IMHO, that's something you can delegate to github as uncovered lines are marked in 'Files changed' tab.

rousseldenis:
@joshkreud Could you just squash commits that are less relevant ?

At least those ones: 

image

rousseldenis (approved):
Code review

PR #1580, opened 2022-12-13
[16.0][ADD] stock_route_mto

rousseldenis:


sbejaoui (approved):
LGTM (code review)

adrienpeiffer (approved):
Code review and functionnal tests

lmignon (approved):
Code review and functionnal tests
Nitpicking, IMO it should be better to always display the mto field  by default into the route tree.

PR #1581, opened 2022-12-15
[16.0][ADD] Add sale_stock_available_info_popup,stock_available,stock_available_immediately modules in v16

bvanjeelharia:
…vailable_immediately modules


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

And only one module per PR.

PR #1585, opened 2022-12-19
[14.0][FIX] stock_available: dict.record.type is undefined

ferran-S73:
Ever since [this](https://github.com/odoo/odoo/commit/322a89e16fe41e64b8d944327b15587105a323f9) was merged on base, the field `type` stopped being available in the view.

~~Here I add it back to avoid getting an error when display kanban view for `product.template`.~~
Here I change it to use the new `show_on_hand_qty_status_button` field.

### DIFF
--- a/stock_available/views/product_template_view.xml
+++ b/stock_available/views/product_template_view.xml
@@ -83,7 +83,7 @@
         <field name="arch" type="xml">
             <xpath expr="//div[@name='product_lst_price']" position="after">
                 <div
-                    t-if="record.type.raw_value == 'product'"
+                    t-if="record.show_on_hand_qty_status_button.raw_value"
                 >Available to Promise: <field name="immediately_usable_qty" /> <field
                         name="uom_id"
                     /></div>


pedrobaeza (approved):

rousseldenis:
Mmmmh, not sure this is the good place as it could lead to duplicate field if people do the same elsewhere

pedrobaeza:
Not really, Denis. It's the kanban view, that pre-declares all the variables in first place for catching data. It doesn't matter if it's duplicated. Please check a full kanban view structure.

rousseldenis:
> Not really, Denis. It's the kanban view, that pre-declares all the variables in first place for catching data. It doesn't matter if it's duplicated. Please check a full kanban view structure.

Yes, indeed, but not so clean anyway

pedrobaeza:
I don't see it a special problem, but if you have a better idea...

rousseldenis:
> I don't see it a special problem, but if you have a better idea...

Yes: Don't insert the type field. Replace instead the condition like in Odoo core for availability fields by the new show_on_hand_qty_status_button core one.

stock_available/views/product_template_view.xml
```diff
@@ -81,6 +81,9 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="stock.product_template_kanban_stock_view" />
         <field name="arch" type="xml">
+            <xpath expr="//kanban" position="inside">
+                <field name="type" />
+            </xpath>
             <xpath expr="//div[@name='product_lst_price']" position="after">
                 <div
                     t-if="record.type.raw_value == 'product'"
```
rousseldenis:
```suggestion
                    t-if="record.show_on_hand_qty_status_button.raw_value"
```

ferran-S73: I actually see this reasonable, I'll change it

ferran-S73: Change applied @rousseldenis

rousseldenis (approved):

PR #1507, opened 2022-10-11
[MIG] stock_available_immediately: Migration to 15.0

glitchov:

PR #1588, opened 2022-12-22
[15.0][MIG] stock_available_immediately: Migration to 15.0

victoralmau:
Superseed https://github.com/OCA/stock-logistics-warehouse/pull/1507

Migration to 15.0

@Tecnativa TT36990


rousseldenis:
@victoralmau Don't forget that fair play here is to ask first the original author if we can switch. Thanks

pedrobaeza:
We asked on November the 14th for the needed changes with no answer: https://github.com/OCA/stock-logistics-warehouse/pull/1507#issuecomment-1314043641, so I think it's already fair play.

victoralmau:
Tests are now green, could you review it?

rousseldenis:
> We asked on November the 14th for the needed changes with no answer: [#1507 (comment)](https://github.com/OCA/stock-logistics-warehouse/pull/1507#issuecomment-1314043641), so I think it's already fair play.

Yes. I don't like putting in code something that is for tests only.

rousseldenis:
Better is to put this one in rebel module as it is modifying normal field value

pedrobaeza:
I don't like such technique and prefer this one, and it's reasonable to put such code. Tests are code as well, so we are not doing something exceptional. Putting rebel modules complicates everything. If you prefer to do it, please do it on your own modules. We will stick to this one.

rousseldenis:
> If you prefer to do it, please do it on your own modules. We will stick to this one.

This is not how community is working. It's worth the common good that should enforce. Rebel modules is an OCA flow, so we should use it and not putting code that is not necessary. Don't do that.

pedrobaeza:
Rebel modules is a tool like the one we are using. You can't enforce such one that brings a lot of more load to the CI and wait times. This is only adding 3 lines of code.

rousseldenis:
> Rebel modules is a tool like the one we are using. You can't enforce such one that brings a lot of more load to the CI and wait times. This is only adding 3 lines of code.

We don't use Travis anymore that was used to be very long. Github actions jobs are very fast compared to (3 minutes here). So, this is not an argument.

As already said, adding code that is not necessary for a particular context is not wanted as for quality we want on OCA.

Moreover, you can see in the whole history of available modules that we have been involved in their developments.

pedrobaeza:
4 minutes x 2 = 8 minutes

vs

3 lines of code.

We prefer the second one, sorry.

pedrobaeza:
And it's not only that, but then 2 separate commits of PO updates that will generate as well 2 new builds on merge, apart from other problems like the one faced in https://github.com/OCA/product-attribute/pull/1169#issuecomment-1341518663

chienandalu (approved):
Aside from discussions :+1: for me

pedrobaeza (approved):
/ocabot migration stock_available_immediately
/ocabot merge nobump

PR #1589, opened 2022-12-22
[14.0][FIX] stock_available_unreserved: dict.record.type is undefined

zamberjo:
Same error as: https://github.com/OCA/stock-logistics-warehouse/pull/1585

### DIFF
--- a/stock_available_unreserved/views/product_view.xml
+++ b/stock_available_unreserved/views/product_view.xml
@@ -35,9 +35,11 @@
         <field name="inherit_id" ref="stock.product_template_kanban_stock_view" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='qty_available']/parent::div" position="after">
-                <div t-if="record.type.raw_value == 'product'">Unreserved: <field
-                        name="qty_available_not_res"
-                    /> <field name="uom_id" /></div>
+                <div
+                    t-if="record.show_on_hand_qty_status_button.raw_value"
+                >Unreserved: <field name="qty_available_not_res" /> <field
+                        name="uom_id"
+                    /></div>
             </xpath>
         </field>
     </record>


LoisRForgeFlow (changes requested):
This doesn't seem to be working. Just tested it in runboat:

image

LoisRForgeFlow:
BTW, in the original PR it is not described what is being fixed, could post here the steps to reproduce what you are fixing?

Thanks!

zamberjo:
Hi @LoisRForgeFlow ! 

For reproduce, with [odoo](https://github.com/odoo/odoo/commit/322a89e16fe41e64b8d944327b15587105a323f9) updated (or [OCB)](https://github.com/OCA/OCB/commit/322a89e16fe41e64b8d944327b15587105a323f9) and stock_available_unreserved installed go to menu `Sales > Product > Products` (product.template) rendering the Kanban view gives an error as the `type` field is no longer available in the view.

LoisRForgeFlow:
@zamberjo I see now, thanks for the pointer, indeed after that Odoo change there is an issue. However, your fix seems to have some problems too, could you try to fix it?

Thanks for contributing.

zamberjo:
It seems that odoo is out of date on the runboat instance.
image

Works with oca/oca-ci image? The last build was 15 days ago (2022-12-07), while the odoo commit was last week (2022-12-16)

pedrobaeza:
@sbidoul which is the frequency of update of the image?

zamberjo:
Also I just noticed that there is a last module with the same error: [stock_free_quantity](https://github.com/OCA/stock-logistics-warehouse/blob/14.0/stock_free_quantity/views/product_template_view.xml#L44
)
@LoisRForgeFlow If you see it correct I will do another PR fixing it

LoisRForgeFlow:
@zamberjo Yes, I think the fix is correct, you can already open the PR for the last one. In order to merge though, I would like to understand what's the procedure to deal with this cases in the OCA CI - let's wait @sbidoul input.

sbidoul:
> which is the frequency of update of the image?

Nightly but they the builds have been failing due to the chrome installation. Not had time to investigate yet.

zamberjo:
Hi @LoisRForgeFlow the image seems to have been updated:
[image
](https://github.com/OCA/oca-ci/pkgs/container/oca-ci%2Fpy3.6-odoo14.0)
so you can try it again

---

Edit:
I have recreated the instance. When testing it, it gave an error since both modules (`stock_free_quantity` and `stock_available_unreserved`) are installed. You would have to uninstall the `stock_free_quantity` module to test this PR.

LoisRForgeFlow (approved):
Thanks for following up. LGTM :+1:

LoisRForgeFlow:
@zamberjo Is the fix needed in version 15? if so, could you forward port this?

Issue #1578, opened 2022-12-12
16.0 - stock_move_location : not working in inventory adjustment


AungKoKoLin1997:
The name of the module that has a bug.
 -  stock_move_location

**Affected versions**: 16.0

Description:
 - There are two issues when move to location in inventory adjustment. The first one is when we change the "Quantitiy to move" and press enter.Then, it causes the virtual__xxx issue because of many2many field with one2many widget in wizard form and add line using (0,0, values) in default_get function. And the second one is when click "IMMEDIATE TRANSFER" or "PLANNED TRANSFER" button. This issue is caused when we don't enable unit of measure in setting and then face a trouble at float_compare because rounding is got from product_uom_id and there is no product_uom_id in this case.
 
Step to reproduce for first issue:

 - Go to Inventory > Operations > Inventory adjustments
 - Select a product and select action "move location" (you can see wizard)
 - Change the "Quantitiy to move" and click enter.
 
Step to reproduce for second issue:

 - Go to Inventory > Operations > Inventory adjustment
 - Select a product and select action "move location" (you can see wizard)
 - Click "IMMEDIATE TRANSFER" or "PLANNED TRANSFER" button.

Please see this reference for first issue:
https://drive.google.com/file/d/1PV1QwupWy__9CZg8XrDq14SnsBEiRMO4/view?usp=share_link

AungKoKoLin1997:
For first issue, we can fix changing widget from one2many_list to one2many for the versions before v16.0. Unfortunately, it doesn't work in v16.0. Many2many field with one2many widget doesn't support for create="0" in tree tag of v16.0. I think this is the bug for v16.0.

AungKoKoLin1997:
@rousseldenis @pedrobaeza Can you please give me your thoughts on first issue?

pedrobaeza:
Uhm, I don't completely understand the issue as I don't know the internals of the module, but hearing many2many scratches me. Why not using a real one2many for such popup?

PR #743, opened 2019-11-06
[12.0][MIG] stock_move_location

JoanSForgeFlow:
Also improves lines creation in the wizard.
cc @Eficent

PR #1592, opened 2022-12-27
[16.0][FIX] stock_move_location

AungKoKoLin1997:
Use One2many instead of Many2many for popup of move location.

Before this commit, we got issues when we move location from inventory adjust and press enter button. And, even we add create="0" in tree tag of many2many field there is unnecessary popup when we move location from Inventory > Operations > Move from location... and press enter in the last record of many2many tree view.

Using many2many field with one2many_list widget can cause unnecessary things and I think there is no reason we must use many2many field.
Related issues - https://github.com/OCA/stock-logistics-warehouse/issues/1578

### DIFF
--- a/stock_move_location/wizard/stock_move_location.xml
+++ b/stock_move_location/wizard/stock_move_location.xml
@@ -41,7 +41,6 @@
                         <field
                             name="stock_move_location_line_ids"
                             nolabel="1"
-                            widget="one2many_list"
                             mode="tree,kanban"
                             colspan="2"
                         >
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -47,11 +47,10 @@ class StockMoveLocationWizard(models.TransientModel):
         required=True,
         domain=lambda self: self._get_locations_domain(),
     )
-    stock_move_location_line_ids = fields.Many2many(
+    stock_move_location_line_ids = fields.One2many(
+        "wiz.stock.move.location.line",
+        "move_location_wizard_id",
         string="Move Location lines",
-        comodel_name="wiz.stock.move.location.line",
-        column1="move_location_wiz_id",
-        column2="move_location_line_wiz_id",
     )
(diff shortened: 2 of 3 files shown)


stock_move_location/wizard/stock_move_location.xml
```diff
@@ -41,7 +41,7 @@
                         <field
                             name="stock_move_location_line_ids"
                             nolabel="1"
-                            widget="one2many_list"
+                            widget="one2many"
```
pedrobaeza: You can remove the widget in this case.

pedrobaeza (commented):
@rousseldenis do you know which was the reason to put m2m in first place?

yostashiro (approved):
Functional test. 👍

pedrobaeza:
I'm waiting for @rousseldenis' answer. It seems legit, but I want to know the initial reason. Maybe we are missing something. He seems to be on holidays, so let's wait until he comes back.

rousseldenis:
> I'm waiting for @rousseldenis' answer. It seems legit, but I want to know the initial reason. Maybe we are missing something. He seems to be on holidays, so let's wait until he comes back.

I'm currently in holidays. I don't know the reason. Have you looked into commit history?

rousseldenis:
@pedrobaeza it has been done during 12.0 migration.

https://github.com/OCA/stock-logistics-warehouse/pull/743/commits/6b6768dd7bc0ec2adc932d3c9a4c0d72e7543ed4#diff-16b2657901a71cbb31d1be47b3d52e991bd7be9ed2e98545defa4cac435e0cf6

@JoanSForgeFlow could you explain why?

pedrobaeza:
Thanks for investigating. I think it's legit to revert it back, as the change seems arbitrary, and this can be even backported:

/ocabot merge minor

yostashiro:
@AungKoKoLin1997 Can you please backport this to the previous versions with actual UX issue (IIUC, there is no UX issue in 12.0, although it originated there)?

PR #1593, opened 2022-12-27
[15.0][ADD] stock_valuation_layer_accounting_date

AungKoKoLin1997:

@qrtl


stock_valuation_layer_accounting_date/tests/test_stock_valuation_layer_accounting_date.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@tagged("post_install", "-at_install")
+class TestStockValuationStandard(TestStockValuationStandard):
+    def setUp(self):
+        super(TestStockValuationStandard, self).setUp()
+
+    @freeze_time("2022-12-02 23:00:00", tz_offset=9)
+    def test_svl_accounting_date_real_time(self):
+        self.product1.product_tmpl_id.categ_id.property_valuation = "real_time"
```
yostashiro:
```suggestion
        self.product1.categ_id.property_valuation = "real_time"
```

stock_valuation_layer_accounting_date/tests/test_stock_valuation_layer_accounting_date.py
```diff
@@ -0,0 +1,41 @@
...
+        account_move.button_draft()
+        account_move.name = "/"
+        account_move.date = "2022-11-30"
+        account_move.action_post()
+        self.assertEqual(valuation_layer.accounting_date, date(2022, 11, 30))
+
+    def test_svl_accounting_date_manual_periodic(self):
+        # Not using freeze_time() in this test since it cannot be applied to create_date
+        # without a hack.
+        self.product1.product_tmpl_id.categ_id.property_valuation = "manual_periodic"
```
yostashiro:
```suggestion
        self.product1.categ_id.property_valuation = "manual_periodic"
```

yostashiro (approved):
Code review.  The module has been in production use for a while.

PR #1600, opened 2023-01-02
[14.0][IMP] stock_move_location: change location domain

RabbitJon-S73:
Use current user company instead of default to get the domain of locations


rousseldenis (approved):
Trivial change. Thanks

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions if already migrated the module.

PR #1601, opened 2023-01-02
[15.0][MIG]stock_change_qty_reason

GuillemCForgeFlow:
Adapt the module to the changes done in v15.0. As the two models used in v14.0 for the Inventory Adjustment were removed (`stock.inventory` and `stock.inventory.line`), we'll use the `stock_inventory` module added in the OCA (https://github.com/OCA/stock-logistics-warehouse/pull/1512) and propagate the reasons from the `stock.inventory`, `stock.quant` and finally `stock.move`.
@ForgeFlow


MiquelRForgeFlow (changes requested):
FYI I have made a new migration of this module in https://github.com/OCA/stock-logistics-warehouse/pull/1608, because the expected functionality can be maintained without the need to use stock_inventory.

Thus, this PR requires more refactor if you want this to be a complementary module for people who want this functionality WITH stock.inventory models. Otherwise, close it and treat it as superseded.

GuillemCForgeFlow:
> FYI I have made a new migration of this module in #1608, because the expected functionality can be maintained without the need to use stock_inventory.
> 
> Thus, this PR requires more refactor if you want this to be a complementary module for people who want this functionality WITH stock.inventory models. Otherwise, close it and treat it as superseded.

I agree. Closing this.

PR #1602, opened 2023-01-04
[14.0][FIX] stock_free_quantity, get only active variants to compute free_qty on template

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/stock_free_quantity/models/product_template.py
+++ b/stock_free_quantity/models/product_template.py
@@ -10,7 +10,7 @@ class ProductTemplate(models.Model):
         prod_available = super()._compute_quantities_dict()
         for template in self:
             free_qty = 0
-            for p in template.with_context(active_test=False).product_variant_ids:
+            for p in template.product_variant_ids:
                 free_qty += p.free_qty
             prod_available[template.id]["free_qty"] = free_qty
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
Yes, the former code was strange

JordiMForgeFlow (approved):

stock_free_quantity/models/product_template.py
```diff
@@ -10,7 +10,7 @@ def _compute_quantities_dict(self):
         prod_available = super()._compute_quantities_dict()
         for template in self:
             free_qty = 0
-            for p in template.with_context(active_test=False).product_variant_ids:
```
LoisRForgeFlow, line 13: @florian-dacosta as original author, do you have more insights in the need to include archived variants here?

hparfr: @florian-dacosta ?

florian-dacosta: Hello,
Sorry, I missed the first ping.

I don't remember no. 
I think it does not change a lot because archived product should not have stock level equal to 0.
But anyway, ok for the change on my side

PR #1606, opened 2023-01-12
[15.0][IMP] stock_secondary_unit: Compatibility with product variants

sergio-teruel:
cc @Tecnativa TT40323

### DIFF
--- a/stock_secondary_unit/views/stock_move_views.xml
+++ b/stock_secondary_unit/views/stock_move_views.xml
@@ -15,6 +15,7 @@
                     domain="['|', ('product_id', '=', product_id),
                                 '&amp;', ('product_tmpl_id.product_variant_ids', 'in', [product_id]),
                                          ('product_id', '=', False)]"
+                    attrs="{'readonly': [('state', '!=', 'draft'), ('id', '!=', False)]}"
                     options="{'no_create': True}"
                 />
             </field>
@@ -36,6 +37,7 @@
                     domain="['|', ('product_id', '=', product_id),
                                 '&amp;', ('product_tmpl_id.product_variant_ids', 'in', [product_id]),
                                          ('product_id', '=', False)]"
+                    attrs="{'readonly': [('state', '!=', 'draft'), ('id', '!=', False)]}"
                     options="{'no_create': True}"
                 />
             </field>
--- a/stock_secondary_unit/views/stock_move_views.xml
+++ b/stock_secondary_unit/views/stock_move_views.xml
@@ -12,7 +12,9 @@
                 <field name="secondary_uom_qty" />
                 <field
                     name="secondary_uom_id"
-                    domain="[('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
+                    domain="['|', ('product_id', '=', product_id),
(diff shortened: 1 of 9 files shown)


sergio-teruel:
ping  @CarlosRoca13 @carlosdauden

stock_secondary_unit/views/stock_move_views.xml
```diff
@@ -12,7 +12,9 @@
                 <field name="secondary_uom_qty" />
                 <field
                     name="secondary_uom_id"
-                    domain="[('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
+                    domain="['|', ('product_id', '=', product_id),
+                                '&amp;', ('product_tmpl_id.product_variant_ids', 'in', [product_id]),
+                                         ('product_id', '=', False)]"
```
CarlosRoca13, line 17:
```suggestion
                    domain="['|', ('product_id', '=', product_id),
                                ('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
```



If the first part of the domain is evaluated as True, I think the second part of the domain is not checked because True or whatever is equal to True, so in case the product_id is filled, it would not check this second part of the domain and it would only check in case product_id = False.

stock_secondary_unit/views/stock_move_views.xml
```diff
@@ -31,7 +33,9 @@
...
```
CarlosRoca13, line 39:
```suggestion
                    domain="['|', ('product_id', '=', product_id),
                                '&amp;', ('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
```

stock_secondary_unit/views/stock_picking_views.xml
```diff
@@ -18,7 +18,9 @@
                 />
                 <field
                     name="secondary_uom_id"
-                    domain="[('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
+                    domain="['|', ('product_id', '=', product_id),
+                                '&amp;', ('product_tmpl_id.product_variant_ids', 'in', [product_id]),
+                                         ('product_id', '=', False)]"
```
CarlosRoca13, line 23:
```suggestion
                    domain="['|', ('product_id', '=', product_id),
                                ('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
```

stock_secondary_unit/views/stock_picking_views.xml
```diff
@@ -33,7 +35,9 @@
...
```
CarlosRoca13, line 40:
```suggestion
                    domain="['|', ('product_id', '=', product_id),
                                ('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
```

pedrobaeza:
@sergio-teruel what about @CarlosRoca13's comments?

stock_secondary_unit/views/stock_move_views.xml
pedrobaeza: It's correct this way, as `product_id` is always going to have a value, so if the product variant is filled in the secondary unit, it has to match. If not, the double condition of the second part is avoiding to match on other secondary units having a specific variant.

pedrobaeza (approved):
/ocabot merge major

PR #1607, opened 2023-01-12
[14.0][FIX] stock_move_location: wrong destination location

RabbitJon-S73:
When you change `Destination Location`, lines are visually updated but actually behind the field remains the same, so the stock move lines are wrong created. If we force recreate lines it works

For reproducing it: 

1. Go to Inventory > Operations > Move from location... 
2.  Change the `Destination location` to any other. 
3. Press `Inmediate transfer`.
4.  Check that the stock move lines generated in the picking are wrong because location destination isn't the one we chosed.

Function `get_putaway_strategy` is now called `_get_putaway_strategy`.

### DIFF
--- a/stock_move_location/wizard/stock_move_location.xml
+++ b/stock_move_location/wizard/stock_move_location.xml
@@ -64,11 +64,7 @@
                                     readonly="1"
                                     force_save="1"
                                 />
-                                <field
-                                    name="destination_location_id"
-                                    readonly="1"
-                                    force_save="1"
-                                />
+                                <field name="destination_location_id" />
                                 <field
                                     name="lot_id"
                                     domain="[('product_id', '=', product_id)]"
--- a/stock_move_location/tests/test_move_location.py
+++ b/stock_move_location/tests/test_move_location.py
@@ -258,3 +258,17 @@ class TestMoveLocation(TestsCommon):
         second_line = wizard.stock_move_location_line_ids[1]
         second_line.product_id = False
         self.assertEqual(second_line._get_available_quantity(), (0, 0))
+
+    def test_wizard_different_destinations(self):
+        """
+        Create a picking whose line destinations are differents. The first line is sent
+        to the origin location.
+        """
+        wizard = self._create_wizard(self.internal_loc_1, self.internal_loc_2)
+        wizard.onchange_origin_location()
+        wizard.stock_move_location_line_ids[0].write(
+            {"destination_location_id": self.internal_loc_1.id}
+        )
+        wizard.action_move_location()
+        locations = self.internal_loc_1 + self.internal_loc_2
+        self.assertEqual(wizard.picking_id.move_line_ids.location_dest_id, locations)
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -155,8 +155,11 @@ class StockMoveLocationWizard(models.TransientModel):
 
     @api.onchange("destination_location_id")
     def _onchange_destination_location_id(self):
-        for line in self.stock_move_location_line_ids:
-            line.destination_location_id = self.destination_location_id
+        if self.env.context.get("active_model", False) == "stock.quant":
+            for line in self.stock_move_location_line_ids:
+                line.destination_location_id = self.destination_location_id
+        else:
+            self.create_lines()
 
     def _clear_lines(self):
         self.stock_move_location_line_ids = False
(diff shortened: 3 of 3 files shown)


stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -22,7 +22,9 @@ class StockMoveLocationWizardLine(models.TransientModel):
         string="Origin Location", comodel_name="stock.location"
     )
     destination_location_id = fields.Many2one(
-        string="Destination Location", comodel_name="stock.location"
+        string="Destination Location",
+        comodel_name="stock.location",
+        related="move_location_wizard_id.destination_location_id",
```
rousseldenis: Isn't this too restrictive as we maybe want to update some lines from the default value ?

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -100,7 +102,7 @@ def _get_move_line_values(self, picking, move):
         self.ensure_one()
         location_dest_id = (
             self.move_location_wizard_id.apply_putaway_strategy
-            and self.destination_location_id.get_putaway_strategy(self.product_id).id
+            and self.destination_location_id._get_putaway_strategy(self.product_id).id
```
rousseldenis: From my first comment, this won't work as the destination location will always be the wizard one.

stock_move_location/wizard/stock_move_location_line.py
RabbitJon-S73: Finally I'm doing the field computed because it wasn't applying correctly the putaway strategies. The functionality is the same as before since this line field was readonly

stock_move_location/wizard/stock_move_location_line.py
```diff
@@ -22,7 +39,9 @@ class StockMoveLocationWizardLine(models.TransientModel):
         string="Origin Location", comodel_name="stock.location"
     )
     destination_location_id = fields.Many2one(
-        string="Destination Location", comodel_name="stock.location"
+        string="Destination Location",
+        comodel_name="stock.location",
+        compute="_compute_destination_location",
```
rousseldenis: I would say it should be stored, readonly False in order to allow user change one line if desired

rousseldenis: Missing readonly=False

RabbitJon-S73: I couldn't implement the readonly=False funtionality because compute override the destination location in every line when picking is created, and doesn't take the correct value. I try to do that later.

RabbitJon-S73:
@rousseldenis could this be merged?

rousseldenis:
@RabbitJon-S73 IMHO, this is a change in the desired behavior. With this, you cannot change a particular destination location on a particular line.

RabbitJon-S73:
@rousseldenis I reverted the `compute` since I couldn't find the way to make it work. Instead, if we clean lines and recompute, the onchange is applied correctly and we still can edit each line destination

RabbitJon-S73:
@rousseldenis how do you find this fix/imp now?

RabbitJon-S73:
@rousseldenis @etobella could this be merged?

rousseldenis:
@RabbitJon-S73 Tests are red

stock_move_location/wizard/stock_move_location.xml
```diff
@@ -64,11 +64,7 @@
                                     readonly="1"
                                     force_save="1"
                                 />
-                                <field
-                                    name="destination_location_id"
-                                    readonly="1"
-                                    force_save="1"
-                                />
+                                <field name="destination_location_id" force_save="1" />
```
rousseldenis: If you remove readonly, you can remove force_save

RabbitJon-S73: Done!

RabbitJon-S73:
@rousseldenis

RabbitJon-S73:
@pedrobaeza @rousseldenis can we merge please?

rousseldenis:
@RabbitJon-S73 Could you rebase ?

RabbitJon-S73:
@rousseldenis :green_apple:

rousseldenis:
> @rousseldenis 🍏

Yes I wanted to test on runboat :-)

rousseldenis (approved):
Code and functional review.

rousseldenis:
@RabbitJon-S73 Do you have time to do the test case ?

RabbitJon-S73:
@rousseldenis I really thougt that was working but not. I changed the code and now works fine. I've done many tests and that is the only way I found to let the user choose each destination location in line. I also add a test of this case

sergio-teruel (approved):
Thanks... 

I thinlk that this module must be reviewed.... in all version... but this is other scope... 
@RabbitJon-S73 Can you review the v15 behavior..??

PR #1608, opened 2023-01-12
[15.0][MIG] stock_change_qty_reason

MiquelRForgeFlow:
Standard migration (without using stock_inventory module).


GuillemCForgeFlow (changes requested):
You should extend the method `_get_inventory_fields_write` for the `stock.quant` model adding the two new fields. Otherwise, when applying the inventory adjustment it will fail.

MiquelRForgeFlow:
@GuillemCForgeFlow ready

GuillemCForgeFlow (approved):
thank you 😄

AaronHForgeFlow:
Why the reason is in both, the quants and the stock move lines?

MiquelRForgeFlow:
@AaronHForgeFlow In the quant, to be able to temporarily introduce the reason in the adjustment view. In the move line, because it's were it ends saved (inventory_history).

rousseldenis:
> @AaronHForgeFlow In the quant, to be able to temporarily introduce the reason in the adjustment view. In the move line, because it's were it ends saved (inventory_history).

@MiquelRForgeFlow Shouldn't it be located in stock move instead? As stock.move.line seems too volatile for me to be as safe as wanted.

MiquelRForgeFlow:
@rousseldenis inventory history is handled by move lines, not moves, see [odoo code here](https://github.com/odoo/odoo/blob/e83f83fb8180683793a4030f91c246deaec5308b/addons/stock/models/stock_quant.py#L361).

Have you tested functionally the module?

1st step:
Selection_727
2nd step:
Selection_728
3rd step:
Selection_729

rousseldenis:
> @rousseldenis inventory history is handled by move lines, not moves, see [odoo code here](https://github.com/odoo/odoo/blob/e83f83fb8180683793a4030f91c246deaec5308b/addons/stock/models/stock_quant.py#L361).
> 
> Have you tested functionally the module?
> 
> 1st step: Selection_727 2nd step: Selection_728 3rd step: Selection_729

Yes of course. But inventories entry point is stock.move:

https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_quant.py#L603

So, I would have put the reason fields on that level and relateds on move lines

MiquelRForgeFlow:
> So, I would have put the reason fields on that level and relateds on move lines

@rousseldenis Yes, I could do that. Doable. But, in my opinion, what you propose doesn't improve anything substantially, it's practically the same. I prefer in stock move lines for continuity reasons. In this module, in v14, you could have a different reason for each inventory line, independently of the global reason of the stock inventory. Although in v15 each inventory stock move only has a move line, I wanted to maintain that feature of v14 in some way, and thus, deciding to put the reason in the stock move line. Imagine if a custom module enables doing adjustments to two products at the same time, and thus, generating a move with two stock move lines. Then, what if one may want a different reason for each move line?

rousseldenis:
> s to two products at the same time, and thus, generating a move with two stock move lines.

It's impossible as it will generate one move per product.

But ok, let's go forward with that. I wanted to highlight the fact that quite every stock reporting is done on move level and people should be aware to get the information from the line.

MiquelRForgeFlow:
Ok, thanks, could you approve? :)

rousseldenis (approved):

AaronHForgeFlow (approved):
Code review

JoanSForgeFlow (approved):

rousseldenis:
@MiquelRForgeFlow Ready ?

MiquelRForgeFlow:
Yes

PR #1613, opened 2023-01-20
[16.0][MIG] account_move_line_stock_info: Migration to 16.0

BT-atornero:


account_move_line_stock_info/tests/test_account_move_line_stock_move.py
```diff
@@ -121,18 +114,16 @@ def _create_picking(self, picking_type, location, location_dest):
                 "picking_type_id": picking_type.id,
                 "location_id": location.id,
                 "location_dest_id": location_dest.id,
-                "move_lines": [
+                "move_line_ids": [
```
rousseldenis: That is not the same.

Change to 'move_ids' instead

BT-atornero: do you mean 'move_lines'? because 'move_ids' does not seem to exist

rousseldenis: Of course it exists : https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L346

rousseldenis: move_lines has been renamed to move_ids : https://github.com/odoo/odoo/commit/3a1473c75c5fabfa9b060c77eed729f74f14b559

BT-atornero:
there is another PR with the migration of this module https://github.com/OCA/stock-logistics-warehouse/pull/1645

PR #1614, opened 2023-01-20
[14.0][FIX] stock_available: field must be present in view before being able to access raw_value

cybernexus:
Go to Sales > Products kanban view and a hard error shows that raw_value is not available on undefined. This is because the field in question is not in the kanban view. This pull request adds the field and sets it to invisible.

### DIFF
--- a/stock_available/__manifest__.py
+++ b/stock_available/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Stock available to promise",
-    "version": "14.0.1.1.0",
+    "version": "14.0.1.1.1",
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Hi, thanks for the contribution. Please don't change the version manually. It will be done on bot merge command.

Is this problem reproducible in runboat?

jbaudoux (changes requested):
The field is already there:
https://github.com/odoo/odoo/blob/14.0/addons/stock/views/product_views.xml#L156

I think you are not up to date with odoo core

PR #1618, opened 2023-01-25
[14.0][IMP] stock_request: Change the label string to make it consistent

victoralmau:
Change the label string to make it consistent

@Tecnativa TT41343

### DIFF
--- a/stock_request/views/stock_request_order_views.xml
+++ b/stock_request/views/stock_request_order_views.xml
@@ -4,7 +4,7 @@
         <field name="name">stock.request.order.tree</field>
         <field name="model">stock.request.order</field>
         <field name="arch" type="xml">
-            <tree string="Stock requests">
+            <tree>
                 <field name="name" />
                 <field name="warehouse_id" groups="stock.group_stock_multi_locations" />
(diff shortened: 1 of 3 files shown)


stock_request/views/stock_request_order_views.xml
```diff
@@ -4,7 +4,7 @@
         <field name="name">stock.request.order.tree</field>
         <field name="model">stock.request.order</field>
         <field name="arch" type="xml">
-            <tree string="Stock requests">
+            <tree string="Stock request Orders">
```
pedrobaeza: The string in tree/form is useless, so remove them instead.

JordiBForgeFlow (approved):

stock_request/views/stock_request_order_views.xml
```diff
@@ -16,7 +16,7 @@
         <field name="name">stock.request.order.form</field>
         <field name="model">stock.request.order</field>
         <field name="arch" type="xml">
-            <form string="Stock request">
+            <form string="Stock request Order">
```
pedrobaeza: I said both in tree/form

pedrobaeza (approved):
/ocabot merge minor

PR #1619, opened 2023-01-26
[14.0][FIX] stock_request: Change state fields to display translated by language

victoralmau:
Change state fields to display translated by language

@Tecnativa TT41343

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -5,25 +5,12 @@ from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools import float_compare
 
-REQUEST_STATES = [
-    ("draft", "Draft"),
-    ("open", "In progress"),
-    ("done", "Done"),
-    ("cancel", "Cancelled"),
-]
-
 
 class StockRequest(models.Model):
     _name = "stock.request"
     _description = "Stock Request"
     _inherit = "stock.request.abstract"
     _order = "id desc"
-
-    def __get_request_states(self):
-        return REQUEST_STATES
-
-    def _get_request_states(self):
-        return self.__get_request_states()
 
     def _get_default_requested_by(self):
         return self.env["res.users"].browse(self.env.uid)
@@ -34,7 +21,12 @@ class StockRequest(models.Model):
 
     name = fields.Char(states={"draft": [("readonly", False)]})
     state = fields.Selection(
-        selection=_get_request_states,
+        selection=[
+            ("draft", "Draft"),
+            ("open", "In progress"),
+            ("done", "Done"),
+            ("cancel", "Cancelled"),
+        ],
         string="Status",
         copy=False,
         default="draft",
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -25,7 +25,9 @@ class StockRequestOrder(models.Model):
         return res
 
     def __get_request_order_states(self):
-        return self.env["stock.request"]._get_request_states()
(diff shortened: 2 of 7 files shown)


stock_request/models/stock_request.py
```diff
@@ -5,25 +5,20 @@
...
     _description = "Stock Request"
     _inherit = "stock.request.abstract"
     _order = "id desc"
 
-    def __get_request_states(self):
-        return REQUEST_STATES
-
-    def _get_request_states(self):
-        return self.__get_request_states()
+    def _get_states(self):
```
chienandalu: Why changing the method name?

stock_request/models/stock_request.py
```diff
@@ -5,25 +5,20 @@
...
+        return [
+            ("draft", _("Draft")),
```
rousseldenis: You cannot include translations like that in selections. I have to check, but I've solved this in another way

stock_request/models/stock_request.py
```diff
@@ -34,7 +29,7 @@ def _get_expected_date():
 
     name = fields.Char(states={"draft": [("readonly", False)]})
     state = fields.Selection(
-        selection=_get_request_states,
+        selection="_get_states",
```
rousseldenis: @victoralmau I think you should use a lambda method instead

rousseldenis: @victoralmau Please follow @chienandalu comments and change this to lambda? Thanks

victoralmau:
Changes done according to another approach.

stock_request/models/stock_request_order.py
```diff
@@ -42,7 +41,7 @@ def _get_default_requested_by(self):
         default="/",
     )
     state = fields.Selection(
-        selection=_get_request_order_states,
+        selection="_get_request_order_states",
```
rousseldenis: Could you change to lambda too ?

stock_request_submit/models/stock_request.py
```diff
@@ -2,17 +2,17 @@
...
 
-    def __get_request_states(self):
-        states = super().__get_request_states()
-        if not ("submitted", "Submitted") in states:
+    def _get_request_states(self):
+        states = super()._get_request_states()
+        if not ("submitted", _("Submitted")) in states:
             states.insert(
-                states.index(("draft", "Draft")) + 1, ("submitted", "Submitted")
+                states.index(("draft", _("Draft"))) + 1, ("submitted", _("Submitted"))
```
chienandalu: In any case, if this is the reason of this implementation, there's already a way to it without such overhead:

https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html?highlight=selection_add#odoo.fields.Selection
```
selection = [('a', 'A'), ('b', 'B')]
selection_add = [('c', 'C'), ('b',)]
> result = [('a', 'A'), ('c', 'C'), ('b', 'B')]
```

Another issue is that you need to mantain the uninstall hook, wich is also covered with `ondelete` parameter.

rousseldenis: @chienandalu This is not available in v14

chienandalu: Ok, then :sweat_smile:

pedrobaeza: `selection_add` is in v14 AFAIK, and the uninstall thing is independent from using a selection method.

victoralmau: Ok, changes done using this approach (`selection_add`) and simplifying everything.

chienandalu (approved):
:+1:

pedrobaeza (approved):
/ocabot merge minor

PR #1621, opened 2023-01-27
[15.0][ADD] stock_request_separate_picking

ps-tubtim:
This module automatically creates a new procurement group when creating a stock request/stock request order. Therefore one stock request/stock request order creates one new picking.

**Settings**
Selection_196


stock_request_separate_pickings/models/stock_request_abstract.py
```diff
@@ -0,0 +1,24 @@
...
+    _inherit = "stock.request.abstract"
+
+    procurement_group_id = fields.Many2one(
+        copy=False,
+    )
+
+    @api.model
+    def create(self, vals):
+        res = super().create(vals)
+        if not res.procurement_group_id:
```
rousseldenis: @ps-tubtim As this break stock_request behavior, maybe a parameter to activate it (on company with res.config.settings) should be great

pedrobaeza:
Please don't use plurals in module name.

PatthamaCH (approved):
ok

stock_request_separate_pickings/models/stock_request_abstract.py
ps-tubtim: @rousseldenis Thank you for your suggestion. I already add the **Allow Separate Picking** setting. Could you review this module again?

ps-tubtim:
> Please don't use plurals in module name.

@pedrobaeza I already renamed this module to **stock_request_separate_picking**.  What do you think?

rousseldenis (approved):

PR #1626, opened 2023-02-01
[16.0][ADD] stock_picking_volume

lmignon:
Compute and store the volume of a picking.

For some processes, you need to know the volume of your stock picking. This could
be useful for example to know the device to use to process it. An other use case
could be to know how many trucks you need to deliver your products.

By default, the delivery module from Odoo will compute the picking weight.
Unfortunately, it will not compute the volume. This module will add the volume
computation to the picking and to the stock move. This information will be
stored in the database and will be available on the picking form view and
on the stock move form view.

cc @rousseldenis @jbaudoux


rousseldenis (approved):
Code review. LGTM

jbaudoux (approved):

simahawk (approved):
LG

simahawk:
@lmignon next time,  please, keep the commit msg short ;)

simahawk:
> @lmignon next time, please, keep the commit msg short ;)

The decription of the PR should be included in the commit description -> you'll get a reason to keep the msg short and the PR description will be filled automatically. Just saying.... ;P

PR #1629, opened 2023-02-01
[16.0][ADD] stock_picking_volume_packaging

lmignon:
* [x] includes #1626 
* [x] depends on https://github.com/OCA/stock-logistics-warehouse/pull/1626

This modules improves the way the volume is computed on a stock.move by
taking into account the volume of the potential product packaging
that can be picked to fulfill the move.

The potential packaging that can be picked for a move are computed by the
module *stock_packaging_calculator*. Thanks to this module we can compute the
best distribution of the packaging to uses to fulfill a specific quantity of a
product. (This information is important for the picking operators to minimize
the number of manipulations to do. Even if this information is not available
into the Odoo UI, The *Shopfloor* addon takes advantage of it to propose
the best picking strategy to the user).

By default the volume information is not available on the product packaging.
Hopefully the module *product_packaging_dimension* provides this information.

Since the volume information is not a mandatory field on the product packaging
when we ask for the best distribution of the packaging, packaging without volume
information are ignored. In this way we ensure that the volume of the packaging
is only taken into account when it's relevant otherwise we fallback on the
volume of the products.

cc @rousseldenis @jbaudoux


jbaudoux (approved):
LGTM
Needs rebase after dependency is merged

rousseldenis:
@lmignon As dependencies are merged, could you rebase and remove test-requirements.txt commit ?

jbaudoux:
> @lmignon As dependencies are merged, could you rebase and remove test-requirements.txt commit ?

and make first commit line short :)

lmignon:
@rousseldenis @jbaudoux rebase done...

rousseldenis (approved):

jbaudoux:
@lmignon But not the split of the long commit message on the first line :(

lmignon:
@jbaudoux Done

Comment not related to the current RP:
I agree whe have to pay attention to our commit messages. Nevertheless I'm not prepared to waste time rewriting my commits to meet everyone's expectations. We don't all share the same vision of how to manage our commits. Some people like to keep track of the reasoning they have, others would like 1 PR = 1 commit.

PR #1633, opened 2023-02-08
[14.0][FIX] stock_request_submit: Consistency of string with method

victoralmau:
Consistency of string with method

@Tecnativa TT41533


pedrobaeza (commented):
I don't get why changing the button action and text and doing it more convoluted instead of just changing the visibility of the existing button.

victoralmau:
> I don't get why changing the button action and text and doing it more convoluted instead of just changing the visibility of the existing button.

Ok, the change you indicate is already applied (the same one that already existed in the form view of `stock.request`).

pedrobaeza (approved):

chienandalu (approved):

PR #1634, opened 2023-02-08
[FIX] stock_available_base_exclude_location: Fix excluded locations domains

paradoxxxzero:
I think there is a copy paste error in the `product.product._get_domain_locations_new` domains override

_Courtesy of Olivier Nibart_


rousseldenis:
@paradoxxxzero Indeed :sweat_smile: . Thanks!

rousseldenis (approved):

PR #1637, opened 2023-02-13
[14.0][FIX] stock_request: use stock request expected_date as deadline_date

petrus-v:
In this PR I've done 3 things, let me now It you prefer 3 distincts PRs:

* https://github.com/OCA/stock-logistics-warehouse/commit/5f40bc0f997c18b73e81f1049cfb47291a2f1a58: improves unit test to make sure `assetRaises` realy test what it expected to test
* https://github.com/OCA/stock-logistics-warehouse/commit/dddabb3003781f83a80cf45791428df5f45fc3e9: do not force default expected date on stock.request if provide on creation without stock.request.order
* use the expected date as deadline_date on procurement creation to avoid merging `stock.move` from different stock.request with different expected date. I wonder the impact on running prod env and wonder if I should do something less impacting ?

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -19,8 +19,7 @@ class StockRequest(models.Model):
     def _get_default_requested_by(self):
         return self.env["res.users"].browse(self.env.uid)
 
-    @staticmethod
-    def _get_expected_date():
+    def _get_expected_date(self):
         return fields.Datetime.now()
 
     name = fields.Char(states={"draft": [("readonly", False)]})
@@ -48,6 +47,7 @@ class StockRequest(models.Model):
     expected_date = fields.Datetime(
         "Expected Date",
         index=True,
+        default=_get_expected_date,
         required=True,
         readonly=True,
         states={"draft": [("readonly", False)]},
@@ -406,8 +406,6 @@ class StockRequest(models.Model):
         if "order_id" in upd_vals:
             order_id = self.env["stock.request.order"].browse(upd_vals["order_id"])
             upd_vals["expected_date"] = order_id.expected_date
-        else:
-            upd_vals["expected_date"] = self._get_expected_date()
         return super().create(upd_vals)
 
     def unlink(self):
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -330,6 +330,7 @@ class StockRequest(models.Model):
             "date_planned": max(
                 [self.expected_date, datetime.combine(self.lead_days_date, time.min)]
             ),
+            "date_deadline": self.expected_date,
             "warehouse_id": self.warehouse_id,
             "stock_request_allocation_ids": self.id,
             "group_id": group_id or self.procurement_group_id.id or False,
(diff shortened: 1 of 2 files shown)


stock_request/models/stock_request.py
```diff
@@ -283,6 +280,7 @@ def _prepare_procurement_values(self, group_id=False):
         """
         return {
             "date_planned": self.expected_date,
+            "date_deadline": self.expected_date,
```
petrus-v, line 333: I would prefer send date instead date time but it really depends on business here :thinking:

petrus-v: In fact that this is an other matter that should be address from an other module, this proposed fix is similar from what's  done in sale_stock module while selling goods, date_deadline  is set from date_order which is a datetime

petrus-v: done by https://github.com/OCA/stock-logistics-workflow/pull/1226

petrus-v:
@hparfr, @Kev-Roche review welcome here !

stock_request/models/stock_request.py
```diff
@@ -48,6 +44,7 @@ def _get_expected_date():
     expected_date = fields.Datetime(
         "Expected Date",
         index=True,
+        default=lambda self: fields.Datetime.now(),
```
rousseldenis: @petrus-v I would have put this in a function in order to allow override.

stock_request/tests/test_stock_request.py
```diff
@@ -451,7 +492,7 @@ def test_stock_request_order_validations_06(self):
         expected_date = fields.Datetime.now()
         child_expected_date = "2015-01-01"
         vals = {
-            "company_id": self.company_2.id,
+            "company_id": self.main_company.id,
```
rousseldenis, line 497: Why this change ?

stock_request/models/stock_request.py
petrus-v: I agree there is no reason for this change !...

stock_request/tests/test_stock_request.py
petrus-v: I probably think this `test_stock_request_order_validations_06` was the same test as [test_stock_request_order_validations_05](https://github.com/OCA/stock-logistics-warehouse/pull/1637/files#diff-aea988ceb44324ce797bd2eaf94bc771996f7db0ef6308404d2bf431570ec55cR430), but maybe not... I'll probably restore this test and create a new one changing the doc-string seems to be outdated.

Thanks

stock_request/models/stock_request.py
petrus-v: hum.. I mean I'll use the existing method !

stock_request/tests/test_stock_request.py
petrus-v: I've rename the test and change the docstring, I confirm that it was testing the same assertion as `test_stock_request_order_validations_05` !

stock_request/models/stock_request.py
petrus-v: done

damdam-s (approved):

PR #1639, opened 2023-02-15
[14.0][FIX] stock_reserve: Avoid creation of pickings when confirming stock reservations

victoralmau:
Avoid creation of pickings when confirming stock reservations.

Example use case:
- Create a stock reservation.
- Click on the "Reserve" button

Before this change, a delivery picking was created (unnecessary in my opinion), now, NO picking is created.

@Tecnativa TT41080

### DIFF
--- a/stock_reserve/models/stock_reserve.py
+++ b/stock_reserve/models/stock_reserve.py
@@ -122,8 +122,9 @@ class StockReservation(models.Model):
         A date until which the product is reserved can be specified.
         """
         self.write({"date_deadline": fields.Datetime.now()})
-        self.mapped("move_id")._action_confirm(merge=False)
-        self.mapped("move_id.picking_id").action_assign()
+        # Don't call _action_confirm() method to prevent assign picking
+        self.mapped("move_id").write({"state": "confirmed"})
+        self.mapped("move_id")._action_assign()
         return True
 
     def release_reserve(self):
@@ -132,8 +133,6 @@ class StockReservation(models.Model):
         """
         moves = self.mapped("move_id")
         moves._action_cancel()
-        # HACK: For avoiding to accumulate all reservations in the same picking
-        moves.write({"picking_id": False})
         return True
 
     def _get_state_domain_release_reserve(self, mode):
--- a/stock_reserve/tests/test_stock_reserve.py
+++ b/stock_reserve/tests/test_stock_reserve.py
@@ -33,9 +33,11 @@ class TestStockReserve(common.TransactionCase):
     def test_reservation_and_reservation_release(self):
         reservation_1 = self._create_stock_reservation(6)
         reservation_1.reserve()
+        self.assertFalse(reservation_1.picking_id)
         self.assertEqual(self.product.virtual_available, 4)
         reservation_2 = self._create_stock_reservation(1)
         reservation_2.reserve()
+        self.assertFalse(reservation_2.picking_id)
         self.assertEqual(self.product.virtual_available, 3)
         reservation_1.release_reserve()
         self.assertEqual(self.product.virtual_available, 9)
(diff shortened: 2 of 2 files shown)


rousseldenis:
@victoralmau Could you explain a little bit the use case ?

victoralmau:
> @victoralmau Could you explain a little bit the use case ?

Ok, sorry, I have updated the initial description with the use case.

rousseldenis:
Mmmh, I would say:

- The advantage of the picking is that all reserved products are grouped into the same one:
image

- This is a breaking change as people maybe want this feature.
- Nevertheless, I'm never in favor of passing parameters through context. For that purpose, I would have created a module that:
- - Adds a field on stock move: - no_picking (module name: stock_move_no_picking ?)
- - Then, on picking assignation, filter the moves that has that field True and bypass assignation
- Add a configuration that enable that feature or not in stock_reserve module

rousseldenis:
@yvaucher @gurneyalex @guewen  What do you think ?

pedrobaeza:
The picking shouldn't be in any case a delivery order. There may be a picking type for reservations. Are all the reservations across time put on the same picking? If so, that's terrible.

rousseldenis:
> The picking shouldn't be in any case a delivery order. There may be a picking type for reservations. Are all the reservations across time put on the same picking? If so, that's terrible.

Indeed, a new picking type should be great.

Using the standard flow, it is populated until Unreserve, so yes, it can be big.

I see this, so there is maybe more than what is done here :

https://github.com/OCA/stock-logistics-warehouse/blob/14.0/stock_reserve/models/stock_reserve.py#L136

pedrobaeza:
And why not simply having them as `stock.move`? They are searchable through several menus, and also traceability is not lost, so I think we don't really need a picking.

rousseldenis:
> And why not simply having them as `stock.move`? They are searchable through several menus, and also traceability is not lost, so I think we don't really need a picking.

Indeed, this is sufficient to manage the reservations.

At least, my suggestions about the context are still valid

victoralmau:
Reviewing what you suggest, IMO I think it is adding too much complexity and code (add a new module and configuration to enable or disable the feature) for something that we clearly need to inhibit (auto-creation of pickings with stock reservations).

stock_reserve/models/stock_reserve.py
```diff
@@ -122,8 +122,10 @@ def reserve(self):
         A date until which the product is reserved can be specified.
         """
         self.write({"date_deadline": fields.Datetime.now()})
-        self.mapped("move_id")._action_confirm(merge=False)
-        self.mapped("move_id.picking_id").action_assign()
+        self.mapped("move_id").with_context(skip_assign_picking=True)._action_confirm(
```
pedrobaeza: I think this is simpler: just don't call `_action_confirm`. It's not doing anything we want if I don't oversee anything.

victoralmau: Ok, we don't call `_action_confirm()` but we need to at least confirm the stock moves for later in `_action_assign()` method.

rousseldenis:
> Reviewing what you suggest, IMO I think it is adding too much complexity and code (add a new module and configuration to enable or disable the feature) for something that we clearly need to inhibit (auto-creation of pickings with stock reservations).

It was because that context key can appear in several same approach modules.
Adding a new one can be shared in other flows.

stock_reserve/models/stock_move.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _assign_picking(self):
```
pedrobaeza: You can now remove this.

pedrobaeza (approved):
Now there's no context keys

chienandalu (approved):

rousseldenis (approved):

PR #1466, opened 2022-08-10
[14.0] stock_helper: add get_closest_warehouse method to stock.location

mt-software-de:
This PR adds the method get_closest_warehouse to stock.location
This will return the closest parent warehouse of the location

PR #1640, opened 2023-02-15
[FW] stock_helper: add location.get_closest_warehouse

rousseldenis:
Forwardport of :

- #1466 

On a multi warehouse configuration
where a warehouse view_location is parent of an other warehouse's view_location this will return the closest warehouse of a location

Not like the get_warehouse method (odoo core code) which returns from all parent warehouses found
the first one ordered by the sequence

### DIFF
--- a/stock_helper/README.rst
+++ b/stock_helper/README.rst
@@ -49,12 +49,14 @@ Authors
 ~~~~~~~
 
 * Camptocamp
+* Michael Tietz (MT Software)
 
 Contributors
 ~~~~~~~~~~~~
 
 * Guewen Baconnier <guewen.baconnier@camptocamp.com>
 * Christopher Ormaza <chris.ormaza@forgeflow.com>
+* Michael Tietz (MT Software) <mtietz@mt-software.de>
 
 Maintainers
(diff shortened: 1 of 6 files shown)


rousseldenis:
@mt-software-de

rousseldenis:
@mt-software-de I think the warehouse_id field is now well computed and is the closest warehouse on a location

https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_location.py#L151

mt-software-de:
> @mt-software-de I think the warehouse_id field is now well computed and is the closest warehouse on a location
> 
> https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_location.py#L151

I don't think so because, there is no sorting happening on the parent_path, otherwise it depends on how the warehouses are returned here https://github.com/odoo/odoo/blob/076e48251ca502f14258dd0ba3a94b65fe2ecb14/addons/stock/models/stock_location.py#L152

mt-software-de:
We don't need this anymore. 
I have created a PR for odoo.
https://github.com/odoo/odoo/pull/118128

PR #1641, opened 2023-02-16
[15.0][MIG] stock_orderpoint_manual_procurement: Migration to version 15.0

ernesto-garcia-tecnativa:
@tecnativa TT37320


ernesto-garcia-tecnativa:
@pedrobaeza @victoralmau please review

chienandalu (commented):
I wonder if this module features overlap with the already existing in core where you can force a procurement, analyze the stock forecast...:

image

What do you think @pedrobaeza ?

pedrobaeza:
I'm afraid it's not exactly the same, as Odoo feature is only real demand, and not based on a past period, but this generator extrapolates the amounts of the given range for having the corresponding reordering rules.

It should be migrated, but several things:

1. The README is very outdated, even mentioning a non-existing security group. Please update it.
2. There's an ACL that I don't like too much, because it's very broad. Please review why it's needed, and find an alternative.
3. Please be sure the created reordering rules are not overwritten when entering again on the replenishment screen.

ernesto-garcia-tecnativa:
@pedrobaeza please review

pedrobaeza:
Oh, I'm ashamed. I have totally mixed this module with `stock_orderpoint_generator`. Indeed @chienandalu was right, and this one is not needed, as now Odoo includes the possibility of doing the same.

The only pain points:

- Reordering rules are created automatically by Odoo. We have to check the persistence of the manual ones. There's a filter "Trigger: Manual", so it seems OK, but not 100% sure.
- There's no specific security group for doing that.

For now, let's close this and merge it with `stock` in OpenUpgrade, and check if there's no problem in the customer with these pain points.

PR #1646, opened 2023-02-20
[16.0][ADD] stock_scrap_location_default

rousseldenis:


stock_scrap_location_default/views/res_config_settings.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="arch" type="xml">
+            <div id="track_product_location" position="inside">
+                <div class="o_setting_left_pane">
+                    <field name="scrap_default_location_id" />
+                </div>
+                <div class="o_setting_right_pane">
+                    <label for="scrap_default_location_id" />
+
+                </div>
+            </div>
```
lmignon: @rousseldenis The UI is ugly an unusable..

```suggestion
            <div id="track_product_location" position="after">
                <div class="col-12 col-lg-6 o_setting_box" id="scrap_default_location">
                    <div class="o_setting_left_pane">
                    </div>
                    <div class="o_setting_right_pane">
                        <label for="scrap_default_location_id"/>
                        <div class="content-group">
                            <field name="scrap_default_location_id"/>
                        </div>
                    </div>
                </div>
            </div>
```

lmignon (changes requested):

rousseldenis:
@lmignon Done

lmignon (approved):

lmignon:
@rousseldenis Can you reopen this one?

rousseldenis:
@jbaudoux

stock_scrap_location_default/models/stock_scrap.py
```diff
@@ -0,0 +1,22 @@
...
+    def _get_default_scrap_location_id(self) -> StockLocation:
+        """
+        Returns the default scrap location defined on company level
+        """
+        if self.env.company.scrap_default_location_id:
+            return self.env.company.scrap_default_location_id
+        return super()._get_default_scrap_location_id()
+
+    # This is necessary as Odoo still uses (for v<=16) static method for default
+    scrap_location_id = fields.Many2one(default=_get_default_scrap_location_id)
```
jbaudoux, line 22: Better to call the class method
```suggestion
    scrap_location_id = fields.Many2one(default=lambda r: r._get_default_scrap_location_id())
```

stock_scrap_location_default/models/stock_scrap.py
```diff
@@ -0,0 +1,22 @@
...
+
+class StockScrap(models.Model):
+
+    _inherit = "stock.scrap"
+
+    def _get_default_scrap_location_id(self) -> StockLocation:
+        """
+        Returns the default scrap location defined on company level
+        """
+        if self.env.company.scrap_default_location_id:
```
jbaudoux, line 17: This is not multi-company friendly as you are not using the company of the record

jbaudoux (commented):
Can you add a roadmap note that this module is not multi-company, nor multi-warehouse friendly. Base module is not either.
It would have been cleaner to set the default location on the warehouse and not on the company.

stock_scrap_location_default/models/stock_scrap.py
rousseldenis: The record at this step is being created (not yet). So, the scrap will be created with the company_id of the Environment. I think this will work in multi-company env.

stock_scrap_location_default/models/stock_scrap.py
rousseldenis: See comment above.

jbaudoux: You are not overriding the standard method as it is not bound to the class. You are overriding the field definition. This doesn't prevent you to call properly the class method.

stock_scrap_location_default/models/stock_scrap.py
rousseldenis: @jbaudoux ROADMAP added

jbaudoux (approved):

PR #1647, opened 2023-02-22
[14.0][FIX] stock_request: Set with cancelled state the stock moves that are cancelled.

victoralmau:
Set with cancelled state the stock moves that are cancelled.

Example use case:

- Create a stock.request (or stock.request.order)
- Confirm.
- Cancel the stock move.
- stock.request and stock.request.order must be set with status cancelled.

@Tecnativa TT41827

### DIFF
--- a/stock_request/models/stock_move.py
+++ b/stock_request/models/stock_move.py
@@ -69,7 +69,7 @@ class StockMove(models.Model):
 
     def _action_cancel(self):
         res = super()._action_cancel()
-        self.mapped("allocation_ids.stock_request_id").check_done()
+        self.mapped("allocation_ids.stock_request_id").check_cancel()
         return res
 
     def _action_done(self, cancel_backorder=False):
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -236,12 +236,19 @@ class StockRequest(models.Model):
     def action_cancel(self):
         self.sudo().mapped("move_ids")._action_cancel()
         self.write({"state": "cancel"})
+        self.mapped("order_id").check_cancel()
         return True
 
     def action_done(self):
         self.write({"state": "done"})
         self.mapped("order_id").check_done()
         return True
+
+    def check_cancel(self):
+        for request in self:
+            if request._check_cancel_allocation():
+                request.write({"state": "cancel"})
+                request.mapped("order_id").check_cancel()
 
     def check_done(self):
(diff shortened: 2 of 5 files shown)


stock_request/models/stock_move.py
```diff
@@ -69,7 +69,7 @@ def copy_data(self, default=None):
 
     def _action_cancel(self):
         res = super()._action_cancel()
-        self.mapped("allocation_ids.stock_request_id").check_done()
+        self.mapped("allocation_ids.stock_request_id").check_cancel()
```
rousseldenis, line 72: @kittiu Could we have your advice here as this is you who introduced this during 14.0 migration ? Thanks

stock_request/models/stock_request.py
```diff
@@ -235,14 +235,23 @@ def action_draft(self):
 
     def action_cancel(self):
         self.sudo().mapped("move_ids")._action_cancel()
-        self.write({"state": "cancel"})
+        self._action_cancel()
         return True
 
+    def _action_cancel(self):
+        self.write({"state": "cancel"})
```
pedrobaeza, line 238: Why an extra method?

stock_request/models/stock_request_order.py
```diff
@@ -236,9 +236,12 @@ def action_draft(self):
 
     def action_cancel(self):
         self.mapped("stock_request_ids").action_cancel()
-        self.write({"state": "cancel"})
+        self._action_cancel()
         return True
 
+    def _action_cancel(self):
```
pedrobaeza: The same. It can be done in the same method and you can overwrite such method.

stock_request/models/stock_request.py
victoralmau: It is not really necessary, but by using the same technique as the `action_done()` method.

stock_request/models/stock_request_order.py
victoralmau: It is not possible because `stock.request.order` does not inherit from `stock.request`.

pedrobaeza: I mean in `action_cancel` inside `stock.request.order`. Don't do `action_cancel` and `_action_cancel`.

victoralmau: Ok, `_action_cancel()` method is already removed.

pedrobaeza (approved):

petrus-v (approved):

PR #1653, opened 2023-03-09
[16.0][ADD] stock_picking_product_interchangeable

geomer198:
This module implements "interchangeable" products. If an interchangeable product is not available in stock in requested quantity its substitute products will be added to picking to complete the order.

Eg there is an "English Breakfast" tea and "Breakfast in England" tea which are absolutely the same inside and have the same price. Only package labels are different. And our customers don't care which of them will be actually delivered.

NB: Interchangeable products substitute each other. Eg if "English Breakfast" can substitutes "Breakfast in England" then "Breakfast in England" can substitute "English Breakfast".

### DIFF
--- a/stock_picking_product_interchangeable/demo/product_demo.xml
+++ b/stock_picking_product_interchangeable/demo/product_demo.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo noupdate="1">
+
+    <record id="res_partner_bob" model="res.partner">
+        <field name="name">Bob</field>
+        <field name="email">bob@example.com</field>
+    </record>
+
+    <record id="product_template_plate" model="product.template">
+        <field name="name">Plate</field>
+        <field name="detailed_type">product</field>
+    </record>
+
+    <record id="product_template_napkin" model="product.template">
+        <field name="name">Napkin</field>
+        <field name="detailed_type">product</field>
+    </record>
+
+    <record
+        id="product_template_napkin_attribute_line"
+        model="product.template.attribute.line"
+    >
+        <field name="product_tmpl_id" ref="product_template_napkin" />
+        <field name="attribute_id" ref="product.product_attribute_2" />
+        <field
+            name="value_ids"
+            eval="[(4, ref('product.product_attribute_value_3')), (4, ref('product.product_attribute_value_4'))]"
+        />
+    </record>
+
+    <record id="product_product_chopsticks" model="product.product">
+        <field name="name">Chopsticks</field>
+        <field name="detailed_type">product</field>
+    </record>
+
+    <record id="product_product_fork" model="product.product">
+        <field name="name">Fork</field>
+        <field name="detailed_type">product</field>
+    </record>
+
+    <record id="product_product_spoon" model="product.product">
+        <field name="name">Spoon</field>
+        <field name="detailed_type">product</field>
+    </record>
+
+    <record id="product_product_knife" model="product.product">
+        <field name="name">Knife</field>
+        <field name="detailed_type">product</field>
+        <field
+            name="product_interchangeable_ids"
+            eval="[(4,ref('product_product_fork')), (4, ref('product_product_spoon'))]"
+        />
+    </record>
+
+</odoo>
--- a/stock_picking_product_interchangeable/models/product.py
+++ b/stock_picking_product_interchangeable/models/product.py
@@ -1,0 +1,66 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    product_tmpl_interchangeable_ids = fields.Many2many(
+        comodel_name="product.product",
+        compute="_compute_product_tmpl_interchangeable_ids",
+        inverse="_inverse_product_tmpl_interchangeable_ids",
+    )
+
+    def _compute_product_tmpl_interchangeable_ids(self):
+        """Compute interchangeable products"""
+        for rec in self:
+            rec.product_tmpl_interchangeable_ids = (
+                rec.product_variant_ids.product_interchangeable_ids
+            )
+
+    def _inverse_product_tmpl_interchangeable_ids(self):
+        """Set new interchangeable product"""
+        for rec in self:
+            rec.product_variant_id.product_replaces_ids = (
+                rec.product_tmpl_interchangeable_ids
+            )
+
+
+class Product(models.Model):
+    _inherit = "product.product"
+
+    product_interchangeable_ids = fields.Many2many(
+        comodel_name="product.product",
+        string="Replaces",
+        help="Products that can be substituted by current product",
+        inverse="_inverse_product_interchangeable_ids",
+        compute="_compute_product_interchangeable_ids",
+    )
+
+    product_replaces_ids = fields.Many2many(
+        comodel_name="product.product",
+        string="Replaces",
+        relation="product_substitute_rel",
+        column1="product_id",
+        column2="product_replaced_id",
+        help="Products that can be substituted by current product",
+    )
+    product_replaced_by_ids = fields.Many2many(
+        comodel_name="product.product",
+        string="Replaces",
+        relation="product_substitute_rel",
+        column1="product_replaced_id",
+        column2="product_id",
+        help="Products that can substitute current current product",
+    )
+
+    def _compute_product_interchangeable_ids(self):
+        """Compute interchangeable products"""
+        for rec in self:
+            rec.product_interchangeable_ids = (
+                rec.product_replaces_ids | rec.product_replaced_by_ids
+            ) - rec
+
+    def _inverse_product_interchangeable_ids(self):
+        """Set new interchangeable product"""
+        for rec in self:
+            rec.product_replaces_ids = rec.product_interchangeable_ids
--- a/stock_picking_product_interchangeable/models/stock_move.py
+++ b/stock_picking_product_interchangeable/models/stock_move.py
@@ -1,0 +1,108 @@
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _prepare_interchangeable_products(self, mode, for_qty):
+        """
+        Preparing products for current product replacement
+        :param mode: stock.picking.type substitute_products_mode value
+        :param for_qty: product qty for replacement
+        :return: list of tuple [(product_obj, count), ...]
+        """
+        qty = for_qty
+        if qty >= 0:
+            return False
+        products_info = []
+        for product in self.product_id.product_interchangeable_ids:
+            available_qty = product.immediately_usable_qty
+            if available_qty > 0 > qty:
+                product_qty = abs(qty) if available_qty + qty >= 0 else available_qty
+                products_info.append((product, product_qty))
+                qty += product_qty
+        if (mode == "all" and qty == 0) or mode == "any":
+            return products_info
+        return False
+
+    def _create_stock_move_interchangeable_products(self, products_info):
+        """
+        Creates stock.move records for replacement product
+        :param products_info: struct list of tuple [(product_obj, count), ...]
+        :return: Stock Move object
+        """
+        stock_move_obj = self.env["stock.move"]
+        if not products_info:
+            return stock_move_obj
+        return stock_move_obj.create(
+            [
+                {
+                    "picking_id": self.picking_id.id,
+                    "name": product.display_name,
+                    "product_id": product.id,
+                    "product_uom_qty": qty,
+                    "location_id": self.location_id.id,
+                    "location_dest_id": self.location_dest_id.id,
+                    "company_id": self.company_id.id,
+                }
+                for product, qty in products_info
+            ]
+        )
+
+    def _interchangeable_stock_move_filter(self):
+        """
+        Filter for applying interchangeable behavior for stock.move
+        :return: True/False
+        """
+        type_ = self.picking_type_id
+        mode = type_.substitute_products_mode
+        skip_behavior = not (mode and type_.code == "outgoing")
+        return not (skip_behavior or self.picking_id.pass_interchangeable)
+
+    def _add_note_interchangeable_picking_note(self, products_info, qty):
+        """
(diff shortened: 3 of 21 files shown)


rousseldenis:
@geomer198 thanks for this. Have you explored https://github.com/OCA/e-commerce/tree/14.0/product_template_multi_link as base module for the flow you want to implement.

As I suggest to use a new link 'Replaces' between products.

ivs-cetmix:
> @geomer198 thanks for this. Have you explored https://github.com/OCA/e-commerce/tree/14.0/product_template_multi_link as base module for the flow you want to implement.
> 
> As I suggest to use a new link 'Replaces' between products.

@rousseldenis thank you for the feedback! Yes we have checked that module. However our module is more “low level” because it targets not only e-commerce but every operation where products are involved. Eg Sales Orders, POS Orders or even bare pickings.
Eg if we are preparing an internal transfer from our stock to one of our shops.

rousseldenis:
> > @geomer198 thanks for this. Have you explored https://github.com/OCA/e-commerce/tree/14.0/product_template_multi_link as base module for the flow you want to implement.
> > As I suggest to use a new link 'Replaces' between products.
> 
> @rousseldenis thank you for the feedback! Yes we have checked that module. However our module is more “low level” because it targets not only e-commerce but every operation where products are involved. Eg Sales Orders, POS Orders or even bare pickings. Eg if we are preparing an internal transfer from our stock to one of our shops.

Yes, but you should not be feared of the repo's name.

It has no dependency on e-commerce feature. The module is really done as a base module.
IMHO you should use it (I know your use case as I have the same - in an old project).

The drawback of that module is the dependency on sale module, but I think this was only to have a menu entry point. So, that can be improved.

ivs-cetmix:
> > > @geomer198 thanks for this. Have you explored https://github.com/OCA/e-commerce/tree/14.0/product_template_multi_link as base module for the flow you want to implement.
> > > As I suggest to use a new link 'Replaces' between products.
> > 
> > 
> > @rousseldenis thank you for the feedback! Yes we have checked that module. However our module is more “low level” because it targets not only e-commerce but every operation where products are involved. Eg Sales Orders, POS Orders or even bare pickings. Eg if we are preparing an internal transfer from our stock to one of our shops.
> 
> Yes, but you should not be feared of the repo's name.
> 
> It has no dependency on e-commerce feature. The module is really done as a base module. IMHO you should use it (I know your use case as I have the same - in an old project).
> 
> The drawback of that module is the dependency on sale module, but I think this was only to have a menu entry point. So, that can be improved.

Yup I totally agree with you when it comes to extending existing modules instead of developing a new one. And we are always trying to follow that way.
However those modules are too different:
1. This module works at the "stock.move" level.
2. It works with product variants not with product template. Because sometimes you can replace a product variant with another variant of the sae template.

What can be reasonable to do in the future is to make a cross module that might add a new link type and inherit from both of these modules.

rousseldenis:
> > > > @geomer198 thanks for this. Have you explored https://github.com/OCA/e-commerce/tree/14.0/product_template_multi_link as base module for the flow you want to implement.
> > > > As I suggest to use a new link 'Replaces' between products.
> > > 
> > > 
> > > @rousseldenis thank you for the feedback! Yes we have checked that module. However our module is more “low level” because it targets not only e-commerce but every operation where products are involved. Eg Sales Orders, POS Orders or even bare pickings. Eg if we are preparing an internal transfer from our stock to one of our shops.
> > 
> > 
> > Yes, but you should not be feared of the repo's name.
> > It has no dependency on e-commerce feature. The module is really done as a base module. IMHO you should use it (I know your use case as I have the same - in an old project).
> > The drawback of that module is the dependency on sale module, but I think this was only to have a menu entry point. So, that can be improved.
> 
> Yup I totally agree with you when it comes to extending existing modules instead of developing a new one. And we are always trying to follow that way. However those modules are too different:
> 
> 1. This module works at the "stock.move" level.
> 2. It works with product variants not with product template. Because sometimes you can replace a product variant with another variant of the sae template.
> 
> What can be reasonable to do in the future is to make a cross module that might add a new link type and inherit from both of these modules.

In fact, its little brother exists : https://github.com/OCA/e-commerce/tree/14.0/product_variant_multi_link

But, nevertheless, let's move forward on this. We'll see how to integrate that later.

ivs-cetmix:
> 
> In fact, its little brother exists : https://github.com/OCA/e-commerce/tree/14.0/product_variant_multi_link
> 
> But, nevertheless, let's move forward on this. We'll see how to integrate that later.

Great, thank you! Would appreciate your code review as well)

stock_picking_product_interchangeable/demo/product_demo.xml
aayartsev, line 1: https://github.com/OCA/stock-logistics-warehouse/blob/0318408308d42eee9b7e9aada449f11da0cc11d3/stock_picking_product_interchangeable/demo/product_demo.xml#L22

https://github.com/OCA/stock-logistics-warehouse/blob/0318408308d42eee9b7e9aada449f11da0cc11d3/stock_picking_product_interchangeable/demo/product_demo.xml#L46

I think it is better to use (4, id, 0) triplet, because if i will change this field over UI and then update module, i will lost my data.

stock_picking_product_interchangeable/models/product.py
aayartsev, line 1: https://github.com/OCA/stock-logistics-warehouse/blob/0318408308d42eee9b7e9aada449f11da0cc11d3/stock_picking_product_interchangeable/models/stock_move.py#L41
May be it is better for performance to create list with values and call create onсe?

aayartsev (commented):
Watch my notes. Another parts for me is good.

stock_picking_product_interchangeable/demo/product_demo.xml
geomer198: Resolved

stock_picking_product_interchangeable/models/product.py
geomer198: Resolved

aayartsev (approved):
It is good.

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+        type_ = moves.picking_type_id
+        mode = type_.substitute_products_mode
+        skip_behavior = not (mode and type_.code == "outgoing")
+        if skip_behavior or moves.picking_id.pass_interchangeable:
+            return moves
+        move_ids = moves.filtered(lambda m: m.product_id.product_interchangeable_ids)
+        new_moves = self.env["stock.move"]
+        for move in move_ids:
+            qty = move.product_id.immediately_usable_qty
+            products, products_qty = move._prepare_interchangeable_products(mode, qty)
```
GabbasovDinar: can we calc `interchangeable_products_qty` as sum(product_info[1] for product_info in products_info) and thereby simplify the method `_prepare_interchangeable_products` it seems that this method, in addition to preparing data for a particular product, should not do summation, which is used only in one place, what do you think?

GabbasovDinar: I suggest renaming the variable that assigns the value to the `_prepare_interchangeable_products` method, since it does not return a recordset, but information about the product and its quantity. for example:
`products_info = move._prepare_interchangeable_products(mode, qty)`

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+                }
+                for product, qty in products
+            ]
+        )
+
+    def _action_confirm(self, merge=True, merge_into=False):
+        moves = super(StockMove, self)._action_confirm(
+            merge=merge, merge_into=merge_into
+        )
+        type_ = moves.picking_type_id
```
GabbasovDinar: What happens if moves with different pickings are confirmed?

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+                    "name": product.display_name,
+                    "product_id": product.id,
+                    "product_uom_qty": qty,
+                    "location_id": self.location_id.id,
+                    "location_dest_id": self.location_dest_id.id,
+                    "company_id": self.company_id.id,
+                }
+                for product, qty in products
+            ]
+        )
```
GabbasovDinar:
```suggestion
    def _create_stock_move_interchangeable_products(self, products_info):
        """
        Creates stock.move records for replacement product
        :param products_info: struct list of tuple [(product_obj, count), ...]
        :return: Stock Move object
        """
        stock_move_obj = self.env["stock.move"]
        if not products_info:
            return stock_move_obj
        return stock_move_obj.create(
            [
                {
                    "picking_id": self.picking_id.id,
                    "name": product.display_name,
                    "product_id": product.id,
                    "product_uom_qty": qty,
                    "location_id": self.location_id.id,
                    "location_dest_id": self.location_dest_id.id,
                    "company_id": self.company_id.id,
                }
                for product, qty in products_info
            ]
        )
```

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+            return False, False
+        products = []
+        interchangeable_products_qty = 0
+        for product in self.product_id.product_interchangeable_ids:
+            available_qty = product.immediately_usable_qty
+            if available_qty > 0 > qty:
+                product_qty = abs(qty) if available_qty + qty >= 0 else available_qty
+                products.append((product, product_qty))
+                qty += product_qty
+                interchangeable_products_qty += product_qty
```
GabbasovDinar: I think you can move the calculation of `interchangeable_products_qty` from this method

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+        if (mode == "all" and qty == 0) or mode == "any":
+            return products, interchangeable_products_qty
```
GabbasovDinar: and rename products to products_info, because it is not products but a tuple with data for each product

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _prepare_interchangeable_products(self, mode, for_qty):
+        """
+        Preparing products for current product replacement
+        :param move_id: stock.move record
```
GabbasovDinar: there is no such parameter

GabbasovDinar (changes requested):
Thanks for the PR, please check a few of my comments

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

stock_picking_product_interchangeable/models/stock_move.py
```diff
@@ -50,26 +45,64 @@ def _create_stock_move_interchangeable_products(self, products):
                     "location_dest_id": self.location_dest_id.id,
                     "company_id": self.company_id.id,
                 }
-                for product, qty in products
+                for product, qty in products_info
             ]
         )
 
+    @api.model
```
GabbasovDinar: why? self -> stock.move(), move -> stock.move(1)

you can use self intead of move variable

GabbasovDinar (changes requested):

stock_picking_product_interchangeable/models/stock_move.py
geomer198: Resolved

aayartsev (approved):

GabbasovDinar (approved):
Not tested, code LGTM

Aldeigja (approved):
Functional OK

Aldeigja:
@OCA/logistics-maintainers can you take a look?

simahawk:
As @rousseldenis suggests this can be done w/ product links. We did it at sale level by adding a new type "replacement" and by replacing the product in the SO directly. This way you don't have to deal w/ the rest of the stack and if you sell products the SO will reflect this change for the customer (we still have to rework this in a reusable module).
I think you can integrate this later so I'm not blocking but I'm curious to know more about your use case. 
I guess you don't care about sales, right?

@rousseldenis regarding the link modules: we should probably move them out of e-commerce repo.

rousseldenis:
> @rousseldenis regarding the link modules: we should probably move them out of e-commerce repo.

@simahawk Of course

ivs-cetmix:
> As @rousseldenis suggests this can be done w/ product links. We did it at sale level by adding a new type "replacement" and by replacing the product in the SO directly. This way you don't have to deal w/ the rest of the stack and if you sell products the SO will reflect this change for the customer (we still have to rework this in a reusable module). I think you can integrate this later so I'm not blocking but I'm curious to know more about your use case. I guess you don't care about sales, right?
> 
> @rousseldenis regarding the link modules: we should probably move them out of e-commerce repo.

Yes you are right. As I mentioned above in our use case we are replacing products at a stock level. Not at the sales one.
Eg you have same products one of them is has a red package and another one is packed in a green box. 
So when a customer orders eg 10 of "red" but we have only 6 of them we can just pick another 4 of green and complete the order. There will be still 10 of red on both SO and invoice but in our case it's not an issue.

PR #1656, opened 2023-03-16
[13.0][IMP] stock_demand_estimate: enhance views

LoisRForgeFlow:
* Add a graph view.
* Add handy filters: (not) expired; group by date, location and/or product.
* Relabel `product_qty` as "Real Quantity" was really confusing.

@ForgeFlow

PR #1659, opened 2023-03-17
[14.0][FW-PORT] stock_demand_estimate_matrix: remove action extension

BernatPForgeFlow:
no need to modify action, once the tree view is set as editable it will work as itended, no need to remove form from the view modes.

Origin: https://github.com/OCA/stock-logistics-warehouse/pull/1656


LoisRForgeFlow (approved):
Thanks!

PR #1661, opened 2023-03-20
[16.0][MIG] stock_inventory_revaluation_adjustment

Murtaza-OSI:


dreispt (commented):
@Murtaza-OSI Where is the commit history?

PR #1665, opened 2023-03-22
]16.0][MIG] stock_archive_constraint

Urvisha-OSI:


rousseldenis:
@Urvisha-OSI Thanks for this. Could you put pre-commit modifications in a separate commit as explained there : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0 ?

dreispt:
Ping @Urvisha-OSI

Bart-dh:
@Urvisha-OSI ?

dreispt (approved):

stock_archive_constraint/models/product_template.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.constrains("active")
+    def _check_active_stock_archive_constraint(self):
```
sebastienbeau, line 11: Looking at the code. It seem that this code is dead from a long time.
- the method : _check_active_stock_archive_constraint does not exist on product.product
- the file is not imported in the __init__.py

I think this file can be dropped 


Note: it was already the case in V14.
Note 2: It will be interesting to check what happen when we archive the product template.

Bart-dh:
@Urvisha-OSI

PR #1666, opened 2023-03-22
]16.0][MIG] stock_inventory

Urvisha-OSI:


rousseldenis:
@Urvisha-OSI Thanks for this.

Could you put pre-commit modifications (like setup) in a separate commit as explained here https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0 ?

stock_inventory/tests/test_stock_inventory.py
```diff
@@ -55,14 +55,14 @@ def setUp(self):
...
-                "warehouse_id": self.location_src.id,
+                "location_id": self.location_src.id,
             }
         )
         self.location2 = self.location_model.create(
             {
                 "name": "Location 2",
                 "usage": "internal",
-                "location_id": self.location_src.id,
+                "location_id": self.location_dst.id,
```
rousseldenis, line 65: Why changing location 2 (internal) to a customer location ???

rousseldenis: @Urvisha-OSI

AlexPForgeFlow:
Hi @Urvisha-OSI, when doing the #1849, tests failed because in stock_inventory is trying to modify a field that is not allowed by odoo base. In the stock_quant.py file of your module I would add:
     def _get_inventory_fields_write(self):
         return super()._get_inventory_fields_write() + ["to_do"]
To make sure that you can modify this field, I have test it locally and it works you can see the changes for v15 https://github.com/OCA/stock-logistics-warehouse/pull/1848/commits/be501a4600e5671fad9b2e724e68960e6af39713.

Also I did a fix in v15, see the commit (https://github.com/ForgeFlow/stock-logistics-warehouse/commit/264bc7c93f128339f72db834df9d59146e4a926f) also it is present in the #1848 , can you add the changes?

rousseldenis:
@Urvisha-OSI I will supersedes this with fixes by @AlexPForgeFlow

PR #1667, opened 2023-03-22
[16.0][MIG] stock_inventory_exclude_sublocation

Urvisha-OSI:


rousseldenis:
/ocabot migration stock_inventory_exclude_sublocation

@Urvisha-OSI Thanks for this. Could you include the PR dependency in your PR description and embed it following : https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

This could help review. Thanks

dhara-wibtec:
Hi @Urvisha-OSI , I've gone through your development for the version-16.0, based on which the boolean "exclude_sublocation" is not used anywhere in the view because the view is incorrectly inherited as that view is not present in the version 16.

Adding further, the methods that you have used in the python file is not present in the "odoo addons", Could you just let me know the source of those functions so that I can use it in my environment, Correctly !
Also you can provide a brief knowledge that how you maintained to develop the logic.

Thanks in advance. Your help will be appreciated.

PR #1676, opened 2023-03-24
[16.0][MIG] stock_inventory_line_product_cost

Urvisha-OSI:


stock_inventory_line_product_cost/README.rst
```diff
@@ -19,3 +19,4 @@ Contributors
 ------------
 
 * Open Source Integrators <contact@opensourceintegrators.com>
+* Urvisha Desai <udesai@opensourceintegrators.com>
```
zamberjo, line 22: Do not modify the Readme directly, as it is self-generated, but create the README folder.
SEE: https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

zamberjo (commented):
Technical review seems ok, can you look at the README and do a rebase to get the runboat build done so someone can review it functionally?

stock_inventory_line_product_cost/README.rst
rousseldenis: @zamberjo That's only the case when readme fragments are available. This is not the case

PR #1677, opened 2023-03-24
[16.0][MIG] stock_location_empty

Urvisha-OSI:


HaraldPanten:
@rousseldenis @LoisRForgeFlow @pedrobaeza isn't that module needed for V15?

In V14 it was included in core, but it changed in V15 (as we don't have stock inventory).

We can do it for V15.

WDYT?

PR #1681, opened 2023-03-30
[16.0][ADD] stock_picking_commercial_partner

lmignon:
When you organize the work into your warehouse you could need to search for pickings related to a specific commercial entity. This use case could be motivated by the fact that pickings for the same commercial entity are usually delivered together.

This addon add on the picking form the 'Commercial Entity' field the related partner is part of. This field is readonly and comes from the partner information. It also allows you to quickly search for pickings related to a specific commercial entity and to group by this field in the picking tree view.


rousseldenis (approved):
Code and functional review

jbaudoux (approved):
LG

PR #1684, opened 2023-04-07
[14.0][FIX] product_route_profile: Wrong inverse profile id

bosd:
- fix _inverse_route_ids to store the selected route_profile_id


bosd:
@awisky Can you check this

bosd:
superseeded by #1688

awisky:
@bosd I checked it, it has the same issue we are trying to solve.
I think I found a workaround, as we have the issue during the creation of the product, I avoid the inverse method with a flag. 
I couldn't reopen the PR so I created another here, which looks fine
https://github.com/OCA/stock-logistics-warehouse/pull/1689

PR #1686, opened 2023-04-10
[16.0][ADD] stock_valuation_layer_accounting_date

kanda999:

@qrtl


AungKoKoLin1997 (approved):

rinaldifirdaus (approved):

rousseldenis (approved):
LGTM

rousseldenis:
@kanda999 Could you rebase ?

kanda999:
@rousseldenis 
I rebased it but there seems to be an error somewhere unrelated to this PR. I will rebase again when the error is fixed.

PR #1690, opened 2023-04-14
[15.0][ADD] stock_request: Allow to define if a route can be selected in a stock request line

alan196:
Add new feature to allow manual selection of stock location routes on stock requests. This is particularly useful for companies that do not want to define a default stock for any route and need to manually select the routes they require. This feature includes the addition of a new field, "stock_request_selectable," to the stock.location.route model and a view change to allow users to toggle this option on the form view.


stock_request/models/stock_request_abstract.py
```diff
@@ -125,7 +125,7 @@ def _compute_route_ids(self):
                 )
                 routes_by_warehouse[warehouse.id] |= route
         for record in self:
-            routes = route_obj
+            routes = route_obj.search([("stock_request_selectable", "=", True)])
```
rousseldenis, line 128: @alan196 

As the query will be executed each time for each record and as the domain is trictly equivalent, put that search outside the loop.

Moreover I would have changed the code here under to manipulate a set of ids for performances reasons, then browse the ids when doing the filtered.

rousseldenis (changes requested):
At least first change.

@alan196 Could you add a test case that will take into account the new field ?

PR #1695, opened 2023-04-21
[MIG] stock_reserve_sale: Migration to 16.0

XanderDeJaegere:
[[MIG]stock_reserve_sale: migrate from v13.0 to v16.0](https://github.com/OCA/stock-logistics-warehouse/pull/1695/commits/4ff4e11ca24bf32bbf95a92517a175b7fa0f112a)

PR #1486, opened 2022-09-15
[MIG] stock_reserve_sale 15.0

flachica:
Dependencies:

https://github.com/OCA/stock-logistics-warehouse/pull/1485

PR #1347, opened 2022-01-10
[14.0][MIG] stock_reserve_sale

mt-software-de:
This is the migration for stock_reserve_sale
there is an other migration for stock_reserve https://github.com/OCA/stock-logistics-warehouse/pull/1346

PR #1696, opened 2023-04-25
[14.0] mig stock_reserve_sale

rvalyi:
replaces https://github.com/OCA/stock-logistics-warehouse/pull/1347 
the history is preserved and I re-built the migration commit from @mt-software-de (preserving his work) and bugfix (specially https://github.com/OCA/stock-logistics-warehouse/pull/1347#pullrequestreview-1005776726 ) that is not found in the v15 (https://github.com/OCA/stock-logistics-warehouse/pull/1486) and v16 (https://github.com/OCA/stock-logistics-warehouse/pull/1695) migration PRs.

cc @gdgellatly @rousseldenis @victoralmau @XanderDeJaegere @flachica


stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+    )
+
+    def release_all_stock_reservation(self):
+        line_ids = [line.id for order in self for line in order.order_line]
+        lines = self.env["sale.order.line"].browse(line_ids)
+        lines.release_stock_reservation()
+        return True
+
+    def action_reserve_all_lines(self):
+        RESERVE = self.env["sale.stock.reserve"]
```
hparfr, line 51:
```suggestion
        Reserve = self.env["sale.stock.reserve"]
```

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+        for rec in self.filtered(lambda s: s.is_stock_reservable):
+            reserve = RESERVE.with_context(
```
hparfr, line 53:
```suggestion
            reserve = Reserve.with_context(
```

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+        compute="_compute_stock_reservation",
+        readonly=True,
+        store=True,
+        string="Can Have Stock Reservations",
+    )
+
+    def release_all_stock_reservation(self):
+        line_ids = [line.id for order in self for line in order.order_line]
+        lines = self.env["sale.order.line"].browse(line_ids)
+        lines.release_stock_reservation()
```
hparfr, line 47:
```suggestion
        self.order_line.release_stock_reservation()
```

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+            wh_routes = warehouse.route_ids
+            wh_route_ids = [route.id for route in wh_routes]
+            domain = [
+                "|",
+                ("warehouse_id", "=", warehouse.id),
+                ("warehouse_id", "=", False),
+                ("route_id", "in", wh_route_ids),
+            ]
+
+            rules = StockRule.search(domain, order="route_sequence, sequence")
```
hparfr, line 149:
```suggestion
            rules = StockRule.search(domain, order="route_sequence, sequence", limit=1)
```

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if rules:
+            fields.first(rules)
+        return False
```
hparfr, line 153:
```suggestion
        return rules
```

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+                    )
+                    % (order.name)
+                )
+        return super().unlink()
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_line_rule(self):
```
hparfr, line 122: looks like dead code, it always return False

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+    @api.depends("state", "product_id.route_ids", "product_id.type")
+    def _compute_is_stock_reservable(self):
+        for line in self:
+            reservable = False
+            if (
+                not (
+                    line.state not in ("draft", "sent")
+                    or line._get_procure_method() == "make_to_order"
+                    or not line.product_id
+                    or line.product_id.type == "service"
```
hparfr, line 171: not ...or  .. or not... a bit hard to read

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+    )
+    is_stock_reservable = fields.Boolean(
+        compute="_compute_is_stock_reservable", readonly=True, string="Can be reserved"
+    )
+    is_readonly = fields.Boolean(compute="_compute_is_readonly", store=False)
+
+    def release_stock_reservation(self):
+        reserv_ids = [reserv.id for line in self for reserv in line.reservation_ids]
+        reservations = self.env["stock.reservation"].browse(reserv_ids)
+        reservations.release_reserve()
```
hparfr, line 196:
```suggestion
        self.reservation_ids.release_reserve()
```

hparfr (commented):
Code read, not tested.

`    def _get_line_rule(self):` I may miss something but it looks like dead code.

Some minor remarks about style.

stock_reserve_sale/models/sale.py
gdgellatly: > 

@hparfr Are you sure? We run similar and we were sure we needed a rebrowse here.

hparfr: Not; I didn't run the code.

Did you investigate why you need to rebrowse here ? Does it need some cache invalidation somewhere else ?

gdgellatly: Not overly we only run this in v13 and tried and we got odd results, it just seemed strange to assert that a rebrowse wasn't needed when we always had one before, it isn't introduced. Though you must have inside info

hparfr: This code doesn't look like v14 code. 

As you know, back in the days, the ORM was not as performant as now and chained expressions were not possible. We had to write *for in a for* (like the hardtoread double `for` in the comprehension list above) or `mapped()` chain ( *somerecord.mapped('somefield_id').mapped('someother_ids')...*) it's not needed anymore.

In theory, we can expect the same result with both version of the code.

gdgellatly: I disagree. Both with the example and the conclusion. This isn't special to v14, or even as far back as v6 you could use that syntax albeit with a few extra args in the function. I think where you have a migration, in review it is especially important to apply the principles of Chesterton's fence. This module was written by an accomplished dev, it went through multiple years of reviews, it deliberately included a rebrowse, and until we know why that rebrowse was there, and whether it is still required, it would seem extremely risky to say "In theory they would do the same" because we know they don't always, hence why sometimes we rebrowse.

hparfr: There is no advance in this discussion. 

I *suggest* to rewrite this part in more recent style to improve the code quality: readability, aligned with other part of the v14 code, and performance.

You says in v13 the rebrowse was necessary. I trust you for the v13 code. I did not run it.

But since we both did not run this code in v14, we can't decide if the rebrowse is still really necessary anymore.

gdgellatly: I'm not saying you are wrong, but I know for me personally, if I see a rebrowse, that is an unusual thing to do in any version. I would want to know why it was there in the first place before suggesting it be removed. I even tried to find out, and that isn't easy with the way we port commits and that it isn't noted anywhere.  However, I just saw, we are using #1347 in a v14 production instance, which does have the rebrowse, for some months now. And still I don't know if it is necessary, but I know that it works and is subtly different to the suggestion.  We have some really bad instances lately of OCA modules breaking on seemingly minor changes and refactoring.

mourad-ehm:
Hi @rvalyi, Thanks for migration. 
Could you accept commit suggestions of @hparfr ?

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+        string="Has Stock Reservations",
+    )
+    is_stock_reservable = fields.Boolean(
+        compute="_compute_stock_reservation",
+        readonly=True,
+        store=True,
+        string="Can Have Stock Reservations",
+    )
+
+    def release_all_stock_reservation(self):
```
hugo-cordoba, line 44: Good morning. 

I have read the conversation between @gdgellatly and @hparfr, and I decided to make a functionally test for this PR and debug the function 'release_stock_reservation' to see how the rebrowse affects. The conclusion is the following:


Captura de pantalla de 2023-06-06 12-30-10
Captura de pantalla de 2023-06-06 12-33-14

in both ways the same result is obtained, therefore, I see more coherent to leave the function only with the line `'self.reservation_ids.release_reserve()'` instead of making the rebrowse.

The rest of the code, except for the improvements proposed by @hparfr in the previous comments, I see it well both functional and coded.

hugo-cordoba:
What is your opinion about that @rvalyi? Do you plan to modify it?

hugo-cordoba:
Hi @gdgellatly @rvalyi @hparfr, can you take a look at the comment above? Let's see if together, we can move forward on this issue and push the v16 migration.

rvalyi:
hello, I'll make my best to fix this till the end of the week.

hugo-cordoba:
@rvalyi perfect! thank you

hugo-cordoba:
Hi @rvalyi  ! Are you going to go ahead with this? Can I develop the migration to version 16 and when you have this I will add the improvements?

jbaudoux:
This module was initially a POC we did 10 years ago on v7.0 at Camptocamp. After all those years and looking back at this module, I don't understand anymore why it's not behaving the same way than a SO confirmation placing a delivery move(s) through a procurement run. 

With @mt-software-de and Camptocamp, we did this new approach https://github.com/OCA/stock-logistics-warehouse/pull/1700 . I find it simpler as you don't need this stock.reserve model, it supports bom kits, respects stock route configuration, and has a better compatibility with stock like the virtual reservations as a real move is placed (cf https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release)

I would be interested to know why you would prefer to stick to this module instead of going to the new proposed approach?

gdgellatly:
> This module was initially a POC we did 10 years ago on v7.0 at Camptocamp. After all those years and looking back at this module, I don't understand anymore why it's not behaving the same way than a SO confirmation placing a delivery move(s) through a procurement run.
> 
> With @mt-software-de and Camptocamp, we did this new approach #1700 . I find it simpler as you don't need this stock.reserve model, it supports bom kits, respects stock route configuration, and has a better compatibility with stock like the virtual reservations as a real move is placed (cf https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release)
> 
> I would be interested to know why you would prefer to stick to this module instead of going to the new proposed approach?

The primary difference is that this is an extension module to stock_reserve, so while the linked module will solve the sale case, it does not solve the general reserve case AFAICT.

stock_reserve_sale/models/sale.py
```diff
@@ -0,0 +1,251 @@
...
+                "|",
+                ("warehouse_id", "=", warehouse.id),
+                ("warehouse_id", "=", False),
+                ("route_id", "in", wh_route_ids),
+            ]
+
+            rules = StockRule.search(domain, order="route_sequence, sequence")
+
+        if rules:
+            fields.first(rules)
```
rousseldenis, line 152:
```suggestion
            return fields.first(rules)
```

rousseldenis:
@rvalyi What's the status of this ?

rvalyi:
> @rvalyi What's the status of this ?

hehe that's the project Dual... we passed to you, so I guess now they pay you do deal with it eventually... They were hardly paying us to take care of all this, so after loosing a lot of time with them I personally didn't fight to keep this customer...

PR #1698, opened 2023-04-25
[IMP] stock_request: Use lead dates

etobella:
Without the change, when you do a purchase, the planned date is, by default, the current date, so it tries to create Purchase Orders with a previous date.
With the change, it reviews the lead times of the product in order to get the bigger date.

Follows the same logic used on odoo
https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_orderpoint.py#L153-L161

@LoisRForgeFlow


sanchonuria (approved):
LGTM!

LoisRForgeFlow (approved):
Makes sense, code review :+1:

kluna1998 (approved):

PR #1700, opened 2023-04-26
[14.0][ADD] stock reserve sale with outgoing moves

mt-software-de:
This addon makes it possible to reserve a sale order, before it is confirmed.
It uses therefore a procurement.run


mt-software-de:
cc @jbaudoux 
I need a better name for this addon, do you have an idea?

jbaudoux:
As it is a feature offered in sales, I would place it in https://github.com/OCA/sale-workflow and call it `sale_stock_prebook`

mt-software-de:
This PR will be closed and moved to sale-workflow
https://github.com/OCA/sale-workflow/pull/2624

PR #1702, opened 2023-05-05
[15.0][FIX] stock_move_location: No read on actions

pedrobaeza:
The permissions on actions has been restricted on this version. Use _for_xml_id helper that was built for not requiring such permissions.

@Tecnativa TT42955


pedrobaeza:
This can be fast-tracked as the regular leftover on migrations:

/ocabot merge patch

PR #1528, opened 2022-10-29
[16.0][MIG] stock_quant_manual_assign

alexis-via:
Fix bug #1527  when using a unit of measure on stock.move.line which is different from product's unit of measure
Fix bug #1460 set "Others reserved" as readonly

PR #1707, opened 2023-05-16
[16.0][MIG] `stock_quant_manual_assign`

yankinmax:
Open this PR to fast-track questions and comments asked on original migration PR created by @alexis-via here:
- [ ] https://github.com/OCA/stock-logistics-warehouse/pull/1528


yankinmax:
@rousseldenis @sebalix I opened this PR and considered your suggestions:
1. Bugs are splitted to separate commits.
2. We don't need to run tests with `@tagged("post_install", "-at_install")`

@pedrobaeza can this PR replace the one opened by @alexis-via to finally merge it?

jcadhoc:
Hello @yankinmax !

I`m tryng to use this module in v16 but i have the following error: https://watch.screencastify.com/v/GBE9JageLOKSYH3EOqGX

Can you check it please? In the build is not working.

yankinmax:
hey @jcadhoc I'll check, thanks

jcadhoc:
@yankinmax Thank you!!

stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -128,7 +139,6 @@ class AssignManualQuantsLines(models.TransientModel):
     assign_wizard = fields.Many2one(
         comodel_name="assign.manual.quants",
         string="Move",
-        required=True,
```
yankinmax: we don't need here `required` as it forces an error when `_onchange_selected` is played
the field `assign_wizard` will be anyway propagated with the value, because one2many creation is handled

yankinmax:
@jcadhoc issue is fixed, you can check and pls submit your approve in case it's ok

stock_quant_manual_assign/wizard/assign_manual_quants.py
pedrobaeza: It should be required, as we should assure DB integrity on one2many tables, so any other problem is because something is incorrect in other place. Maybe the problem is to have such field in the child model definition.

yankinmax: @pedrobaeza thanks for quick response on this
Actually, I can't reproduce the problem locally with demo database.
What I definitely checked is the fact `assign_wizard` is filled with the value when wizard is created.
But, it's empty before `assign_quants` (Confirm button) is called, so it forces an error on runboat build.
Locally it's not the case and works properly.
Then, my guess to fix the issue with `Command.create`, just refactors the one2many creation.
So, I have no idea for the moment what can cause a problem.

rousseldenis (approved):
Code review.

@yankinmax Could you improve your commit messages including module name in them ([IMP] stock_quant_manual_assign: (...)) ? Thanks

stock_quant_manual_assign/wizard/assign_manual_quants.py
yankinmax: @jcadhoc I've checked different implementations of one2many lines for a wizard.
The code is definitely ok.
IDK why there is an issue with the `assign_wizard` field being checked for required when onchange `selected` called.
It should be checked on creation.

SilvioC2C (approved):
LGTM overall

ivantodorovich (approved):
LG

stock_quant_manual_assign/wizard/assign_manual_quants.py
BT-jmichaud: @yankinmax I tested the version merged in 16.0 and got the same error:
https://github.com/OCA/stock-logistics-warehouse/issues/1736

yankinmax: @BT-jmichaud I've tried functional testing several times on real database and demo database. Locally I don't have such issue. On the runbot it exists. I have no idea why. If you know where the problem comes from please open a PR to fix that.
The way one2many field on the wizard is filled with data is ok.

PR #1710, opened 2023-05-18
[14.0][IMP] stock_request_purchase: Propagate the cancellation of related purchases

victoralmau:
Propagate the cancellation of related purchases

@Tecnativa TT43414


stock_request_purchase/models/stock_request.py
```diff
@@ -38,6 +39,13 @@ def _check_purchase_company_constrains(self):
                 )
             )
 
+    def action_cancel(self):
+        res = super().action_cancel()
```
pedrobaeza: Remember to put docstrings:

```suggestion
    def action_cancel(self):
        """Propagate the cancellation to the generated purchase orders."""
        res = super().action_cancel()
```

pedrobaeza (approved):
Thanks for adding extra tests for the use case.

pilarvargas-tecnativa (approved):
thanks :+1:

PR #1711, opened 2023-05-18
[16.0][ADD] stock_move_actual_date

AungKoKoLin1997:
This module adds actual date in picking and scrap and pass the value to stock move, stock move line and accounting date of SVL's journal entry. It also adds actual date in SVL, enabling reporting based on this field.

When a user makes the transfer for the products with "real time" inventory valuation, effective date in transfer couldn't be selected.
And Accounting date for stock valuation journal entry will be the same with effective date.
So, for the users who want to specify each transfer's actual accounting date needs to reset to draft manually and
change accounting date. This module intends to get rid of this inconvenience by adding actual date in transfer and scrap and then pass to journal entry.

@qrtl QT4850 QT 5388


stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,50 @@
...
+        store=True,
+    )
+
+    @api.depends("date", "accounting_date")
+    def _compute_accounting_date(self):
+        # 'force_period_date' context is assigned when user sets accounting date in
+        # inventory adjustment
+        force_period_date = self._context.get("force_period_date")
+        if force_period_date:
+            self.write({"accounting_date": force_period_date})
```
rousseldenis: @AungKoKoLin1997 Never doing a write in a compute !

stock_picking_accounting_date/tests/test_stock_picking_accounting_date.py
```diff
@@ -0,0 +1,55 @@
...
+
+from odoo.addons.stock_account.tests.test_stockvaluation import TestStockValuation
+
+
+class TestStockValuation(TestStockValuation):
+    def setUp(self):
+        super(TestStockValuation, self).setUp()
+
+    def test_stock_picking_accounting_date(self):
+        self.product1.categ_id.property_cost_method = "fifo"
```
yostashiro:
```suggestion
```
I guess this assignment doesn't affect the test.  Can we ensure that the valuation is `real_time` instead?

yostashiro (commented):
@AungKoKoLin1997 Can you please add another test to ensure that accounting date set in inventory adjustment is not interfered by this module?

stock_picking_accounting_date/tests/test_stock_picking_accounting_date.py
```diff
@@ -0,0 +1,87 @@
...
+            [
+                ("location_id", "=", self.stock_location.id),
+                ("product_id", "=", self.product.id),
+            ]
+        )
+        inventory_quant.inventory_quantity = 200.00
+        accounting_date = date.today() + timedelta(days=3)
+        inventory_quant.accounting_date = accounting_date
+        inventory_quant.with_context(
+            skip_exceeded_discrepancy=True
```
yostashiro: I don't think we should be doing this as this context doesn't belong to a module in the dependency chain. It should ideally be handled by adjusting the `stock_inventory_discrepancy` module or by setting rogue modules.

stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,51 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    accounting_date = fields.Date(
+        compute="_compute_accounting_date",
+        store=True,
+    )
+
+    @api.depends("date", "accounting_date")
```
rousseldenis: @AungKoKoLin1997 Shouldn't it be 'picking_id.accounting_date' ???

stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,51 @@
...
+    def _compute_accounting_date(self):
+        # 'force_period_date' context is assigned when user sets accounting date in
+        # inventory adjustment
+        force_period_date = self._context.get("force_period_date")
```
rousseldenis: I don't really like computed stored fields based on context values....

stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,51 @@
...
+        if force_period_date:
+            for rec in self:
+                rec.accounting_date = force_period_date
+        else:
+            for rec in self:
+                if rec.picking_id.accounting_date:
+                    rec.accounting_date = rec.picking_id.accounting_date
+                    continue
+                rec.accounting_date = fields.Datetime.context_timestamp(self, rec.date)
```
yostashiro: Let's remove the condition based on the context here, as @rousseldenis suggests.
```suggestion
        for rec in self:
            if rec.picking_id.accounting_date:
                rec.accounting_date = rec.picking_id.accounting_date
                continue
            rec.accounting_date = fields.Datetime.context_timestamp(self, rec.date)
```

stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,51 @@
...
+    ):
+        am_vals = super(StockMove, self)._prepare_account_move_vals(
+            credit_account_id,
+            debit_account_id,
+            journal_id,
+            qty,
+            description,
+            svl_id,
+            cost,
+        )
```
yostashiro:
```suggestion
        am_vals = super(StockMove, self)._prepare_account_move_vals(
            credit_account_id,
            debit_account_id,
            journal_id,
            qty,
            description,
            svl_id,
            cost,
        )
        # i.e. Inventory adjustments with accounting date
        if self._context.get("force_period_date"):
            self.write({"accounting_date": self._context["force_period_date"]})
            return am_vals
```

yostashiro (changes requested):

stock_picking_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,50 @@
...
+
+    accounting_date = fields.Date(
+        compute="_compute_accounting_date",
+        store=True,
+    )
+
+    @api.depends("date", "picking_id.accounting_date")
+    def _compute_accounting_date(self):
+        # 'force_period_date' context is assigned when user sets accounting date in
+        # inventory adjustment
```
yostashiro: Oops, missed these lines.
```suggestion
```

stock_picking_accounting_date/tests/test_stock_picking_accounting_date.py
```diff
@@ -0,0 +1,85 @@
...
+        inventory_quant = self.env["stock.quant"].search(
+            [
+                ("location_id", "=", self.stock_location.id),
+                ("product_id", "=", self.product.id),
+            ]
+        )
+        inventory_quant.inventory_quantity = 200.00
+        accounting_date = date.today() + timedelta(days=3)
+        inventory_quant.accounting_date = accounting_date
+        inventory_quant.with_context().action_apply_inventory()
```
yostashiro:
```suggestion
        inventory_quant.action_apply_inventory()
```

stock_picking_accounting_date/tests/test_stock_picking_accounting_date.py
```diff
@@ -0,0 +1,85 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import date, timedelta
+
+from odoo.addons.stock_account.tests.test_stockvaluation import TestStockValuation
+
+
+class TestStockValuation(TestStockValuation):
+    def setUp(self):
+        super(TestStockValuation, self).setUp()
```
rousseldenis: Maybe switch to classmethod

stock_picking_accounting_date/models/stock_picking.py
```diff
@@ -0,0 +1,23 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    accounting_date = fields.Date(
+        help="Accounting date for stock valuation journal entry.",
+    )
+    show_accounting_date = fields.Boolean(compute="_compute_show_accounting_date")
+
+    def _compute_show_accounting_date(self):
+        for pick in self:
```
rousseldenis:
```suggestion
        self.show_accounting_date = False
        for pick in self:
```

stock_picking_actual_date/models/stock_scrap.py
```diff
@@ -0,0 +1,45 @@
...
+        compute="_compute_is_editable_actual_date", string="Is Editable"
+    )
+
+    def _compute_is_editable_actual_date(self):
+        for rec in self:
+            rec.is_editable_actual_date = False
+            if rec.state != "done" or self.env.user.has_group(
+                "stock.group_stock_manager"
+            ):
+                rec.is_editable_actual_date = True
```
yostashiro: Do you think it's a good idea to create a mixin and let stock.picking and stock.scrap inherit it?

stock_picking_actual_date/models/stock_scrap.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _prepare_move_values(self):
+        vals = super()._prepare_move_values()
+        if self.actual_date:
+            vals["actual_date"] = self.actual_date
+        return vals
```
yostashiro: Do we need this when `actual_date` of stock.move is a compute field?

stock_picking_actual_date/models/stock_scrap.py
```diff
@@ -0,0 +1,45 @@
...
+                account_moves = rec.move_id.account_move_ids
+                if not account_moves:
+                    continue
+                account_moves.button_draft()
+                for am in account_moves:
+                    am = am.with_context(skip_date_sequence_check=True)
+                    am.date = am.stock_move_id.actual_date
+                    if not am._sequence_matches_date():
+                        am.name = False
+                account_moves.action_post()
```
yostashiro: Let's not repeat the same logic in different places. Please create a method in account.move to handle this.

stock_picking_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,66 @@
...
+        return self.env.context.get("tz") or self.env.user.tz or "UTC"
+
+    @api.depends("date", "picking_id.actual_date", "scrap_ids.actual_date")
+    def _compute_actual_date(self):
+        tz = self._get_timezone()
+        for rec in self:
+            if rec.scrapped:
+                if rec.scrap_ids and rec.scrap_ids[0].actual_date:
+                    rec.actual_date = rec.scrap_ids[0].actual_date
+                continue
```
yostashiro: This implementation is incorrect as it won't set actual_date on the stock move when there is no actual_date set in scrap.

I think the following should do.
```suggestion
            if rec.scrap_ids.actual_date:
                rec.actual_date = rec.scrap_ids.actual_date
                continue
```

yostashiro (changes requested):

stock_picking_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _get_timezone(self):
+        return self.env.context.get("tz") or self.env.user.tz or "UTC"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("actual_date"):
+                self = self.with_context(actual_date=vals["actual_date"])
+        return super().create(vals_list)
```
yostashiro: A bit tedious but somethings like this?
```suggestion
    moves = self.env["stock.move"]
    vals_list_with_actual_date = []
    vals_list_without_actual_date = []
    for vals in vals_list:
        if vals.get("actual_date"):
            vals_list_with_actual_date.append(vals)
        else:
            vals_list_without_actual_date.append(vals)
    for vals in vals_list_with_actual_date:
        move = super().with_context(actual_date=vals["actual_date"]).create(vals)
        moves += move
    if vals_list_without_actual_date:
        moves += super().create(vals_list_without_actual_date)
    return moves
```

stock_picking_actual_date/models/stock_scrap.py
```diff
@@ -0,0 +1,27 @@
...
+class StockScrap(models.Model):
+    _name = "stock.scrap"
+    _inherit = ["stock.scrap", "actual.date.mixin"]
+
+    def _prepare_move_values(self):
+        vals = super()._prepare_move_values()
+        if self.actual_date:
+            vals["actual_date"] = self.actual_date
+        return vals
+
```
yostashiro:
```suggestion
```

stock_picking_actual_date/models/stock_scrap.py
```diff
@@ -0,0 +1,27 @@
...
+        res = super().write(vals)
+        if "actual_date" in vals:
+            for rec in self:
+                if rec.state != "done":
+                    continue
+                account_moves = rec.move_id.account_move_ids
+                if not account_moves:
+                    continue
+                account_moves._update_accounting_date()
+        return res
```
yostashiro:
```suggestion
        return res

    def do_scrap(self):
        for scrap in self:
            scrap = scrap.with_context(actual_date=scrap.actual_date)
            super(StockScrap, scrap).do_scrap()
        return True
```

stock_picking_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,82 @@
...
+        # Since there is no 'scrap_ids' value at the time of creation of the move,
+        # the scrap and move are linked only after the move is done as per the design
+        # of the 'do_scrap()' method.
+        for vals in vals_list:
+            if vals.get("actual_date") and vals.get("scrapped"):
+                # Expect only one stock move in vals_list if it is from scrap.
+                self = self.with_context(actual_date=vals["actual_date"])
+                return super(StockMove, self).create([vals])
+        return super().create(vals_list)
+
```
yostashiro:
```suggestion
```

yostashiro (commented):
@AungKoKoLin1997 My suggestion should work as far as I've checked. Am I missing something?

AungKoKoLin1997:
> My suggestion should work as far as I've checked. Am I missing something?

@yostashiro You are right. It seems like I missed something. My apologies.

stock_picking_actual_date/tests/test_stock_picking_actual_date.py
```diff
@@ -0,0 +1,88 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import date, timedelta
+
+from odoo.addons.stock_account.tests.test_stockvaluation import TestStockValuation
+
+
+class TestStockValuation(TestStockValuation):
+    def setUp(self):
+        super(TestStockValuation, self).setUp()
```
yostashiro: We should isolate our tests.

yostashiro (commented):
@AungKoKoLin1997 Can you please increase the test coverage.

I'm also thinking that the name `stock_move_actual_date` may be more appropriate.  What do you think?

stock_move_actual_date/tests/test_stock_move_actual_date.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+        # Update actual_date after done
+        scrap.actual_date = date(2024, 8, 11)
+        self.assertEqual(scrap.move_id.actual_date, date(2024, 8, 11))
+        self.assertEqual(
+            scrap.move_id.stock_valuation_layer_ids.account_move_id.date,
+            date(2024, 8, 11),
+        )
+
+        # Test inventory adjustment with actual date
```
yostashiro: Put this into a dedicated test method to make it easier to follow.

stock_move_actual_date/tests/test_stock_move_actual_date.py
```diff
@@ -0,0 +1,142 @@
...
+        cls.product_1 = cls.env["product.product"].create(
+            {
+                "name": "Test Product",
+                "type": "product",
+                "categ_id": product_category.id,
+            }
+        )
+        cls.supplier_location = cls.env.ref("stock.stock_location_suppliers")
+        cls.stock_location = cls.env.ref("stock.stock_location_stock")
+        cls.env.user.tz = "Asia/Tokyo"
```
yostashiro: For clarity of the test scenario, it would be better to put this inside `test_stock_move_without_actual_date_from_picking_or_scrap()` which is the only test this setting matters.

stock_move_actual_date/tests/test_stock_move_actual_date.py
```diff
@@ -0,0 +1,140 @@
...
+    def test_inventory_adjustment_actual_date(self):
+        _, _ = self.create_picking()
+        inventory_quant = self.env["stock.quant"].search(
+            [
+                ("location_id", "=", self.stock_location.id),
+                ("product_id", "=", self.product_1.id),
+            ]
+        )
+        inventory_quant.inventory_quantity = 20.0
+        inventory_quant.accounting_date = date(2024, 7, 1)
```
yostashiro: Why don't we just directly create a quant?

AungKoKoLin1997: There was no reason. Your suggestion is more clear. I updated.

stock_move_actual_date/tests/test_stock_move_actual_date.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        scrap.action_validate()
+        return scrap
+
+    def test_stock_move_actual_date(self):
+        receipt, move = self.create_picking(date(2024, 9, 1))
+        self.assertEqual(move.actual_date, date(2024, 9, 1))
+        self.assertEqual(
+            move.stock_valuation_layer_ids.account_move_id.date, date(2024, 9, 1)
+        )
```
yostashiro: `move.stock_valuation_layer_ids` could involve adjustment layers and therefore is not desirable to be relied upon.

```suggestion
        self.assertEqual(move.account_move_ids.date, date(2024, 9, 1))
```

Please apply the same to the rest.

yostashiro (approved):
Code review.

@AungKoKoLin1997 Please squash commits once it's good to do so.

stock_move_actual_date/hooks.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
```
kanda999:
```suggestion
from odoo.tools.sql import column_exists
```
We don't need openupgradelib?

AungKoKoLin1997: Thanks! It is valid.

kanda999 (approved):
Functional review and Code review.

AungKoKoLin1997:
@rousseldenis Can you please review this PR?

stock_move_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,68 @@
...
+        compute="_compute_actual_date",
+        store=True,
+    )
+
+    def _get_timezone(self):
+        return self.env.context.get("tz") or self.env.user.tz or "UTC"
+
+    @api.depends("date", "picking_id.actual_date", "scrap_ids.actual_date")
+    def _compute_actual_date(self):
+        tz = self._get_timezone()
```
rousseldenis, line 35:
```suggestion
        tz = self._get_timezone()
        context_actual_date = self.env.context.get("actual_date")
```

stock_move_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,68 @@
...
+        self,
+        credit_account_id,
+        debit_account_id,
+        journal_id,
+        qty,
+        description,
+        svl_id,
+        cost,
+    ):
+        am_vals = super(StockMove, self)._prepare_account_move_vals(
```
rousseldenis:
```suggestion
        am_vals = super()._prepare_account_move_vals(
```

stock_move_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,68 @@
...
+            credit_account_id,
+            debit_account_id,
+            journal_id,
+            qty,
+            description,
+            svl_id,
+            cost,
+        )
+        # i.e. Inventory adjustments with actual date
+        if self._context.get("force_period_date"):
```
rousseldenis:
```suggestion
        if self.env.context.get("force_period_date"):
```

rousseldenis (commented):
Code review. Some minor changes

stock_move_actual_date/models/stock_move.py
AungKoKoLin1997: Done!

stock_move_actual_date/models/stock_move.py
AungKoKoLin1997: Done!

AungKoKoLin1997:
The tests failed is related with https://github.com/odoo/odoo/issues/197556.

stock_move_actual_date/models/actual_date_mixin.py
```diff
@@ -16,10 +16,12 @@ class ActualDateMixin(models.AbstractModel):
         compute="_compute_is_editable_actual_date", string="Is Editable"
     )
 
+    def _allow_editable(self):
```
yostashiro: Let's make the method name more self-explanatory.
```suggestion
    def _check_actual_date_editable(self):
```

AungKoKoLin1997: @yostashiro 
I'm not satisfied with the current design as it allows editing in unintended models (e.g., `mrp.production`) if they inherit this `mixin` model. `actual_date` in mrp.production should only be editable by an `MRP Manager` or when the state is not `canceled` or `done`. Currently, a user with both` Stock Manager` and` MRP User` roles can still edit `actual_date`, and the same issue exists in `stock.picking`.

Instead of adding user group checks in the `mixin` model, we should inherit the method in each related model and apply specific group checks. However, this approach introduces redundancy (e.g., checking Stock Manager in both `stock.picking` and `stock.scrap` for the same logic). 
Do you have any idea on this?

AungKoKoLin1997: As we discussed, I added the new group and applied the changes.

stock_move_actual_date/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+* Go to Settings > Users & Companies > Groups.
+* Open Editable Actual Date and add the users who are allowed to edit the actual date
```
yostashiro:
```suggestion
* Open 'Modify Actual Date' and add the users who are allowed to edit the actual date
```

stock_move_actual_date/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+Use this module's function in the following situations:
+
+* If you are late in processing a transfer or scrap in Odoo and wish to record the
+  transaction with the actual transfer date, fill in the Actual Date field in the picking or
+  scrap form. The Actual Date of the picking or scrap is then propagated to its
+  corresponding stock moves and stock move lines, and is also passed to the journal
+  entry as the date.
+* You can also update the Actual Date of a completed picking or scrap if you belong to
+  the "Editable Actual Date" group. This operation updates the date of the related journal
```
yostashiro:
```suggestion
  the 'Modify Actual Date' group. This operation updates the date of the related journal
```

stock_move_actual_date/tests/test_stock_move_actual_date.py
```diff
@@ -0,0 +1,148 @@
...
+                "name": "10 in",
+                "location_id": self.supplier_location.id,
+                "location_dest_id": self.stock_location.id,
+                "product_id": self.product_1.id,
+                "product_uom_qty": 10.0,
+                "price_unit": 10,
+                "move_line_ids": [
+                    (
+                        0,
+                        0,
```
yostashiro:
```suggestion
                    Command.create(
```

stock_move_actual_date/models/res_currency.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import models
+
+
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    def _convert(self, from_amount, to_currency, company, date, round=True):
+        date = self.env.context.get("actual_date") or date
+        return super()._convert(from_amount, to_currency, company, date, round)
```
yostashiro:
```suggestion
        return super()._convert(from_amount, to_currency, company, date, round=round)
```

yostashiro (commented):
Some minor adjustments.

yostashiro:
@AungKoKoLin1997 Please also apply a temporary fix to the test with a link to the reported issue in odoo/odoo.

AungKoKoLin1997:
> Please also apply a temporary fix to the test with a link to the reported issue in odoo/odoo.

Done!

yostashiro (approved):
Code review. LGTM.

nobuQuartile (approved):
Functional review. LGTM.

AungKoKoLin1997:
@rousseldenis Can you please review PR?

stock_move_actual_date/models/actual_date_mixin.py
```diff
@@ -0,0 +1,69 @@
...
+        return res
+
+    def write(self, vals):
+        res = super().write(vals)
+        move_field_name = self._get_stock_move_field_name()
+        if "actual_date" in vals or move_field_name and move_field_name in vals:
+            for rec in self:
+                moves = rec._get_stock_moves()
+                moves.write({"actual_date_source": rec.actual_date})
+                if rec.state != "done":
```
yostashiro: We should technically do something like this, IMO.
```suggestion
                if rec.state != self._get_done_state():
```

stock_move_actual_date/models/actual_date_mixin.py
```diff
@@ -0,0 +1,69 @@
...
+                    continue
+                account_moves = moves.account_move_ids
+                if not account_moves:
+                    continue
+                account_moves._update_accounting_date()
+        return res
+
+    def _check_actual_date_editable(self):
+        self.ensure_one()
+        return self.state not in ["done", "cancel"] or self.env.user.has_group(
```
yostashiro: Given that we don't really care about the cancelled documents.
```suggestion
        return self.state not self._get_done_state() or self.env.user.has_group(
```

stock_move_actual_date/models/stock_move.py
```diff
@@ -0,0 +1,81 @@
...
+                continue
+            rec.actual_date = fields.Date.context_today(
+                self.with_context(tz=tz), rec.date
+            )
+
+    def _action_done(self, cancel_backorder=False):
+        moves = super()._action_done(cancel_backorder)
+        # i.e. Inventory adjustments with actual date
+        if self.env.context.get("force_period_date"):
+            self.write({"actual_date": self.env.context["force_period_date"]})
```
yostashiro: This may be more appropriate.
```suggestion
            self.write({"actual_date_source": self.env.context["force_period_date"]})
```

marcelsavegnago (approved):
LGTM

nobuQuartile (approved):
LGTM
Functional Review

yostashiro (approved):
This module should be in good shape now, after some refactoring to make it more extensible. 👍

stock_move_actual_date/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    def _compute_actual_date(self):
+        for rec in self:
+            account_move = rec.account_move_id
+            if not account_move:
+                rec.actual_date = rec.stock_move_id.actual_date
+                continue
+            if account_move and account_move.state == "posted":
+                rec.actual_date = account_move.date
+                continue
```
yostashiro: To ensure that `actual_date` is always set even when there is neither an account move nor a stock move associated (e.g., records from valuation method change from manual to real time for zero-cost products).

```suggestion
            if account_move and account_move.state == "posted":
                rec.actual_date = account_move.date
                continue
            if rec.stock_move_id.actual_date:
                rec.actual_date = rec.stock_move_id.actual_date
                continue
```

stock_move_actual_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,12 @@
+This module adds an Actual Date field to the stock picking, stock scrap,
+stock move, and stock move line models. This field allows users to
+record the actual date on which a stock transfer or stock scrap took
+place, in case the transaction in Odoo is processed after the fact. It
+also adds an Actual Date field to the Stock Valuation Layer model,
```
yostashiro:
```suggestion
place, in case the transaction in Odoo is processed after the fact.

It also adds an Actual Date field to the Stock Valuation Layer model,
```

yostashiro (changes requested):

yostashiro (approved):
LGTM.

nobuQuartile:
LGTM
Functional review.

nobuQuartile (approved):

stock_move_actual_date/models/actual_date_mixin.py
```diff
@@ -0,0 +1,70 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ActualDateMixin(models.AbstractModel):
+    _name = "actual.date.mixin"
+
+    actual_date = fields.Date(
+        tracking=True,
```
yostashiro, line 11:
```suggestion
        tracking=True,
        copy=False,
```
Otherwise, backorders keep the actual date.

yostashiro (commented):
I guess we should move this to stock-logistics-workflow as well according to https://github.com/OCA/stock-logistics-warehouse/pull/2320#pullrequestreview-2839395777.

AungKoKoLin1997:
This module is moved to https://github.com/oca/stock-logistics-workflow
https://github.com/OCA/stock-logistics-workflow/pull/1941

PR #1716, opened 2023-05-22
[15.0][IMP] stock_request_purchase: Propagate the cancellation of related purchases

victoralmau:
FWP from 14.0: https://github.com/OCA/stock-logistics-warehouse/pull/1710

Propagate the cancellation of related purchases

@Tecnativa TT43414

### DIFF
--- a/stock_request_purchase/models/stock_request.py
+++ b/stock_request_purchase/models/stock_request.py
@@ -1,4 +1,5 @@
 # Copyright 2017-20 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2023 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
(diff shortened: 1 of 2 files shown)


stock_request_purchase/models/stock_request.py
```diff
@@ -38,6 +39,14 @@ def _check_purchase_company_constrains(self):
                 )
             )
 
+    def action_cancel(self):
+        """Propagate the cancellation to the generated purchase orders."""
+        res = super().action_cancel()
+        self.sudo().purchase_ids.filtered(
```
pedrobaeza, line 45: What happens if several products from other stock request/MTO/etc are in the same purchase order? I think you should check that the PO is only from this stock request.

pedrobaeza (approved):
Patch to be picked to v14

pilarvargas-tecnativa (approved):
thanks :)

PR #1718, opened 2023-05-22
[14.0][IMP] stock_request_purchase: Filter purchases that are linked only to requests

victoralmau:
Filter purchases that are linked only to requests

@Tecnativa TT43414


pedrobaeza (approved):
Fast-tracking to amend previous patch:

/ocabot merge patch

PR #1720, opened 2023-05-23
[14.0][ADD] stock_reserve_area

mariadforgeflow:
Allows to create reservation areas to manage stock globally on them. 
For example, each area can represent a warehouse and it can be useful to make sure that no other moves created later in other locations from the same Reserve Area take the products.
In pickings that move a product out of the source location area, before making the reservation in the location, the product needs to be available and reserved in the area.

cc @JordiBForgeFlow 

TODO:

* [x] Create by default reservation area for each warehouse
* [x] When a new warehouse is created, create the corresponding reservation area
* [x] tests
* [x] Allow one location to be in more than one reserve area as long as they are concentric


rousseldenis:
@mariadforgeflow Are you aware of this ?

https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_location_orderpoint

These two modules seem to cover the same usage no ?

mariadforgeflow:
Hi @rousseldenis, I think that the purpose of both modules is slightly different. In this case, we just want to reserve in an area (for example, each area can represent a warehouse)  to make sure that no other moves created later in other locations from the same Reserve Area take the products. I was also planning to add a smart button in the Reserve Area to be able to track easily the reserved products in them.

rousseldenis:
> Hi @rousseldenis, I think that the purpose of both modules is slightly different. In this case, we just want to reserve in an area (for example, each area can represent a warehouse) to make sure that no other moves created later in other locations from the same Reserve Area take the products. I was also planning to add a smart button in the Reserve Area to be able to track easily the reserved products in them.

Yes, but that can be solved :

- Create your reserve area under the view_location_id of your warehouse and not lot_stock_id. So, the global stock will take into account that stock. But if the stock rules point to lot_stock_id (standard config), no move can take the reserve stock.
- Using the stock_location_orderpoint module, you can create an orderpoint on Stock/Shelf 1 for instance that will be refilled by Reserve/Reserve 1

jbaudoux:
The move source location limits where the reservation can occur (i.e. on any child location). When the reservation is placed, the other moves won't steal the reservation. This is the standard behavior and so I don't understand what you mean with

> it can be useful to make sure that no other moves created later in other locations from the same Reserve Area take the products. In pickings that move a product out of the source location area, before making the reservation in the location, the product needs to be available and reserved in the area.

If the area contains multiple source locations to evaluate, then you can use this module to list the source locations of the area and even give the reservation rules for each of the locations of the area:
https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_reserve_rule

JordiBForgeFlow:
@rousseldenis @jbaudoux This module resolves the following problem.

We have customer demand in WH, and yet we don't want to necessary reserve the stock for that demand in a particular location upfront. The delivery orders would place demand on WH/Output for example, where there's no stock.

Then, the warehouse people can decide to fulfill that demand later by organizing pickings from WH/Stock to WH/Output.

However, if you try to do that without this module, you cannot guarantee to the customer of a particular sales order that they will get the products they ordered.

With this module, as soon as the delivery order is created, an area level reservation (typicall area will be the whole warehouse) will take place, regardless if the local reservation has taken place yet. That way we guarantee that new orders will not 'steal' from the previous ones.

This would replace https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release (at least partially). The problem with stock_available_to_promise_release is IMHO that manufacturing orders can 'steal' the stock.  The module stock_reserve_area resolves the problem of guaranteeing that the stock will be reserved to whoever asks for it first within an area (warehouse), regardless if it is a manufacturing order, or a picking.

There are other topics that the module stock_available_to_promise_release does that are not included in this module (and seem useful, but not strictly related with the reservation topic).

cc @guewen @simahawk

stock_reserve_area/models/stock_quant.py
```diff
@@ -0,0 +1,56 @@
...
+            >= 0.0
+            else 0.0
+        )
+
+    def _gather_reserve_area(self, product_id, reserve_area_id):
+        self.env["stock.quant"].flush(["reserve_area_id", "available_quantity"])
+        quants = self
+        for location_id in reserve_area_id.location_ids:
+            if reserve_area_id.is_location_in_area(location_id):
+                quants |= self._gather(product_id, location_id)
```
jbaudoux: This will likely be slow. Collect the ids and recreate the recordset by calling browse(ids) at the end

stock_reserve_area/models/stock_quant.py
```diff
@@ -0,0 +1,56 @@
...
+                [
+                    ("reserve_area_id", "=", reserve_area_id.id),
+                    ("product_id", "=", product_id.id),
+                ]
+            )
+            .mapped("area_reserved_availability")
+        )
+        available_quantity = sum(quants.mapped("quantity")) - sum(
+            area_reserved_quantity
+        )
```
jbaudoux: Pay attention that if you have lots of quants and/or lots of moves, that a lot of data to fetch to compute the sum. Maybe at some point, you'll want to replace it by read_group

stock_reserve_area/models/stock_reserve_area.py
```diff
@@ -0,0 +1,18 @@
...
+    name = fields.Char()
+    active = fields.Boolean(default=True)
+    company_id = fields.Many2one("res.company")
+    location_ids = fields.One2many("stock.location", "reserve_area_id")
+
+    def is_location_in_area(self, location):
+        child_locations = self.env["stock.location"].search(
+            [("id", "child_of", self.location_ids.ids)]
+        )
+        return location in child_locations
```
jbaudoux: That's inefficient.
Add the location in the domain or use parent_path

jbaudoux:
@JordiBForgeFlow Thanks for the clarification. I now understand your approach.

Can you update the README to make it better understandable.

It will need some tweaks to make it scalable.

JordiBForgeFlow:
@jbaudoux Yes, we will update as per your suggestions. We appreciate your feedback. What could be interesting is to see how this module could work with https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release. I see them as kind of complementary on some parts. For example, the part of releasing the deliveries so that they would trigger the chained transfers was interesting.

rousseldenis:
> @jbaudoux Yes, we will update as per your suggestions. We appreciate your feedback. What could be interesting is to see how this module could work with https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release. I see them as kind of complementary on some parts. For example, the part of releasing the deliveries so that they would trigger the chained transfers was interesting.

@JordiBForgeFlow Yes, that one will be in your interest maybe : https://github.com/OCA/wms/tree/14.0/stock_release_channel

stock_reserve_area/views/stock_reserve_area_views.xml
```diff
@@ -0,0 +1,88 @@
...
+                                name="company_id"
+                                groups="base.group_multi_company"
+                                options="{'no_create': True}"
+                            />
+                            </group>
+                            <group>
+                                <field
+                                name="location_ids"
+                                widget="many2many"
+                                domain="[('reserve_area_id', '=', False), ('usage', '=', 'internal')]"
```
JordiBForgeFlow: Why not include the view locations? Otherwise you cannot inclue, for example, the entire WH

JordiBForgeFlow: Also, seems better to put the domain in the field itself instead of the view.

stock_reserve_area/models/stock_quant.py
mariadforgeflow: @jbaudoux thanks for you suggestions. In this one, I don't understand what you mean. Quants are returned by the _gather method that takes the location as an argument. You mean doing 
`quants += self._gather(product_id, location_id).ids` ?

jbaudoux: yes
```
quant_ids += self._gather(product_id, location_id).ids
...
quants = self.browse(quant_ids)
```

stock_reserve_area/models/stock_reserve_area.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockReserveArea(models.Model):
+    _name = "stock.reserve.area"
+
+    name = fields.Char()
+    active = fields.Boolean(default=True)
+    company_id = fields.Many2one("res.company")
```
JordiBForgeFlow: Add a default

stock_reserve_area/models/stock_reserve_area.py
```diff
@@ -0,0 +1,23 @@
...
+    location_ids = fields.One2many(
+        "stock.location",
+        "reserve_area_id",
+        domain="[('reserve_area_id', '=', False),"
```
JordiBForgeFlow: Add to the domain the locations that belong to the company indicated in the reserve area

stock_reserve_area/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_stock_reserve_area_stock_user,access_stock_reserve_area_stock_user,model_stock_reserve_area,stock.group_stock_user,1,1,1,1
```
JordiBForgeFlow: Only stock manager should be able to create edit or delete the reserve area

mariadforgeflow:
@JordiBForgeFlow @jbaudoux, could you please review the changes I made? I have populated the reserve_area_id field in all child locations of the locations specified in the Reserve Area.

I have implemented as well:
 - Create by default reservation area for each warehouse
 - When a new warehouse is created, create the corresponding reservation area

JordiBForgeFlow:
@jgrandguillaume I think that it would be great to talk with you and @jbaudoux about this module, and the relationship with https://github.com/OCA/wms/tree/14.0/stock_available_to_promise_release.

MiquelRForgeFlow (approved):

JoanSForgeFlow (approved):
Technical Review, LGTM!

stock_reserve_area/models/stock_location.py
```diff
@@ -0,0 +1,23 @@
...
+        res = super().create(vals)
+        if res.location_id.reserve_area_id:
+            res.reserve_area_id = res.location_id.reserve_area_id
+        return res
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("active") or vals.get("location_id"):
+            self.reserve_area_id = self.location_id.reserve_area_id
+        return res
```
LoisRForgeFlow: Why not making this field a compute stored one with the proper api.depends?

stock_reserve_area/models/stock_move.py
```diff
@@ -0,0 +1,182 @@
...
+                        "picking Reservation Area."
+                    )
+                )
+
+    @api.constrains("is_out_area_move")
+    def check_is_out_area_move(self):
+        for rec in self:
+            if (
+                rec.picking_id
+                and rec.picking_id.is_out_area_picking != rec.is_out_area_move
```
LoisRForgeFlow: An api.constrains with a condition in 2 computed (non-stored) fields, why? is this code even triggered?

stock_reserve_area/models/stock_move.py
```diff
@@ -0,0 +1,182 @@
...
+                )
+
+    @api.depends("location_id")
+    def _compute_reserve_area_id(self):
+        for move in self:
+            reserve_areas = self.env["stock.reserve.area"].search([])
+            for reserve_area in reserve_areas:
+                if reserve_area.is_location_in_area(move.location_id):
+                    move.reserve_area_id = reserve_area
+                    break
```
LoisRForgeFlow: what is the performance of this on installation? it could probably be improved by having a map of all relevant locations (a set of all the locations in the moves) to reservation area, then asignation is quicker.

stock_reserve_area/models/stock_quant.py
```diff
@@ -0,0 +1,57 @@
...
+    @api.depends("location_id")
+    def _compute_reserve_area_id(self):
+        for quant in self:
+            reserve_areas = self.env["stock.reserve.area"].search([])
+            for reserve_area in reserve_areas:
+                if reserve_area.is_location_in_area(quant.location_id):
+                    quant.reserve_area_id = reserve_area
+                    break
+
+    def _get_reserve_area_available_quantity(self, product_id, reserve_area_id):
```
LoisRForgeFlow, line 11: Wouldn't this method make more sense in the stock.reserve.area model directly?

stock_reserve_area/models/stock_reserve_area.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2023 ForgeFlow SL.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockReserveArea(models.Model):
+    _name = "stock.reserve.area"
```
LoisRForgeFlow, line 8: I miss an action to see current reservations information. It can be generic and the be also accesible from other places, for instance: the stock moves (operations tab in picking)?, maybe a smart button in the picking?

stock_reserve_area/readme/CONTRIBUTING.rst
LoisRForgeFlow, line 1: CONTRIBUTORS.rst

LoisRForgeFlow (commented):
Technical review, some comments below

stock_reserve_area/models/stock_quant.py
mariadforgeflow: I did it in here because it is like the _get_available_quantity form this model but specific for the reserve area.

PR #1722, opened 2023-05-24
[14.0][IMP] stock_request: Add option to use stock available

victoralmau:
Add option to use stock available

@Tecnativa TT43413

### DIFF
--- a/stock_request/models/res_company.py
+++ b/stock_request/models/res_company.py
@@ -12,3 +12,6 @@ class ResCompany(models.Model):
     stock_request_allow_virtual_loc = fields.Boolean(
         string="Allow Virtual locations on Stock Requests"
     )
+    stock_request_check_available_first = fields.Boolean(
+        string="Check available stock first"
+    )
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -259,7 +259,7 @@ class StockRequest(models.Model):
 
     def _action_confirm(self):
         self._action_launch_procurement_rule()
-        self.write({"state": "open"})
+        self.filtered(lambda x: x.state != "done").write({"state": "open"})
 
     def action_confirm(self):
         self._action_confirm()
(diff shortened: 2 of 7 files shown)


stock_request/models/res_company.py
```diff
@@ -12,3 +12,6 @@ class ResCompany(models.Model):
     stock_request_allow_virtual_loc = fields.Boolean(
         string="Allow Virtual locations on Stock Requests"
     )
+    stock_request_check_available_stock_first = fields.Boolean(
```
pedrobaeza: Simplify variable name at maximum, as stock is already mentioned:

```suggestion
    stock_request_check_available_first = fields.Boolean(
```

stock_request/models/stock_request.py
```diff
@@ -258,8 +258,9 @@ def check_order_picking_policy(self):
             raise ValidationError(_("The picking policy must be equal to the order"))
 
     def _action_confirm(self):
+        self._action_use_stock_available()
```
pedrobaeza: Call this method only for the SR that have it checked.

stock_request/models/stock_request.py
```diff
@@ -340,6 +341,49 @@ def _prepare_procurement_values(self, group_id=False):
...
+        move._action_done()
+        return move
+
+    def _action_use_stock_available(self):
+        allocation_model = self.env["stock.request.allocation"]
+        stock_move_model = self.env["stock.move"].sudo()
+        for request in self.filtered(
+            lambda x: x.company_id.stock_request_check_available_stock_first
+        ):
+            if request.product_id.sudo().qty_available >= request.product_uom_qty:
```
pedrobaeza: You are not passing location as context for restricting the available quantity. I also understand that you need to take into account other SR allocations for reducing this quantity.

stock_request/models/stock_request.py
```diff
@@ -340,6 +341,49 @@ def _prepare_procurement_values(self, group_id=False):
     def _skip_procurement(self):
         return self.state != "draft" or self.product_id.type not in ("consu", "product")
 
+    def _prepare_stock_move(self):
+        return {
```
pedrobaeza, line 340: This phantom move will be garbage and will corrupt statistics like forecasting, so we should try to get another way. And if not, at least `location_id` and `location_dest_id` should be both `location.id` in my understanding.

I know that allocations require a move to be linked, but I think we can relax such requirement. The only place where it's used is for checking its state for putting the opened qty.

pedrobaeza (changes requested):

victoralmau:
I have made some changes and added more use cases.

I believe that stock move is necessary, and in some cases several, for example when you have several child locations (the stock you need may be entirely in one child location, or you may need to use several locations).

victoralmau:
Ping @pedrobaeza  and @chienandalu

victoralmau:
All changes are done, can you check it again?

stock_request/models/stock_request.py
```diff
@@ -362,6 +394,14 @@ def _action_launch_procurement_rule(self):
             if float_compare(qty, request.product_qty, precision_digits=precision) >= 0:
                 continue
 
+            # If stock is available we use it and we do not execute rule
+            if request.company_id.stock_request_check_available_first:
+                quants = quant_model._gather(request.product_id, request.location_id)
+                qty_available = sum(quants.mapped("available_quantity"))
+                if qty_available >= request.product_uom_qty:
```
rousseldenis: Please use float_compare

stock_request/models/stock_request.py
```diff
@@ -259,7 +259,7 @@ def check_order_picking_policy(self):
 
     def _action_confirm(self):
         self._action_launch_procurement_rule()
-        self.write({"state": "open"})
+        self.filtered(lambda x: x.state != "done").write({"state": "open"})
```
pedrobaeza, line 262: Isn't this a computed field now?

victoralmau: Not yet, it is related to https://github.com/OCA/stock-logistics-warehouse/pull/1822

stock_request/models/stock_request.py
```diff
@@ -362,6 +394,21 @@ def _action_launch_procurement_rule(self):
             if float_compare(qty, request.product_qty, precision_digits=precision) >= 0:
                 continue
 
+            # If stock is available we use it and we do not execute rule
+            if request.company_id.stock_request_check_available_first:
+                quants = quant_model._gather(request.product_id, request.location_id)
```
pedrobaeza: This is still not valid, as you have to make as many moves as possible locations. That's why I told you about first getting qty_available from product with the proper context, and then do the moves with `_gather`.

victoralmau: Changes done to take into account quants and create the necessary moves.

pedrobaeza (approved):

carlosdauden (approved):

carlosdauden:
Please rebase

victoralmau:
Rebase done.

chienandalu (approved):

PR #1730, opened 2023-05-30
[14.0][FIX] stock_picking_volume: _compute_volume do not use canceled moves

mt-software-de:


mt-software-de:
@jbaudoux and @rousseldenis could you please do the review?

jbaudoux (approved):
This should be ported to v16 also then

stock_picking_volume/models/stock_picking.py
```diff
@@ -16,10 +16,17 @@ class StockPicking(models.Model):
...
     def _compute_volume(self):
         for picking in self:
-            new_volume = sum(picking.move_lines.mapped("volume"))
+            volume_by_states = defaultdict(lambda: 0)
+            for move in picking.move_lines:
+                volume_by_states[move.state] += move.volume
+            new_volume = 0
+            states = volume_by_states.keys()
+            if states != ["cancel"]:
+                volume_by_states["cancel"] = 0
```
jbaudoux: Add a comment that you exclude canceled lines except for canceled picking

jbaudoux: I find the code complex. Why not something like:
```
moves = picking.move_lines
exclude_cancel = any(m.state != "cancel" for m in moves)
volume = 0
for move in moves:
  if move.state == "cancel" and exclude_cancel:
    continue
  volume += move.volume
```

jbaudoux (changes requested):
tests are failing

stock_picking_volume/models/stock_picking.py
mt-software-de: yes this is much nicer

jbaudoux (approved):

mt-software-de:
ping @lmignon

marcwimmer (approved):

rousseldenis (commented):
@mt-software-de Maybe a test that highlight this change ?

mt-software-de:
> @mt-software-de Maybe a test that highlight this change ?

done. Then let's merge it. :-)

stock_picking_volume/models/stock_picking.py
```diff
@@ -16,12 +16,18 @@ class StockPicking(models.Model):
...
-            new_volume = sum(picking.move_lines.mapped("volume"))
-            if picking.volume != new_volume:
-                picking.volume = new_volume
+            moves = picking.move_lines
+            exclude_cancel = any(m.state != "cancel" for m in moves)
+            volume = 0
+            for move in moves:
+                if move.state == "cancel" and exclude_cancel:
+                    continue
+                volume += move.volume
```
lmignon, line 28:
```suggestion
            volume = 0
            for move in moves:
                if move.state == "cancel":
                    continue
                volume += move.volume
```
May be I miss something but IMO this code could be simplified to avoid a useless loop

lmignon (changes requested):

stock_picking_volume/models/stock_picking.py
jbaudoux: @lmignon no, that's not the same. Purpose is to exclude canceled moves from non-canceled pickings. For canceled pickings, we still compute the volume

lmignon: At least we should have a comment to explain this behavior (And a test for both cases).  If the exclude of cancelled moves from the computed volume makes sense,  I don't see the motivation to still compute the volume of cancelled pickings.  I'm not at ease with the lack of symmetry between the 2 ways of computing the volume.

jbaudoux: > I don't see the motivation to still compute the volume of cancelled pickings.

To not break current behavior. And I find it good to keep totals on a canceled document.

lmignon: I will not block but this a comment and a test is more than welcome to avoid regression and makes this behavior explicit.

mt-software-de: I already added a Test for it.

lmignon: there are no tests for the case where we cancel the picking and want to make sure that in this case, the volume is computed on all the moves.

lmignon:
@pedrobaeza Why do you've merged this one? My last comment is not addressed. There is no comment nor tests to help to understand some part of the code that could be removed if you are not aware of the UC behind.

pedrobaeza:
There were 2 approvals and @mt-software-de put this comment: https://github.com/OCA/stock-logistics-warehouse/pull/1730#issuecomment-1606252663

lmignon:
By reading the code, you could see that this comment did not cover the request made. Today I can delete a line of code without breaking the tests and I challenge anyone reading the code to find out why it was written as it is and not as I naively suggested changing it.

pedrobaeza:
Sorry, as I acted on good faith without reading code.

lmignon:
I know you acted in good faith. :smile:  It's just that there's always a risk when merging if you haven't taken part in the review or checked yourself that all the requests have been addressed by carrying out a final review. Thank you for your work and I'hope @mt-software-de will add the missing test and comment in a new PR.

lmignon:
@pedrobaeza Reading once again the tests, I found one to cover the specific case I mentioned https://github.com/OCA/stock-logistics-warehouse/pull/1730/files#diff-152d3848b60c000d420d5a24e6cd15bda12cfb2febbd0705ee6618635ede79d7R151 Everything is OK.

pedrobaeza:
I'm glad to hear that.

jbaudoux:
@lmignon You wrote that test yourself 2 months ago https://github.com/OCA/stock-logistics-warehouse/blame/14.0/stock_picking_volume/tests/test_stock_picking_volume.py#L151
:)

PR #1737, opened 2023-06-02
[14.0][ADD] module stock_inventory_fix_reservation

bguillot:
This module is used to re assign move after validation of an inventory


rousseldenis:
@bguillot This deserves a good README. And I would have renamed it 'stock_inventory_reassign_move' as 'fix' is a little bit too vague.

sebalix:
Note: I think this module introduces the same feature than [`stock_move_free_reservation_reassign`](https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_move_free_reservation_reassign)

PR #1754, opened 2023-06-14
[ADD][16.0] new module stock_quant_import

PicchiSeba:
This PR aims to add a new module making it easier to import stock data


stock_quant_import/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows to update stock availability using a file.
+
+Load a csv containing the new data of stock information.
```
rousseldenis: @PicchiSeba Could you describe in a USAGE file, what content (what should the file look like - columns etc...) ?

rousseldenis (changes requested):
LGTM. Apart a better module usage description

PicchiSeba:
I'm trying to figure out why tests are OK on my machine but they're failing on OCA's

PicchiSeba:
@rousseldenis 

It seems  that `stock_inventory_discrepancy` and `stock_quant_import` conflict with each other in the tests.

The module `stock_inventory_discrepancy` adds an additional modal to confirm `stock.quant` changes.

I added a way to detect that module in the tests, if it's installed some necessary steps are executed.

stock_quant_import/tests/test_stock_import.py
```diff
@@ -0,0 +1,139 @@
...
+        cls.is_stock_inv_discrepancy_installed = bool(
+            cls.env["ir.module.module"].search(
+                [
+                    ("name", "=", "stock_inventory_discrepancy"),
+                    ("state", "=", "installed"),
+                ]
+            )
+        )
+        if cls.is_stock_inv_discrepancy_installed:
+            cls.ConfirmDiscrepancyWiz = cls.env["confirm.discrepancy.wiz"]
```
HekkiMelody, line 30: comment(non-blocking): I think an easier way to avoid the issue would have been to add a context variable

https://github.com/OCA/stock-logistics-warehouse/blob/16.0/stock_inventory_discrepancy/models/stock_quant.py#L55-L57

More importantly: does the module work from the UI even with `stock_inventory_discrepancy` installed?

PicchiSeba:
With the last force-pushed commit I added the maintainers

HekkiMelody (approved):
Code review, LGTM. We have also been using this successfully in production for a couple of months now.

stock_quant_import/tests/test_stock_import.py
PicchiSeba: 1. That was my first attempt actually, but setting that flag to True didn't seem to work (in the tests at least)

2. It works with the other module installed. The only difference is an additional step required to apply stock amounts

PR #1741, opened 2023-06-06
[14.0][FIX] stock_request: set stock request order state to done when partially delivered

OriolMForgeFlow:

PR #1755, opened 2023-06-15
[15.0][FIX] stock_request: set stock request order state to done when parti…

OriolMForgeFlow:
Forwardport #1741


LoisRForgeFlow (approved):
Ty!

PR #1756, opened 2023-06-15
[13.0][FIX] stock_request: set stock request order state to done when partially delivered

OriolMForgeFlow:
Backwardport #1741


LoisRForgeFlow (approved):
Thanks!

PR #1760, opened 2023-06-16
[16.0][MIG] stock_free_quantity: Migration to 16.0

EvaSForgeFlow:
Correcting tests.

PR #1761, opened 2023-06-16
[16.0][MIG] stock_pull_list: Migration to 16.0

EvaSForgeFlow:
Correcting tests.

Depends on #1760


rousseldenis:
@LoisRForgeFlow @EvaSForgeFlow Could we consider this could go too to the stock-logistics-availability repository ?

LoisRForgeFlow:
@rousseldenis Yes, not a problem for us. BTW and for future reference, is there a place where it is documented the logic to select the proper place for logistics modules?

@EvaSForgeFlow @JoanSForgeFlow Could you move this migration to the repository suggested?

rousseldenis:
> @rousseldenis Yes, not a problem for us. BTW and for future reference, is there a place where it is documented the logic to select the proper place for logistics modules?

For that one in particular, there is such a description: 

image


For the other, I have a todo to fill in the missing one :smiley:

LoisRForgeFlow:
> > @rousseldenis Yes, not a problem for us. BTW and for future reference, is there a place where it is documented the logic to select the proper place for logistics modules?
> 
> For that one in particular, there is such a description:
> 
> image
> 
> For the other, I have a todo to fill in the missing one smiley

This module is aimed to create procurements out of this "availability analysis/projection", so the main purpose is not giving availability information but to create transfers in the warehouse. Does it still make sense to move it there?

rousseldenis:
> > > @rousseldenis Yes, not a problem for us. BTW and for future reference, is there a place where it is documented the logic to select the proper place for logistics modules?
> > 
> > 
> > For that one in particular, there is such a description:
> > image
> > For the other, I have a todo to fill in the missing one smiley
> 
> This module is aimed to create procurements out of this "availability analysis/projection", so the main purpose is not giving availability information but to create transfers in the warehouse. Does it still make sense to move it there?

Nope indeed. But maybe more in workflow one no ?

LoisRForgeFlow:
> > > > @rousseldenis Yes, not a problem for us. BTW and for future reference, is there a place where it is documented the logic to select the proper place for logistics modules?
> > > 
> > > 
> > > For that one in particular, there is such a description:
> > > image
> > > For the other, I have a todo to fill in the missing one smiley
> > 
> > 
> > This module is aimed to create procurements out of this "availability analysis/projection", so the main purpose is not giving availability information but to create transfers in the warehouse. Does it still make sense to move it there?
> 
> Nope indeed. But maybe more in workflow one no ?

Between "workflow" and "warehouse" I always doubt, I let you have the final word if you have it more clear.

rousseldenis:
@LoisRForgeFlow @EvaSForgeFlow Keep it here

stock_pull_list/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+The pull list checks the stock situation at the given location and calculates
```
rousseldenis, line 1: My question here: in your flows, do you generate procurements that may generate purchases or just moves from other locations ? @EvaSForgeFlow @LoisRForgeFlow 

Because it makes me think about this : https://github.com/OCA/stock-logistics-orderpoint/pull/3

jbaudoux: It looks like it's serving the same purpose but with a selective approach. https://github.com/OCA/stock-logistics-orderpoint/pull/3 replenish all products with a need on the monitored location either automatically as soon as there is a new need, or manually by launching the replenishment action. From what I see, this one allows to choose which product to replenish in a wizard. Moreover, you can set a date boundary. It's the "select" trigger mode you don't have in https://github.com/OCA/stock-logistics-orderpoint/pull/3
Would be interesting to understand in which context this module is used.

LoisRForgeFlow: @rousseldenis @jbaudoux I have reviewed that module as I did not know it, it is indeed very interesting, thanks for the pointer. The need behind seems pretty similar, however as J-E has pointed out the pull list is based on a wizard in which you can filter more, use different routes and cherry-pick what you want to do now.

The main use case for us, is to create in a controlled way a big picking to move materials to the manufacturing area from the backstock. Another use case is moving products to the output location, or from the backstock to the picking area. In any case the use is always to generate internal transfers within the warehouse.

LoisRForgeFlow:
Superceded by https://github.com/OCA/stock-logistics-warehouse/pull/2093

PR #1766, opened 2023-06-20
[12.0][IMP] stock_request: add route helper field to the stock request order

antonioburic:
- adds a route_id field to the stock request orders, serving as a helper field to set a default value when entering the stock requests

image

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -109,10 +109,42 @@ class StockRequestOrder(models.Model):
         readonly=True,
     )
 
+    route_id = fields.Many2one('stock.location.route', string='Route',
+                               domain="[('id', 'in', route_ids)]",
+                               ondelete='restrict')
+
+    route_ids = fields.Many2many(
+        'stock.location.route', string='Routes',
+        compute='_compute_route_ids',
+        readonly=True,
+    )
+
     _sql_constraints = [
         ('name_uniq', 'unique(name, company_id)',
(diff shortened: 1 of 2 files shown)


stock_request/models/stock_request_order.py
```diff
@@ -109,11 +109,43 @@ def _get_default_requested_by(self):
...
+                parents = record.get_parents().ids
+                record.route_ids = routes.filtered(lambda r: any(
+                    p.location_id.id in parents for p in r.rule_ids))
+
+    def get_parents(self):
+        location = self.location_id
+        result = location
+        while location.location_id:
+            location = location.location_id
+            result |= location
```
rousseldenis, line 146:
```suggestion
      location_ids = set()
      while (...):
            (...)
            location_ids.add(location.id)
       return location.browse(location_ids)

File: stock_request/models/stock_request_order.py [hunk 2]
```diff
@@ -109,11 +109,43 @@ def _get_default_requested_by(self):
...
+    )
+
     _sql_constraints = [
         ('name_uniq', 'unique(name, company_id)',
          'Stock Request name must be unique'),
     ]
 
+    @api.depends('warehouse_id', 'location_id')
+    def _compute_route_ids(self):
+        route_obj = self.env['stock.location.route']
```
rousseldenis, line 129:
```suggestion
        route_browse = self.env['stock.location.route'].browse()
```

PR #1701, opened 2023-04-26
[16.0][MIG] stock_secondary_unit

remihb:
With changes introduced by #1606 for 15.0 (Compatibility with product variants)

Wait for https://github.com/OCA/product-attribute/pull/1325 to be merged

PR #1769, opened 2023-06-21
[16.0] [MIG] stock_secondary_unit

BT-aleonard:
- The PR https://github.com/OCA/stock-logistics-warehouse/pull/1701 is not attended, so I replaced it with this PR.


BT-aleonard:
@sergio-teruel can we replace this PR in the issues file?

rousseldenis:
/ocabot migration stock_secondary_unit.

@BT-aleonard It's usually better to add the link to the depending PR in this one description. Thanks

BT-dmoreno (approved):

BT-aleonard:
The PR https://github.com/OCA/stock-logistics-warehouse/pull/1701 is not attended, so I replaced it with this PR.

BT-aleonard:
@sergio-teruel Can you please review this PR, thanks.

yajo (approved):

rousseldenis (approved):
Code review

stock_secondary_unit/models/stock_move.py
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    product_uom_qty = fields.Float(
+        store=True,
+        readonly=False,
+        compute="_compute_product_uom_qty",
+        copy=True,
+        precompute=True,
+    )
+
```
yajo, line 22: Please add this code here:

```suggestion
    secondary_uom_qty = fields.Float(
        compute="_compute_secondary_uom_qty",
        precompute=True,
        store=True,
    )

```

stock_secondary_unit/models/stock_move.py
```diff
@@ -0,0 +1,70 @@
...
+class StockMoveLine(models.Model):
+    _inherit = ["stock.move.line", "product.secondary.unit.mixin"]
+    _name = "stock.move.line"
+    _secondary_unit_fields = {"qty_field": "qty_done", "uom_field": "product_uom_id"}
+
+    qty_done = fields.Float(
+        store=True, readonly=False, compute="_compute_qty_done", precompute=True
+    )
+
```
yajo, line 47: Also here:

```suggestion
    secondary_uom_qty = fields.Float(
        compute="_compute_secondary_uom_qty",
        precompute=True,
        store=True,
    )

```

yajo (changes requested):
After double-checking, you can see in the CI you can see you have these warnings:

```log
2023-06-22 07:27:17,318 245 WARNING odoo py.warnings: /opt/odoo/odoo/fields.py:802: UserWarning: Field stock.move.product_uom_qty cannot be precomputed as it depends on non-precomputed field stock.move.secondary_uom_qty
2023-06-22 07:27:17,322 245 WARNING odoo py.warnings: /opt/odoo/odoo/fields.py:802: UserWarning: Field stock.move.line.qty_done cannot be precomputed as it depends on non-precomputed field stock.move.line.secondary_uom_qty
```

Follow the suggestions below to fix the situation, please.

sergio-teruel:
Minor commente... I hava commented in this PR https://github.com/OCA/sale-workflow/pull/2645 you can add the precompute attribute in mixin model in product_secondary_unit module

yajo:
I was unsure about that because a lot of modules depend on it. I'm afraid it could break other behaviors unexpectedly...

sergio-teruel:
Hi @BT-aleonard Can you take into account this PR https://github.com/OCA/stock-logistics-warehouse/pull/1606 from V15??

OpenCode (approved):

stock_secondary_unit/models/stock_move.py
```diff
@@ -0,0 +1,70 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            move = self.env["stock.move"].browse(vals.get("move_id", False))
+            if move.secondary_uom_id:
+                uom = self.env["uom.uom"].browse(vals["product_uom_id"])
+                factor = move.secondary_uom_id.factor * uom.factor
+                move_line_qty = vals.get("product_uom_qty", vals.get("qty_done", 0.0))
```
acsonefho, line 55: This `product_uom_qty` doesn't exist in v`16.0`
Should be replaced by `reserved_uom_qty`

https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_move.py#L1449

acsonefho (changes requested):

stock_secondary_unit/report/report_deliveryslip.xml
```diff
@@ -0,0 +1,44 @@
...
+            </th>
+        </xpath>
+    </template>
+    <template
+        id="stock_report_delivery_has_serial_move_line"
+        inherit_id="stock.stock_report_delivery_has_serial_move_line"
+    >
+        <xpath expr="//span[@t-field='move_line.product_id']/.." position="after">
+            <td>
+                <span t-field="move_line.secondary_uom_qty" />
```
acsonefho, line 39: If your `stock.picking` is validated, the table header is displayed (`"Secondary Qty"`) but the layout is not correct, it's shifted. In others states, the layout is correct.
image

acsonefho (changes requested):

rousseldenis:
@BT-aleonard Could you attend comments ? Moreover, this deserves some little tests. Thanks

pedrobaeza:
Please include #1606

pedrobaeza:
Please cherry-pick #1948

rousseldenis:
@BT-dmoreno Maybe someone still active to continue this ?

stock_secondary_unit/models/stock_move.py
Rad0van, line 55: I am trying to wrap my head around this because whatever I do the lines end up with 0 secondary qty and None secondary uom id. Why? Because move.secondary_uom_id is None. I cannot find any piece of code that would ever be attempting to write any value there. And I have looked and tested versions 15 and 14 as well. Any insights on this?

Rad0van: So when I confirm Purchase Order and have a look at associated stock picking this is what I get always. Or am I expecting wrongly to have secondary qty and uom displayed here?
image

Rad0van: In my opinion stock.move is missing something like this:
```
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            product = self.env["product.product"].browse(vals.get("product_id", False))
            if product:
                vals.update(
                    {
                        "secondary_uom_id": product.stock_secondary_uom_id.id
                        or product.product_tmpl_id.stock_secondary_uom_id.id,
                    }
                )
        return super().create(vals_list)

```

After that the result is what (at least) I expect:
image

stock_secondary_unit/models/stock_move.py
```diff
@@ -0,0 +1,70 @@
...
+        self._onchange_helper_product_uom_for_secondary()
+
+    @api.model
+    def _prepare_merge_moves_distinct_fields(self):
+        """Don't merge moves with distinct secondary units"""
+        distinct_fields = super()._prepare_merge_moves_distinct_fields()
+        distinct_fields += ["secondary_uom_id"]
+        return distinct_fields
+
+
```
Rad0van, line 38: This ensures calculation of secondary qty. See comments in main thread.

```suggestion

        @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            product = self.env["product.product"].browse(vals.get("product_id", False))
            if product:
                vals.update(
                    {
                        "secondary_uom_id": product.stock_secondary_uom_id.id
                        or product.product_tmpl_id.stock_secondary_uom_id.id,
                    }
                )
        return super().create(vals_list)

```

jcadhoc:
Hello!
What is the state of this pr? I tried and it works fine more or less but i have an apreciation to make.

Is it active?

pilarvargas-tecnativa:
Please include https://github.com/OCA/stock-logistics-warehouse/pull/2080

jbaudoux:
Replaced by https://github.com/OCA/stock-logistics-warehouse/pull/1985

jbaudoux:
> Please include #2080

@pilarvargas-tecnativa Can you make a fwd port of that, I'm merging what was already done so far

pilarvargas-tecnativa:
> @pilarvargas-tecnativa Can you make a fwd port of that, I'm merging what was already done so far

Done! Thanks!

https://github.com/OCA/stock-logistics-warehouse/pull/2171

stock_secondary_unit/models/stock_move.py
rrebollo: Hello, I'm in a similar situation to yours. After reviewing [OCA/product-attribute#377](https://github.com/OCA/product-attribute/pull/377), I gained some insights, but I still suspect that we might both have incorrect expectations. For example, your suggestion doesn't account for the fact that secondary unit fields can be optional.

I believe the issue lies here:  
- [`product_secondary_unit_mixin.py#L38`](https://github.com/OCA/product-attribute/blob/439e1559e873d0445c323e5a31f0360fdaa88e46/product_secondary_unit/models/product_secondary_unit_mixin.py#L38)  
- [`product_template.py#L18-L23`](https://github.com/OCA/product-attribute/blob/439e1559e873d0445c323e5a31f0360fdaa88e46/product_secondary_unit/models/product_template.py#L18-L23)  

The use of `@api.model` ensures that `self` will always be the model, not a record. Consequently, `secondary_oum_ids` will always evaluate to `None`, which might be causing the problem.

The fact that everyone here seems to ignore your concerns makes me think we might be misunderstanding the addon's purpose or intended usage.

ping: @sergio-teruel @lk-eska @rousseldenis @jbaudoux :face_with_head_bandage:

rrebollo: Take a look at this: [Commit 35374e39586c](https://github.com/OCA/stock-logistics-warehouse/commit/35374e39586c53b8fa6933f393eae7e0e2d06e54). There's no doubt the behavior is intentional. Why? 🤔 ... That was never explained.

PR #1775, opened 2023-06-23
[15.0][MIG] product_route_profile

bosd:
Standard migration.


bosd:
Generated the icon with command:
```
oca-gen-addon-icon --addon-dir=. --format png --icon-set-params "fa fa-route" --icon-extra='{"fa fa-box fa-lg": {"position": "absolute", "top": "100px", "left": "5px", "font-size": "28px"}}' --icon-color yellow
```

bosd:
@pedrobaeza pre-commit was broken here..
But all good now. :+1:

bosd:
@francesco-ooops Can you please review?

francesco-ooops (commented):
Hi @bosd sorry but I don't understand well what this module does. I can only see a "Route Profile" menu where I can create a record and assign a name and nothing else

bosd:
> Hi @bosd sorry but I don't understand well what this module does. I can only see a "Route Profile" menu where I can create a record and assign a name and nothing else

After you created a route profile, It can be assigned to a product.
It is usefull to have a products route configured with one click.
Specifically when multiple routes can be applied to a product.


e.g. with one click, the correct values are set according to the defined route_profile.
It can be set as mandatory. to prevent users erros selecting the wron routes.

StefanRijnhart (approved):
Thanks! Using this as the basis of the migration to 17.0 at https://github.com/OCA/stock-logistics-warehouse/pull/2097

product_route_profile/models/product_template.py
```diff
@@ -28,7 +28,7 @@ class ProductTemplate(models.Model):
     @api.depends("route_profile_id", "force_route_profile_id")
     @api.depends_context("company")
     def _compute_route_ids(self):
-        for rec in self:
+        for rec in self.sudo():
```
rousseldenis, line 31: @bosd Why is this needed ?

bosd: I think it was needed to pass the tests.

Dranyel-Bosd (approved):
LGTM :100:

bosd:
Can we get a merge :pray:

StefanRijnhart:
@francesco-ooops @rousseldenis Is this PR ok for you now?

product_route_profile/models/product_template.py
rousseldenis: @bosd I still need an answer on this as just putting this for tests is not wanted.

bosd: Good question. I don't remember, it has been a while :)
Maybe better to do functional test it without this. To see if it only is relevant for the tests. Or also required to achieve the functionality.

StefanRijnhart: Given that different routes may apply depending on the current company, it might be the case that there are route profiles (with an empty company_id) that provide routes for different companies. In that case it makes sense to escalate privileges when copying routes so that routes are copied even if the user does not have access to the all the companies that the routes belong to.

francesco-ooops:
> > Hi @bosd sorry but I don't understand well what this module does. I can only see a "Route Profile" menu where I can create a record and assign a name and nothing else
> 
> After you created a route profile, It can be assigned to a product. It is usefull to have a products route configured with one click. Specifically when multiple routes can be applied to a product.
> 
> e.g. with one click, the correct values are set according to the defined route_profile. It can be set as mandatory. to prevent users erros selecting the wron routes.

@bosd great, this should go in the docs then :)

etobella:
@rousseldenis can you confirm that we can proceed?

rousseldenis (approved):

etobella (approved):

PR #1781, opened 2023-06-29
[15.0][IMP] stock_inventory: do not delete inv adjustment groups if in process or done

AaronHForgeFlow:
I think it is weird you can delete inventory adjustments groups that are in progress or completed, I think they should be kept for traceability to what was done.
@ForgeFlow 
@DavidJForgeFlow

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -1,5 +1,5 @@
 from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo.exceptions import UserError, ValidationError
 from odoo.osv import expression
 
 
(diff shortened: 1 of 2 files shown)


stock_inventory/models/stock_inventory.py
```diff
@@ -214,3 +214,14 @@ def _check_one_product_in_product_selection(self):
                             " you are only able to add one product."
                         )
                     )
+
+    def unlink(self):
+        for adjustment in self:
+            if adjustment.state != "draft":
+                raise UserError(
+                    _(
+                        "You can only delete inventory adjustmnets groups in"
```
```suggestion
                        "You can only delete inventory adjustments groups in"
```

rousseldenis (approved):
Makes sense as it was the case in v<=14

https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_inventory.py#L83

LoisRForgeFlow (approved):
LGTM!

Maybe you could also add some ondelete="retrict" for stock.move.lines.inventory_id?

AaronHForgeFlow:
@LoisRForgeFlow  Nice suggestion, thanks

PR #1787, opened 2023-07-10
[15.0][FIX] stock_orderpoint_generator: fix test test_errors

pilarvargas-tecnativa:
The Size attribute is already defined in the point_of_sale module so when both modules are installed an error occurs in the tests. This is solved by changing the name of the attribute in the tests of the stock_orderpoint_generator module.

image

cc @Tecnativa TT41548

@pedrobaeza @victoralmau please review

### DIFF
--- a/stock_orderpoint_generator/tests/test_orderpoint_generator.py
+++ b/stock_orderpoint_generator/tests/test_orderpoint_generator.py
@@ -15,7 +15,9 @@ class TestOrderpointGenerator(TransactionCase):
         cls.orderpoint_model = cls.env["stock.warehouse.orderpoint"]
         cls.orderpoint_template_model = cls.env["stock.warehouse.orderpoint.template"]
         cls.product_model = cls.env["product.product"]
-        cls.attr = cls.env["product.attribute"].create({"name": "Size"})
+        cls.attr = cls.env["product.attribute"].create(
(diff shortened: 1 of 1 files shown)


stock_orderpoint_generator/tests/test_orderpoint_generator.py
```diff
@@ -15,7 +15,7 @@ def setUpClass(cls):
         cls.orderpoint_model = cls.env["stock.warehouse.orderpoint"]
         cls.orderpoint_template_model = cls.env["stock.warehouse.orderpoint.template"]
         cls.product_model = cls.env["product.product"]
-        cls.attr = cls.env["product.attribute"].create({"name": "Size"})
+        cls.attr = cls.env["product.attribute"].create({"name": "Test attribute"})
```
pedrobaeza: For avoiding more collapsings:

```suggestion
        cls.attr = cls.env["product.attribute"].create({"name": "stock_orderpoint_generator attribute"})
```

pilarvargas-tecnativa: thanks!

pedrobaeza (approved):
Fast-tracking as a test resiliency only:

/ocabot merge patch

PR #1788, opened 2023-07-11
16.0 mig stock reserve sale

hugo-cordoba:
[MIG] stock_reserve_sale: migrate from v13.0 to v16.0

Hi. 

After trying to push the developments in [version 14](https://github.com/OCA/stock-logistics-warehouse/pull/1696) and [version 16](https://github.com/OCA/stock-logistics-warehouse/pull/1695), and receiving no feedback, I have decided to create a PR and take responsibility for the migration. 

I am open to any kind of suggestions.

I have included improvements in the module that I find useful and improve the module at user and usability level.

We will find a button to access the reservations associated to the order and the reservation slip. They will only be visible if the quotation has reservations.

Within the reservations, we will also be able to access the associated reservation delivery note, the reservation movement and even release (undo) the reservation.

A button is added in the header, along with the rest of the buttons, to cancel the reservation made. When pressed, it will cancel the reservation movements and the reservation delivery note, so the corresponding smart buttons will no longer be visible.

The button to reserve stock will only be visible if there are units to be reserved. If all units have already been reserved, the button will disappear.


stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+                    is_stock_reservable = True
+            if sale.state not in ("draft", "sent"):
+                is_stock_reservable = False
+            sale.is_stock_reservable = is_stock_reservable
+            sale.has_stock_reservation = has_stock_reservation
+
+    has_stock_reservation = fields.Boolean(
+        compute="_compute_stock_reservation",
+        readonly=True,
+        multi="stock_reservation",
```
gdgellatly, line 34:
```suggestion
```

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+        store=True,
+        string="Has Stock Reservations",
+    )
+    is_stock_reservable = fields.Boolean(
+        compute="_compute_stock_reservation",
+        readonly=True,
+        multi="stock_reservation",
```
gdgellatly, line 41:
```suggestion
```

gdgellatly (commented):
Just initial install errors - functional review to come

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+        store=True,
+        string="Can Have Stock Reservations",
+    )
+    reserves_count = fields.Integer(compute="_compute_reserves_count")
+    all_lines_reserved = fields.Boolean(
```
gdgellatly, line 46: this will need a compute_sudo

`2023-08-14 21:04:18,128 1 WARNING mig odoo.modules.registry: sale.order: inconsistent 'compute_sudo' for computed fields: reserves_count, all_lines_reserved`

stock_reserve_sale/view/sale.xml
```diff
@@ -0,0 +1,140 @@
...
+            <xpath
+                expr="//field[@name='order_line']/tree/field[@name='price_subtotal']"
+                position="after"
+            >
+                <field name="reservation_ids" invisible="1" />
+                <field name="is_stock_reservable" invisible="1" />
+                <button
+                    name="%(action_sale_stock_reserve)d"
+                    type="action"
+                    icon="fa-lock"
```
gdgellatly, line 72: `2023-08-14 21:09:09,611 1 WARNING mig odoo.addons.base.models.ir_ui_view: A button with icon attribute (fa-lock) must have title in its tag, parents, descendants or have text`

stock_reserve_sale/view/sale.xml
```diff
@@ -0,0 +1,140 @@
...
+                    attrs="{'invisible': [('is_stock_reservable', '=', False)]}"
+                />
+                <button
+                    name="release_stock_reservation"
+                    type="object"
+                    icon="fa-undo"
```
gdgellatly, line 78: `2023-08-14 21:09:09,612 1 WARNING mig odoo.addons.base.models.ir_ui_view: A button with icon attribute (fa-undo) must have title in its tag, parents, descendants or have text`

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+            line.is_stock_reservable = reservable
+
+    @api.depends("order_id.state", "reservation_ids")
+    def _compute_is_readonly(self):
+        for line in self:
+            line.is_readonly = (
+                len(line.reservation_ids) > 0 or line.order_id.state != "draft"
+            )
+
+    reservation_ids = fields.One2many(
```
rousseldenis, line 72: Could you put fields on top ?

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+        block_on_reserve = ("product_id", "product_uom", "type")
+        update_on_reserve = ("price_unit", "product_uom_qty")
+        keys = set(vals.keys())
+        test_block = keys.intersection(block_on_reserve)
+        test_update = keys.intersection(update_on_reserve)
+        if test_block:
+            for line in self:
+                if not line.reservation_ids:
+                    continue
+                raise except_orm(
```
rousseldenis, line 96: except_orm has been deprecated a long time ago. Could you change to UserError ?

stock_reserve_sale/model/stock_reserve.py
```diff
@@ -10,10 +10,26 @@ class StockReservation(models.Model):
         "sale.order.line", string="Sale Order Line", ondelete="cascade", copy=False
     )
     sale_id = fields.Many2one(
-        "sale.order", string="Sale Order", related="sale_line_id.order_id"
+        "sale.order", string="Sale Order", store=True, related="sale_line_id.order_id"
     )
 
     def release_reserve(self):
         for rec in self:
             rec.sale_line_id = False
```
rousseldenis, line 18: `self.update({"sale_line_id": False})`

hugo-cordoba:
Thanks for the feedback @gdgellatly. In the afternoon, when I have some free time, I'll check it out!!!

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+        compute="_compute_reserves_count", store=True, default=False
+    )
+
+    def release_all_stock_reservation(self):
+        line_ids = [line.id for order in self for line in order.order_line]
+        lines = self.order_line.browse(line_ids)
```
FrancoMaxime, line 52: Hello, thanks for your works.
I think we can optimize this function.
[EDIT 11/09] Adds @gdgellatly suggestion instead of self.mapped("order_line")
```suggestion
        lines = self.order_line
```

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+    def action_view_reserves_stock_picking(self):
+        stock_picking = ""
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "stock.action_picking_tree_all"
+        )
+        stock_picking = (
+            self.env["stock.picking"]
+            .search([("origin", "=", self.name)])
+            .filtered(lambda a: a.state not in "cancel")
+        )
```
FrancoMaxime, line 96: I think we can make a search with both conditions.
```suggestion
        stock_picking = (
            self.env["stock.picking"]
            .search([
                ("origin", "=", self.name),
                ("state", "!=", "cancel")
            ])
        )
```

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+        ]
+        rules = StockRule.search(
+            [("route_id", "in", product_route_ids)],
+            order="route_sequence, sequence",
+            limit=1,
+        )
+        if not rules:
+            warehouse = self.order_id.warehouse_id
+            wh_routes = warehouse.route_ids
+            wh_route_ids = [route.id for route in wh_routes]
```
FrancoMaxime, line 30: We can optimize this
```suggestion
            wh_route_ids = warehouse.route_ids.ids
```

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+        "stock.reservation", "sale_line_id", string="Stock Reservation", copy=False
+    )
+    is_stock_reservable = fields.Boolean(
+        compute="_compute_is_stock_reservable", readonly=True, string="Can be reserved"
+    )
+    is_readonly = fields.Boolean(compute="_compute_is_readonly", store=False)
+
+    def release_stock_reservation(self):
+        reserv_ids = [reserv.id for line in self for reserv in line.reservation_ids]
+        reservations = self.env["stock.reservation"].browse(reserv_ids)
```
FrancoMaxime, line 82: We can optimize
```suggestion
        reservations = self.mapped("reservation_ids")
```

stock_reserve_sale/model/stock_reserve.py
```diff
@@ -0,0 +1,35 @@
...
+        stock_picking = ""
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "stock.action_picking_tree_all"
+        )
+        if self.sale_id:
+            stock_picking = (
+                self.env["stock.picking"]
+                .search([("origin", "=", self.sale_id.name)])
+                .filtered(lambda a: a.state not in "cancel")
+            )
```
FrancoMaxime, line 31: We can add both conditions in the search
```suggestion
            stock_picking = (
                self.env["stock.picking"]
                .search([
                    ("origin", "=", self.sale_id.name),
                    ("state", "!=", "cancel"),
                ])
            )
```

FrancoMaxime (commented):
Maybe we could improve these remark

stock_reserve_sale/model/sale.py
gdgellatly: Generally a != condition for a state field in search is a bad idea. The problem with these is often state fields are indexed and the other conditions aren't. But because of the generally bad distribution of states it is usually faster if you aren't using '=' for a state which is a small percentage, the queries can be quite bad. The search would have prefetched anyway, the cancelled records will be minimal. IDK, I feel like I would want testing on big databases first as one of the performance bottlenecks we always look for is exactly this suggestion. It is almost always better to reverse it to a state in type query or use filtered.

stock_reserve_sale/model/sale_order_line.py
gdgellatly: why mapped? surely just `self.reservation_ids`

stock_reserve_sale/model/sale.py
gdgellatly: not just `self.order_line`?

stock_reserve_sale/model/stock_reserve.py
gdgellatly: As above, I think a suggestion like this which could have big performance implications should at least be tested. I can do it, just not right now.

stock_reserve_sale/model/sale.py
gdgellatly: Just looking at this, the filtered clause is wrong anyway, it should be != instead of not in. But as it happens, in testing the suggested query on a database with 1m pickings performed basically identically as the original BEFORE the filtered clause and planned identically using standard Odoo indexes, so I think it is a good suggestion.

stock_reserve_sale/model/sale.py
FrancoMaxime: If self is a recordset, i don't remember if odoo gives all order_line for all record without a mapped.

stock_reserve_sale/model/sale.py
FrancoMaxime: @gdgellatly Thanks for your tests

stock_reserve_sale/model/sale.py
rousseldenis: @FrancoMaxime In v16, you can call the relation field on multi recordset

stock_reserve_sale/model/sale_order_line.py
FrancoMaxime: See comment https://github.com/OCA/stock-logistics-warehouse/pull/1788#discussion_r1321060295

stock_reserve_sale/model/sale.py
```diff
@@ -41,13 +42,63 @@ def _compute_stock_reservation(self):
...
+            ),
+            fields=["sale_id"],
+            groupby=["sale_id"],
+        )
+        lines = self.order_line.filtered(
+            lambda l: l.product_id.detailed_type != "service"
+        )
+        for order in self:
+            if reserve_ids:
+                order.reserves_count = reserve_ids[0]["sale_id_count"]
```
rousseldenis, line 71: @HugoCordobaLeal Is that correct to take only the first one ???

stock_reserve_sale/model/sale.py
FrancoMaxime: @gdgellatly Have you tried with a condition like "state", "in", "[....]" ? Did they give the same result (in time)?

stock_reserve_sale/model/sale.py
FrancoMaxime: @rousseldenis Thanks

stock_reserve_sale/model/sale.py
gdgellatly: Doesn't really matter I think. The existing code is clearly a mistake. But also the planner will have statistics on uniqueness and origin field is indexed so regardless, for this query, I think no matter the state condition, it will pick the origin index.

jappi00:
@HugoCordobaLeal are you working on this? If you want I can help on this to implement the last change requests.

stock_reserve_sale/wizard/sale_stock_reserve.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class SaleStockReserve(models.TransientModel):
+    _name = "sale.stock.reserve"
+    _description = "Sale Stock Reserve"
+
+    @api.model
+    def _default_location_id(self):
+        return self.env["stock.reservation"].get_location_from_ref(
+            "stock.stock_location_stock"
```
acsonefho, line 13: That'll work only if you're connected using the default company (ID 1).
Should be something like:

```
    @api.model
    def _default_location_id(self):
        domain = [
            '|',
            ('company_id', '=', self.env.company.id),
            ('company_id', '=', False),
        ]
        return self.env["stock.warehouse"].search(domain, limit=1).lot_stock_id
```

stock_reserve_sale/wizard/sale_stock_reserve.py
```diff
@@ -0,0 +1,150 @@
...
+        self.ensure_one()
+        close = {"type": "ir.actions.act_window_close"}
+        active_model = env.context.get("active_model")
+        active_ids = env.context.get("active_ids")
+        if not (active_model and active_ids):
+            return close
+
+        if active_model == "sale.order":
+            sales = env["sale.order"].browse(active_ids)
+            line_ids = [line.id for sale in sales for line in sale.order_line]
```
acsonefho, line 144: Simplest way to read that:
`line_ids = sales.order_line.ids`

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def release_all_stock_reservation(self):
+        line_ids = [line.id for order in self for line in order.order_line]
+        lines = self.order_line.browse(line_ids)
+        lines.release_stock_reservation()
+        return True
+
+    def _compute_reserves_count(self):
+        reserve_ids = self.env["stock.reservation"]._read_group(
+            domain=expression.AND(
```
acsonefho, line 58: useless.
`domain=[("sale_id", "in", self.ids)]`

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _get_line_rule(self):
+        """Get applicable rule for this product
+        Reproduce get suitable rule from procurement
+        to predict source location
+        """
+        StockRule = self.env["stock.rule"]
+        product = self.product_id
+        product_route_ids = [
+            x.id for x in product.route_ids + product.categ_id.total_route_ids
```
acsonefho, line 20: More readable:
`product_route_ids = (product.route_ids + product.categ_id.total_route_ids).ids`

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+            wh_routes = warehouse.route_ids
+            wh_route_ids = [route.id for route in wh_routes]
+            domain = [
+                "|",
+                ("warehouse_id", "=", warehouse.id),
+                ("warehouse_id", "=", False),
+                ("route_id", "in", wh_route_ids),
+            ]
+            rules = StockRule.search(domain, order="route_sequence, sequence")
+        if rules:
```
acsonefho, line 38: This `if` section should be inside of the previous `if` section.
Because you already have a `limit=1`.
And this is useless: you should add a `limit=1` in the `search(...)` on the line before.

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+            fields.first(rules)
+        return False
```
acsonefho, line 40: This function should return a rule (`_get_line_rule(...)`) but return `False`.

So you're looking for a rule but you never `return` it.

And if you really don't find a rule, you should return an empty recordset.

stock_reserve_sale/model/sale_order_line.py
```diff
@@ -0,0 +1,139 @@
...
+    reservation_ids = fields.One2many(
+        "stock.reservation", "sale_line_id", string="Stock Reservation", copy=False
+    )
+    is_stock_reservable = fields.Boolean(
+        compute="_compute_is_stock_reservable", readonly=True, string="Can be reserved"
+    )
+    is_readonly = fields.Boolean(compute="_compute_is_readonly", store=False)
+
+    def release_stock_reservation(self):
+        reserv_ids = [reserv.id for line in self for reserv in line.reservation_ids]
```
acsonefho, line 81: More readable:
`reserv_ids = self.reservation_ids.ids`
But it's completely useless as you browse them just after.
You should do this:
`reservations = self.reservation_ids`

acsonefho (changes requested):
Thanks for this migration :smiley: 
:warning: Some importants changes/fixes are required before merging this PR

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        lines = self.order_line.filtered(
+            lambda l: l.product_id.detailed_type != "service"
+        )
+        for order in self:
+            if reserve_ids:
+                order.reserves_count = reserve_ids[0]["sale_id_count"]
+            else:
+                order.reserves_count = 0
+            if order.reserves_count == len(lines):
```
acsonefho, line 74: How is this possible?
That can not happens in a multi-recordset computation.
If this compute is called on a recordset with more than one `sale.order`, this will be never True.

acsonefho (changes requested):

stock_reserve_sale/model/sale.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def action_view_reserves_stock_picking(self):
+        stock_picking = ""
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "stock.action_picking_tree_all"
+        )
+        stock_picking = (
+            self.env["stock.picking"]
+            .search([("origin", "=", self.name)])
+            .filtered(lambda a: a.state not in "cancel")
```
acsonefho, line 95: This is not correct at all.
The value of state could be "draft", "cancel",...

As is, this mean:
`"draft" in ("c", "a", "n", "c", "e", "l")` ? => `True/False`

It should be:
```
# Use limit=1 because we use the record with .id later
stock_picking = self.env["stock.picking"].search([("origin", "=", self.name), ("state", "!=", "cancel")], limit=1)
```

acsonefho (changes requested):

stock_reserve_sale/model/stock_reserve.py
```diff
@@ -0,0 +1,35 @@
...
+    def action_view_reserves_stock_picking_reservation(self):
+        stock_picking = ""
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "stock.action_picking_tree_all"
+        )
+        if self.sale_id:
+            stock_picking = (
+                self.env["stock.picking"]
+                .search([("origin", "=", self.sale_id.name)])
+                .filtered(lambda a: a.state not in "cancel")
```
acsonefho, line 30: This is not correct. Like my previous comment

acsonefho (changes requested):

stock_reserve_sale/model/stock_reserve.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+        if stock_picking:
+            view_id = self.env.ref("stock.view_picking_form").id
+            action.update(views=[(view_id, "form")], res_id=stock_picking.id)
+            return action
```
acsonefho, line 35: This should be outstide of this `if` section.

acsonefho (changes requested):

rousseldenis:
@username-hugo Do you plan to attend comments ?

rousseldenis:
@username-hugo

astirpe:
I'm working on the migration of this module to V17: #2004

rousseldenis:
@hugo-cordoba

PR #1790, opened 2023-07-11
[14.0][IMP] stock_picking_volume: Move _get_volume_for_qty to product.product

mt-software-de:


mt-software-de:
cc @jbaudoux @raumschmiede-joshuaL

stock_picking_volume/models/stock_move.py
```diff
@@ -23,27 +23,10 @@ def _compute_volume(self):
             qty = move.product_uom_qty
             if move.state in ("partially_available", "assigned"):
                 qty = move.reserved_availability
-            new_volume = move._get_volume_for_qty(qty=qty)
+            new_volume = move.product_id._get_volume_for_qty(qty, move.product_uom)
```
lmignon, line 26: @mt-software-de We can move the compute method to the product.product model but we could still keep the method `_get_volume_for_qty` defined on the stock.move forwarding the call to the product.  In this way we keep an extension point if we want to adapt the way the weight is computed according to information on the stock.move and we don't break the current api.

mt-software-de: The compute method is still defined on stock.move. 
We can keep _get_volume_for_qty as it is on the stock.move but i can't see any advantage of that. 
Also, the development status beta would allow changing the current API. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst#id2 

If we would keep it, i would change it to something like this.
image

stock_picking_volume/models/product_product.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = "product.product"
+
+    def _get_volume_for_qty(self, qty, from_uom=None):
+        """Return the volume for the given qty.
+
+        This method is meant to be inherited to change the volume
+        computation for a specific product.
+
+        qty: float quantity to compute the volume for.
+
```
jbaudoux, line 18: missing the new from_uom in the docstring

lmignon (approved):

stock_picking_volume/models/stock_move.py
mt-software-de: Ok then i will leave it as it is

rousseldenis (approved):
Code review

jbaudoux (approved):

PR #1792, opened 2023-07-12
[FIX]stock_quant_manual_assign: bug with boolean

jcadhoc:


jcadhoc:
Hi,

I've made this PR because this module is not working fine, just testing it with any transfer the error of "Create/update: a mandatory field is not set" appears and works bad. We found out that the widget "boolean_toggle" are not working fine in the wizard, so we remove it and works fine.

BT-jmichaud (approved):
Yes it works with this fix. Tested!
@jcadhoc Thank you!

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -12,9 +12,9 @@
                         <field name="package_id" groups="stock.group_tracking_lot" />
                         <field name="owner_id" groups="stock.group_tracking_owner" />
                         <field name="location_id" />
-                        <field name="on_hand" force_save="1" />
-                        <field name="reserved" force_save="1" />
-                        <field name="selected" widget="boolean_toggle" />
```
pedrobaeza, line 17: Why removing the widget?

BT-jmichaud: With the widget, there is an error message "Create/update: a mandatory field is not set". I tested new changes without removing the widget because I was curious too.

pedrobaeza: And then why remove the `force_save` switch? Anyway, the bug seems to be in the widget itself, or in the form missing a field. This is only a workaround removing an interesting feature.

BT-jmichaud: I did not test without force_save. 

But as far as I saw, the toggle and force_save never existed in older versions. It was added in version 16.0 and never really worked with this module. I don't know why.

BT-jmichaud:
Can someone please approve this PR? The actual module is really not working.

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
AungKoKoLin1997: I don't think `force_save` leads to issue. As far as I checked, I don't see the process that remove `assign_wizard` field when we change boolean field. So, I think the bug seems to be in the widget.

ioans73:
Hi, what is the status of this PR?

hrahuu:
@jcadhoc @pedrobaeza I wonder why this fix never got committed to 16.0? The bug exists still in 16.0 and without this fix is unusable.

BT-jmichaud:
> @jcadhoc @pedrobaeza I wonder why this fix never got committed to 16.0? The bug exists still in 16.0 and without this fix is unusable.

Completely agree.

AungKoKoLin1997:
Can this PR reopen? We need this fix.

rousseldenis:
> Hi,
> 
> I've made this PR because this module is not working fine, just testing it with any transfer the error of "Create/update: a mandatory field is not set" appears and works bad. We found out that the widget "boolean_toggle" are not working fine in the wizard, so we remove it and works fine.

The problem is not the wizard boolean widget. Your error message is not complete. The field that is not filled in is 'Move'. So, in fact, when you click on the boolean widget, you do a write() on the wizard line. You should ask yourself why is the move_id field not filled in ?

AungKoKoLin1997:
> You should ask yourself why is the move_id field not filled in ?

@rousseldenis  I am doing functional test and I don't think this is about move_id. This is about the connection of between one2many field(`quants_lines`) in `assign.manual.quants` and inverse field(`assign_wizard`) in `assign.manual.quants.lines`.

Screenshot 2024-05-13 at 16 38 29

AungKoKoLin1997 (approved):
I think removing `force_save` is not related with the issue but I am not against it because it is not necessary.

rousseldenis:
@jcadhoc @AungKoKoLin1997 @BT-jmichaud The real solution is to add this option to the widget:

`options="{'autosave': False}"`

Check this : https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/views/fields/boolean_toggle/boolean_toggle_field.js#L21

AungKoKoLin1997:
> The real solution is to add this option to the widget:
> 
> `options="{'autosave': False}"`

@rousseldenis Thank you. I can confirm this is the real solution.

AungKoKoLin1997:
It seems like the author is not currently active, and we need this fix ASAP because without it, the feature is unusable.
So, I have superseded it with https://github.com/OCA/stock-logistics-warehouse/pull/2035.

@jcadhoc 
If you want to continue with this PR, I will close mine.

PR #1797, opened 2023-07-17
[14.0][FIX] stock_location_package_restriction for backorder

TDu:
Only check the restriction for moves actually being done.

### DIFF
--- a/stock_location_package_restriction/models/stock_move.py
+++ b/stock_location_package_restriction/models/stock_move.py
@@ -31,8 +31,9 @@ class StockMove(models.Model):
         location_packages = {
             g["location_id"][0]: set(g["package_id"]) for g in quants_grouped
         }
+        lines_being_processed = self.move_line_ids.filtered(lambda line: line.qty_done)
         for location, move_lines in groupby(
-            self.move_line_ids, lambda m: m.location_dest_id
+            lines_being_processed, lambda m: m.location_dest_id
         ):
             if not location.package_restriction:
                 continue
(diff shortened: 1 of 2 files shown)


stock_location_package_restriction/models/stock_move.py
```diff
@@ -31,8 +31,9 @@ def _check_location_package_restriction(self):
         location_packages = {
             g["location_id"][0]: set(g["package_id"]) for g in quants_grouped
         }
+        lines_being_processed = self.move_line_ids.filtered(lambda line: line.qty_done)
```
rousseldenis: @TDu Is this working when not using the detailed operations and quantity_done directly on stock.move ?

TDu: Thanks for your review and the answer is no, it is not working...

jbaudoux (approved):
LGTM

stock_location_package_restriction/models/stock_move.py
jbaudoux: It should also work as there are always move lines. See https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L358

TDu: My quick test was changing the unit test to update the quantity done  on the stock.move instead of the move line.
And it failed with 
`A package is mandatory on the location BWH/Stock/TestLocation2. You cannot move the product Wood without a package`
Which does not seem related to your question.

TDu:
I have added options on the check function that I need from another PR.
No changes in functionality.

TDu:
Rebased

jbaudoux:
@rousseldenis Good for you ?

rousseldenis (approved):

PR #1807, opened 2023-07-28
[IMP] stock_helper: Stop stock.rule's _run_pull via domain given by context

mt-software-de:
Tests are missing

### DIFF
--- a/stock_helper/models/__init__.py
+++ b/stock_helper/models/__init__.py
@@ -1,1 +1,2 @@
 from . import stock_location
+from . import stock_rule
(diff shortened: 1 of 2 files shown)


rousseldenis:
@mt-software-de I think stock_helper is not a good idea module at the beginning as you can put everything in it and having no idea what it does really. It's not speaking for users and technically borderline.

So, please put this in a separate module.

That's being said, wouldn't this behavior change lead to unwanted side effects ?

mt-software-de:
> @mt-software-de I think stock_helper is not a good idea module at the beginning as you can put everything in it and having no idea what it does really. It's not speaking for users and technically borderline.
> 
> So, please put this in a separate module.
> 
> That's being said, wouldn't this behavior change lead to unwanted side effects ?

yes i can move this.
I can lead to unwanted side effects, if you are using it.
It is needed for sale_stock_prebook
It is nearly the same logic as it is implemented here, https://github.com/OCA/wms/blob/fd917e896659271d1dfa36493510a1b6b0791dd0/stock_available_to_promise_release/models/stock_rule.py#L24
I this PR will be merged i will create also a PR to switch to the new module then

PR #1809, opened 2023-08-03
[ADD] product_uom_merge

JasminSForgeFlow:
Module for merge duplicate uoms

cc @ForgeFlow


AaronHForgeFlow (commented):
Tested with done stock moves. Works good. Thanks!

AaronHForgeFlow (approved):
Approved as green now and functional tests done

base_uom_merge/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Base Product UOMs Merge",
+    "summary": "Merge duplicate product UOMs",
+    "version": "15.0.1.0.0",
+    "license": "LGPL-3",
```
AaronHForgeFlow:
```suggestion
    "license": "AGPL-3",
```
based on https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py

JasminSForgeFlow: Updated, Thanks

base_uom_merge/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+We can merge duplicates products into single product
```
OriolMForgeFlow:
```suggestion
We can merge duplicates UOMs into single UOM.
```

base_uom_merge/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+Select uoms from same cateogry then click on Merge UOMs in Action menu.
```
OriolMForgeFlow:
```suggestion
Select uoms from same category then click on Merge UOMs in Action menu.
```

base_uom_merge/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Product UOMs Merge",
+    "summary": "Merge duplicate product UOMs",
+    "version": "15.0.1.0.0",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "category": "Warehouse Management",
```
OriolMForgeFlow: I would change the category field, as this module is not exclusively for Warehouse Management. It can be used from the Sales application without requiring the use of a warehouse, as it no depends to "stock"

Maybe change to category "Product".

OriolMForgeFlow (commented):
Some small comments

Functional review: it works properly

base_uom_merge/readme/DESCRIPTION.rst
JasminSForgeFlow: Updated, Thanks

base_uom_merge/readme/USAGE.rst
JasminSForgeFlow: Updated, Thanks

base_uom_merge/__manifest__.py
JasminSForgeFlow: Update category as per dependent module Product's category.

Thanks for suggestion

base_uom_merge/README.rst
```diff
@@ -0,0 +1,3 @@
+===================
+Base Products Merge
```
OriolMForgeFlow: Maybe change to Base Product UOMs Merge

base_uom_merge/tests/test_base_uom_merge.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2023 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import UserError
+from odoo.tests.common import TransactionCase
+
+
+class TestMergeUOM(TransactionCase):
+    def setUp(self):
```
OriolMForgeFlow: Change to setUpClass according to the migration rules to v15

base_uom_merge/README.rst
JasminSForgeFlow: Updated, Thanks

base_uom_merge/tests/test_base_uom_merge.py
JasminSForgeFlow: Updated, Thanks

OriolMForgeFlow (approved):
LGTM!

rousseldenis:
@JasminSForgeFlow My 10k€ question: is this should be here or in product-attribute repository as it does not depends on stock ?

hbrunn (changes requested):
you should either forbid merging UoMs with different factors too, or add code that converts the quantity of quants. Otherwise you'll get wrong stock levels, as quants always refer to the product's standard UoM. Maybe a big red warning also suffices for cases where stock is wrong anyways and the merging is the first step to get it right?

Imagine you have product A with UoM km and product B with UoM m. If you now go and merge km into m, your stock levels of product A will be off by factor 1000.

I'm working on a [generic merge addon](https://github.com/OCA/server-ux/pull/625) where my customer has been bitten by a very similar issue when merging products, there also quants suddenly refer to a different UoM and are potentially off by the factor difference of the merged products' UoMs.

My fix is to convert quantities before merging: https://github.com/OCA/server-ux/pull/625/files#diff-b166ff54d6536b547b2dbad5be3664f2b7926c93fda8cbb3ddf38e312037aa78R23

@rousseldenis that would then also answer the 10k question, if you want to implement the conversion here, you'd need a dependency on stock.

rousseldenis:
> @rousseldenis that would then also answer the 10k question, if you want to implement the conversion here, you'd need a dependency on stock.

@hbrunn I haven't look at the implementation yet :sweat_smile:  but indeed this needs to be more coercive. In fact, your module depends on stock but could be split (supplierinfo in product / quant in stock ones).

hbrunn:
@JasminSForgeFlow I had to implement UoM support for the PR above anyways, so it might be just a little more work to simply lift this to v15 (I expect you just have to change the version number and some field names), then you get all the checks and conversions I talked about for free. And the possibility to merge countries, states and whatever else users tend to duplicate by misspelling things in many2one fields with inline create activated.
Note this is *not* meant to bully you into doing this, I'll just as happily review an update here where you add the conversion/check/warning. I just think you/your customers would get much more options this way for free.

@rousseldenis let's discuss stuff related to my PR there such as not to hijack the discussion here :-)

JasminSForgeFlow:
Closing PR

PR #1820, opened 2023-08-24
[FIX] stock_reserve - Notes field gets squashed into unreadable width due to v16 quirk

chrisandrewmann:
Notes field is not properly readable due to squashed width.

image

### DIFF
--- a/stock_reserve/view/stock_reserve.xml
+++ b/stock_reserve/view/stock_reserve.xml
@@ -69,8 +69,8 @@
                             <field name="location_id" />
                             <field name="location_dest_id" />
                         </group>
-                        <group name="note" string="Notes">
-                            <field name="note" nolabel="1" />
+                        <group name="note" colspan="2">
+                            <field name="note" />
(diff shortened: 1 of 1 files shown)


stock_reserve/view/stock_reserve.xml
```diff
@@ -69,8 +69,8 @@
                             <field name="location_id" />
                             <field name="location_dest_id" />
                         </group>
-                        <group name="note" string="Notes">
-                            <field name="note" nolabel="1" />
+                        <group name="note" colspan="2">
```
hparfr, line 72: I agree for the removing of `nolabel=1`

why removing the `string="Notes"` group name ?

hparfr (approved):

stock_reserve/view/stock_reserve.xml
chrisandrewmann: It's superfluous as the field is also called notes. So it would show notes twice.

PR #1821, opened 2023-08-24
[ADD][14.0] procurement_auto_create_group_by_product

jbaudoux:
Pick by product by generating one picking per product on the procurement run.
Configurable on the rule.
A procurement group per product will be generated.

I refactored a bit procurement_create_group to make it hookable

The use case it to be able a pick by product instead of by SO. So if 5 SO are for the same product, you have only one move for the total quantity.

Initially we had configured one fixed procurement group on the rule and all moves where added to the same picking. Users where working on the move lines. This was causing performance issue each time a new move was inserted as there are lot of moves and odoo recomputes the check_entire_pack on all moves which is slow. This was also causing concurrent updates when 2 users working on different products were extracting & validating a move line from that same picking.
With this new module, we greatly increase the amount of pickings but we reduce a lot the recomputation when a new move is added as there should be only one move in the picking, and we reduce to the minimum the risk of concurrent updates (it remains only if 2 users work on the same product on 2 different locations).

cc @mt-software-de @LoisRForgeFlow @BernatPForgeFlow


procurement_auto_create_group/models/procurement_group.py
```diff
@@ -13,13 +13,6 @@ def _get_rule(self, product_id, location_id, values):
         rule = super()._get_rule(product_id, location_id, values)
         # If there isn't a date planned in the values it means that this
         # method has been called outside of a procurement process.
-        if (
-            rule
-            and not values.get("group_id")
```
sebalix, line 18: Are you sure we want to remove this check? This could have side-effects on existing installations no?

procurement_auto_create_group/models/stock_rule.py
```diff
@@ -16,15 +16,18 @@ def _onchange_group_propagation_option(self):
         if self.group_propagation_option != "propagate":
             self.auto_create_group = False
 
+    def _get_auto_procurement_group(self, product_id):
+        group_data = self._prepare_auto_procurement_group_data(product_id)
```
sebalix: If `product_id` is a recordset, I would name it `product` here (for the `_get_rule` method we cannot as it's a std method).

procurement_auto_create_group/models/stock_rule.py
```diff
@@ -16,15 +16,18 @@ def _onchange_group_propagation_option(self):
...
         new_move_vals = super()._push_prepare_move_copy_values(move_to_copy, new_date)
         if self.auto_create_group:
-            group_data = self._prepare_auto_procurement_group_data()
-            group = self.env["procurement.group"].create(group_data)
+            group = self._get_auto_procurement_group(move_to_copy.product_id)
             new_move_vals["group_id"] = group.id
         return new_move_vals
 
-    def _prepare_auto_procurement_group_data(self):
+    def _prepare_auto_procurement_group_data(self, product_id):
```
sebalix:
```suggestion
    def _prepare_auto_procurement_group_data(self, product):
```

procurement_auto_create_group_by_product/models/stock_rule.py
```diff
@@ -0,0 +1,23 @@
...
+    auto_create_group_by_product = fields.Boolean(string="Procurement Group by Product")
+
+    def _get_auto_procurement_group(self, product_id):
+        if self.auto_create_group_by_product:
+            if product_id.auto_create_procurement_group_ids:
+                return fields.first(product_id.auto_create_procurement_group_ids)
+        return super()._get_auto_procurement_group(product_id)
+
+    def _prepare_auto_procurement_group_data(self, product_id):
+        result = super()._prepare_auto_procurement_group_data(product_id)
```
sebalix:
```suggestion
    def _prepare_auto_procurement_group_data(self, product):
        result = super()._prepare_auto_procurement_group_data(product)
```

procurement_auto_create_group_by_product/models/stock_rule.py
```diff
@@ -0,0 +1,23 @@
...
+        if self.auto_create_group_by_product:
+            result["product_id"] = product_id.id
```
sebalix:
```suggestion
            result["product_id"] = product.id
```

procurement_auto_create_group/models/procurement_group.py
sebalix: Following your commit message it's what you want, but just to be sure about the possible regressions.

jbaudoux: I could not find out why it was there. I hope it is safe to remove it @LoisRForgeFlow @JordiBForgeFlow 
Otherwise it prevents the feature to work in a pull flow coming from a SO.

sebalix:
@jbaudoux pre-commit to fix

jbaudoux:
rebased

jbaudoux:
ping @mt-software-de for review

jbaudoux:
@mt-software-de @sebalix @TDu Need review

procurement_auto_create_group_by_product/models/procurement_group.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    product_id = fields.Many2one("product.product", index=True)
```
mt-software-de, line 10: Do we need a constraint? To ensure that there is only one group per product?

procurement_auto_create_group_by_product/models/stock_rule.py
```diff
@@ -0,0 +1,23 @@
...
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    auto_create_group_by_product = fields.Boolean(string="Procurement Group by Product")
+
+    def _get_auto_procurement_group(self, product):
+        if self.auto_create_group_by_product:
+            if product.auto_create_procurement_group_ids:
+                return fields.first(product.auto_create_procurement_group_ids)
```
mt-software-de, line 38: Maybe we should add a lock, to ensure there is only one group created in case of auto_by_product?
```suggestion
            if product.auto_create_procurement_group_ids:
                return fields.first(product.auto_create_procurement_group_ids)
            else:
                 # lock to ensure there is only one group created
```

mt-software-de (approved):
Minor remarks, to ensure only one group per product.

procurement_auto_create_group/models/procurement_group.py
LoisRForgeFlow: Indeed, this is a bit risky. From our side, we have reviewed and we shouldn't be affected by the change as we never use it for SO rules. However, there might be other installations in which this can lead to a change in the behavior.

My 2 cents.

AaronHForgeFlow: Just in case, I think you can reduce the risk by putting the if condition in a method and doing super in the new module, removing the condition that is not wanted.

jbaudoux: If you don't see a breaking change, let's consider it a bug fix.
Note that I'm not changing anything in regards of the above comment regarding date planned.

The configuration of the rule should prevail. If it is configured to create a new procurement group, then it should do it. Independently of how the rule is run.

procurement_auto_create_group_by_product/models/stock_rule.py
mt-software-de: @jbaudoux could you merge https://github.com/jbaudoux/stock-logistics-warehouse/pull/1 ?

sebalix (approved):

jbaudoux:
@mt-software-de I have done this improvement https://github.com/OCA/stock-logistics-warehouse/pull/1821/commits/3213a30314e8c04c070c4eabc00de3d78dd4b492

mt-software-de (approved):
LGTM

PR #1822, opened 2023-08-25
[14.0][IMP] stock_request: Convert the state field of stock.request.order as compute store to complete all use cases

victoralmau:
Convert the `state` field of `stock.request.order` as compute store to complete all use cases

This change prevents `stock.request.order` from remaining in the `open` state when they have already finished.

@Tecnativa TT44811

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -51,6 +51,8 @@ class StockRequestOrder(models.Model):
         index=True,
         readonly=True,
         tracking=True,
+        compute="_compute_state",
+        store=True,
     )
     requested_by = fields.Many2one(
         "res.users",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why not converting the field to computed, and put all the logic inside one method? That way, it's more difficult to have unhandled flows.

victoralmau:
Changed the approach to define the `state` field as a compute store.

stock_request/models/stock_request_order.py
```diff
@@ -144,6 +146,20 @@ def _get_default_requested_by(self):
...
+    def _compute_state(self):
+        for item in self:
+            if not item.stock_request_ids or all(
+                x.state == "draft" for x in item.stock_request_ids
+            ):
+                item.state = "draft"
+            elif all(x.state == "cancel" for x in item.stock_request_ids):
+                item.state = "cancel"
+            elif all(x.state in ("done", "cancel") for x in item.stock_request_ids):
+                item.state = "done"
```
pedrobaeza: "Cache" the states for not having to query them several times:

```suggestion
            states = item.stock_request_ids.mapped("state")
            if not item.stock_request_ids or all(x == "draft" for x in states):
                item.state = "draft"
            elif all(x == "cancel" for x in states):
                item.state = "cancel"
            elif all(x in ("done", "cancel") for x in states):
                item.state = "done"
```

pedrobaeza (approved):

sergio-teruel (approved):

PR #1823, opened 2023-08-28
[MIG+REF] stock_move_packaging_qty: migrate to 16.0

yajo:
Migration:
- Code and DB migrations to use upstream field `product_packaging_id`, introduced in https://github.com/odoo/odoo/pull/68654.
- Test class moved to `post_install`, to let it pass if `sale_stock` is installed too.
- Satisfy new linters.

---

[REF] stock_move_packaging_qty: closer match upstream UX

Without this patch, a user that wasn't able to edit a picking product qty would be able to edit the packaging qty and produce a product qty update (which they shouldn't be able to do).

Now, the field is hidden or readonly when the `product_uom_qty` field is too.

Also, the inverse method of `product_packaging_qty` was only being triggered when saving. That doesn't fit well with v16, where saving is pseudo-automatic. Thus, the inverse is reused as an onchange.

To make it worse, the inverse method had a lot of validations. If they were meaningful, they should have been constraints instead. But they really didn't make any sense because they were testing impossible scenarios. For example, they were testing you didn't update the field when there's no packaging, but that's fixed now with a dynamic readonly attribute; also they were testing that the packaging qty != 0, but there's an upstream sql constraint that forbids that.

There was a check that warned about impossible combinations of packages and UoM qtys. Now that everything is properly automated, that scenario is impossible. Thus, the check is removed.

Summarizing, the module should be more usable and consistent now.

---

@moduon MT-3694 @MateuGForgeFlow


stock_move_packaging_qty/models/stock_move.py
```diff
@@ -22,14 +16,14 @@ class StockPicking(models.Model):
...
+        "product_qty", "product_uom", "product_packaging_id", "product_packaging_id.qty"
     )
     def _compute_product_packaging_qty(self):
         for move in self:
             if (
-                not move.product_packaging
+                not move.product_packaging_id
                 or move.product_qty == 0
-                or move.product_packaging.qty == 0
+                or move.product_packaging_id.qty == 0
```
rousseldenis, line 25: This should use float_is_zero() as digits precision is used.

stock_move_packaging_qty/models/stock_move.py
```diff
@@ -0,0 +1,111 @@
...
+        return res
+
+    def _check_package(self):
+        default_uom = self.product_id.uom_id
+        pack = self.product_packaging_id
+        qty = self.product_qty
+        q = default_uom._compute_quantity(pack.qty, self.product_uom)
+        if qty and q and round(qty % q, 2):
+            newqty = qty - (qty % q) + q
+            return {
```
rousseldenis: I would have changed this behavior as onchanges become more and more obsolete.

rousseldenis (commented):
First code review

stock_move_packaging_qty/models/stock_move.py
yajo: Makes sense, but I prefer to keep the migration diff smaller.

stock_move_packaging_qty/models/stock_move.py
yajo: Then maybe it's better to remove precision digits. Upstream doesn't use them:

- https://github.com/odoo/odoo/blob/92ad45547a9d9bb77041d925e7c748bb6d964ca2/addons/sale/models/sale_order_line.py#L188-L191
- https://github.com/odoo/odoo/blob/92ad45547a9d9bb77041d925e7c748bb6d964ca2/addons/purchase/models/purchase.py#L982

WDYT?

fcvalgar:
The type and quantity is correctly propagated to the delivery note but I have detected an error generated in the delivery note part.

The error occurs in the onchange when we try to edit a delivery note such as an internal movements delivery note.

Captura desde 2023-08-28 11-44-54

image

Error 
Captura desde 2023-08-28 11-40-55

```
  File "/opt/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1815, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 42, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 448, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/odoo/models.py", line 6510, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/opt/odoo/odoo/models.py", line 6221, in _onchange_eval
    method_res = method(self)
  File "/mnt/data/odoo-addons-dir/stock_move_packaging_qty/models/stock_move.py", line 83, in onchange_quantity
    res = super().onchange_quantity()
AttributeError: 'super' object has no attribute 'onchange_quantity'

The above server error caused the following client error:
null



```

stock_move_packaging_qty/models/stock_move.py
rousseldenis: Ok, so, at least put it in ROADMAP.rst

rousseldenis:
@yajo Fixing the onchange, could you also add tests to cover code ?

See https://app.codecov.io/gh/OCA/stock-logistics-warehouse/pull/1823?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=OCA

yajo:
Fixed, and tests added.

rafaelbn:
@yajo , tested in runbot, recorded 3 minutes video.

Just to be sure:

- Package qty is readonly , is it intended? I doubt.
- Package qty is not update until you click save, so it's confusing. Best would be calculated after changing from the field. Check video.

https://github.com/OCA/stock-logistics-warehouse/assets/8736623/94503dee-469c-4326-8b27-329b0acb779b

Let me know 😄 

Thank you! 🥳

stock_move_packaging_qty/tests/test_stock_move_packaging_qty.py
```diff
@@ -0,0 +1,60 @@
...
+        self.assertEqual(move.product_uom_qty, 15.0)
+
+    def test_product_uom_qty_change(self):
+        picking_f = Form(self.env["stock.picking"])
+        picking_f.partner_id = self.partner
+        picking_f.picking_type_id = self.env.ref("stock.picking_type_out")
+        with picking_f.move_ids_without_package.new() as move_f:
+            move_f.product_id = self.product
+            self.assertEqual(move_f.product_uom_qty, 1)
+            self.assertEqual(move_f.product_packaging_qty, 0)
```
rousseldenis, line 57: @yajo For your test fail, is the test user in group packaging ?

yajo: Oh thanks! I was becoming crazy.

rousseldenis: de nada :wink:

yajo:
Finally green! ✅

fcvalgar:
The calculation of the package and the package units seems to me to be correct.

Like @rafaelbn, I am missing flexibility in the delivery.

image

image

yajo:
As you see, upstream Odoo doesn't allow you to modify kg when the picking is related to a sales order. For the same reason, you cannot change the amount of packages (you'd be modifying the kg as a side effect).

fcvalgar (approved):
LGTM, thak you @yajo

rafaelbn (approved):
Tested in Staging as runboat of this PRs doen't works 👍🏼 

Ready to merge

rafaelbn:
@rousseldenis do you want to have a look before merging?

PR #1825, opened 2023-08-28
[14.0][IMP] stock_request_purchase: Change invisible attr of the action_view_stock_request action of purchase.order

victoralmau:
Changes done:
- [x] Change invisible attr of the `action_view_stock_request` action of `purchase.order`
- [x] Add `_get_stock_requests()` method to get all stock requests from action (allows inheritance by other modules)

@Tecnativa TT43297


rousseldenis (approved):
As the other one

chienandalu (approved):

PR #1830, opened 2023-09-06
[14.0][IMP] stock_reserve: Define the correct quantity available based on reservations

victoralmau:
Define the correct quantity available based on reservations

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT43413

### DIFF
--- a/stock_reserve/models/product.py
+++ b/stock_reserve/models/product.py
@@ -1,6 +1,8 @@
 # Copyright 2013 Camptocamp SA - Guewen Baconnier
+# Copyright 2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.tools.float_utils import float_round
 
 
 class ProductTemplate(models.Model):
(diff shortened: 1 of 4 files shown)


stock_reserve/models/product.py
```diff
@@ -36,6 +38,13 @@ class ProductProduct(models.Model):
     reservation_count = fields.Float(
         compute="_compute_reservation_count", string="# Sales"
     )
+    reserve_qty = fields.Float(
```
rousseldenis, line 41:
```suggestion
    reserved_qty = fields.Float(
```

stock_reserve/models/product.py
```diff
@@ -46,6 +55,30 @@ def _compute_reservation_count(self):
...
+            item.reserve_qty = float_round(
+                data.get(item.id, 0.0), precision_rounding=item.uom_id.rounding
+            )
+
+    @api.depends("reserve_qty")
+    def _compute_quantities(self):
+        """Reduce from available the reserved quantity."""
+        super()._compute_quantities()
+        for item in self:
+            item.qty_available -= item.reserve_qty
```
rousseldenis, line 77: This should go into a stock_available_* module

pedrobaeza:
This solution is not what we are looking for.

PR #1832, opened 2023-09-07
[14.0][FIX] stock_request: Change to free_qty field to use stock

victoralmau:
We need to use the `free_qty` field because it is the one that has the correct value if for example we have stock reservations.

@Tecnativa TT43413

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -404,7 +404,7 @@ class StockRequest(models.Model):
                     float_compare(
                         request.product_id.sudo()
                         .with_context(location=request.location_id.id)
-                        .qty_available,
+                        .free_qty,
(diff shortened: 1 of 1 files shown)


stock_request/models/stock_request.py
```diff
@@ -399,12 +399,14 @@ def _action_launch_procurement_rule(self):
                 continue
 
             # If stock is available we use it and we do not execute rule
+            # We need to use the free_qty field because it is the one that has the
```
pedrobaeza: No need of this explanation in code, as the field is self-explanatory. The comment is for the commit message.

pedrobaeza (approved):

rousseldenis (approved):
Makes sense.

A good test should be great as this is a critical flow

PR #1833, opened 2023-09-08
[IMP] stock_move_packaging_qty: display in PDF reports

yajo:
Directly display qty on delivery slip.

flameshot_2023-09-08_11-06_1

In the case of picking operations report, compute the qty based on the one found in the `stock.move.line` and the `product.packaging` ratio.

flameshot_2023-09-08_11-05

@moduon MT-3774


fcvalgar (approved):
LGTM, thank you @yajo

PR #1831, opened 2023-09-06
[14.0][IMP] stock_reserve_rule: add full bin strategy

TDu:
The full empty bin rule is like the empty bin rule, but it should take everything. That means there can not be any pre existing reservations.

ref.: rau-206

PR #1834, opened 2023-09-08
[14.0][IMP] stock_reserve_rule: add full lot strategy

geomer198:
New removal strategy is added.


aayartsev (approved):
Looks good

francesco-ooops (approved):
functional ok!

@OCA/logistics-maintainers good for merge?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -133,6 +133,7 @@ class StockReserveRuleRemoval(models.Model):
             ("default", "Default Removal Strategy"),
             ("empty_bin", "Empty Bins"),
             ("packaging", "Full Packaging"),
+            ("by_lot", "By lot"),
```
jbaudoux: The strategy name is not explicit

jbaudoux (changes requested):
Can you clarify what you try to achieve ? What are your real use cases ?

I have the impression that you try to minimize the amount of locations from which you reserve

elvise:
> Can you clarify what you try to achieve ? What are your real use cases ?
> 
> I have the impression that you try to minimize the amount of locations from which you reserve

Hi @jbaudoux thanks for review! 
The final goal is to have a configuration on which lot to reserve.
For example, for some fabrics we only reserve lots that meet the reserved quantity without shipping several lots to reach the required quantity.

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @jbaudoux any suggestions ?

CC @francesco-ooops

jbaudoux: What about `unique_lot` following your use case?

francesco-ooops: @jbaudoux lots are unique by definition :) what about single_lot ?

elvise: @jbaudoux @francesco-ooops we can follow the example of the packaging (Full Packaging), what do you think about “Full Lot”?

francesco-ooops: @elvise it can be misleading since this IMP also allows to reserve a lot with qty > requested qty, so the "full lot" would not be reserved. I think "Single lot" is the most fitting definition

elvise: > @elvise it can be misleading since this IMP also allows to reserve a lot with qty > requested qty, so the "full lot" would not be reserved. I think "Single lot" is the most fitting definition

ok let’s go 
@jbaudoux do you agree with @francesco-ooops ?

jbaudoux: Yeah, single lot sounds perfect

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,21 @@ class StockReserveRuleRemoval(models.Model):
...
+            ("lot_qty_eq_reserved_qty", "Lot qty = reserved qty"),
+        ],
+        string="Reserve lot if",
+    )
+
+    tolerance_margin = fields.Selection(
+        selection=[("percentage", "Percentage"), ("absolute", "Absolute value")]
+    )
+    tolerance_from = fields.Integer()
+    tolerance_to = fields.Integer()
```
jbaudoux: This is not used, can you drop it?

elvise: @geomer198 @aayartsev can you fix it ?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,21 @@ class StockReserveRuleRemoval(models.Model):
...
     )
 
+    reserve_lot = fields.Selection(
+        selection=[
+            ("lot_qty_gt_reserved_qty", "Lot qty > reserved qty"),
+            ("lot_qty_gte_reserved_qty", "Lot qty >= reserved qty"),
+            ("lot_qty_eq_reserved_qty", "Lot qty = reserved qty"),
+        ],
+        string="Reserve lot if",
+    )
```
jbaudoux: Do you need this complexity? Or you can always apply lot_qty_gte_reserved_qty ?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +313,37 @@ def is_greater_eq(value, other):
...
+    def _apply_strategy_by_lot(self, quants):
+        need = yield
+        product = fields.first(quants).product_id
+        comparison = {
+            "lot_qty_gt_reserved_qty": lambda lot_qty, need_qty: lot_qty > need_qty,
+            "lot_qty_gte_reserved_qty": lambda lot_qty, need_qty: lot_qty >= need_qty,
+            "lot_qty_eq_reserved_qty": lambda lot_qty, need_qty: lot_qty == need_qty,
+        }.get(self.reserve_lot)
+        if product.tracking == "lot" or comparison is not None:
+            for location, location_quants in quants._group_by_location():
```
jbaudoux: Can you clarify if your strategy is a single lot per location or a single lot ?
Location A; Lot2 5 units
Location B: Lot1 3 units + Lot2 4 units
If you need 8 units, do you accept Lot1 from B & Lot2 from A. Or you want only Lot2 from A & B?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +313,37 @@ def is_greater_eq(value, other):
...
+                lot_ids_with_quantity = {
+                    group["lot_id"][0]: group["quantity"] for group in grouped_quants
+                }
+                location_quantity = sum(location_quants.mapped("quantity")) - sum(
+                    location_quants.mapped("reserved_quantity")
+                )
+                lot_id = False
+                for rec_id, product_qty in lot_ids_with_quantity.items():
+                    if comparison(product_qty, need):
+                        lot_id = rec_id
```
jbaudoux: You need to respect default removal strategy (FIFO, FEFO, ...)

jbaudoux:
> @OCA/logistics-maintainers good for merge?

Please also drop a review on https://github.com/OCA/stock-logistics-warehouse/pull/1831 :)

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @jbaudoux In this case no lots will be reserved, as it is not possible to reserve lots with qty < the requested qty.

CC @francesco-ooops

jbaudoux: Ah yes, you're right. On purpose?

elvise: > Can you clarify if your strategy is a single lot per location or a single lot ? 

@jbaudoux Single LOT
if the user wants to limit this behavior by location, he can do that with the widget domain in the rule.

CC @francesco-ooops

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @jbaudoux we need these options because there are different scenarios

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > You need to respect default removal strategy (FIFO, FEFO, ...)

What do you mean?

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @geomer198 keep in mind this one

geomer198: > @geomer198 keep in mind this one

@elvise Resolved)

stock_reserve_rule/models/stock_reserve_rule.py
elvise: > > You need to respect default removal strategy (FIFO, FEFO, ...)
> 
> What do you mean?

@jbaudoux any suggestions ?

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: then I don't understand why you group the quants by location

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +307,41 @@ def is_greater_eq(value, other):
...
+            "lot_qty_eq_reserved_qty": lambda lot_qty, need_qty: lot_qty == need_qty,
+        }.get(self.reserve_lot)
+
+        if product.tracking == "lot" or comparison is not None:
+            for location, location_quants in quants._group_by_location():
+                grouped_quants = self.env["stock.quant"].read_group(
+                    [
+                        ("lot_id", "!=", False),
+                        ("location_id", "in", location.ids),
+                        ("product_id", "=", product.id),
```
jbaudoux: Also filter out negative quants

`("quantity", ">", 0),`

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: make a `quants._group_by_lot()` keeping quants sorting

elvise: @geomer198 please check it

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @geomer198  please check it

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @geomer198  please check it

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @geomer198 any news ?

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > @geomer198 please check it

Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: @jbaudoux I think use `quants._group_by_lot()` not solution because strategy need return `location.id`. This function return `lot` record and `quants` records that have many locations maybe.  That is why I use `group_by_location` and than do read_group for search `quants` by location and product.

elvise:
@francesco-ooops @geomer198 looks tolerance doesn’t work :(

geomer198:
> @francesco-ooops @geomer198 looks tolerance doesn’t work :( 

Yes, I only added fields without functional.

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: You could change what the method returns to provide the arguments required by `_update_reserved_quantity`.
You could even return a list and, in the caller, loop on the results to call `_update_reserved_quantity` on each element.
You anyway need to return the lot to ensure you reserve that selected lot in case there are multiple lots on the location.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,32 @@ class StockReserveRuleRemoval(models.Model):
...
     )
 
+    reserve_lot = fields.Selection(
+        selection=[
+            ("lot_qty_eq_requested_qty", "Lot qty = requested qty"),
+            ("lot_qty_gt_requested_qty", "Lot qty > requested qty"),
+            ("lot_qty_lt_requested_qty", "Lot qty < requested qty"),
+        ],
+        string="Requested lot if",
+    )
```
jbaudoux: Can you elaborate why you need all these?
Finding a single lot should compare quantity with `>=`
Finding an exact lot should compare quantity with '='. Maybe this should be a different strategy.
I don't understand `>` or `<`

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +324,88 @@ def is_greater_eq(value, other):
...
+                <= lot
+                < n - rounding
+                if lower_limit > 0.0
+                else lot < n,
+                **requested_eq,
+            }
+        # Default Behavior
+        return {
+            "lot_qty_gt_requested_qty": lambda lot_qty, need_qty: lot_qty > need_qty,
+            "lot_qty_lt_requested_qty": lambda lot_qty, need_qty: lot_qty < need_qty,
```
jbaudoux: Don't bind the tolerance to the lot strategy as this tolerance could also apply to the `packaging` strategy. Like reserve on a location containing a pallet if you need more than 80% of a pallet.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,32 @@ class StockReserveRuleRemoval(models.Model):
...
+
+    tolerance_margin = fields.Selection(
+        selection=[
+            ("percentage", "Percentage (%)"),
+            ("absolute", "Absolute Value"),
+        ],
+    )
+
+    tolerance_upper_limit = fields.Float(string="Tolerance (upper limit)")
+    tolerance_lower_limit = fields.Float(string="Tolerance (lower limit)")
```
jbaudoux: Add help

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: Hi @jbaudoux , this is a use case where it's important that all quants dispatched to customer are from the same lot (specific textile product processing), so the idea is to use a sequence of different rules to be granular on lot selection priority, eg: 

- first, look for a lot with exact qty as requested
- if not found, look for a lot with qty higher than requested, with a 10% tolerance
- if not found, look for a lot with qty higher than requested, with a 20% tolerance
- if not found, check if there's a lot with less qty than requested, with a 10% tolerance. We will check with customer if a lower qty can work for them. 

This is why we removed >= and =< as there was less clarity on lot selection; although I agree there is no difference between using "=" and ">=, tolerance: 0%"

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux we're reworking the PR according to your comments, thanks for those. 

Do you think "tolerance" should be available for all existing strategies? Or at least "Default removal strategy" should be excluded?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,62 @@ class StockReserveRuleRemoval(models.Model):
...
+            ("percentage", "Percentage (%)"),
+            ("absolute", "Absolute Value"),
+        ],
+        string="Tolerance computation",
+    )
+
+    tolerance_requested_value = fields.Float(string="Tolerance value", default=0.0)
+
+    tolerance_display = fields.Char(
+        compute="_compute_tolerance_display", store=True, stirng="Tolerance"
```
GSLabIt:
```suggestion
        compute="_compute_tolerance_display", store=True, string="Tolerance"
```

elvise:
@jbaudoux please take a look :)

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux this was done as requested, please let us know what do you think as you have more experience than us on this removal strategy. Thanks!

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux can we mark this as solved?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux can we mark this as resolved?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @geomer198 what about this?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @geomer198 what about this?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux can we mark this as outdated?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @jbaudoux in a previous version we tried to implement tolerance on "Full packaging" strategy but we could not define a correct behavior due to 

- lack of use case for testing and clarity of flow
- implementation being outside the scope of the PR
- budget constraint related to customer's request

So we removed the feature. It is anyway working as intended for the new "Single lot" strategy so the code is available for further implementation if the need for tolerance on full packaging arises.

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: Solved?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @GSLabIt solved?

stock_reserve_rule/__manifest__.py
```diff
@@ -4,7 +4,7 @@
     "name": "Stock Reservation Rules",
     "summary": "Configure reservation rules by location",
     "version": "14.0.1.2.0",
-    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "author": "Cetmix, Camptocamp, Odoo Community Association (OCA)",
```
francesco-ooops: @geomer198 I think contributors is enough

stock_reserve_rule/readme/DESCRIPTION.rst
```diff
@@ -21,6 +21,9 @@ The included advanced removal strategies are:
 * Full Packaging: tries to remove full packaging (configured on the products)
   first, by largest to smallest package or based on a pre-selected package
   (default removal strategy is then applied for equal quantities).
+* By lot: tries to remove a lot. A specific field allows to set whether to reserve
+  a lot with qty matching requested qty, or a lot with qty >= requested quantity,
+  or only lots with qty > requested qty.
```
francesco-ooops: * Single lot: tries to remove a single lot. This strategy requests to set a tolerance (either in percentage or absolute value) to allow selection of a lot with same qty as requested qty by setting a tolerance value of 0. 
By setting a Tolerance value different from 0, user can create rules to select a lot with qty higher than requested (Tolerance on = Upper Limit) or lower than requested (Tolerance on = Lower Limit). Note that rules with Tolerance value != 0 will not select lots with requested qty = lot qty.

francesco-ooops (approved):

stock_reserve_rule/readme/DESCRIPTION.rst
```diff
@@ -21,6 +21,13 @@ The included advanced removal strategies are:
 * Full Packaging: tries to remove full packaging (configured on the products)
   first, by largest to smallest package or based on a pre-selected package
   (default removal strategy is then applied for equal quantities).
+* Single lot: tries to remove a single lot.
+  This strategy requests to set a tolerance (either in percentage or absolute value)
+  to allow selection of a lot with same qty as requested qty by setting a tolerance value of 0.
+  By setting a Tolerance value different from 0, user can create rules to select a lot with qty
+  higher than requested (Tolerance on = Upper Limit) or lower than requested
+  (Tolerance on = Lower Limit). Note that rules with Tolerance value != 0
+  will not select lots with requested qty = lot qty.
```
francesco-ooops: This strategy requires to select if a tolerance should be applied on lot selection, allowing to select lots with qty higher or lower than qty requested in picking. 
Rules can be applied in sequence, for example, to first check for exact lot qty, then for a lot with qty 5% higher than requested, then for a lot with qty 10% higher than requested, and so on.

If lot selected must have same qty as requested, set "Tolerance on = No tolerance"; otherwise it's possible to select a lot with higher qty (Tolerance on = Upper Limit") or lower qty ("Lower Limit"), either in percentage of qty or absolute value. 
Please note that both "Upper limit" or "Lower limit" rules will not select a lot with qty = requested qty.

francesco-ooops:
@geomer198 please squash commits

@jbaudoux from our side we're done implementing the new features. As per https://github.com/OCA/stock-logistics-warehouse/pull/1834#discussion_r1372809763 we will stop here, but we think the feature is ready to be extended for other strategies.

We look forward to your feedback!

francesco-ooops:
@jbaudoux ping :)

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,92 @@ class StockReserveRuleRemoval(models.Model):
         "When empty, any packaging can be removed.",
     )
 
+    TOLERANCE_LIMIT = [
+        ("no_tolerance", "No tolerance"),
+        ("upper_limit", "Upper Limit"),
+        ("lower_limit", "Lower Limit"),
+    ]
+
+    tolerance_requested_limit = fields.Selection(
```
rousseldenis, line 166: @geomer198 Maybe some help entries for those fields ?

geomer198: Resolved

francesco-ooops: @rousseldenis good now?

GSLabIt (approved):

rousseldenis:
@jbaudoux Could you update your review ?

francesco-ooops:
@geomer198 rebase?

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: According to me, this needs to be addressed. In the other reservation rule, we try to select a location, that's why we group by location. As you try to select a lot (no matter where), you should implement a `quants._group_by_lot()` keeping quants sorting (to respect removal strategy fifo/fefo). I dropped you a comment below on how to approach this.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +388,66 @@ def is_greater_eq(value, other):
...
+            return lambda product_qty, need: product_qty == need
+        computation = self.tolerance_requested_computation
+        if self.tolerance_requested_limit == "upper_limit":
+
+            def wrap(product_qty, need):
+                if computation == "percentage":
+                    return need + rounding < product_qty <= need * (100 + value) / 100
+                # computation == "absolute"
+                else:
+                    return need + rounding < product_qty <= need + value
```
jbaudoux: I don't understand the comparison formulas.
You say in readme `select a lot with higher qty (Tolerance on = Upper Limit")`. I understand that in this case you want a quant that has more stock than the need, i.e. `need <= lot_qty`. You need to use float_compare and provide the rounding.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +388,66 @@ def is_greater_eq(value, other):
                     # compute how much packaging we can get
                     take = (need // pack_quantity) * pack_quantity
                     need = yield location, location_quantity, take, None, None
+
+    @api.model
+    def _get_requested_comparisons(self, rounding):
```
jbaudoux: Make a method `_compare_with_tolerance(self, need, quantity, rounding)` implementing the comparisons. You don't need wrappers.

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: Or I wrongly understood what you try to achieve. Looking at your formula, you try in fact to reserve a full lot. Like we have full packaging and full bin where the quantity matches the need, or where the quantity is a bit more (upper) or less (lower) than the need.
The formulas must be fixed:
* No tolerance: `float_compare(need, quantity, rounding) == 0`
* Upper Limit: lot with higher qty than required: `float_compare(need, quantity, rounding) <= 0 and float_compare(quantity - need, tolerance, rounding) <= 0`

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +388,66 @@ def is_greater_eq(value, other):
...
+        value = self.tolerance_requested_value
```
jbaudoux: rename your variable `value` by `tolerance` for clarity

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,96 @@ class StockReserveRuleRemoval(models.Model):
...
+        ("upper_limit", "Upper Limit"),
+        ("lower_limit", "Lower Limit"),
+    ]
+
+    tolerance_requested_limit = fields.Selection(
+        selection=TOLERANCE_LIMIT,
+        string="Tolerance on",
+        help="Selecting a tolerance limit type"
+        "No tolerance: lot qty should be equal required"
+        "Upper Limit: lot with higher qty than required"
```
jbaudoux: Make help independent of lot
```suggestion
        "Upper Limit: quantity higher than demand but within tolerance"
```

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -296,3 +388,66 @@ def is_greater_eq(value, other):
                     # compute how much packaging we can get
                     take = (need // pack_quantity) * pack_quantity
                     need = yield location, location_quantity, take, None, None
+
+    @api.model
+    def _get_requested_comparisons(self, rounding):
+        value = self.tolerance_requested_value
+        if value == 0.0:
```
jbaudoux:
```suggestion
        if tolerance_requested_limit == "no_tolerance" or float_compare(tolerance, 0, rounding) == 0: 
```

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,96 @@ class StockReserveRuleRemoval(models.Model):
...
+        "Upper Limit: lot with higher qty than required"
+        "Lower Limit: lot with lower qty than required",
+    )
+
+    tolerance_requested_computation = fields.Selection(
+        selection=[
+            ("percentage", "Percentage (%)"),
+            ("absolute", "Absolute Value"),
+        ],
+        string="Tolerance computation",
```
jbaudoux, line 182: Make it required with a default value

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,96 @@ class StockReserveRuleRemoval(models.Model):
         "When empty, any packaging can be removed.",
     )
 
+    TOLERANCE_LIMIT = [
+        ("no_tolerance", "No tolerance"),
+        ("upper_limit", "Upper Limit"),
+        ("lower_limit", "Lower Limit"),
+    ]
```
jbaudoux, line 164: What about having also "Within Limits" ?

(suggestion, non-blocking and non required)

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,96 @@ class StockReserveRuleRemoval(models.Model):
...
+
+    @api.onchange("tolerance_requested_limit")
+    def _onchange_tolerance_requested_limit(self):
+        if self.tolerance_requested_limit == "no_tolerance":
+            self.update(
+                {
+                    "tolerance_requested_computation": "percentage",
+                    "tolerance_requested_value": 0.0,
+                }
+            )
```
jbaudoux: I would drop this. You have tolerance_requested_limit that says what to do. Any value in the other parameters may stay.

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -152,6 +154,96 @@ class StockReserveRuleRemoval(models.Model):
...
 
+    TOLERANCE_LIMIT = [
+        ("no_tolerance", "No tolerance"),
+        ("upper_limit", "Upper Limit"),
+        ("lower_limit", "Lower Limit"),
+    ]
+
+    tolerance_requested_limit = fields.Selection(
+        selection=TOLERANCE_LIMIT,
+        string="Tolerance on",
```
jbaudoux, line 168: make required with first option as default value

stock_reserve_rule/views/stock_reserve_rule_views.xml
```diff
@@ -62,6 +63,19 @@
                                         widget="many2many_tags"
                                         attrs="{'invisible': [('removal_strategy', '!=', 'packaging')]}"
                                     />
+                                    <field
+                                        name="tolerance_requested_limit"
+                                        attrs="{'invisible': [('removal_strategy', '!=', 'single_lot')], 'required': [('removal_strategy', '=', 'single_lot')]}"
+                                    />
+                                    <field
+                                        name="tolerance_requested_computation"
+                                        attrs="{'invisible': [('removal_strategy', '!=', 'single_lot')], 'required': [('removal_strategy', '=', 'single_lot')], 'readonly': [('tolerance_requested_limit', '=', 'no_tolerance')]}"
```
jbaudoux:
```suggestion
                                        attrs="{'invisible': [('tolerance_requested_limit', '=', 'no_tolerance')]}"
```

stock_reserve_rule/views/stock_reserve_rule_views.xml
```diff
@@ -62,6 +63,19 @@
...
+                                    />
+                                    <field
+                                        name="tolerance_requested_value"
+                                        attrs="{'invisible': [('removal_strategy', '!=', 'single_lot')], 'required': [('removal_strategy', '=', 'single_lot')], 'readonly': [('tolerance_requested_limit', '=', 'no_tolerance')]}"
```
jbaudoux:
```suggestion
                                        attrs="{'invisible': [('tolerance_requested_limit', '=', 'no_tolerance')]}"
```

stock_reserve_rule/views/stock_reserve_rule_views.xml
```diff
@@ -62,6 +63,19 @@
                                         widget="many2many_tags"
                                         attrs="{'invisible': [('removal_strategy', '!=', 'packaging')]}"
                                     />
+                                    <field
+                                        name="tolerance_requested_limit"
+                                        attrs="{'invisible': [('removal_strategy', '!=', 'single_lot')], 'required': [('removal_strategy', '=', 'single_lot')]}"
```
jbaudoux: put required on the field, set default and drop required here

stock_reserve_rule/readme/DESCRIPTION.rst
```diff
@@ -21,6 +21,19 @@ The included advanced removal strategies are:
...
+  allowing to select lots with qty higher or lower than qty requested in picking.
+  Rules can be applied in sequence, for example, to first check for exact lot qty,
+  then for a lot with qty 5% higher than requested, then for a lot with qty 10% higher
+  than requested, and so on.
+
+  If lot selected must have same qty as requested, set "Tolerance on = No tolerance";
+  otherwise it's possible to select a lot with higher qty (Tolerance on = Upper Limit")
+  or lower qty ("Lower Limit"), either in percentage of qty or absolute value.
+  Please note that both "Upper limit" or "Lower limit" rules
+  will not select a lot with qty = requested qty.
```
jbaudoux: Why ? Doesn't make sense to me. A tolerance must accept exact value as well.

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: I'm coming back to this naming as I know understand better what you achieved in your code.
For me, single lot means that you don't want to end up with 2 lots in the reservation but you could take any lot that only partially fill the demand.
However, here you want to reserve a lot that has the required quantity within the tolerance. That's similar to what we already have with "full packaging" or "full bin". So we need to rename to "full lot".

rousseldenis:
@geomer198 Could you fix tests and pre-commit ?

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > I would drop this. You have tolerance_requested_limit that says what to do. Any value in the other parameters may stay.

Access rules to field only from view and without backend is not good solution. So I can't drop this.

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > Make a method `_compare_with_tolerance(self, need, quantity, rounding)` implementing the comparisons. You don't need wrappers.

It's not just wrapper, I use closures in my solution. And why wappers is not good solution?

stock_reserve_rule/views/stock_reserve_rule_views.xml
geomer198: Resolved

stock_reserve_rule/views/stock_reserve_rule_views.xml
geomer198: Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: Resolved

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: I thinks is not need remove this functions. In backend must be validate data to, not only in form view.

stock_reserve_rule/views/stock_reserve_rule_views.xml
geomer198: Resolved

stock_reserve_rule/readme/DESCRIPTION.rst
geomer198: Description is changed.

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: Resolved

geomer198:
> @geomer198 Could you fix tests and pre-commit ?

Done, fixed it!

francesco-ooops:
@jbaudoux feedback?

francesco-ooops:
@jbaudoux gentle reminder!

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: I don't understand the benefit of this approach of coding in odoo. You can't override the inner methods.
A plain compare method would be simple and easy to read.

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: This is not resolved

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: But the rule will behave the same with or without those methods. You just loose the previous tolerance config when changing if you want to apply tolerance or not. I'm not fan of this but this is just an opinion.

jbaudoux (changes requested):
There are still issues to be addressed. Main concern is the group by location which does not seems inline with the intention of the module.
Also use float_compare and make a plain compare method.

geomer198:
> There are still issues to be addressed. Main concern is the group by location which does not seems inline with the intention of the module. Also use float_compare and make a plain compare method.

Could you check my changes? Is now ok?

elvise:
> > There are still issues to be addressed. Main concern is the group by location which does not seems inline with the intention of the module. Also use float_compare and make a plain compare method.
> 
> Could you check my changes? Is now ok?

@jbaudoux gentle reminder :)

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -333,3 +398,68 @@ def _apply_strategy_full_bin(self, quants):
...
+        tolerance = self.tolerance_requested_value
+        limit = self.tolerance_requested_limit
+        computation = self.tolerance_requested_computation
+        if limit == "no_tolerance" or float_compare(tolerance, 0, rounding) == 0:
+            return float_compare(need, product_qty, rounding) == 0
+        elif limit == "upper_limit":
+            if computation == "percentage":
+                # need + rounding < product_qty <= need * (100 + tolerance) / 100
+                return (
+                    float_compare(need, product_qty, rounding) == -1
```
jbaudoux: Why don't you accept exact value when you give a tolerance? It should be in the limit defined by the tolerance but if the quantity is exactly the need, it should also be valid, isn't it?
```suggestion
                    float_compare(need, product_qty, rounding) <= 0
```

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -333,3 +398,68 @@ def _apply_strategy_full_bin(self, quants):
...
+                    and self._compare_with_tolerance(need, product_qty, rounding)
+                    and lot_quantity > 0
+                ):
+                    need = (
+                        yield fields.first(lot_quants).location_id,
+                        lot_quantity,
+                        need,
+                        lot,
+                        None,
+                    )
```
jbaudoux: You have selected a lot and it can have multiple quants. I think you need to yield each quant one by one by looping on the quants (they should be already sorted by fifo/fefo) decreasing the need until need is 0 or all quants are consumed.
Can you add a test with this?
Example:
* Loc1, Lot A, 10 units
* Loc2, Lot A, 5 units
Need for 15 units. Lot A is selected

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: Some unit tests on `_compare_with_tolerance` with real figures would be helpful. Just the call to the method without picking.
Tests with inside tolerance, at the lower limit of tolerance, at the upper limit of the tolerance, below tolerance, above tolerance 
* no tolerance
* upper_limit, percentage
* upper_limit, absolute
* lower_limit, percentage
* lower_limit, absolute

stock_reserve_rule/readme/DESCRIPTION.rst
```diff
@@ -21,6 +21,18 @@ The included advanced removal strategies are:
 * Full Packaging: tries to remove full packaging (configured on the products)
   first, by largest to smallest package or based on a pre-selected package
   (default removal strategy is then applied for equal quantities).
+* Single lot: tries to remove a single lot.
```
jbaudoux:
```suggestion
* Full lot: tries to remove a complete lot.
```

jbaudoux:
> > > There are still issues to be addressed. Main concern is the group by location which does not seems inline with the intention of the module. Also use float_compare and make a plain compare method.
> > 
> > Could you check my changes? Is now ok?

Thanks for the work. The code is so much cleaner, easy to read and simple :)

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: @jbaudoux Thank you for your solution! I updated function implementation. Now the strategy looks better!

geomer198:
@jbaudoux Please check my changes.

elvise:
@jbaudoux gentle reminder :)

elvise:
@jbaudoux gentle reminder :)

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -333,3 +398,69 @@ def _apply_strategy_full_bin(self, quants):
...
+                        and self._compare_with_tolerance(need, product_qty, rounding)
+                        and lot_quantity > 0
+                    ):
+                        need = (
+                            yield lot_quant.location_id,
+                            lot_quantity,
+                            need,
+                            lot,
+                            None,
+                        )  # noqa
```
jbaudoux: What about lots that are partially reserved? I think you want to skip them.

`product_qty` can be computed outside loop.

```suggestion
                if not lot:
                    continue
                if any(quant.reserved_quantity in lot_quants):
                    # skip lots that are partially reserved
                    continue
                product_qty = sum(lot_quants.mapped("quantity"))
                if not self._compare_with_tolerance(need, product_qty, rounding):
                    continue
                for lot_quant in lot_quants:
                    lot_quantity = lot_quant.quantity - lot_quant.reserved_quantity
                    need = yield (
                            lot_quant.location_id,
                            lot_quantity,
                            need,
                            lot,
                            None,
                        )  # noqa
```

geomer198: @jbaudoux Could you check my proposed version of the function?

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -333,3 +398,66 @@ def _apply_strategy_full_bin(self, quants):
...
+        need = yield
+        # We take goods only if we empty the bin.
+        # The original ordering (fefo, fifo, ...) must be kept.
+        product = fields.first(quants).product_id
+        if product.tracking == "lot":
+            rounding = product.uom_id.rounding
+            for lot, lot_quants in quants.filtered(
+                lambda quant: quant.product_id.id == product.id
+                and quant.lot_id
+                and not quant.reserved_quantity
```
jbaudoux: You cannot exclude quant with reserved quantities before the group by. If the lot is on 2 locations and has a reservation only on one location, you will not end-up with the right result

geomer198: @jbaudoux I fixed it. I hope now is better.

elvise:
@jbaudoux good now ? :)

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: Now we lost the exclusion. See my proposed solution above

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: @elvise This still needs to be fixed or answered

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > Now we lost the exclusion. See my proposed solution above

I didn't understand this line:
```py
                if any(quant.reserved_quantity in lot_quants):
                    # skip lots that are partially reserved
                    continue
```
In this function only `quants` variable. And why float must be contain in quant's record set?

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: @jbaudoux Your code has an error.

stock_reserve_rule/models/stock_reserve_rule.py
elvise: @francesco-ooops @geomer198 please take care

stock_reserve_rule/models/stock_reserve_rule.py
jbaudoux: if any(quant.reserved_quantity for quant in lot_quants):

jbaudoux:
> @jbaudoux good now ? :)

@elvise @francesco-ooops 
Please write a list of tests instead of keeping asking me if this is now good. Write them in English and ask the developer to implement them. Also run your tests manually. You will then have your answer.

Write unit tests for the tolerance. See https://github.com/OCA/stock-logistics-warehouse/pull/1834#discussion_r1400309998
Write unit test for a lot on multiple locations
Write unit test for anything that has been discussed above and for which you could not tell yourself if it is now good

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: @jbaudoux Done! Please check tests.

elvise: > @jbaudoux Done! Please check tests.

@geomer198 can you add tests with decimals (qty)?

stock_reserve_rule/models/stock_reserve_rule.py
francesco-ooops: @geomer198 was this addressed?

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > > @jbaudoux Done! Please check tests.
> 
> @geomer198 can you add tests with decimals (qty)?

Yes) Done!

stock_reserve_rule/models/stock_reserve_rule.py
geomer198: > @geomer198 was this addressed?

Yes, I added this code.

PR #1836, opened 2023-09-11
[14.0][IMP] stock_request_purchase: Add exceptions from purchase order cancel to stock.request or stock.request.order

victoralmau:
Add exceptions from purchase order cancel to `stock.request` or `stock.request.order`

Steps to reproduce it:

**A: Only with `stock.request`**
- Create a stock request of a product to be purchased.
- Confirm the stock request
- Go to the purchase order and cancel it
- An exception will have been added to the stock.request indicating that the purchase order has been cancelled.

**B: Only with `stock.request.order`**
- Do the same steps as in A but with request order

@Tecnativa TT44811

### DIFF
--- a/stock_request_purchase/models/purchase_order.py
+++ b/stock_request_purchase/models/purchase_order.py
@@ -1,4 +1,5 @@
 # Copyright 2017-20 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2023 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
 from odoo import api, fields, models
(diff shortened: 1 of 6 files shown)


stock_request_purchase/tests/test_stock_request_purchase.py
```diff
@@ -263,3 +267,9 @@ def test_view_actions(self):
         self.assertEqual(action["domain"], "[]")
         self.assertEqual("views" in action.keys(), True)
         self.assertEqual(action["res_id"], order.purchase_ids[0].id)
+
```
pedrobaeza: Don't add empty lines inside a method.

stock_request_purchase/models/purchase_order.py
```diff
@@ -22,6 +23,52 @@ def _compute_stock_request_ids(self):
             rec.stock_request_ids = rec.order_line.mapped("stock_request_ids")
             rec.stock_request_count = len(rec.stock_request_ids)
 
+    def button_cancel(self):
+        super().button_cancel()
+        self._exception_on_stock_request()
+
+    def _get_exception_on_stock_request_info(self):
```
pedrobaeza, line 30: Why do you need so much code for this easy one? And adding an extra method is not needed, as this is not called from several places. Do you need any hook?

victoralmau: This method is needed to cover all use cases:
- Purchase linked to stock requests.
- Purchase linked to a child production linked to a stock request (https://github.com/OCA/stock-logistics-warehouse/pull/1837) **Important**: In this case it will be necessary to call the method before cancellation to continue to have the linked productions.

What do you consider would be the correct way to proceed in this case?

pedrobaeza: That "inherited link" is not something of this module, and the propagation of the cancellation shouldn't be treated at that level. For me this should only handle the direct link.

PR #1842, opened 2023-09-12
[14.0][FIX] stock_move_location: Fix onchange and make editable

RabbitJon-S73:
This bug https://github.com/OCA/stock-logistics-warehouse/pull/1607 also happens in v15. I also took the functionality to edit each line destination location.

In this case, since the onchange on `Operation type` doesn't exists an error happens because there is no `Destination location` on the picking moves. 

Captura desde 2023-09-12 11-39-19


RabbitJon-S73:
@rousseldenis @sergio-teruel

sergio-teruel (approved):

RabbitJon-S73:
Can be merged? @pedrobaeza  @sergio-teruel

PR #1848, opened 2023-09-19
[15.0][FIX] stock_inventory

AlexPForgeFlow:
- Fixed the case that corresponds to a shared location between several companies (as company_id is a related on quant location)
- Fixed allowed fields to be written in stock_quant to add to_do field

### DIFF
--- a/stock_inventory/tests/test_stock_inventory.py
+++ b/stock_inventory/tests/test_stock_inventory.py
@@ -292,6 +292,8 @@ class TestStockInventory(TransactionCase):
             }
         )
         inventory1.action_state_to_in_progress()
+        # Remove company_id from stock-quant to test it works as expected
+        inventory1.stock_quant_ids.write({"company_id": False})
         self.assertEqual(inventory1.stock_quant_ids.ids, [self.quant4.id])
         inventory1.action_state_to_draft()
         self.assertEqual(inventory1.stock_quant_ids.ids, [])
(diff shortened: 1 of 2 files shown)


rousseldenis:
@AlexPForgeFlow Thanks for this. Could you be a little bit more explicit in your commit message.

What are you fixing? And a test would be welcome

AlexPForgeFlow:
I've seen the case when a stock.quant or a stock.move.line has not a company_id associated, in search (stock_quant.py#L21) it returns an empty list and, when trying to access it later it raises exception for index out of range. This is not a major bug, because we only have found it in a local database, but I think ensuring the process is the correct path. 
I have added a new line in the test to make sure it won't happen in the future.

DavidJForgeFlow (approved):
LGTM! Thanks @AlexPForgeFlow!

stock_inventory/tests/test_stock_inventory.py
```diff
@@ -292,6 +292,8 @@ def test_05_category_selection(self):
             }
         )
         inventory1.action_state_to_in_progress()
+        # Remove company_id from stock-quant to test it works as expected
+        inventory1.stock_quant_ids.write({"company_id": False})
```
rousseldenis, line 296: @AlexPForgeFlow In fact, functionally, this case correspond to a shared location between several companies (as company_id is a related on quant location).

I would have added this explanation here and in the commit message (that is too poor)

rousseldenis (approved):

rousseldenis:
Let's merge this as this fixes flow.

/ocabot merge minor

PR #1851, opened 2023-09-21
[MIG] stock_warehouse_orderpoint_stock_info_unreserved to 14.0

sergiocorato:


sergiocorato:
(first check fails only for missing module from pr #1850, which is installed by `test-requirements.txt` file later in the process)

PR #1674, opened 2023-03-23
[16.][MIG] stock_change_qty_reason

Urvisha-OSI:

PR #1854, opened 2023-09-28
[16.0] [MIG] stock_change_qty_reason

paradoxxxzero:
Superseeds #1674

Fix UI view for stock.quant.reason (fix button box, remove extraneous attributes and add explicit label):

20230928_11h35m34s_grim

Also remove past migrations, and improve comments.


rousseldenis:
@paradoxxxzero Thanks for this. Maybe could you split your changes not related to button box commit either in migration commit, either in a new one ? Thanks

stock_change_qty_reason/views/stock_move_line_view.xml
```diff
@@ -0,0 +1,35 @@
...
+                    groups="stock_change_qty_reason.group_qty_reason_preset"
+                />
+            </field>
+            <xpath expr="//field[@name='reason']" position="attributes">
+                <t groups="stock_change_qty_reason.group_qty_reason_preset">
+                    <attribute name="invisible">True</attribute>
+                </t>
+            </xpath>
+        </field>
+    </record>
```
AungKoKoLin1997, line 34: Can we just use like this instead of adding new inherited view?
```suggestion
                <field name="reason" groups="!stock_change_qty_reason.group_qty_reason_preset"/>
                <field
                    name="preset_reason_id"
                    string="Reason"
                    groups="stock_change_qty_reason.group_qty_reason_preset"
                />
            </field>
        </field>
    </record>

```

PR #1858, opened 2023-10-05
[14.0][IMP] stock_location_package_restriction: Change no restriction value

TDu:
The required constraint on the field was causing conflict with other module when running the tests.

So instead of having a specific value the default no restriction behaviour, a falsy value is used.


simahawk (approved):
LG

rousseldenis (approved):
Seems ok

PR #1859, opened 2023-10-06
[14.0][IMP] stock_reserve: Substract Forecasted Quantity from products according to configuration option

victoralmau:
Substract Forecasted Quantity from products according to configuration option.

Example use case:
- Product with stock 10 > Forectasted = 10
- Create an outgoing picking of 10 qty > Forectasted = 0
- Create a reservation of 10 qty + reserve > Forecasted = 0
- Release the reserve > Forecasted = 0

@Tecnativa TT43413


stock_reserve/models/product.py
```diff
@@ -46,6 +46,20 @@ def _compute_reservation_count(self):
             reservations = self.env["stock.reservation"].search(domain)
             product.reservation_count = sum(reservations.mapped("product_qty"))
 
+    def _get_domain_locations_new(
+        self, location_ids, company_id=False, compute_child=True
+    ):
+        """Add location reservation for avoiding reservations to be substracted on
+        the forecasted quantity."""
```
pedrobaeza: Remember how docstrings are put according the PEP:

```suggestion
        """Add location reservation for avoiding reservations to be substracted on
        the forecasted quantity.
        """
```

pedrobaeza (commented):
Seeing the tests, it seems the initial intention is that the quantities to be substracted from the forecasted quantity which I don't see before.

This is something reasonable for reservation in sales quotations, where you want to already have less quantity. There are 2 ways then:

1. Add an option to substract or not, global by company or per reservation (this second one is more flexible, but more difficult to be implemented).
2. Do this for our custom case apart from the standard OCA, as we will have only the case of not substracting.

I think we are going to the point 2 in this occasion.

victoralmau:
Changes done.

stock_reserve/README.rst
```diff
@@ -22,11 +19,11 @@ Stock Reservation
 .. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
     :target: https://translation.odoo-community.org/projects/stock-logistics-warehouse-14-0/stock-logistics-warehouse-14-0-stock_reserve
     :alt: Translate me on Weblate
-.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
-    :target: https://runboat.odoo-community.org/builds?repo=OCA/stock-logistics-warehouse&target_branch=14.0
-    :alt: Try me on Runboat
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/153/14.0
```
pedrobaeza, line 23: You have an outdated README generator. Please update your local maintainer-tools installation

stock_reserve/models/product.py
```diff
@@ -46,6 +46,23 @@ def _compute_reservation_count(self):
             reservations = self.env["stock.reservation"].search(domain)
             product.reservation_count = sum(reservations.mapped("product_qty"))
 
+    def _get_domain_locations_new(
+        self, location_ids, company_id=False, compute_child=True
+    ):
+        """Add location reservation for avoiding reservations to be substracted on
+        the forecasted quantity.
+        """
+        company = self.company_id or self.env.company
```
pedrobaeza, line 55: This is not valid, as products by default don't have company, so this is always going to take the main company, which is not what is needed, specially on things coming from OdooBot. Argument `company_id` is not filled? If any doubt about the company passed, please go back to what I asked to put this on custom.

pedrobaeza (changes requested):

stock_reserve/models/product.py
victoralmau: Ok, taking into account what you indicate, we add this to custom (subtract always) and leave this PR to implement option 1 (more difficult to implement)

pedrobaeza: But it won't be valid if the company is not perfectly guessed.

victoralmau: With option 1 yes: Define for each reservation whether to subtract or not (auto-defining the value of a new field according to the company configuration), but the filter to be performed would be more complex than just by location.

pedrobaeza:
Let's abort this, as it's not easily done.

PR #1868, opened 2023-10-18
[16.0][ADD] stock_valuation_layer_accounting_date

AungKoKoLin1997:

Depends on: https://github.com/OCA/stock-logistics-warehouse/pull/1711
@qrtl


AungKoKoLin1997:
This module has already been merged and it seems like redundant work was done. Closing this PR.

PR #1872, opened 2023-10-26
[FIX] stock_mts_mto_rule: multiple procs same product

ferran-S73:
When launching 2 mts+mto procurements for the same product, if the first one has a greater o equal qty than the second, the second won't run it's mto action. This is because the replenishment move gets confirmed whereas the original moves are still in drat state.

Steps to reproduce the issue:
  - Configure a product with the mts+mto rule and some other rule (in my case it was a manufacturing rule)
  - Launch it's rule twice at the same time (i.e. confirming a sales order with two lines of the same product) without having any stock for the product. First line with 10units, second with 5.
  - Only the procurement for the first line will apply it's mto rule (in my case it only creates one manufacturing order with 10 units).

Debugging the code I found out that when the `_run_split_procurement` method calls `get_mto_qty_to_order` for the second procurement the product `virtual_available` includes the qty that from the manufacturing order created by the first proc (it's move is confirmed) but isn't taking into account the sale's move as it is still in draft. So I think we can take this into account by keeping track of what's being replenished for each product.

### DIFF
--- a/stock_mts_mto_rule/models/stock_rule.py
+++ b/stock_mts_mto_rule/models/stock_rule.py
@@ -36,14 +36,20 @@ class StockRule(models.Model):
                     ) % (rule.name,)
                     raise ValidationError(msg)
 
-    def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
+    def get_mto_qty_to_order(
+        self, product, product_qty, product_uom, values, done_procs=None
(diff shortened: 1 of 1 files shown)


rousseldenis:
@ferran-S73 Could you provide a test case in order to highlight changes ?

PR #1878, opened 2023-11-03
[ADD] stock_quant_estimated_value in v15

AaronHForgeFlow:
This module adds the estimated value field in quant for reporting purposes.
This is useful to determine issues in valuation for FIFO and AVG valuated products. For example, to compare with layers valuation

cc @ForgeFlow


rousseldenis:
@AaronHForgeFlow Could you fix tests ?

stock_quant_estimated_value/models/stock_quant.py
```diff
@@ -0,0 +1,18 @@
...
+
+    currency_id = fields.Many2one(
+        comodel_name="res.currency", string="Currency", related="company_id.currency_id"
+    )
+    estimated_value = fields.Float(compute="_compute_estimated_value", store=True)
+
+    @api.depends("quantity")
+    def _compute_estimated_value(self):
+        for record in self:
+            record.estimated_value = record.quantity * record.product_id.standard_price
```
xavier-bouquiaux, line 18: shouldn't you depends on product_id.standard_price ?

xavier-bouquiaux (commented):
LGTM

stock_quant_estimated_value/models/stock_quant.py
AaronHForgeFlow: yes, I thought it won't work on company dependent fields but it does

xavier-bouquiaux (commented):
LGTM

AaronHForgeFlow:
Actually, I can use the value field. I thought it was not, but it is the same thing. Closing.

PR #1880, opened 2023-11-06
[FIX] fix compute method failed to assign

mohmdlabeeb:
Latest Odoo 15 throws a value error with the current compute method

Error:
ValueError: Compute method failed to assign product.template(<NewId 0x7fb38d3512b0>,).qty_available_not_res

Fix:
Assign qty_available_not_res as 0 for the new product


LoisRForgeFlow:
Solved in https://github.com/OCA/stock-logistics-warehouse/pull/1889

PR #1884, opened 2023-11-09
[14.0][FIX] stock_available_mrp: Use correct bom for computation

mt-software-de:
If a product has siblings and more then one has a bom first(product.bom_ids) could return a bom
for a diferent product, because bom_ids is a one2many field of product.template and this contains all boms of variants Therefor switch to the default methods of odoo
_bom_find and _get_product2bom

PR #1887, opened 2023-11-13
[15.0][FIX] stock_available_mrp: Use correct bom for computation

mt-software-de:
If a product has siblings and more then one has a bom first(product.bom_ids) could return a bom
for a diferent product, because bom_ids is a one2many field of product.template and this contains all boms of variants Therefor switch to the default methods of odoo
_bom_find

Forwardport of https://github.com/OCA/stock-logistics-warehouse/pull/1884
cc @rousseldenis @acsonefho


acsonefho (approved):

mt-software-de:
ping @rousseldenis

rousseldenis (approved):

PR #1889, opened 2023-11-15
[FIX] stock_available_unreserved: qty_available_not_res not computed in all cases.

angelgarciadelachica:
The `qty_available_not_res` field is not computed when creating a new product.template.

To reproduce the error:
Go to Inventory module -> Products -> Products -> Create

The following error pops up:
`ValueError: Compute method failed to assign product.template(<NewId 0x7fd28cf332e0>,).qty_available_not_res`

### DIFF
--- a/stock_available_unreserved/models/product_template.py
+++ b/stock_available_unreserved/models/product_template.py
@@ -21,6 +21,7 @@ class ProductTemplate(models.Model):
     def _compute_product_available_not_res(self):
         for tmpl in self:
             if isinstance(tmpl.id, models.NewId):
+                tmpl.qty_available_not_res = False
                 continue
             tmpl.qty_available_not_res = sum(
                 tmpl.mapped("product_variant_ids.qty_available_not_res")


angelgarciadelachica:
@ValentinVinagre @HaraldPanten

ValentinVinagre (approved):
LGTM 👍🏻

etobella (approved):

stock_available_unreserved/models/product_template.py
```diff
@@ -21,6 +21,7 @@ class ProductTemplate(models.Model):
     def _compute_product_available_not_res(self):
         for tmpl in self:
             if isinstance(tmpl.id, models.NewId):
+                tmpl.qty_available_not_res = False
```
rousseldenis, line 24: @angelgarciadelachica @etobella Really ? Should be `0.0` instead.

ValentinVinagre: If you see it better we can change it. But False is not an inappropriate value either, since in the end Odoo represents it as 0.0

PR #1896, opened 2023-11-28
[14.0][FIX] stock_request_analytic: Adding the correct groups to the views to avoid permissions errors

victoralmau:
Adding the correct groups to the views to avoid permissions errors

@Tecnativa

### DIFF
--- a/stock_request_analytic/views/stock_request_order_views.xml
+++ b/stock_request_analytic/views/stock_request_order_views.xml
@@ -6,6 +6,10 @@
         <field name="name">stock.request.order.form - stock_request_analytic</field>
         <field name="model">stock.request.order</field>
         <field name="inherit_id" ref="stock_request.stock_request_order_form" />
+        <field
+            name="groups_id"
+            eval="[(4, ref('analytic.group_analytic_accounting')),(4, ref('analytic.group_analytic_tags'))]"
+        />
         <field name="arch" type="xml">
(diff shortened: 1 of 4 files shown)


stock_request_analytic/views/stock_request_order_views.xml
```diff
@@ -8,8 +8,16 @@
         <field name="inherit_id" ref="stock_request.stock_request_order_form" />
         <field name="arch" type="xml">
             <div name="button_box" position="inside">
-                <field name="analytic_account_ids" invisible="1" />
-                <field name="analytic_tag_ids" invisible="1" />
+                <field
+                    name="analytic_account_ids"
+                    groups="analytic.group_analytic_accounting"
```
pedrobaeza: Putting groups here in this version doesn't avoid permission errors. It should be at view level. Anyway, removing the analytic group is only for not showing the fields, not for permissions.

pedrobaeza (changes requested):
The changes on analytic seems weird

stock_request_analytic/views/stock_request_order_views.xml
victoralmau: Ok, it is already changed and the corresponding groups are set in the views.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #1900, opened 2023-12-01
[16.0][IMP] stock_reserve default_picking_type

acsonefho:
Better to search for the default `picking_type_id` instead of using an XML id. Because the XML ID could not exist anymore and/or be related to another company.

### DIFF
--- a/stock_reserve/model/stock_reserve.py
+++ b/stock_reserve/model/stock_reserve.py
@@ -108,8 +108,13 @@ class StockReservation(models.Model):
 
     @api.model
     def _default_picking_type_id(self):
-        ref = "stock.picking_type_out"
-        return self.env.ref(ref, raise_if_not_found=False).id
+        domain = [
+            ("code", "=", "outgoing"),
+            "|",
+            ("company_id", "=", self.env.company.id),
+            ("company_id", "=", False),
+        ]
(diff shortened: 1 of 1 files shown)


acsonefho:
@rousseldenis @xavier-bouquiaux Can you review and merge please :pray:

stock_reserve/model/stock_reserve.py
```diff
@@ -108,8 +108,13 @@ def get_location_from_ref(self, ref):
 
     @api.model
     def _default_picking_type_id(self):
-        ref = "stock.picking_type_out"
-        return self.env.ref(ref, raise_if_not_found=False).id
+        domain = [
```
rousseldenis, line 111: @acsonefho I would have changed to the company default warehouse 'out_type_id' field.

acsonefho: I agree but in case of multi-warehouse for a same company?

gurneyalex: There is an implicit order on sequence, id in the search for stock.picking.type, so you can control which one gets selected by default. I'm fine with this. 

The alternative is to rely on the warehouse sequence for a given company, this is also fine. 

Whatever the choice, I would document it in the README.

gurneyalex (approved):

PR #1903, opened 2023-12-12
[16.0][FIX] stock_inventory_discrepancy: fix conflicts and discrepancy lines

RodrigoBM:
if you have one line with conflics and another line with discrepancies, the discrepancy wiz is executed first, which is incorrect, as we will have to correct the conflicts first.

Without Changes:

https://github.com/OCA/stock-logistics-warehouse/assets/6359121/22aa0cb3-cea6-4666-a2fa-4f1ed57ee16b

With Changes:

https://github.com/OCA/stock-logistics-warehouse/assets/6359121/d662c3fd-a317-48f3-b97f-ca3cdb3f109e

### DIFF
--- a/stock_inventory_discrepancy/models/stock_quant.py
+++ b/stock_inventory_discrepancy/models/stock_quant.py
@@ -53,10 +53,16 @@ class StockQuant(models.Model):
             )
 
     def action_apply_inventory(self):
-        if self.env.context.get("skip_exceeded_discrepancy", False):
-            return super().action_apply_inventory()
+        self = self.with_context(skip_apply_inventory=True)
         over_discrepancy = self.filtered(lambda r: r.has_over_discrepancy)
-        if over_discrepancy:
+        res = super().action_apply_inventory()
+        if res:
+            return res
+        if (
+            not self.env.context.get("skip_exceeded_discrepancy", False)
+            and over_discrepancy
+        ):
+            self.inventory_quantity_set = True
             action = self.env["ir.actions.act_window"]._for_xml_id(
                 "stock_inventory_discrepancy.confirm_discrepancy_action"
             )
@@ -66,4 +72,10 @@ class StockQuant(models.Model):
                 active_ids=self.ids,
             )
             return action
+        self = self.with_context(skip_apply_inventory=False)
         return super().action_apply_inventory()
+
+    def _apply_inventory(self):
+        if self._context.get("skip_apply_inventory"):
+            return False
+        return super()._apply_inventory()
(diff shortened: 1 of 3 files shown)


RodrigoBM:
ping @carolinafernandez-tecnativa @pedrobaeza

pedrobaeza:
Hi, Rodrigo, why the conflict arises? Maybe it's better to keep the same order, but making the conflict popup to appear later than the discrepancy one.

RodrigoBM:
> Hi, Rodrigo, why the conflict arises? Maybe it's better to keep the same order, but making the conflict popup to appear later than the discrepancy one.

Hi @pedrobaeza,

I think, in a line of inventory adjustment, that we have conflicts and discrepancies at the same time.

First fix the conflict as this will give us the actual quantity with which we can determine if we have a discrepancy later.

On the other hand, if we accept the discrepancy and then the conflict, depending on what we choose to fix the conflict, we might have a discrepancy again which will not be reported to us.

Sorry for the delay in replying

Regards

pedrobaeza:
Can you tell me how a conflict arises (what are the reasons)?

About the code, there's a early return that makes me suspicious that the discrepancy won't be applied taking that path.

xAdrianCif (approved):
Tech and functional review ok! Thanks!

RodrigoBM:
> Can you tell me how a conflict arises (what are the reasons)?
> 
> About the code, there's a early return that makes me suspicious that the discrepancy won't be applied taking that path.

Without Changes:

- Step 1: We select the lines we want to adjust.
    Selección_658
- Step 2: We start inventory adjustment
    Selección_659
- Step 3: We add the differences
    Selección_661
- Step 4: We create a delivery order with the products that we are making the inventory adjustment. (This step will generate conflicts in the inventory adjustment.)
    Selección_662
- Step 5: As we can see, we have two lines with conflict and one with discrepancy.
    Selección_663
- Step 6: If we apply the inventory setting the wizard will not work.
    Selección_664


With Changes:

- Step 1-5: same steps 
- Step 6: now we correct conflicts first, and depending on the choice, we may have new discrepancies. For this reason it is necessary to correct conflicts first
    Selección_668
- Step 7: we accept discrepancies or review these
    Selección_667

@pedrobaeza , I hope I have solved your question?

jbaudoux:
Here is the fix for the failing test https://github.com/OCA/wms/pull/815

pedrobaeza:
@RodrigoBM now I understand, but I'm still reluctant if one line has conflict and rebased discrepancy thresold at the same time that both are shown. Can you please check?

PR #1906, opened 2023-12-21
[MIG][17.0] scrap_reason_code: Migration to 17.0

john-herholz-dt:


scrap_reason_code/tests/test_scrap_reason_code.py
```diff
@@ -106,17 +106,17 @@ def test_scrap_reason_code(self):
...
         )
         self.assertEqual(
             scrap.scrap_qty,
             5,
             "Scrap quantity has been modified and is not " "correct anymore.",
         )
-        move = scrap.move_id
-        self.assertEqual(move.reason_code_id.id, self.reason_code.id)
+        moves = scrap.move_ids
+        self.assertEqual(moves.mapped("reason_code_id.id"), [self.reason_code.id])
```
rousseldenis, line 117: FYI, Odoo supports mapping through all relational records with dot. So, `moves.reason_code_id.id` will work.

rousseldenis (approved):
LGTM. Code and functional review

scrap_reason_code/tests/test_scrap_reason_code.py
john-herholz-dt: @rousseldenis Thanks for looking into it.
I know that, but isn't there a technical possibility that moves are more than one record and moves.reason_code_id.id might end up in a singleton error?

edit: ...but maybe not in the context of this test, true.

rousseldenis: @john-herholz-dt The only case where you have that is if you put in the chain (without mapped()) another field (Boolean, Selection, Char,...)

gurneyalex (approved):

PR #1907, opened 2023-12-23
[16.0][FIX] stock_quant_manual_assign: fixing issue on immediate transfer

rinaldifirdaus:
When we create immediate transfer, and click on  fill with stock button, odoo will propose the available stock and fill the quantity done.
But when you select the other lot, the lot you selected properly added, but the proposed line with done qty will remains. 
This fix is to address this issue, and also the fix for auto_fill_qty_done error when you change the lot in immediate transfer scenario. last but not least, i add also the logic to assign quantity done for immediate transfer when auto_fill_qty_done is activated.

This versions includes the fix on PR [[FIX]stock_quant_manual_assign: bug with boolean #1792](https://github.com/OCA/stock-logistics-warehouse/pull/1792)

@qrtl

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -60,13 +60,27 @@ class AssignManualQuants(models.TransientModel):
 
     def assign_quants(self):
         move = self.move_id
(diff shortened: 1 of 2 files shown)


rinaldifirdaus:
Closing this PR and revised the content on the #1908 PR.

PR #1914, opened 2024-01-13
[15.0][IMP] stock_packaging_calculator: black, isort, prettier

bosd:
Fixes:
image

### DIFF
--- a/stock_packaging_calculator/models/product.py
+++ b/stock_packaging_calculator/models/product.py
@@ -201,7 +201,7 @@ class Product(models.Model):
             as_string.append(f"{unit_qty} {self.uom_id.name}")
         # We want to avoid line break here as this string
         # can be used by reports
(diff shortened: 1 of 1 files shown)


rvalyi (approved):

pedrobaeza:
This is not correct, as it's by design to put a non breakable space.

bosd:
Previously flake8 and autoflake was broken on 15.0 branch.
After version bump of those modules it generated this error.



> This is not correct, as it's by design to put a non breakable space.

Is a update of flake8 rules needed  to exclude this E321?

Let's track it in: https://github.com/OCA/oca-addons-repo-template/pull/243

bosd:
Mentioned error is generated locally by
`pre-commit run -a -v`

When invoking
`pre-commit run -v`
The error is not shwowing. It is ok on CI here as well.
So closing this one.

PR #1915, opened 2024-01-13
[15.0][FIX] stock-logistics-warehouse: copier Repo update

bosd:
This should fix it :)

**Warning:** I've version bumped this manually. To versions higher then from the repo template.
(As the repo template still did'nt fix the tests)

Maybe some aditional settings might need adjustment for this change.


pedrobaeza (approved):

pedrobaeza:
I don't get anyway why you say it's broken, as CI was green also on the other PR.

PR #1918, opened 2024-01-24
[15.0][ADD] stock_picking_auto_revert

AaronHForgeFlow:
Module to correct mistakes in pickings the fast way. Instead of cancel this module will create a return and then will recreate the issued picking.

Based on https://github.com/OCA/stock-logistics-workflow/pull/593 Originally it was discarded because of the stock_cancel module.

cc @ForgeFlow


rousseldenis:
@AaronHForgeFlow Why having put here and not in workflow repo ? Could you explain the difference with stock_cancel module ?

AaronHForgeFlow:
I am not sure what is the difference between the two repos. It was proposed here because this module was proposed to this repo in v10. But I don't mind to move it.

Stock_cancel module allowed to cancel done pickings, by removing the stock move from the quant history and change the location of the quant: https://github.com/OCA/stock-logistics-workflow/blob/10.0/stock_cancel/models/stock_quant.py It was used to correct mistakes when validating pickings. However, there were challenges like covering the case where the product is not anymore in the destination location. Moreover, after the introduction of stock.valuation.layers it was almost impossible to migrate that module with consistency. At the end we followed a similar approach Odoo uses for correcting mistakes in invoices. that is, do a reverse and recreate the original invoice in draft state so the user can edit it.

rousseldenis:
> t was proposed here because this module was proposed to this repo in v10.

It seems not :sweat_smile: https://github.com/OCA/stock-logistics-workflow/pull/593

> At the end we followed a similar approach Odoo uses for correcting mistakes in invoices. that is, do a reverse and recreate the original invoice in draft state so the user can edit it.

Great

AaronHForgeFlow:
> It seems not 😅 https://github.com/OCA/stock-logistics-workflow/pull/593

:rofl: 

Closed in favor of https://github.com/OCA/stock-logistics-workflow/pull/1491

PR #1919, opened 2024-01-25
[15.0][IMP] stock_request: add route_id to stock.request.order

JoanSForgeFlow:
Implemented a new route_id field in stock.request.order that automatically propagates to all associated stock.request lines. If the stock.request lines later have differing routes, the route_id in the order is left empty to reflect this variance.

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -139,9 +139,86 @@ class StockRequestOrder(models.Model):
         string="Stock requests", compute="_compute_stock_request_count", readonly=True
     )
 
+    route_ids = fields.Many2many(
+        "stock.location.route",
+        string="Routes",
+        compute="_compute_route_ids",
+        readonly=True,
+    )
+
+    route_id = fields.Many2one(
+        "stock.location.route",
+        compute="_compute_route_id",
+        inverse="_inverse_route_id",
+        readonly=True,
+        states={"draft": [("readonly", False)]},
+        store=True,
+        help="The route related to a stock request order",
+        domain="[('id', 'in', route_ids)]",
+    )
+
     _sql_constraints = [
         ("name_uniq", "unique(name, company_id)", "Stock Request name must be unique")
     ]
+
+    @api.depends("warehouse_id", "location_id", "stock_request_ids")
+    def _compute_route_ids(self):
+        route_obj = self.env["stock.location.route"]
+        routes = route_obj.search(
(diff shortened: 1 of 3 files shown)


stock_request/models/stock_request_order.py
```diff
@@ -139,10 +139,42 @@ def _get_default_requested_by(self):
...
 
+    @api.depends("stock_request_ids")
+    def _compute_route_id(self):
+        for order in self:
+            if order.stock_request_ids:
+                first_route = order.stock_request_ids[0].route_id or False
+                for request in order.stock_request_ids:
+                    if request.route_id != first_route:
+                        first_route = False
+                        break
```
JordiMForgeFlow:
```suggestion
                if any(r.route_id != first_route for r in order.stock_request_ids):
```

JoanSForgeFlow: comments applied :)

JordiMForgeFlow (approved):
Code + functional 👍🏼

stock_request/models/stock_request_order.py
```diff
@@ -139,10 +139,78 @@ def _get_default_requested_by(self):
...
+                )
+                routes_by_warehouse[warehouse.id] |= route
+        for record in self:
+            routes = route_obj
+            if record.warehouse_id and routes_by_warehouse.get(record.warehouse_id.id):
+                routes |= routes_by_warehouse[record.warehouse_id.id]
+            parents = record.get_parents().ids
+            record.route_ids = routes.filtered(
+                lambda r: any(p.location_id.id in parents for p in r.rule_ids)
+            )
```
LoisRForgeFlow: This one has a couple of problems:

1. It is duplicating code with the equivalent method of stock.request and making maintenance a bit more difficutl.
2. It is completely ignoring product and product category routes making it have fewer possibilities than what actually would have been correct to propose.

image


Proposed approach: just combine the route_ids of all lines. Pseudo code:

```python
def _compute_route_ids(self)
    for rec in self:
        rec.route_ids = rec.mapped("line_ids.route_ids")
```

Downside: if a route is feasible for one line and not for another line one of the requests might not work. To avoid problems, I would add a constraint on stock.request to check that the relected route_id is among route_ids at the time of validation.

JoanSForgeFlow: comments attended

LoisRForgeFlow (approved):
Works better now :+1: 

/ocabot merge minor

PR #1869, opened 2023-10-20
[MIG] stock_orderpoint_generator to 14.0

sergiocorato:

PR #1920, opened 2024-01-26
[14.0] [ADD] stock_orderpoint_template

paradoxxxzero:
This module helps creating orderpoints automatically using templates.
A template is similar to an orderpoint, but it is not linked to a product.

When a picking is validated, the system will look for templates that match
the product and location of the picking. If there is no orderpoint but a 
template is found, it will create an orderpoint for the product and location
of the template.

The new orderpoint is linked to the template, and a change in the template values
will propagate to the orderpoints. To avoid this, you can unlink the orderpoint
by setting its template to False.


rousseldenis:
@paradoxxxzero Have you looked at https://github.com/OCA/stock-logistics-warehouse/pull/1869 ?

paradoxxxzero:
@rousseldenis ah missed that, but it's not quite the same, here the orderpoints are generated lazily and only on picking/move confirmation (not in the run scheduler cron).

rousseldenis:
@paradoxxxzero yes, but as both modules introduce a template model, IMHO this should either be part of the other module either overrides it

paradoxxxzero:
@rousseldenis indeed this is the same model name, but what IF this module was named `stock_orderpoint_blueprint`? :)

Issue #850, opened 2020-02-20
[13.0] in stock_move_location get_putaway_strategy fails


anderwm:
Odoo Server Error

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 619, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 309, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/usr/lib/python3/dist-packages/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 664, in dispatch
    result = self._call_function(**self.params)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 345, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 93, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 338, in checked_call
    result = self.endpoint(*a, **kw)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 909, in __call__
    return self.method(*args, **kw)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 510, in response_wrap
    response = f(*args, **kw)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1323, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1311, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 395, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 382, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/mnt/extra-addons/stock_move_location/wizard/stock_move_location.py", line 176, in action_move_location
    self._create_moves(picking)
  File "/mnt/extra-addons/stock_move_location/wizard/stock_move_location.py", line 143, in _create_moves
    move = self._create_move(picking, lines)
[...]

PR #1921, opened 2024-01-28
[FIX] Update putaway strategy per https://github.com/OCA/stock-logist…

totuk2:
This will fix `AttributeError: 'stock.location' object has no attribute 'get_putaway_strategy'` as in […ics-warehouse/issues/850](https://github.com/OCA/stock-logistics-warehouse/issues/850)

### DIFF
--- a/stock_move_location/wizard/stock_move_location_line.py
+++ b/stock_move_location/wizard/stock_move_location_line.py
@@ -98,7 +98,7 @@ class StockMoveLocationWizardLine(models.TransientModel):
         self.ensure_one()
         location_dest_id = (
             self.move_location_wizard_id.apply_putaway_strategy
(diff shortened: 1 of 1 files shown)


rousseldenis:
@totuk2 Thanks for this. Is there a test case that passes through this code ?

totuk2:
This part of code is not explicitly tested, however this was more a typing mistake. There is no method called `get_putaway_strategy` in any part of the code, however `_get_putaway_strategy` is defined in [addons/stock/models/stock_location.py](https://github.com/OCA/OCB/blob/11872c923c42120bf0f995d8ee11cca5c6e90b1e/addons/stock/models/stock_location.py#L257) and used widely in `stock` and `mrp` modules.

PR #1922, opened 2024-01-29
[15.0][FIX] stock_reserve: Change web_ribbon text to title to make it translatable

victoralmau:
Change web_ribbon text to title to make it translatable

@Tecnativa

### DIFF
--- a/stock_reserve/view/stock_reserve.xml
+++ b/stock_reserve/view/stock_reserve.xml
@@ -34,7 +34,7 @@
                 <sheet>
                     <widget
                         name="web_ribbon"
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

pedrobaeza (changes requested):
I think the POT exportation is not correct, and it's removing a lot of terms. Please just do the change in the XML. The rest will be done by bots.

victoralmau:
Changes done.

pedrobaeza (approved):
/ocabot merge minor

PR #1558, opened 2022-11-18
[14.0][ADD] stock_warehouse_relationship: add warehouse id on various model

petrus-v:
useful for other modules such as stokc_multi_warehouse_security #1543

PR #1923, opened 2024-01-29
[16][MIG] Migrate stock_warehouse_relationship

florian-dacosta:
I did start from https://github.com/OCA/stock-logistics-warehouse/pull/1558 which was stale.

@petrus-v I guess it may interest you
@lmignon maybe since you had made a review on the 14 PR
@bguillot


stock_warehouse_relationship/i18n/stock_warehouse_relationship copy.pot
```diff
@@ -0,0 +1,83 @@
+# Translation of Odoo Server.
```
petrus-v: not sure if this file is present in the former PR! but can be removed!

stock_warehouse_relationship/tests/test_warehouse_relationship.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2022 Foodles (http://www.foodles.co).
```
petrus-v, line 1: this file should probably import from the __init__.py file above unless practices changes in version 16.0 (but I belives I've probably forget it in former PR as long I'm using pytest to launch them

petrus-v (approved):
Thanks taking care of this migration :+1:

stock_warehouse_relationship/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "summary": "Technical module to add warehouse_id field on various stock.* models",
+    "version": "16.0.1.0.0",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "Akretion, Pierre Verkest, Odoo Community Association (OCA)",
+    "maintainers": ["petrus-v"],
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "stock",
```
petrus-v, line 16: just notice that this module is based on  `stock_location_warehouse`
```suggestion
        "stock",
        "stock_location_warehouse",
```

florian-dacosta: I don't think  `stock_location_warehouse` is necessary in v16.
`warehouse_id` field on `stock.location` is present in `stock` module.

petrus-v: ok great I haven't deep into it on v16!

jbaudoux (approved):

PR #1924, opened 2024-01-29
[14.0][ADD] stock_warehouse_relationship: add warehouse id on various model

petrus-v:
useful for other modules such as [stokc_multi_warehouse_security](https://github.com/OCA/stock-logistics-warehouse/pull/1927)

replace #1558

PR #1927, opened 2024-01-30
[14.0][ADD] stock_multi_warehouse_security

petrus-v:
This PR replace https://github.com/OCA/stock-logistics-warehouse/pull/1543

---

With this module you are able to set a restricted list
of allowed warehouses that user can see and operate with.

This module is inspired from the experiences of
`stock_multi_warehouse_security <https://github.com/akretion/stock-logistics-warehouse/tree/12-muli-wh-security/stock_multi_warehouse_security/>`_
on version 12.0 but has some key differences on user experience:

* In this module there is no new groups, user is able to see allowed warehouses
  only or all if not set.
* So in this module there is no "current warehouse" concept on user (
  in v12.0 that module was based on `base_multi_warehouse
  <https://github.com/akretion/stock-logistics-warehouse/tree/12-base-multi_warehouse/base_multi_warehouse>`_
  which allowed users to switch between warehouses).

As request by review, I've split this initial work in 2 modules:
* #1924 which add warehouse_id in various models
* This PR is adding stock_multi_warehouse_security (with #1924 as dependency: last commit will be removed once merged)


gurneyalex (approved):

stock_multi_warehouse_security/security/stock_security.xml
```diff
@@ -0,0 +1,116 @@
...
+        <field eval="1" name="perm_read" />
+        <field eval="1" name="perm_create" />
+    </record>
+
+    <record id="ir_rule_stock_move_allowed_warehouse" model="ir.rule">
+        <field name="name">Stock moves from allowed Warehouse</field>
+        <field name="model_id" ref="stock.model_stock_move" />
+        <field
+            name="domain_force"
+        >['|',('warehouse_id', '=', False), '|', ('warehouse_id', 'in', user.warehouse_ids.ids), (1 if user.warehouse_ids.ids == [] else 0, "=", 1)]</field>
```
sebalix, line 46: We should consider `picking_type_id.warehouse_id` too, which differs from `warehouse_id` (used for routing).

sebalix: And that's where things start to be complicated. I need to implement this scenario e.g.:

User A having access to WH A only, wants to replenish from WH B (pull), without read access to it, or the user B from WH B wants to replenish on their own the WH A (push), without having access to this warehouse. The check on `<stock.move>.warehouse_id` will forbid such use case.

IMO we should have two different `ir.rule` if it's possible, one checking `picking_type_id.warehouse_id` (the WH the move belongs to), and another one checking `warehouse_id` (the WH that can be used for routing purpose). The last one can be archived if needed to ease some workflows.

sebalix (commented):
I opened https://github.com/OCA/stock-logistics-warehouse/pull/2602 without knowing this PR first, but I think the module `stock_warehouse_security` that was done on 16.0 and 18.0 has less issues.

PR #1929, opened 2024-02-02
[14.0] [IMP] sale_stock_on_hand_popup: open in a popup

renda-dev:
Make the quants view open in a new window instead of the current window.

Also, added me and my team as maintainers of this module.


sale_stock_on_hand_popup/models/res_config_settings.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_show_transit_location_stock_wizard = fields.Boolean(
```
rousseldenis, line 7: This should be added in README too.

sale_stock_on_hand_popup/wizards/product_quant_wizard.py
```diff
@@ -0,0 +1,22 @@
...
+    stock_quant_ids = fields.Many2many(
+        "stock.quant", compute="_compute_stock_quant_ids"
+    )
+
+    @api.depends("product_id")
+    def _compute_stock_quant_ids(self):
+        for rec in self:
+            rec.stock_quant_ids = rec.product_id.stock_quant_ids.filtered(
+                lambda x: x.location_id.usage in ["internal", "transit"]
+                if self.env.user.has_group(
```
rousseldenis, line 18: I'm not sure about this. Couldn't this leads to operational misunderstandings as stock user not in that group won't see the product quantities even if they are there.

renda-dev: They are not technically there, we are excluding quants that are in "transit", if the user doesn't have the group it's still gonna be seeing all of the quants that are in a location of type "internal".

It depends on the company if they want to keep in consideration also quants that are in transit or not.

francesco-ooops (approved):
Functional ok!

sale_stock_on_hand_popup/views/sale_order_views.xml
```diff
@@ -1,5 +1,24 @@
...
+        <field
+            name="inherit_id"
+            ref="sale_stock.view_order_form_inherit_sale_stock_qty"
+        />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/form/group/group/div[@name='ordered_qty']/widget"
+                position="after"
+            >
+                <widget name="stock_on_hand_widget" width="0.1" />
```
HekkiMelody: **issue**: width is missing the UoM

sale_stock_on_hand_popup/views/sale_order_views.xml
```diff
@@ -1,5 +1,24 @@
...
+            </xpath>
+            <xpath expr="//field[@name='order_line']/tree/widget" position="after">
```
HekkiMelody: **issue**: this is risky, as there can be other widgets: could you specify the name of the widget you want to find?

sale_stock_on_hand_popup/views/sale_order_views.xml
```diff
@@ -8,7 +27,7 @@
         <xpath expr="." position="inside">
             <script
                 type="text/javascript"
-                src="/sale_stock_on_hand_popup/static/src/js/qty_at_date_widget.js"
+                src="/sale_stock_on_hand_popup/static/src/js/stock_on_hand_widget.js"
             />
         </xpath>
     </template>
```
HekkiMelody: **quibble(non-blocking)**: This declaration is usually put in its own file.

sale_stock_on_hand_popup/wizards/product_quant_wizard.py
HekkiMelody: Wonder if "it depends on the company" should also translate to a company-related setting rather than a setting for the whole installation...

sale_stock_on_hand_popup/models/res_config_settings.py
```diff
@@ -0,0 +1,11 @@
...
+        "Display also transit location lines in sale line stock popup.",
+        implied_group="sale_stock_on_hand_popup.group_show_transit_location_stock_wizard",
+        default=True,
```
HekkiMelody: **issue**: on a a new installation e.g. runboat, users don't actually have this group, this `default=True` is not sufficient for activating the setting by default.

HekkiMelody (changes requested):
Code review, a few comments

sale_stock_on_hand_popup/models/res_config_settings.py
renda-dev: Thanks for letting me know, I tried that just for testing purpose, not really needed

sale_stock_on_hand_popup/views/sale_order_views.xml
renda-dev: Good point, fixed!

sale_stock_on_hand_popup/views/sale_order_views.xml
renda-dev: Moved to [sale_stock_on_hand_popup/views/assets.xml](https://github.com/OCA/stock-logistics-warehouse/pull/1929/files#diff-9d5e7a9589d7fb833fea11a65ad227cb34ed5414aad4e0caab73f7607b32e1ed) 

Let me know if it looks right to you

sale_stock_on_hand_popup/wizards/product_quant_wizard.py
renda-dev: > Wonder if "it depends on the company" should also translate to a company-related setting rather than a setting for the whole installation...

You are quite right, even tho I would prefer to keep it as it is for now

HekkiMelody (approved):
Code review

francesco-ooops (approved):
Functional still good!

@rousseldenis @pedrobaeza as module authors we forgot to add maintainership on initlial PR, would you mind merging? Thank you!

PR #1930, opened 2024-02-05
[17.0][ADD] stock_mts_mto_rule

milinsoft:
Port from v16 to v17


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0. Check also for the rest of explanations for avoiding red CIs.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #1935, opened 2024-02-07
[16.0] stock_inventory: various fixes and improvements

benwillig:
- fixed the way to link the move to the inventory, loading all move lines linked to a product or a lot for a given location was not right
- fixed computed methods
- do not use ValidationError if not inside a python constrains
- use states field attribute to avoid duplicating visibility condition
- missing ensure_one
- added cancel state to match with old odoo inventory
- do not delete odoo menu but deactivate it instead!
- create empty quants when confirming a manual inventory for products which have no quants
- show quants button even if there was 0 to allow creating new ones

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -7,6 +7,9 @@ class InventoryAdjustmentsGroup(models.Model):
     _name = "stock.inventory"
     _description = "Inventory Adjustment Group"
     _order = "date desc, id desc"
+    _inherit = [
+        "mail.thread",
+    ]
 
     name = fields.Char(required=True, default="Inventory", string="Inventory Reference")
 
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow:
Hi @benwillig 

I just saw this PR, it seems we colided with you in v15: https://github.com/OCA/stock-logistics-warehouse/pull/1991

Since, we didn't notice this PR we have generated kind of a conflict when we were thinking about forward porting it to 16.0. Our fault :pray: 

To try to re-align versions and to not make you fix the problem, I propose the following:

- We forward port https://github.com/OCA/stock-logistics-warehouse/pull/1991 as is.
- We rebase your improvements in this PR after that and we help to get it merged.
- Then we backport your improvements to v15 also.

Do you think this plan is fine?

Sorry for the inconvenience

benwillig:
Yes this is fine.

benwillig:
I will create a new PR based on https://github.com/OCA/stock-logistics-warehouse/pull/2017

PR #1936, opened 2024-02-08
[16.0][ADD] stock_picking_dimension

MarinaAForgeFlow:
Depends on PR: https://github.com/OCA/stock-logistics-warehouse/pull/1937


stock_picking_dimensions/models/stock_picking.py
```diff
@@ -0,0 +1,66 @@
...
+                    p.picking_height, meter_uom
+                )
+
+                volume_m3 = m_length * m_width * m_heigth
+                volume = meter3_uom._compute_quantity(volume_m3, p.volume_uom_id)
+            p.volume = volume
+
+    @api.depends("dimension_uom_id")
+    def _compute_dimension_uom_name(self):
+        name = False
```
ChrisOForgeFlow, line 63: This method should iterate over records, and you can simplify getting directly name of record, because return false
```suggestion
        name = False
```

stock_picking_dimensions/models/stock_picking.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.dimension_uom_id:
```
ChrisOForgeFlow, line 64:
```suggestion
```

stock_picking_dimensions/models/stock_picking.py
```diff
@@ -0,0 +1,66 @@
...
+            name = self.dimension_uom_id.name
```
ChrisOForgeFlow, line 65:
```suggestion
```

stock_picking_dimensions/models/stock_picking.py
```diff
@@ -0,0 +1,66 @@
...
+        self.dimension_uom_name = name
```
ChrisOForgeFlow, line 66:
```suggestion
    for record in self:
        record.dimension_uom_name = record.dimension_uom_id.name
```

ChrisOForgeFlow (changes requested):
LGTM, a little change, field dimension_uom_name can be a related field, with store=True if is necessary

LoisRForgeFlow (commented):
Avoid plurals in module names to adhere to [OCA guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#L32).

PR #1937, opened 2024-02-08
[16.0][IMP] stock_picking_volume: allow changing the volume uom

MarinaAForgeFlow:


lmignon (approved):
LGTM (Code review only)

stock_picking_volume/models/stock_picking.py
```diff
@@ -11,18 +11,42 @@ class StockPicking(models.Model):
...
-            if picking.volume != new_volume:
-                picking.volume = new_volume
+            volume_converted = default_uom._compute_quantity(
+                new_volume, self.volume_uom_id
+            )
+            if picking.volume != volume_converted:
+                picking.volume = volume_converted
+
+    @api.onchange("volume_uom_id")
+    def _onchange_volume_uom(self):
```
rousseldenis, line 55: @MarinaAForgeFlow Is this really needed and couldn't be included in compute ?

MarinaAForgeFlow: It is needed in case the user has typed a manual volume after the computation.  If the user changes the uom, we would want to convert that current (manual) value of the volume.

If I add the volume_uom_id to the depend of the _compute_volume method, this would not convert the manually typed amount, it would recompute the volume and then convert that computed volume to the new uom.

That is why I did a separated onchange method, to allow to convert the manually typed in amounts. But now I see that it could be included in the compute of the volume_uom_name. That is what that is what you meant?

PR #1939, opened 2024-02-08
[14.0][FIX] product_route_profile fix view

bguillot:
Hello , 

the route_ids field must be read-only in the view to avoid triggering the inverse method, which recalculates the route_profile_id field when the force_route_profile_id field is filled in.

### DIFF
--- a/product_route_profile/views/product_template.xml
+++ b/product_route_profile/views/product_template.xml
@@ -11,6 +11,12 @@
             <xpath expr="//field[@name='route_ids']/parent::div" position="attributes">
                 <attribute name="attrs">{'invisible': True}</attribute>
             </xpath>
+            <xpath expr="//field[@name='route_ids']" position="attributes">
+                <attribute name="readonly">1</attribute>
+            </xpath>
+            <xpath expr="//label[@for='route_ids']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
             <xpath expr="//group[@name='operations']" position="inside">
                 <field
                     name="route_profile_id"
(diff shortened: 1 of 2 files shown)


product_route_profile/views/product_template.xml
```diff
@@ -11,6 +11,12 @@
             <xpath expr="//field[@name='route_ids']/parent::div" position="attributes">
                 <attribute name="attrs">{'invisible': True}</attribute>
             </xpath>
+            <xpath expr="//field[@name='route_ids']" position="attributes">
+                <attribute name="readonly">1</attribute>
+            </xpath>
+            <xpath expr="//label[@for='route_ids']" position="attributes">
```
rousseldenis, line 17: @bguillot Agree for the first, but should we hide the result routes from profile ?

bguillot: It is already hidden, I only hide the label because it is confusing to have the label but empty

StefanRijnhart:
Thanks! Can you amend your commit to something like
```
[FIX] product_route_profile: prevent triggering inverse method

Set route_ids as readonly to avoid triggering inverse method and changing
the route_profile_id when adding force_route_profile_id
```
(is how I'm cherrypicking this into the 17.0 migration at https://github.com/OCA/stock-logistics-warehouse/pull/2097)

rousseldenis:
@bguillot Could you attend @StefanRijnhart comments? Thanks

bguillot:
sorry, I missed the comments.

It's done

StefanRijnhart (approved):

Kev-Roche (approved):

PR #1941, opened 2024-02-12
stock_orderpoint_generator: add fields trigger and route_id

aayartsev:


stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -80,6 +80,22 @@ class OrderpointTemplate(models.Model):
         "scheduled action for every product in this list.",
     )
     auto_last_generation = fields.Datetime(string="Last Automatic Generation")
+    trigger = fields.Selection(
+        selection=[
+            ("auto", "Auto"),
+            ("manual", "Manual"),
```
rousseldenis, line 87: @aayartsev Could you explain the aim of this ? As manual triggering for orderpoint creation... is manual creation

aayartsev: @rousseldenis Value of this field will be copy to record wich will be created from this template.

geomer198 (approved):
LGTM!

stock_orderpoint_generator/models/orderpoint_template.py
rousseldenis: @aayartsev Could you add your changes in the USAGE.rst ? Thanks

dessanhemrayev: Hi @rousseldenis  )
I have added our changes to USAGE.rst
please check)

PR #1945, opened 2024-02-16
[15.0][ADD] stock_move_purchase_uom

MarinaAForgeFlow:


stock_move_purchase_uom/models/stock_move.py
```diff
@@ -0,0 +1,15 @@
+from odoo import api, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    @api.onchange("product_id", "picking_type_id")
+    def _onchange_product_id(self):
+        super()._onchange_product_id()
```
JordiMForgeFlow:
```suggestion
        res = super()._onchange_product_id()
```

stock_move_purchase_uom/models/stock_move.py
```diff
@@ -0,0 +1,15 @@
...
+        if self.picking_type_id and self.picking_type_id.use_purchase_uom:
+            self.product_uom = self.product_id.with_context(
+                lang=self._get_lang()
+            ).uom_po_id.id
+
+        return
```
JordiMForgeFlow:
```suggestion
        return res
```

stock_move_purchase_uom/models/stock_move.py
```diff
@@ -0,0 +1,15 @@
+from odoo import api, models
```
LoisRForgeFlow, line 3: You should add the copyrights in all python files excluding __init__'s

LoisRForgeFlow (commented):
LGTM

Can you squash all your commits?

rousseldenis:
My question regarding this is: isn't it possible to use product packaging in this case ?

LoisRForgeFlow:
> My question regarding this is: isn't it possible to use product packaging in this case ?

We also evaluated that option, but in this case UoM was really embedded in many processes and therefore this proposal was easier to introduce.

LoisRForgeFlow (approved):
LGTM

/ocabot merge nobump

PR #1949, opened 2024-02-22
[14.0][FIX] stock_request: Avoid set stock request orders empty

victoralmau:
Avoid set stock request orders empty.

@Tecnativa TT47959

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -320,6 +320,16 @@ class StockRequestOrder(models.Model):
                 )
             )
 
+    @api.constrains("stock_request_ids")
+    def _check_location_empty_stock_request_ids(self):
+        if any(not request.stock_request_ids for request in self):
+            raise ValidationError(
+                _(
+                    "It is not possible to set empty stock request orders (maybe "
+                    "you should cancel it)."
+                )
+            )
+
     @api.model
     def _create_from_product_multiselect(self, products):
         if not products:
--- a/stock_request/i18n/es.po
+++ b/stock_request/i18n/es.po
@@ -9,8 +9,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 11.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-07-06 07:32+0000\n"
-"PO-Revision-Date: 2023-07-06 09:32+0200\n"
+"POT-Creation-Date: 2024-02-23 08:52+0000\n"
+"PO-Revision-Date: 2024-02-23 09:53+0100\n"
 "Last-Translator: Víctor Martínez <victor.martinez@tecnativa.com>\n"
 "Language-Team: Spanish (https://www.transifex.com/oca/teams/23907/es/)\n"
 "Language: es\n"
@@ -422,6 +422,19 @@ msgid "Is Follower"
 msgstr "Es Seguidor"
 
 #. module: stock_request
+#: code:addons/stock_request/models/stock_request_order.py:0
+#, fuzzy, python-format
+#| msgid ""
+#| "It is not possible to set empty stock request orders (maybe\n"
+#| "                    you should cancel it)."
+msgid ""
+"It is not possible to set empty stock request orders (maybe you should "
+"cancel it)."
+msgstr ""
+"No es posible dejar vacío un pedido de existencias (quizás deberías "
+"cancelarlo)."
+
+#. module: stock_request
 #: model_terms:ir.ui.view,arch_db:stock_request.stock_request_order_form
 msgid "Items"
 msgstr "Elementos"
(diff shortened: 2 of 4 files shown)


pedrobaeza (commented):
I see more logical to not allow to confirm empty requests, like sales and purchase orders.

victoralmau:
This changes does not allow to confirm, only to cancel (and change them to draft) if they are empty.

pedrobaeza:
OK, seeing the definition of the field, it's a computed readonly=True, so you should modify `_compute_state` instead, for achieving what you want to do. There's no sense in having an empty request in other state except `draft`. If you get it, it's twisting the ORM writing a value in the field.

rousseldenis (approved):
Code review

rousseldenis:
@victoralmau Didn't see @pedrobaeza comment. Indeed. The compute seems to be ok, but the depends misses a trigger on 'stock_request_ids' (without state).

victoralmau:
Summary of changes done in this PR:
- Modified `_compute_state()` method to **only** synchronize state when there are `stock_request_ids` set (remember that depends is only `stock_request_ids.state`).
- Allow to cancel empty stock request orders.
- Allow to change to draft empty stock request orders.
- Add a constrains that prevents to set empty stock request orders in state other than Draft or Cancel.
- Test to check all the explained use cases.

stock_request/models/stock_request_order.py
```diff
@@ -148,9 +148,12 @@ def _get_default_requested_by(self):
 
     @api.depends("stock_request_ids.state")
     def _compute_state(self):
-        for item in self:
+        """Synch the status according to the status of stock_request_ids.
+        If no stock_request_ids is set we leave the status 'free' (could be draft
+        or cancel)."""
+        for item in self.filtered("stock_request_ids"):
             states = item.stock_request_ids.mapped("state")
-            if not item.stock_request_ids or all(x == "draft" for x in states):
```
pedrobaeza: I think this was correct before.

stock_request/models/stock_request_order.py
```diff
@@ -244,11 +247,17 @@ def action_confirm(self):
         return True
 
     def action_draft(self):
-        self.mapped("stock_request_ids").action_draft()
+        if self.stock_request_ids:
+            self.mapped("stock_request_ids").action_draft()
+        else:
+            self.state = "draft"
```
pedrobaeza: The problem is here: you can't set `state` field directly: it's a computed readonly=True.

stock_request/models/stock_request_order.py
```diff
@@ -244,11 +247,17 @@ def action_confirm(self):
...
         return True
 
     def action_cancel(self):
-        self.mapped("stock_request_ids").action_cancel()
+        if self.stock_request_ids:
```
pedrobaeza: By definition, an empty request will never be cancel.

pedrobaeza (changes requested):

stock_request/models/stock_request_order.py
victoralmau: Example of use case:
- Create an empty stock request order.
- Check that it is an "_error_" and we don't need to add `stock_request_ids`.

Current behavior:
- It is not possible to cancel it. It will remain as a draft "_forever_".
- It is not possible to  delete it. Only users with the Stock Request Manager permission can do it .

What change do you suggest to cover this use case?

pedrobaeza: > Create an empty stock request order.

Don't allow to save an empty request order then.

victoralmau: Ok, another use case.

- Create a stock request order with several stock_request_ids and save.
- Remove `stock_request_ids` from the stock request order. 
- It is not possible to cancel the empty stock request order.

Also in this case should we not allow to save empty stock request order? The process in that case should be just to cancel it, right?

pedrobaeza: 2 options:

- Auto-remove empty request orders when they are "emptied".
- Allow to remove empty request orders by any user authorized to add them.

victoralmau:
Changed to a simpler approach, not allowing to set empty stock request orders.

stock_request/models/stock_request_order.py
```diff
@@ -320,6 +320,16 @@ def _check_location_company(self):
                 )
             )
 
+    @api.constrains("stock_request_ids")
+    def _check_location_empty_stock_request_ids(self):
+        if any(not request.stock_request_ids for request in self):
+            raise ValidationError(
+                _(
+                    """It is not possible to set empty stock request orders (maybe """
+                    """you should cancel it)."""
```
pedrobaeza: No need of triple quotes:

```suggestion
                    "It is not possible to set empty stock request orders (maybe "
                    "you should cancel it)."
```

pedrobaeza (commented):
We still need to remove the code that set directly the state on the object, as the field is a computed readonly.

victoralmau:
> We still need to remove the code that set directly the state on the object, as the field is a computed readonly.

The code you indicate no longer exists, now this PR only adds the constrains in stock request orders.

pedrobaeza (approved):
/ocabot merge patch

PR #1950, opened 2024-02-23
[FIX] stock_mts_mto_rule: Don't use current company to create route

rousseldenis:
As the current company will be used to create route demo data, this will prevent creating new companies in tests.

Set the route as shared between companies.


rousseldenis:
@florian-dacosta

pedrobaeza (approved):

rousseldenis:
/ocabot merge patch

As making tests failing in 16.0

PR #1953, opened 2024-02-28
[15.0][FIX] stock_request: Avoid set stock request orders empty

victoralmau:
FWP from 14.0: https://github.com/OCA/stock-logistics-warehouse/pull/1949

Avoid set stock request orders empty.

@Tecnativa TT47959

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -317,6 +317,16 @@ class StockRequestOrder(models.Model):
                 )
             )
 
+    @api.constrains("stock_request_ids")
+    def _check_location_empty_stock_request_ids(self):
+        if any(not request.stock_request_ids for request in self):
+            raise ValidationError(
+                _(
+                    "It is not possible to set empty stock request orders (maybe "
+                    "you should cancel it)."
+                )
+            )
+
     @api.model
     def _create_from_product_multiselect(self, products):
         if not products:
--- a/stock_request/i18n/es.po
+++ b/stock_request/i18n/es.po
@@ -9,9 +9,9 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 11.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-01-26 16:12+0000\n"
-"PO-Revision-Date: 2023-03-03 09:06+0000\n"
-"Last-Translator: Leonardo J. Caballero G. <leonardocaballero@gmail.com>\n"
+"POT-Creation-Date: 2024-02-23 08:52+0000\n"
+"PO-Revision-Date: 2024-02-23 09:53+0100\n"
+"Last-Translator: Víctor Martínez <victor.martinez@tecnativa.com>\n"
 "Language-Team: Spanish (https://www.transifex.com/oca/teams/23907/es/)\n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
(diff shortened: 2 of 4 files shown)


stock_request/models/stock_request_order.py
```diff
@@ -317,6 +317,16 @@ def _check_location_company(self):
                 )
             )
 
+    @api.constrains("stock_request_ids")
+    def _check_location_empty_stock_request_ids(self):
+        if any(not request.stock_request_ids for request in self):
+            raise ValidationError(
+                _(
+                    "It is not possible to set empty stock request orders (maybe "
+                    "you should cancel it)."
```
pedrobaeza, line 326: AFAIK, you can't cancel the order, so the message is misleading.

victoralmau: I think it is possible. Create a stock request order with lines and save. Try to delete lines and save (error, if you want that you should cancel it).

stock_request/models/stock_request_order.py
```diff
@@ -317,6 +317,16 @@ def _check_location_company(self):
                 )
             )
 
+    @api.constrains("stock_request_ids")
```
rousseldenis, line 320: IMHO, this is too restricting. As you maybe want to begin a stock request before filling it in. I'm not in favor of this.

As always, maybe with a configuration parameter.

pedrobaeza: There's no sense in having an empty one, Denis. If you want to do it later, just do it later. Stock request orders are literally a grouping over stock requests.

stock_request/models/stock_request_order.py
pedrobaeza: No, Víctor, you can only cancel orders with requests, not empty ones, according to the computed field.

stock_request/models/stock_request_order.py
rousseldenis: Yes, but this seems a little bit restrictive. Nothing in Odoo restricts void purchases, sales, pickings... This is not a fix but a major change

pedrobaeza: They are not exactly the same type of container. This solution comes after the discussion in #1949, as the less bad option. Please check that one.

rousseldenis: Yes, I didn't follow the end conversation and you merged fast.

So, this refrains people to void a draft request order and then reflll it in after ?

I don't get the use case when you need to constrains.

pedrobaeza: @victoralmau can you comment the steps to reproduce the problem? Basically it's to remove the orders in an existing one.

victoralmau: Related to https://github.com/OCA/stock-logistics-warehouse/pull/1953#discussion_r1506122197

Maybe an explanation will help to understand it better.

ejemplo

If we have a stock request order already created with lines, we can delete some of them, but not all. If we want to delete all of them, the process will be to cancel the stock request order.

rousseldenis: Ughhh, yes but this is the change you introduce. Not the current behaviour. :sweat_smile:

victoralmau: The current behavior is as indicated in the previous PR, deleting all the lines of a stock request order would cause it to be set in "_draft_" status but could not be cancelled, causing confusion.

rousseldenis: I don't see confusion remaining in draft mode in that case. As I said in the other PR, the compute should be changed instead introducing a new constraint.

victoralmau: In the other PR, the possibility of changing the compute was raised and discarded. The status of the stock request order is and must be that of the lines that contain it, and to avoid not being able to cancel an empty stock request order, precisely this change is added.

PR #1958, opened 2024-03-15
[14.0][IMP] stock_mts_mto_rule: add split option

TDu:
With this change a new option is available on the stock rule. It allows to use the MTS rule only if the full quantity is avaiable.

### DIFF
--- a/stock_mts_mto_rule/models/stock_rule.py
+++ b/stock_mts_mto_rule/models/stock_rule.py
@@ -15,6 +15,16 @@ class StockRule(models.Model):
     )
     mts_rule_id = fields.Many2one("stock.rule", string="MTS Rule", check_company=True)
     mto_rule_id = fields.Many2one("stock.rule", string="MTO Rule", check_company=True)
+    mts_quantity_rule = fields.Selection(
+        string="MTS quantity rule",
+        selection=[("available", "Available"), ("full", "No Split")],
+        default="available",
+        help=(
+            "With `Available` the MTS rule will be used for any available quantity.\n"
+            "With`No Split` the MTS rule is only used, if the full requested quantity "
+            "is available."
+        ),
+    )
 
     @api.constrains("action", "mts_rule_id", "mto_rule_id")
     def _check_mts_mto_rule(self):
@@ -26,6 +36,10 @@ class StockRule(models.Model):
                     ) % (rule.name,)
                     raise ValidationError(msg)
 
+    @api.model
+    def _get_available_quantity_mts_mto_rule(self, product):
+        return product.virtual_available
+
     def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
         self.ensure_one()
         precision = self.env["decimal.precision"].precision_get(
@@ -33,7 +47,7 @@ class StockRule(models.Model):
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
+        virtual_available = self._get_available_quantity_mts_mto_rule(product_location)
         qty_available = product.uom_id._compute_quantity(virtual_available, product_uom)
         if float_compare(qty_available, 0.0, precision_digits=precision) > 0:
             if (
@@ -41,6 +55,8 @@ class StockRule(models.Model):
                 >= 0
             ):
                 return 0.0
+            elif self.mts_quantity_rule == "full":
+                return product_qty
             else:
                 return product_qty - qty_available
         return product_qty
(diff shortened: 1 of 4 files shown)


stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -26,14 +35,18 @@ def _check_mts_mto_rule(self):
...
+
     def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
         self.ensure_one()
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
+        virtual_available = self._get_available_quantity_mts_mto_rule(product_location)
```
mt-software-de, line 50: Could we simplify this computation?
And adding the check for the mts_mto_split_quantity_rule here instead of in _run_split_procurement?

```suggestion
        virtual_available = self._get_available_quantity_mts_mto_rule(product_location)
        qty_available = product.uom_id._compute_quantity(virtual_available, product_uom)
        qty_mto = product_qty - qty_available
        if float_compare(qty_mto, 0, precision_digits=precision) <= 0:
            return 0
        if rule.mts_mto_split_quantity_rule == "full":
            return product_qty
        return qty_mto
```

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -15,6 +15,15 @@ class StockRule(models.Model):
...
     mto_rule_id = fields.Many2one("stock.rule", string="MTO Rule", check_company=True)
+    mts_mto_split_quantity_rule = fields.Selection(
+        string="MTS/MTO split rule",
+        selection=[("diff", "Difference"), ("full", "No Split")],
+        default="diff",
+        help=(
+            "With `Difference` the MTS rule will be used for any available quantity.\n"
+            "With`No Split` only if the full quantity is avaiable."
+        ),
+    )
```
mt-software-de:
```suggestion
    mts_quantity_rule = fields.Selection(
        string="MTS quantity rule",
        selection=[("available", "Available"), ("full", "No Split")],
        default="available",
        help=(
            "With `Available` the MTS rule will be used for any available quantity.\n"
            "With`No Split` the MTS rule is only used, if the full requested quantity is available"
        ),
    )
```

mt-software-de (changes requested):

stock_mts_mto_rule/models/stock_rule.py
TDu: Updated

stock_mts_mto_rule/models/stock_rule.py
TDu: I am not sure of this.
We don't have access to `rule` in this function, and I don't like changing the signature of an existing function.
And it also makes the diff bigger.

florian-dacosta:
@TDu Isn't it the goal of the native procure method `mts_else_mto` ?
I did not check this recently, but I had in mind the stock_mts_mto_rule module was only still usefull for the case we want to split between mts and mto when there is an available qty in stock but is not enough for the whole procurement. Else, native Odoo system do the job. Maybe I am missing something though

TDu:
> @TDu Isn't it the goal of the native procure method `mts_else_mto` ? I did not check this recently, but I had in mind the stock_mts_mto_rule module was only still usefull for the case we want to split between mts and mto when there is an available qty in stock but is not enough for the whole procurement. Else, native Odoo system do the job. Maybe I am missing something though

I don't think I can fix my need with the `mts_else_mto` method. Because I can not choose the next procurement method, in my case if the stock is not available I want to call the dropshipping rule.

stock_mts_mto_rule/models/stock_rule.py
mt-software-de: we have access to the rule it is self.

mt-software-de: yes it makes the diff bigger, but i think it makes it much more understandable.
The method _run_split_procurement would not change. 

Because we only need to change the returned qty.

TDu: > we have access to the rule it is self.

oh :blush:  you are right

mmequignon (approved):

stock_mts_mto_rule/models/stock_rule.py
mt-software-de: @TDu do you want to simplify the computation? :-)

mt-software-de: ping @TDu ?

TDu: Thanks for your review. But I would like to keep the computation as it is, now.

mt-software-de (approved):
LGTM

jbaudoux (approved):

jbaudoux:
@TDu Please squash commits

TDu:
Squashed

jbaudoux:
@TDu I forgot to ask you to also drop changes to generated files (README.rst & index.html)to ease fwd port

PR #1959, opened 2024-03-15
[17.0][MIG] stock_helper

CarlosVForgeFlow:
Migration to 17.0

@ForgeFlow


stock_helper/tests/test_location_source_from_route.py
```diff
@@ -37,4 +37,4 @@ def test_get_source_location_from_route(self):
         source_location = location._get_source_location_from_route(
             route, "make_to_stock"
         )
-        self.assertFalse(source_location)
+        self.assertEqual(source_location, self.supplier_loc)
```
LoisRForgeFlow, line 40: Can you explain this change? If a test is failing, typically means that some behavior has changed.

rousseldenis:
@LoisRForgeFlow @jbaudoux An improvement I requested was to refactor the module name in order to avoid putting a lot of things inside it without context.

stock_location_sublocation_of / stock_location_from_route /... is better

or at least stock_helper_location_sublocation / stock_helper_location_from_route if we want to keep the 'helper' term

stock_helper/tests/test_location_source_from_route.py
CarlosVForgeFlow: The thing is that the method from the stock module
`    @api.model
    def _get_rule(self, product_id, location_id, values):
        """ Find a pull rule for the location_id, fallback on the parent
        locations if it could not be found.
        """
        result = self.env['stock.rule']
        location = location_id
        while (not result) and location:
            domain = self._get_rule_domain(location, values)
            result = self._search_rule(values.get('route_ids', False), values.get('product_packaging_id', False), product_id, values.get('warehouse_id', False), domain)
            location = location.location_id
        return result`

Has now been changed to
` @api.model
    def _get_rule(self, product_id, location_id, values):
        """ Find a pull rule for the location_id, fallback on the parent
        locations if it could not be found.
        """
        result = self.env['stock.rule']
        location = location_id
        while (not result) and location:
            domain = self._get_rule_domain(location, values)
            result = self._search_rule(values.get('route_ids', False), values.get('product_packaging_id', False), product_id, values.get('warehouse_id', location.warehouse_id), domain)
            location = location.location_id
        return result`

And therefore it doesn't return `False` for the warehouse_id anymore.

LoisRForgeFlow: Fair enough, thanks for pointing the difference :+1:

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
> @LoisRForgeFlow @jbaudoux An improvement I requested was to refactor the module name in order to avoid putting a lot of things inside it without context.
> 
> stock_location_sublocation_of / stock_location_from_route /... is better
> 
> or at least stock_helper_location_sublocation / stock_helper_location_from_route if we want to keep the 'helper' term

Right now it is only 2 helpers and none of them makes or generates any problem if not used. I don't know if such atomization would provide real value in this case.

My 2 cents

DavidJForgeFlow (approved):

PR #1964, opened 2024-03-18
[16.0][FIX] stock_move_location: Move Line destination would stay the default

### LINKED ISSUE #1963: [16.0] stock_move_location: wrong destination location
## Module
stock_move_location

## Describe the bug
The created Stock Move has a correct destination location set, but the stock move lines, are not updated with the same destination location.

## To Reproduce
**Affected versions**:
16.0

1. On the Internal Transfers Operation type enable "Show Move On hand stock?"
2. On the Inventory Overview, click on the "Move On Hand" button on the Internal Transfers
3. Modify the Source Location and the Destination Location
4. Create a Planned Transfer

**Expected behavior**
The stock should move to the destination location.
Is is being moved to the same origin location.

**Additional context**
image

image

dreispt:
Closes https://github.com/OCA/stock-logistics-warehouse/issues/1963

### DIFF
--- a/stock_move_location/wizard/stock_move_location.xml
+++ b/stock_move_location/wizard/stock_move_location.xml
@@ -106,8 +106,14 @@
                                                     name="lot_id"
                                                     groups="stock.group_production_lot"
                                                 />
-                                                <field name="origin_location_id" />
-                                                <field name="destination_location_id" />
+                                                <field
+                                                    name="origin_location_id"
+                                                    force_save="1"
+                                                />
+                                                <field
+                                                    name="destination_location_id"
+                                                    force_save="1"
+                                                />
                                             </div>
                                             <div class="o_kanban_record_bottom">
                                                 <div class="oe_kanban_bottom_right">


stock_move_location/wizard/stock_move_location.xml
```diff
@@ -106,8 +106,14 @@
                                                     name="lot_id"
                                                     groups="stock.group_production_lot"
                                                 />
-                                                <field name="origin_location_id" />
-                                                <field name="destination_location_id" />
+                                                <field
+                                                    name="origin_location_id"
+                                                    force_save="1"
```
rousseldenis, line 111: @dreispt Are you sure this is necessary as force_save should be used if field is readonly

dreispt: I am sure: these values are changed though a form `onchange`, but the changes are not saved because it is read only.
The `force_save` is already there for the list view, but I found it is also needed for the kanban view.

PR #1966, opened 2024-03-20
[16.0][IMP] stock_inventory_discrepancy: add configuration for enabling discrepancy check

AungKoKoLin1997:
This PR aims to resolve the compatibility issue between this module and other inventory adjustment related modules.
Related: https://github.com/OCA/stock-logistics-warehouse/pull/1711

@qrtl QT4850

### DIFF
--- a/stock_inventory_discrepancy/models/__init__.py
+++ b/stock_inventory_discrepancy/models/__init__.py
@@ -1,5 +1,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from . import res_company
+from . import res_config_settings
 from . import stock_quant
 from . import stock_warehouse
 from . import stock_location
--- a/stock_inventory_discrepancy/models/res_config_settings.py
+++ b/stock_inventory_discrepancy/models/res_config_settings.py
@@ -1,0 +1,16 @@
+# Copyright 2024 Quartile
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    inventory_discrepancy_enable = fields.Boolean(
+        related="company_id.inventory_discrepancy_enable",
+        readonly=False,
+        string="Inventory Discrepancy Control",
+        help="Block validation of the inventory adjustment if discrepancy exceeds "
+        "the threshold.",
+    )
--- a/stock_inventory_discrepancy/models/stock_quant.py
+++ b/stock_inventory_discrepancy/models/stock_quant.py
@@ -76,6 +76,8 @@ class StockQuant(models.Model):
     def action_apply_inventory(self):
         if self.env.context.get("skip_exceeded_discrepancy", False):
             return super().action_apply_inventory()
+        if not self.env.company.inventory_discrepancy_enable:
+            return super().action_apply_inventory()
         over_discrepancy = self.filtered(lambda r: r.has_over_discrepancy)
         if over_discrepancy:
             action = self.env["ir.actions.act_window"]._for_xml_id(
--- a/stock_inventory_discrepancy/views/res_config_settings_view.xml
+++ b/stock_inventory_discrepancy/views/res_config_settings_view.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_stock_config_settings" model="ir.ui.view">
+        <field name="name">res.config.settings - inventory_discrepancy</field>
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="stock.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='warning_info']" position="after">
+                <div
+                    class="col-12 col-lg-6 o_setting_box"
+                    id="inventory_discrepancy_enable"
+                >
+                    <div class="o_setting_left_pane">
+                        <field name="inventory_discrepancy_enable" />
+                    </div>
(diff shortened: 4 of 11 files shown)


stock_inventory_discrepancy/models/stock_quant.py
```diff
@@ -53,7 +53,13 @@ def _compute_has_over_discrepancy(self):
...
     def action_apply_inventory(self):
-        if self.env.context.get("skip_exceeded_discrepancy", False):
+        # Get the system parameter configuration
+        param = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("stock_inventory_discrepancy.inventory_discrepancy_enable")
+        )
+        if self.env.context.get("skip_exceeded_discrepancy", False) or param == "0":
             return super().action_apply_inventory()
```
yostashiro: For early returns.
```suggestion
        if self.env.context.get("skip_exceeded_discrepancy", False):
            return super().action_apply_inventory()
        inventory_discrepancy_enable = (
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("stock_inventory_discrepancy.inventory_discrepancy_enable")
        )
        if inventory_discrepancy_enable == "0":
            return super().action_apply_inventory()
```

yostashiro (approved):
Code review.

kanda999 (approved):
Code review: LGTM

AungKoKoLin1997:
@OCA/stock-logistics-warehouse-maintainers
Can you please review this PR?

rousseldenis (approved):
Code review.

@LoisRForgeFlow @JordiBForgeFlow

JordiBForgeFlow:
@AungKoKoLin1997 From a usability perspective I think that it would be much better to add a setting in res.config.settings to enable / disable it, and document it in the readme of the module. A checkbox in res.config.settings would then update the parameter that you have proposed.

LoisRForgeFlow:
> @AungKoKoLin1997 From a usability perspective I think that it would be much better to add a setting in res.config.settings to enable / disable it, and document it in the readme of the module. A checkbox in res.config.settings would then update the parameter that you have proposed.

I agree

yostashiro:
> @AungKoKoLin1997 From a usability perspective I think that it would be much better to add a setting in res.config.settings to enable / disable it, and document it in the readme of the module. A checkbox in res.config.settings would then update the parameter that you have proposed.

@JordiBForgeFlow @LoisRForgeFlow The purpose of this update was more to pass tests from another module rather than providing additional config option to users, and as such I personally thought updating res.config.settings wouldn't be necessary (wasn't sure if there would be some users who'd like to disable the feature after installing the module with config setting when there are already other ways to bypass the check with adjusting location/warehouse settings).  Having said that, if the community opts for having a field in res.config.settings, why not!

rousseldenis:
> The purpose of this update was more to pass tests from another module rather than providing additional config option to users

Be careful. Adding code just for tests should trigger 'warnings'. But, that said, adding a configuration that is accessible for users is always a good process to enable/disable a feature.

For several reasons:

- To be able to disable process in case of bug or... without technical requirements.
- To be able to deploy a solution without enabling it immediately

yostashiro:
Thank you. @AungKoKoLin1997 Can you please follow up to add a configuration?

AungKoKoLin1997:
@LoisRForgeFlow @JordiBForgeFlow  I added the configuration from a usability perspective. Can you please review it?

stock_inventory_discrepancy/models/res_config_settings.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    inventory_discrepancy_enable = fields.Boolean(
+        string="Inventory Discrepancy Control",
+        help="Enable to show line discrepancies in inventory and block validation "
+        "if the discrepancy exceeds the threshold.",
```
yostashiro:
```suggestion
        help="Block validation of the inventory adjustment if discrepancy exceeds the threshold.",
```
Since the setting doesn't affect the UI.

stock_inventory_discrepancy/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,26 @@
...
+                    class="col-12 col-lg-6 o_setting_box"
+                    id="inventory_discrepancy_enable"
+                >
+                    <div class="o_setting_left_pane">
+                        <field name="inventory_discrepancy_enable" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label for="inventory_discrepancy_enable" />
+                        <div class="text-muted">
+                            Enable to show line discrepancies in inventory and block validation if the discrepancy exceeds the threshold.
```
yostashiro: Same as above.

yostashiro (approved):
Code review. 👍

AungKoKoLin1997:
@rousseldenis @LoisRForgeFlow @JordiBForgeFlow 
I have update the PR and I believe it is in good state.
Can you please review?

AungKoKoLin1997:
@kanda999 Could you please review?

AungKoKoLin1997:
@rousseldenis I changed the approach like we did for https://github.com/OCA/stock-logistics-workflow/pull/1760.
Can you please review?

stock_inventory_discrepancy/models/__init__.py
```diff
@@ -3,3 +3,5 @@
 from . import stock_quant
 from . import stock_warehouse
 from . import stock_location
+from . import res_config_settings
+from . import res_company
```
yostashiro: I'd suggest putting them in alphabetical order unless there is a reason not to.

yostashiro (approved):
Code review.

rousseldenis (approved):

PR #1967, opened 2024-03-21
[17.0][MIG] account_move_line_stock_info

CarlosVForgeFlow:
Migration to 17.0

@ForgeFlow


account_move_line_stock_info/views/stock_move_view.xml
```diff
@@ -18,17 +23,60 @@
...
             >
-                <group name="account_move_lines_grp" string="Journal Items">
-                    <field
-                        name="account_move_line_ids"
-                        readonly="1"
-                        nolabel="1"
-                        colspan="2"
-                    />
-                </group>
+                <form string="Stock Moves">
```
JordiBForgeFlow, line 29: @CarlosVForgeFlow why are you doing this? You are breaking the inheritance, and the account_move_line_ids should only be added as a field to the form view.

CarlosVForgeFlow: The form view "Stock Moves" that this xpath inherited from has been eliminated in Odoo 17.0. I added it here for the view to be the same as before.

StefanRijnhart: How about showing them in the tree-view instead, with the many2many_tags widget (with optional="hide") plus a button to open them?

StefanRijnhart:
Superseded in https://github.com/OCA/stock-logistics-warehouse/pull/2160. On validated transfers, a simplified stock move form pops up when clicking on any of the operation lines. I'm removing the whole form view that this PR inserts at that point and adding the journal items to the simplified form.
image

PR #1977, opened 2024-04-03
[15.0][IMP] stock_move_location: fill picking_type_id depends on origin_location_id

AungKoKoLin1997:
This improvement commit ensures the operation type is determined by the original location, aiming to prevent the misuse of incorrect operation types in scenarios where multiple internal transfer options exist.

@qrtl QT4470

### DIFF
--- a/stock_move_location/wizard/stock_move_location.py
+++ b/stock_move_location/wizard/stock_move_location.py
@@ -7,25 +7,12 @@ from itertools import groupby
 
 from odoo import api, fields, models
 from odoo.fields import first
+from odoo.osv import expression
 
 
 class StockMoveLocationWizard(models.TransientModel):
     _name = "wiz.stock.move.location"
     _description = "Wizard move location"
-
-    def _get_default_picking_type_id(self):
-        company_id = self.env.context.get("company_id") or self.env.user.company_id.id
-        return (
-            self.env["stock.picking.type"]
-            .search(
-                [
-                    ("code", "=", "internal"),
-                    ("warehouse_id.company_id", "=", company_id),
-                ],
-                limit=1,
-            )
-            .id
-        )
 
     origin_location_disable = fields.Boolean(
         compute="_compute_readonly_locations",
@@ -54,7 +41,9 @@ class StockMoveLocationWizard(models.TransientModel):
         column2="move_location_line_wiz_id",
     )
     picking_type_id = fields.Many2one(
-        comodel_name="stock.picking.type", default=_get_default_picking_type_id
+        compute="_compute_picking_type_id",
+        comodel_name="stock.picking.type",
+        readonly=False,
     )
     picking_id = fields.Many2one(
         string="Connected Picking", comodel_name="stock.picking"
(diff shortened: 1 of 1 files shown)


stock_move_location/wizard/stock_move_location.py
```diff
@@ -138,6 +138,20 @@ def _prepare_wizard_move_lines(self, quants):
 
     @api.onchange("origin_location_id")
     def _onchange_origin_location_id(self):
+        if self.origin_location_id:
```
rousseldenis: Why not making the picking type computed in order to get the default value in the same function (and as onchanges tend to be less used) ?

stock_move_location/wizard/stock_move_location.py
```diff
@@ -54,7 +40,9 @@ def _get_default_picking_type_id(self):
         column2="move_location_line_wiz_id",
     )
     picking_type_id = fields.Many2one(
-        comodel_name="stock.picking.type", default=_get_default_picking_type_id
+        compute="_compute_picking_type_id",
```
rousseldenis, line 44: @AungKoKoLin1997 The best is to use store=True too but maybe this is sufficent

AungKoKoLin1997: @rousseldenis I think we don't need store=True in here because it is in transient model and this attribute will do nothing.

rousseldenis: Transient are stored in DB

AungKoKoLin1997: @rousseldenis Yes. You are right. Thanks for the pointer. But I will not add this attribute in here because picking_type_id is not used in any domain or search.

rousseldenis:
I'm not sure about the flow.

@jbaudoux Any advice here ?

stock_move_location/wizard/stock_move_location.py
```diff
@@ -75,6 +63,35 @@ def _compute_readonly_locations(self):
...
+        for rec in self:
+            rec.picking_type_id = (
+                self.env["stock.picking.type"]
+                .search(
+                    domain,
+                    limit=1,
+                )
+                .id
+            )
+            if rec.origin_location_id:
```
jbaudoux: You can improve the code to not always do the first search
```
picking_type = False
if rec.origin_location_id:
  picking_type = search(...)
if not picking_type:
  picking_type = search(...)
```

What about taking a picking type that has as default_location_src_id a parent of origin_location_id ?

jbaudoux (approved):

stock_move_location/wizard/stock_move_location.py
```diff
@@ -75,6 +63,32 @@ def _compute_readonly_locations(self):
                 rec.origin_location_disable = True
                 rec.destination_location_disable = True
 
+    @api.depends("origin_location_id")
+    def _compute_picking_type_id(self):
+        company_id = self.env.context.get("company_id") or self.env.user.company_id.id
```
rousseldenis, line 70: @AungKoKoLin1997 If you depends on company context key, add @api.depends_context("company")

stock_move_location/wizard/stock_move_location.py
```diff
@@ -75,6 +63,32 @@ def _compute_readonly_locations(self):
...
+                        domain, limit=1
+                    )
+                    # Move up to the parent location if no picking type found
+                    location_id = not picking_type and location_id.location_id or False
+            if not picking_type:
+                domain = [
+                    ("code", "=", "internal"),
+                    ("warehouse_id.company_id", "=", company_id),
+                ]
+                picking_type = self.env["stock.picking.type"].search(domain, limit=1)
```
yostashiro: Non blocking. Just to make it a bit shorter.
```suggestion
            picking_type = self.env["stock.picking.type"]
            base_domain = [
                ("code", "=", "internal"),
                ("warehouse_id.company_id", "=", company_id),
            ]
            if rec.origin_location_id:
                location_id = rec.origin_location_id
                while location_id and not picking_type:
                    domain = [("default_location_src_id", "=", location_id.id)]
                    domain = expression.AND([base_domain, domain])
                    picking_type = picking_type.search(domain, limit=1)
                    # Move up to the parent location if no picking type found
                    location_id = not picking_type and location_id.location_id or False
            if not picking_type:
                picking_type = picking_type.search(base_domain, limit=1)
```

rousseldenis (approved):
Code review

stock_move_location/wizard/stock_move_location.py
AungKoKoLin1997: We cant use this `picking_type = self.env["stock.picking.type"]` because if the first `if` fails to assign picking_type, the search of second `if` will raise error.

yostashiro (approved):
Code review + functional test in runboat.

AungKoKoLin1997:
@rousseldenis Can we get this PR merged?

kanda999 (approved):
Functioanl review:  It worked as expected.

PR #1984, opened 2024-04-08
stock_location_product_restriction: correct location creation from UI

blaskurain:
When creating a new location using Odoo's web interface any call to _compute_restriction_violation was raising an exception running SQL because self.ids was an empty list. Ensure that the list provided to the SQL sentence is never empty by injecting an otherwise transparent value.


rousseldenis:
@blaskurain There is already https://github.com/OCA/stock-logistics-warehouse/pull/1803/files

Please review that one instead.

PR #1988, opened 2024-04-09
[IMP] stock_move_packaging_qty: report done packaging qtys

yajo:
The title sounds like a simple task, but it requires many changes:

- Store done pkg qtys in `stock.move.line` records.
- The qtys are not computed. Instead, the user must write them by hand. This happens because many times the packaging just indicates an approximate content, but is not always exact until reaching this point of the process, where both the done qtys (both UoM and packagings) are measured and get real values.
- Added compute and inverse on `stock.move` for simple scenarios where one move = one move line.
- Add all this on all reports and views.

This change also opens the door to be able to get a real inventory measured in product packagings. That's not implemented in this MVP, but the idea is recorded.

@moduon MT-5531


stock_move_packaging_qty/tests/test_stock_move_packaging_qty.py
```diff
@@ -59,3 +66,19 @@ def test_product_uom_qty_change(self):
             move_f.product_packaging_id = self.packaging
             self.assertEqual(move_f.product_uom_qty, 5)
             self.assertEqual(move_f.product_packaging_qty, 1)
+        picking = picking_f.save()
+        self.assertEqual(picking.state, "draft")
+        picking.action_set_quantities_to_reservation()
```
Shide, line 72: Assign picking before this line

```suggestion
        picking.action_assign()
        picking.action_set_quantities_to_reservation()
```

Shide (changes requested):

stock_move_packaging_qty/__manifest__.py
```diff
@@ -8,9 +8,11 @@
     "website": "https://github.com/OCA/stock-logistics-warehouse",
     "category": "Warehouse",
     "depends": ["stock"],
+    "development_status": "Alpha",
```
rousseldenis: Be careful as if there are some depending modules on this, they are becoming Alpha too (and possible errors in pre-commit).

stock_move_packaging_qty/models/stock_move.py
```diff
@@ -31,6 +37,18 @@ def _compute_product_packaging_qty(self):
...
 
+    @api.depends(
+        "move_line_ids.product_packaging_qty_done",
+        "move_line_nosuggest_ids.product_packaging_qty_done",
+    )
+    def _compute_product_packaging_qty_done(self):
+        """Get the sum of done packaging qtys from move lines."""
+        for move in self:
+            lines = move._get_move_lines()
+            move.product_packaging_qty_done = sum(
```
rousseldenis, line 49: IMHO a read_group with aggregate function is better.

stock_move_packaging_qty/models/stock_move.py
```diff
@@ -62,3 +97,26 @@ def _get_single_package_uom_qty(self):
         return self.product_packaging_id.product_uom_id._compute_quantity(
             self.product_packaging_id.qty, self.product_uom
         )
+
+    def _set_quantities_to_reservation(self):
+        """Add packaging qtys when clicking on "Set Quantities"."""
+        result = super()._set_quantities_to_reservation()
+        for line in self.move_line_ids:
+            if line.qty_done != line.reserved_uom_qty:
```
rousseldenis: Please use float_compare

rousseldenis (changes requested):

rousseldenis:
@yajo Could you check tests also ?

stock_move_packaging_qty/__manifest__.py
yajo: AFAICS there are no others depending on this one. However I'll drop this change, as I don't know if it's in a different status for the original authors.

stock_move_packaging_qty/models/stock_move.py
yajo: That would be a performance improvement if I had a domain to search for lines, because then I could just group all the logic in a single query.

In the case at hand, I need to loop over moves to call `_get_move_lines()` once per move. At this point, all data should be already prefetched by the ORM.

The computing done by python is very light (just a sum). Probably lighter than forcing one extra query per loop. So I don't think that would improve performance here. Actually, it could make it worse.

In any case, if we see later that there's an impact, it'll be easier to fine-tune after getting real performance graphs.

stock_move_packaging_qty/tests/test_stock_move_packaging_qty.py
```diff
@@ -59,3 +66,20 @@ def test_product_uom_qty_change(self):
             move_f.product_packaging_id = self.packaging
             self.assertEqual(move_f.product_uom_qty, 5)
             self.assertEqual(move_f.product_packaging_qty, 1)
+        picking = picking_f.save()
+        self.assertEqual(picking.state, "draft")
+        picking.action_assign()
+        picking.action_set_quantities_to_reservation()
```
Shide, line 72: No need to test if `action_set_quantities_to_reservation()` is done correctly. Uses another fields that are not established in picking/move and the default values maybe differ from you expect.

```suggestion
        picking.action_set_quantities_to_reservation()
        picking.move_line_ids[0].qty_done = picking.move_line_ids[0].reserved_uom_qty
```

Shide (changes requested):

yajo (commented):
All attended, thanks. Could you review again?

Shide (approved):

Gelojr (approved):
LGTM Nice Job @yajo 
https://www.loom.com/share/179c2eef8bc842d2b9b9b9a14b49b84e?sid=53ed3b2c-2dca-4302-bea9-b679d5168669

rafaelbn (approved):
👍🏼 Great improvement!

rafaelbn:
@rousseldenis @LoisRForgeFlow can we merge? 😄 ❤️

rousseldenis (approved):

PR #1989, opened 2024-04-11
[15.0][IMP] stock_request: Change rounding method to HALF-UP in quantities computation

JoanSForgeFlow:
Switched the rounding method from "UP" (if not indicated default is "UP")  to "HALF-UP" in the `qty_in_progress` computation to ensure more accurate roundings.

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -162,9 +162,15 @@ class StockRequest(models.Model):
             done_qty = abs(other_qty - incoming_qty)
             open_qty = sum(request.allocation_ids.mapped("open_product_qty"))
             uom = request.product_id.uom_id
-            request.qty_done = uom._compute_quantity(done_qty, request.product_uom_id)
+            request.qty_done = uom._compute_quantity(
+                done_qty,
+                request.product_uom_id,
+                rounding_method="HALF-UP",
+            )
             request.qty_in_progress = uom._compute_quantity(
-                open_qty, request.product_uom_id
+                open_qty,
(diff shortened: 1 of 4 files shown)


AaronHForgeFlow:
Maybe the same in request.qty_done and request.qty_cancelled ?

JoanSForgeFlow:
You are right @AaronHForgeFlow , I've just applied what you suggest

AaronHForgeFlow (approved):
LGTM

JordiMForgeFlow (approved):
code review LGTM

MiquelRForgeFlow (approved):

LoisRForgeFlow (changes requested):
There are more usage in other files (examples: https://github.com/OCA/stock-logistics-warehouse/blob/01364900f3280a5020ea7dc6e23c85b7bd57208d/stock_request/models/stock_request_allocation.py#L71 or https://github.com/OCA/stock-logistics-warehouse/blob/01364900f3280a5020ea7dc6e23c85b7bd57208d/stock_request/models/stock_request_abstract.py#L34), any reason to change some usages and others not?

One more request, I think it would be great to depict the issue with a test case if possible.

JoanSForgeFlow:
@LoisRForgeFlow comments attended

JoanSForgeFlow:
The error in the test are caused by another module. I have done a fix: https://github.com/OCA/stock-logistics-warehouse/pull/2026

LoisRForgeFlow (approved):
Let's move this one forward 

/ocabot merge patch

PR #1991, opened 2024-04-11
[15.0][FIX/IMP] stock_inventory: new small features and ux improvements

DavidJForgeFlow:
Allow to create stock.quant making the quants smart button not invisible when quantity is equal to 0. (Removing one of the invisible conditions).

Include exclude sublocations flag in stock inventory to be able to only do an adjustment to the parent location. 

Adds a global flag that automatically sets stock.inventory to done when all quants are done. (Cases with barcode will benefit from this.)

Add company to adjustments to avoid multi-company errors. 

Adds more information fields to tree view. 

Also adds 'Assign to' in stock inventory that propagates to quants. Also propagates the date field.

@ForgeFlow

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -8,9 +8,28 @@ class InventoryAdjustmentsGroup(models.Model):
     _description = "Inventory Adjustment Group"
     _order = "date desc, id desc"
 
-    name = fields.Char(required=True, default="Inventory", string="Inventory Reference")
+    name = fields.Char(
+        required=True,
+        default="Inventory",
+        string="Inventory Reference",
+        states={"draft": [("readonly", False)]},
+        readonly=True,
+    )
 
-    date = fields.Datetime(default=lambda self: fields.Datetime.now())
+    date = fields.Datetime(
+        default=lambda self: fields.Datetime.now(),
+        states={"draft": [("readonly", False)]},
+        readonly=True,
+    )
+
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        readonly=True,
+        index=True,
+        states={"draft": [("readonly", False)]},
+        default=lambda self: self.env.company,
+        required=True,
+    )
 
     state = fields.Selection(
         [("draft", "Draft"), ("in_progress", "In Progress"), ("done", "Done")],
@@ -22,7 +41,10 @@ class InventoryAdjustmentsGroup(models.Model):
     )
 
     location_ids = fields.Many2many(
-        "stock.location", string="Locations", domain="[('usage', '=', 'internal')]"
+        "stock.location",
+        string="Locations",
+        domain="[('usage', '=', 'internal'), "
+        "'|', ('company_id', '=', company_id), ('company_id', '=', False)]",
     )
 
     product_selection = fields.Selection(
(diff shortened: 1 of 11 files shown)


stock_inventory/models/stock_inventory.py
```diff
@@ -99,11 +124,17 @@ def _get_quants(self, locations):
...
-        ]
+        return (
+            [
+                ("location_id", "in", locations.mapped("id")),
+            ]
+            if self.exclude_sublocation
+            else [
+                "|",
+                ("location_id", "in", locations.mapped("id")),
+                ("location_id", "child_of", locations.child_ids.ids),
```
LoisRForgeFlow: child_of will also include the parent location, so `("location_id", "child_of", locations.child_ids.ids)` will sufice in this case

stock_inventory/models/stock_inventory.py
```diff
@@ -146,13 +177,17 @@ def _get_domain_category_quants(self, base_domain):
...
+
     def action_state_to_in_progress(self):
         active_rec = self.env["stock.inventory"].search(
             [
                 ("state", "=", "in_progress"),
                 "|",
-                ("location_ids", "in", self.location_ids.mapped("id")),
-                ("location_ids", "in", self.location_ids.child_ids.ids),
+                ("location_ids", "in", self.location_ids.ids),
+                ("location_ids", "child_of", self.location_ids.ids),
```
LoisRForgeFlow, line 202: same here ("location_id", "child_of", locations.child_ids.ids)

stock_inventory/models/stock_inventory.py
```diff
@@ -37,15 +47,22 @@ class InventoryAdjustmentsGroup(models.Model):
         required=True,
     )
 
-    product_ids = fields.Many2many("product.product", string="Products")
+    product_ids = fields.Many2many(
+        "product.product", string="Products", domain="[('company_id', '=', company_id)]"
```
LoisRForgeFlow: consider also products without company, as they are global to all companies.

stock_inventory/models/stock_inventory.py
```diff
@@ -12,6 +12,14 @@ class InventoryAdjustmentsGroup(models.Model):
 
     date = fields.Datetime(default=lambda self: fields.Datetime.now())
 
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        readonly=True,
+        index=True,
+        states={"draft": [("readonly", False)]},
+        default=lambda self: self.env.company,
```
LoisRForgeFlow, line 30: Either make this required (not sure if some process need this to allow blank company) or consider that company here can be false in the domains applied below to quants, products, locations, etc.

LoisRForgeFlow (changes requested):
Overall looks great, some comments below.

Also, I would like to add one more request. The reference of the inventory adjustment (group) should be passed to the stock move line generated from the count of related quants.

image

I think you can concat the reference with the standard meesages, e.g.:

- "This is my cool reference: Product Quantity Updated"
- "This is my cool reference: Product Quantity Confirmed"

Also, consider that the count could have been made from a barcode scanner using the barcode app, in such case the reference should also be kept.

DavidJForgeFlow:
@LoisRForgeFlow The reference now also contains the name of the Adjustment Group. For the changes of child_ids they're all done! And lastly the multi-company ones are done except for making the company_id required. I think it isn't a field mandatory here and I prefer to don't make it now required to don't affect the previous behavior.

LoisRForgeFlow (approved):
All good now!

JoanSForgeFlow (approved):
Code and functional review, works as expected! :+1:

LoisRForgeFlow:
@DavidJForgeFlow could you fwport to v16?

PR #1995, opened 2024-04-16
[15.0][MIG] stock_cycle_count

ArnauCForgeFlow:
Migration to 15.0

PR #1997, opened 2024-04-17
[15.0][IMP]stock_cycle_count: several improvements

ArnauCForgeFlow:
Depends on: [https://github.com/OCA/stock-logistics-warehouse/pull/1995](https://github.com/OCA/stock-logistics-warehouse/pull/1995)

This PR include the following improvements, se commits for more details.

### DIFF
--- a/stock_cycle_count/models/stock_cycle_count.py
+++ b/stock_cycle_count/models/stock_cycle_count.py
@@ -117,19 +117,10 @@ class StockCycleCount(models.Model):
             data = rec._prepare_inventory_adjustment()
             inv = self.env["stock.inventory"].create(data)
             if rec.company_id.auto_start_inventory_from_cycle_count:
-                inv.prefill_counted_quantity = (
-                    rec.company_id.inventory_adjustment_counted_quantities
-                )
                 try:
                     inv.action_state_to_in_progress()
                 except Exception as e:
                     _logger.info("Error when beginning an adjustment: %s", str(e))
-
-                if inv.prefill_counted_quantity == "zero":
-                    inv.stock_quant_ids.write({"inventory_quantity": 0})
-                else:
-                    for quant in inv.stock_quant_ids:
-                        quant.write({"inventory_quantity": quant.quantity})
         self.write({"state": "open"})
         return True
 
--- a/stock_cycle_count/models/stock_cycle_count.py
+++ b/stock_cycle_count/models/stock_cycle_count.py
@@ -84,17 +84,6 @@ class StockCycleCount(models.Model):
         readonly=True,
     )
 
-    def write(self, vals):
-        result = super().write(vals)
-        if "responsible_id" in vals and not self.env.context.get("no_propagate"):
-            stock_inventory_records = self.mapped("stock_adjustment_ids")
-            for record in stock_inventory_records:
-                if record.responsible_id.id != vals["responsible_id"]:
-                    record.with_context(no_propagate=True).write(
-                        {"responsible_id": vals["responsible_id"]}
-                    )
-        return result
-
     @api.depends("stock_adjustment_ids")
     def _compute_inventory_adj_count(self):
         for rec in self:
--- a/stock_cycle_count/models/stock_cycle_count.py
+++ b/stock_cycle_count/models/stock_cycle_count.py
@@ -1,9 +1,12 @@
 # Copyright 2017-18 ForgeFlow S.L.
 #   (http://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import logging
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
 
 
 class StockCycleCount(models.Model):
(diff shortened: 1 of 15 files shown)


stock_cycle_count/models/stock_inventory.py
```diff
@@ -76,6 +62,27 @@ def _domain_cycle_count_candidate(self):
             ("location_id", "in", self.location_ids.ids),
         ]
 
+    def _calculate_inventory_accuracy(self):
+        for inv in self:
+            sml = self.env["stock.move.line"].search(
```
rousseldenis: Mmmh, IMHO this has to be refined has multiple searches will be done...

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -114,7 +117,11 @@ def action_create_inventory_adjustment(self):
                 inv.prefill_counted_quantity = (
                     self.company_id.inventory_adjustment_counted_quantities
                 )
-                inv.action_state_to_in_progress()
+                try:
+                    inv.action_state_to_in_progress()
+                except Exception as e:
```
rousseldenis, line 122: Isn't it a too broad Exception ? 

Moreover, shouldn't we manage the exception in a more elegant manner than 'just' logging ?

ArnauCForgeFlow: Hey @rousseldenis 

You might be right. I implemented it this way because we don't want the cron job to stop running due to a validation error from a location that already has an inventory adjustment in progress.

Do you have any suggestions for a better approach?

Thanks.

LoisRForgeFlow (approved):
Code review and functional test :+1: 

I think this is a good first batch on improvements to merge

JoanSForgeFlow (approved):
Functional and code review!

stock_cycle_count/models/stock_cycle_count.py
```diff
@@ -66,6 +84,17 @@ class StockCycleCount(models.Model):
         readonly=True,
     )
 
+    def write(self, vals):
+        result = super().write(vals)
+        if "responsible_id" in vals and not self.env.context.get("no_propagate"):
+            stock_inventory_records = self.mapped("stock_adjustment_ids")
+            for record in stock_inventory_records:
+                if record.responsible_id.id != vals["responsible_id"]:
+                    record.with_context(no_propagate=True).write(
```
rousseldenis: I don't really like such code. Don't you have another mean to avoid this ?

rousseldenis: Why not modifying stock_inventory to change the responsible_id fields to computed field (stored/readonly=False) and override the compute here ?

rousseldenis: @ArnauCForgeFlow

JoanSForgeFlow: @rousseldenis, we agree with your proposal and have applied it.
Thank you

LoisRForgeFlow:
@rousseldenis friendly reminder here, can you update your last review? @JoanSForgeFlow has attended your request

stock_cycle_count/models/stock_quant.py
```diff
@@ -0,0 +1,47 @@
...
+                [
+                    ("product_id", "=", rec.product_id.id),
+                    ("lot_id", "=", rec.lot_id.id),
+                    "|",
+                    ("location_id", "=", rec.location_id.id),
+                    ("location_dest_id", "=", rec.location_id.id),
+                ],
+                order="create_date asc",
+            ).filtered(
+                lambda x: not x.company_id.id
```
rousseldenis: Doesn't it correspond to normal record rule ?

rousseldenis: In other words, is the `filtered` necessary ?

stock_cycle_count/models/stock_move_line.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 ForgeFlow S.L.
+#   (http://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    line_accuracy = fields.Float(
```
rousseldenis, line 10: I'm wondering if accuracy represent the accuracy on all related move lines or only for inventory related move lines ?

@ArnauCForgeFlow

ArnauCForgeFlow: Hey @rousseldenis  this line_accuracy is only for inventory related move lines

stock_cycle_count/models/stock_quant.py
ArnauCForgeFlow: @rousseldenis You are correct, since we are searching for stock move lines by location, and the locations already belong to a company, it is not needed this filter. I'll make the change.

Thanks

JoanSForgeFlow (approved):

stock_cycle_count/models/stock_quant.py
LoisRForgeFlow: @rousseldenis @ArnauCForgeFlow  it is not the normal record rule, it is filtering based on the quant company, the normal record rules filters based on the users companies, we cannot rely on the user having the proper companies.

That been said, I think instead of filtering, it can be done with a domain that includes a couple of company_id clauses.

ArnauCForgeFlow:
Hey @rousseldenis just a friendly reminder, could you please update your review?

LoisRForgeFlow:
Let's move forward

/ocabot merge major

LoisRForgeFlow:
@ArnauCForgeFlow can you update the 16.0 fwport (https://github.com/OCA/stock-logistics-warehouse/pull/2169) to be fully aligned with this one?

PR #2002, opened 2024-04-19
[16.0][FIX] stock_move_packaging_qty: Replace ml with move_line

EmilioPascual:
Fix: Replace ml with move_line. In original report odoo use move_line.

@Shide @yajo please review.

@moduon MT-5825


yajo (approved):
Obvious fix, thanks!

/ocabot merge patch

PR #2008, opened 2024-04-24
[14.0][ADD] stock_available_stock_mts_mto_rule

TDu:
This is a glue module between `stock_mts_mto_rule` and `stock_avaiable`. It will use the `immediately_usable_qty` field to compute the split for the mts mto rule, instead of the `virtual_available` value.


mt-software-de (approved):
Thanks. 
LGTM

jbaudoux (approved):

jbaudoux:
@rousseldenis Shouldn't this module be named `stock_mts_mto_rule_stock_available` as it is extending `stock_mts_mto_rule` with `stock_available` concept? Then it can remains in this repo and not move to stock-logistics-availability repo which I find weird...

stock_available_stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import api, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    @api.model
+    def _get_available_quantity_mts_mto_rule(self, product):
+        product = product.with_context(location=False)
```
mt-software-de: Sorry this is not correct. We should not remove the location from the context.
Otherwise all of the stock locations are used, to compute the qty. 
```suggestion
```

mt-software-de (changes requested):

mt-software-de (approved):

jbaudoux:
@TDu Can you squash fixup commit? Thanks

TDu:
@jbaudoux  Commit have been squashed...this should be ready for another check and merge maybe ?

PR #2010, opened 2024-04-24
[15.0][FIX] stock_inventory: only set to done adjustment if there is an adjustment

JoanSForgeFlow:
[FIX] stock_inventory: only set to done adjustment if there is an adjustment

to avoid this error

```
Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5238, in ensure_one
    _id, = self._ids
ValueError: not enough values to unpack (expected 1, got 0)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
    result = request.dispatch()
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 698, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 368, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 357, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 921, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 546, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1324, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1316, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 471, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 456, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/auto/addons/stock_barcode/models/stock_quant.py", line 89, in action_validate
    res = quants.action_apply_inventory()
  File "/opt/odoo/auto/addons/stock/models/stock_quant.py", line 366, in action_apply_inventory
    self._apply_inventory()
[...]

PR #2017, opened 2024-05-02
[16.0][IMP] stock_inventory: various fixes

DavidJForgeFlow:
Forward port of #1991 #2010.

Also adds some improvements from #1935 and supersede it.

From the last PR some changes are colliding with the ones applied in 15.0. I've merged the first two commits (be31060d1caa82f4023bcdd21a4290b376211855 and d4f380803802278aec47b8404084a287b49f9b3d) and the last two were removed due the conflict with the 15.0 changes. 

The changes removed are the ones affecting to stock.quant, as the versions have changed a lot and some options (like creating quants that did not exist earlier) are now covered. Please take a look and if you miss any change I beg you for creating a new PR adding them!

@benwillig @LoisRForgeFlow

### DIFF
--- a/stock_inventory/models/stock_quant.py
+++ b/stock_inventory/models/stock_quant.py
@@ -59,9 +59,9 @@ class StockQuant(models.Model):
     def _get_inventory_fields_write(self):
         return super()._get_inventory_fields_write() + ["to_do"]
 
-    @api.model
-    def create(self, vals):
-        res = super().create(vals)
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
         if self.env.context.get(
             "active_model", False
         ) == "stock.inventory" and self.env.context.get("active_id", False):
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -7,6 +7,9 @@ class InventoryAdjustmentsGroup(models.Model):
     _name = "stock.inventory"
     _description = "Inventory Adjustment Group"
     _order = "date desc, id desc"
+    _inherit = [
+        "mail.thread",
+    ]
 
     name = fields.Char(
         required=True,
@@ -34,6 +37,7 @@ class InventoryAdjustmentsGroup(models.Model):
     state = fields.Selection(
         [("draft", "Draft"), ("in_progress", "In Progress"), ("done", "Done")],
         default="draft",
+        tracking=True,
     )
 
     owner_id = fields.Many2one(
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -1,6 +1,10 @@
 from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo.exceptions import UserError, ValidationError
 from odoo.osv import expression
+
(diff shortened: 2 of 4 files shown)


LoisRForgeFlow (changes requested):
I cannot chosee products in manual selection now:

image

LoisRForgeFlow:
> I cannot chosee products in manual selection now:
> 
> image

We found that the problem is generated by this other module https://github.com/OCA/stock-logistics-warehouse/blob/16.0/stock_inventory_preparation_filter/views/stock_inventory_view.xml. We'll fix after merging this one.

LoisRForgeFlow (approved):
:+1:

stock_inventory/models/stock_quant.py
```diff
@@ -36,9 +39,32 @@ def _apply_inventory(self):
...
+    @api.model
+    def create(self, vals):
+        res = super().create(vals)
+        if self.env.context.get(
+            "active_model", False
+        ) == "stock.inventory" and self.env.context.get("active_id", False):
+            self.env["stock.inventory"].browse(
+                self.env.context.get("active_id")
+            ).refresh_stock_quant_ids()
+        return res
```
benwillig: Can't we simply add context default values and add related fields in the tree view? Like I did in the previous PR. Also you should use model_create_multi.

DavidJForgeFlow: Hi, I've changed the model_create_multi, however I don't understand exactly what you refer by context default values. You mean creating the quant with default values automatically? 

If you agree, we could merge this PR with the alignment changes and you can create a new PR with your changes added correctly. If not please tell me exaclty what do you mean please. Thanks!

rousseldenis: @DavidJForgeFlow I think this is what it is about:

image

DavidJForgeFlow: I've added the default_to_do as stock_inventory_ids is no longer in quants. I think we need to review if this fields is still needed, IMO it's not longer needed.

LoisRForgeFlow: do you agree @benwillig ?

benwillig: I'm not a fan of this implementation. I wasn't a fan of what I did based on your previous algorithm but it was cleaner in my opinion. I should take time to analyze this and also include the rest of my PR, because here the merge is incomplete so I have things to to that was not planned.

LoisRForgeFlow: @DavidJForgeFlow Based on this response I think it is best to split this PR, between the strict fwport from v15 and the adjustments done by @benwillig in which we could continue this discussion.

I would like to have v15 and v16 aligned to unlock the migration of v16 cycle count.

LoisRForgeFlow:
@DavidJForgeFlow https://github.com/OCA/stock-logistics-warehouse/pull/2029 is merged, can you rebase this one?

benwillig (approved):

JoanSForgeFlow (approved):
Code and functional review. LGTM!

LoisRForgeFlow:
Rebased after issues with tests got fixed in https://github.com/OCA/stock-logistics-warehouse/pull/2087

PR #2018, opened 2024-05-02
[14.0][FIX] stock_orderpoint_origin: fix access error

HekkiMelody:
This commit fixes an AccessError that would happen because the report stock.report_product_product_replenishment references several objects (SOs, Pickings, etc) that the user may not have access to.

For example, the user may only have access to SOs that are assigned to them.


rousseldenis:
@aleuffre IMHO, this is a workaround as this raises a question: if people do not have access to such data, why showing them?

HekkiMelody:
> @aleuffre IMHO, this is a workaround as this raises a question: if people do not have access to such data, why showing them?

The module doesn't show the user the report, but it does use data that's fetched by reusing methods defined in the report to link procurement POs with SOs

PR #2021, opened 2024-05-06
[15.0][IMP] stock_inventory: be able to do adjustments in same location but for different products

JoanSForgeFlow:
Current Restriction is too much restrictive

The main idea is to be able to do adjustments in same location but for different products, also be able to do an inventory adjustment in a sublocation if there is an adjustment on the parent location but doing it "excluding sublocations"


AaronHForgeFlow (approved):
Functional review. It works as expected.

LoisRForgeFlow (approved):
Tested, works as expected :+1:

PR #2025, opened 2024-05-07
[14.0][ADD] stock_serial_no_auto_reserve

thomaspaulb:


stock_serial_no_auto_reserve/models/stock_move.py
```diff
@@ -0,0 +1,24 @@
...
+        will have to use the tag icon from stock_quant_manual_assign"""
+        self_not_serial = self.filtered(
+            lambda m: m.product_id.tracking != "serial"
+            or m.product_id.serial_auto_reserve
+        )
+        if self_not_serial:
+            result = super(
+                StockMove, self_not_serial.with_context(from_assign=True)
+            )._action_assign()
+            move.picking_id._compute_is_available()
```
thomaspaulb: What I meant was:

- `self_not_serial` is all the moves **except** those that have products that are tracked by serial number. On these, we call super.
- We then also calculate `self_with_serial` that has all the other moves, eg the serially tracked ones. These are the ones that skip reservation. On those, instead of reservation, we calculate an is_available boolean. This boolean should be a regular field defined on `stock.move`.
- We then define `is_available` as a stored computed field on `stock.picking`, and let it depend on move_ids_without_package.state as well as move_ids_without_package.is_available. The computation will then be along the lines of "loop all the moves, check if state is assigned, if not, check if state is waiting and is_available = true. If one of these conditions is true for all move lines, then is_available for the picking should turn out True. Otherwise, False.

What you're doing here is a different approach which might in theory work, but in its current form I'm seeing some possible flaws:

- The call here to the compute function is only called when `self_not_serial` has values, so it will not be called for those pickings that have serially-tracked products only.
- I don't see `move` being defined anywhere as a variable

ddejong-therp: Yeah about `move`, I had simply forgotten to push my latest commit, but I have now.

Only computing is_available when tracking is not serial or auto-reservation is enabled, was a mistake indeed.
I'll change it.  I was wondering for a moment whether it would be proper to only calculate `is_available` when it is not being assigned, but I think that will be proper indeed, because when it is being assigned it can just be set to `False` right?

I was thinking about doing the same thing yesterday on stock.move: adding an `is_available` field there too, and have stock.picking depend on those. It's just that I took the (in my mind) shorter route because of the time pressure I felt.

thomaspaulb: I would say whenever action_assign is called on any move that belongs to a certain picking, we always need to recalculate is_available boolean of that picking, because it may have changed. Depending on the calculation we make it True or False, but we always decide something.

stock_serial_no_auto_reserve/models/stock_move.py
```diff
@@ -0,0 +1,42 @@
...
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    is_available = fields.Boolean(
+        "Is available but not assigned", compute="_compute_is_available", store=True
+    )
```
thomaspaulb: So I think the reason why the installation of the module is slow, and why you needed the init script, is because of this computed field.

Some points to note:

- It doesn't make sense to have stored computed fields in Odoo at all, if you don't specify an `api.depends` clause for them. What Odoo will do in the background for such fields is, for each of the `api.depends` members, it creates a trigger on `write`, and it calls the computation function when triggered. Currently, you don't have any such triggers defined, so the computation will never trigger. The only two cases where this field is being written to is 1. when you write to it manually, which you're doing in `_action_assign` below, and when you first install the module and the field is created.

- My point then being, why does this need to be a stored computed field at all? Just execute the code that does the computation on `action_assign`, that's the only trigger we need. You can then also dispense with the init script, because no computation will run anymore upon install (only the one for pickings, but that one should be fast enough)

stock_serial_no_auto_reserve/models/stock_move.py
```diff
@@ -0,0 +1,42 @@
...
+        manual stock.move.line creation this is not possible, people
+        will have to use the tag icon from stock_quant_manual_assign"""
+        self_auto_reserve = self.filtered(
+            lambda m: m.product_id.tracking != "serial"
+            or m.product_id.serial_auto_reserve
+        )
+        self_no_auto_reserve = self.filtered(
+            lambda m: m.product_id.tracking == "serial"
+            and not m.product_id.serial_auto_reserve
+        )
```
thomaspaulb: Instead of defining two separate searches, you can also just do `self_no_auto_reserve = self_auto_reserve - self`. Shorter and less prone to bugs

stock_serial_no_auto_reserve/models/stock_move.py
```diff
@@ -0,0 +1,42 @@
...
+        if self_no_auto_reserve:
+            self_no_auto_reserve._compute_is_available()
+        if self_auto_reserve:
+            result = super(
+                StockMove, self_auto_reserve.with_context(from_assign=True)
```
thomaspaulb: This context key was added at some point and had a use, but the use is gone now, so we can remove

thomaspaulb:
This was not a good idea. Not auto-reserving means you don't have any idea whether your shipments are "Ready" or not - if 4 shipments are waiting for the same item, and one is available, which one can be sent out? None of them will be listed as "Ready". Cancelling this attempt, it's better to focus on how to make it insightful for users how to unreserve serials again, also from other pickings.

PR #2029, opened 2024-05-13
[16.0][IMP] stock_inventory: various fixes

DavidJForgeFlow:
Forward port of https://github.com/OCA/stock-logistics-warehouse/pull/1991 https://github.com/OCA/stock-logistics-warehouse/pull/2010. 

Supersedes #2017 

@ForgeFlow

### DIFF
--- a/stock_inventory/views/stock_inventory.xml
+++ b/stock_inventory/views/stock_inventory.xml
@@ -41,7 +41,7 @@
                             name="action_view_inventory_adjustment"
                             class="oe_stat_button"
                             icon="fa-pencil-square-o"
-                            attrs="{'invisible':['|', ('state', 'in', ['draft', 'done']), ('count_stock_quants', '=', 0)]}"
+                            attrs="{'invisible':[('state', 'in', ['draft', 'done'])]}"
                         >
                             <field
                                 name="count_stock_quants_string"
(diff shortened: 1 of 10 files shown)


LoisRForgeFlow (approved):
Ty

JoanSForgeFlow (approved):

LoisRForgeFlow:
I'm merging to align with v15.

/ocabot merge major

stock_inventory/views/stock_inventory.xml
```diff
@@ -150,5 +161,7 @@
         sequence="30"
         action="action_view_inventory_group_form"
     />
-    <delete model="ir.ui.menu" id="stock.menu_action_inventory_tree" />
+    <record id="stock.menu_action_inventory_tree" model="ir.ui.menu">
+        <field name="active" eval="False" />
+    </record>
```
Kev-Roche, line 166: If we already have the module installed, the view does not exist anymore, so we can not access to "stock.menu_action_inventory_tree" and an error occurs. 
@DavidJForgeFlow @LoisRForgeFlow

DavidJForgeFlow: Hi! You can just update the module "stock" and then install / update the "stock_inventory" module!

Kev-Roche: Of course, but depending of the odoo hosting infrastructure, it is not a suitable / clean solution. Returning to the previous line of code or a small migration script will work without effort. what do you think ?

FreezeS: "You can just update the module "stock" and then install / update the "stock_inventory" module!" - I confirm this works but it's not intuitive or documented anywhere except in this thread. A better solution needs to be found. Is it possible to force the upgrade of "stock" before upgrading this module?

PR #2030, opened 2024-05-14
[17.0][MIG] stock_change_qty_reason

JasminSForgeFlow:
Standard Migration

@ForgeFlow


LoisRForgeFlow (commented):
Functional review. Preset reasons work well, the free text reasons are not usable though. With preset reasons disabled, you cannot edit the reason field while doing the inv. adjustment.

Also, not sure if both reasons fields should be displayed when using preset reasons.

image

JasminSForgeFlow:
> Functional review. Preset reasons work well, the free text reasons are not usable though. With preset reasons disabled, you cannot edit the reason field while doing the inv. adjustment.
> 
> Also, not sure if both reasons fields should be displayed when using preset reasons.
> 
> image

The first reason is free text when we select any reason (2nd reason) in line while doing adjustment and click on apply, it will copy reason (2nd field) value to reason (1st field), and reason (2nd field) value set as False. reason(1st field) value is for historical purpose as reason (2nd filed) will be set False when adjustment done

LoisRForgeFlow:
> > Functional review. Preset reasons work well, the free text reasons are not usable though. With preset reasons disabled, you cannot edit the reason field while doing the inv. adjustment.
> > Also, not sure if both reasons fields should be displayed when using preset reasons.
> > image
> 
> The first reason is free text when we select any reason (2nd reason) in line while doing adjustment and click on apply, it will copy reason (2nd field) value to reason (1st field), and reason (2nd field) value set as False. reason(1st field) value is for historical purpose as reason (2nd filed) will be set False when adjustment done

@JasminSForgeFlow Sorry for the late reply.

If you want to keep it make it at least optional=hide, but in general it doesn't look very clean to me to keep both fields displayed. You should either go for preset reasons or not, if you starte with free text, you could do some custom migration or a custom module to show both reasons. In other words, we should think more on clean installations than on past sins or legacy issues.

Also, you haven't answered my second comment: the free text reason code is not editable even without pre-set reasons enabled.

JasminSForgeFlow:
> > > Functional review. Preset reasons work well, the free text reasons are not usable though. With preset reasons disabled, you cannot edit the reason field while doing the inv. adjustment.
> > > Also, not sure if both reasons fields should be displayed when using preset reasons.
> > > image
> > 
> > 
> > The first reason is free text when we select any reason (2nd reason) in line while doing adjustment and click on apply, it will copy reason (2nd field) value to reason (1st field), and reason (2nd field) value set as False. reason(1st field) value is for historical purpose as reason (2nd filed) will be set False when adjustment done
> 
> @JasminSForgeFlow Sorry for the late reply.
> 
> If you want to keep it make it at least optional=hide, but in general it doesn't look very clean to me to keep both fields displayed. You should either go for preset reasons or not, if you starte with free text, you could do some custom migration or a custom module to show both reasons. In other words, we should think more on clean installations than on past sins or legacy issues.
> 
> Also, you haven't answered my second comment: the free text reason code is not editable even without pre-set reasons enabled.

this one is fixed.

LoisRForgeFlow (commented):
I'm not able to write anything in the reason:


https://github.com/user-attachments/assets/2c342dfe-8569-454f-9681-b6e6f4fb136b

JasminSForgeFlow:
> I'm not able to write anything in the reason:
> 
>  2024-09-17.16-52-56.mp4

its fixed, you can edit now

LoisRForgeFlow (approved):
Functional test, all god now :+1: 

Thanks!

PR #2032, opened 2024-05-15
[14][FIX] stock_mts_mto_rule: Source location of initial rule

mt-software-de:
I just was checking the module `stock_mts_mto_rule` in a multi step delivery process, 2 or more delivery steps. 
I was wondering if this addon is capable of handling this. Because in a two step delivery process, the MTS Rule which is by default selected, is the one from `Stock -> Output` shouldn't it be `Output -> Customer` be?
This would than lead to the issue that the wrong source location is used. 

I have to extend the tests for it, to see if it is working or not.
In the meantime, is there anything using this addon in a two or more step delivery process? @OCA/logistics-maintainers @jbaudoux

### DIFF
--- a/stock_mts_mto_rule/models/stock_rule.py
+++ b/stock_mts_mto_rule/models/stock_rule.py
@@ -45,7 +45,7 @@ class StockRule(models.Model):
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
-        src_location_id = self.mts_rule_id.location_src_id.id
+        src_location_id = self.location_src_id.id
         product_location = product.with_context(location=src_location_id)
         virtual_available = self._get_available_quantity_mts_mto_rule(product_location)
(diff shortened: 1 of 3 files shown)


jbaudoux:
cc @florian-dacosta

jbaudoux (approved):
From your explanation this change looks good to me

mt-software-de:
@msteinfeld-sirum as you migrated it to 14.0 and 16.0 could you please have a look at this?
Are you using this addon still?

mt-software-de:
cc @i-vyshnevska

rousseldenis (approved):
Seems better indeed

rousseldenis:
@mt-software-de Could you maybe add a test case ?

mt-software-de:
> @mt-software-de Could you maybe add a test case ?

Yes thats the plan. 😁

jbaudoux (approved):

TDu (approved):

PR #2033, opened 2024-05-16
[15.0][FIX] stock_request: compute_move_ids

JoanSForgeFlow:
We have included the origin moves as well because when there is a route configuration with "Trigger another rule" rules, only the last transfer is linked.

This is confusing since the stock request actually generates a flow of 2 or 3 transfers that should be linked.

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -125,10 +125,21 @@ class StockRequest(models.Model):
         ("name_uniq", "unique(name, company_id)", "Stock Request name must be unique")
     ]
 
+    def _get_all_origin_moves(self, move):
+        all_moves = move
+        if move.move_orig_ids:
+            for orig_move in move.move_orig_ids:
+                all_moves |= self._get_all_origin_moves(orig_move)
+        return all_moves
+
     @api.depends("allocation_ids", "allocation_ids.stock_move_id")
     def _compute_move_ids(self):
         for request in self:
-            request.move_ids = request.allocation_ids.mapped("stock_move_id")
+            move_ids = request.allocation_ids.mapped("stock_move_id")
+            all_moves = self.env["stock.move"]
+            for move in move_ids:
+                all_moves |= self._get_all_origin_moves(move)
+            request.move_ids = all_moves
 
     @api.depends(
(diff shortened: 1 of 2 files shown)


stock_request/models/stock_request.py
```diff
@@ -128,7 +128,9 @@ def _get_expected_date():
     @api.depends("allocation_ids", "allocation_ids.stock_move_id")
     def _compute_move_ids(self):
         for request in self:
-            request.move_ids = request.allocation_ids.mapped("stock_move_id")
+            move_ids = request.allocation_ids.mapped("stock_move_id")
+            all_moves = move_ids | move_ids.mapped("move_orig_ids")
```
rousseldenis: Could you explain why including those ones ?

JoanSForgeFlow: Hi @rousseldenis,

We have included the origin moves as well because when there is a route configuration with "Trigger another rule" rules, only the last transfer is linked. 

This is confusing since the stock request actually generates a flow of 2 or 3 transfers that should be linked.

rousseldenis: Yes but in this case you assume there is only two step route

JoanSForgeFlow: You're absolutely right, I think the best approach is to make it recursive until the last move added has no move_orig_ids

LoisRForgeFlow (approved):
Fucntional review, it has some downsides (not all moves are explicitly linked to the stock request) but the alternative introduces a lot of complexity and a requirement for a refactor of allocations... so I think this is maybe simplistic but add UX value with a small change.

LGMT :+1:

JoanSForgeFlow:
The error in the test is fixed in this PR in stock_orderpoint_move_link: https://github.com/OCA/stock-logistics-warehouse/pull/2107

GuillemCForgeFlow (approved):
code review 👍🏿

LoisRForgeFlow:
Looks like the open review was attended and we have the approvals to move forward:

/ocabot merge patch

LoisRForgeFlow:
@JoanSForgeFlow another one to fwport whenever you can

PR #2035, opened 2024-05-17
[16.0][FIX] stock_quant_manual_assign

AungKoKoLin1997:
Supersede of https://github.com/OCA/stock-logistics-warehouse/pull/1792
This PR is to address the issue that is raised when users toggle the boolean widget in wizard.

@qrtl

### DIFF
--- a/stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
@@ -14,7 +14,11 @@
                         <field name="location_id" />
                         <field name="on_hand" force_save="1" />
                         <field name="reserved" force_save="1" />
-                        <field name="selected" widget="boolean_toggle" />
+                        <field
+                            name="selected"
+                            widget="boolean_toggle"
+                            options="{'autosave': False}"
+                        />
                         <field
                             name="qty"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
```diff
@@ -14,7 +14,11 @@
                         <field name="location_id" />
                         <field name="on_hand" force_save="1" />
                         <field name="reserved" force_save="1" />
```
yostashiro, line 16:
```suggestion
                        <field name="on_hand" />
                        <field name="reserved" />
```

I think we should remove `false_save="1"` to avoid future confusions.

rousseldenis (approved):

stock_quant_manual_assign/wizard/assign_manual_quants_view.xml
pedrobaeza: Why? That's a different thing

yostashiro: My bad, I misunderstood the history. No, no need to remove `false_save="1"`.

yostashiro (approved):

PR #2036, opened 2024-05-17
[14.0][FIX] stock_reserve: Set the correct _default_picking_type_id value

victoralmau:
Set the correct `_default_picking_type_id` value

Do not use a fixed value because it will not be compatible with multi-company.

@Tecnativa TT49168

### DIFF
--- a/stock_reserve/models/stock_reserve.py
+++ b/stock_reserve/models/stock_reserve.py
@@ -77,7 +77,17 @@ class StockReservation(models.Model):
         # If picking_type_id is present and location_id is not, try to find
         # default value for location_id
         if not res.get("picking_type_id", None):
-            res["picking_type_id"] = self._default_picking_type_id()
+            res["picking_type_id"] = (
+                self.env["stock.picking.type"]
+                .search(
+                    [
+                        ("code", "=", "outgoing"),
+                        ("company_id", "=", res.get("company_id")),
+                    ],
+                    limit=1,
+                )
+                .id
+            )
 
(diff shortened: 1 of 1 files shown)


stock_reserve/models/stock_reserve.py
```diff
@@ -107,8 +107,11 @@ def get_location_from_ref(self, ref):
 
     @api.model
     def _default_picking_type_id(self):
-        ref = "stock.picking_type_out"
-        return self.env.ref(ref, raise_if_not_found=False).id
+        return (
+            self.env["stock.picking.type"]
+            .search([("code", "=", "outgoing")], limit=1)
```
pedrobaeza: Limit to the reserve company, as there can be several allowed companies when doing this.

stock_reserve/models/stock_reserve.py
```diff
@@ -106,9 +108,9 @@ def get_location_from_ref(self, ref):
         return location_id
 
     @api.model
-    def _default_picking_type_id(self):
-        ref = "stock.picking_type_out"
-        return self.env.ref(ref, raise_if_not_found=False).id
+    def _default_picking_type_id(self, company_id):
```
pedrobaeza: This method may be executed directly by ORM, so arguments shouldn't change. Call `default_get` for `company_id` here or remove this extra method.

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2040, opened 2024-05-20
[16.0][ADD] stock_picking_location_check

AungKoKoLin1997:
This module intends to ensure the internal location consistency between the picking and its stock move lines at the time of validation.

@qrtl QT4063


yostashiro (approved):
Code review.

stock_picking_location_check/models/stock_picking.py
```diff
@@ -0,0 +1,47 @@
...
+                continue
+            line_source_locations = [
+                line.location_id
+                for line in pick.move_line_ids
+                if not getattr(line.move_id, "is_subcontract", False)
+            ]
+            self._check_location_consistency(pick.location_id, line_source_locations)
+            self._check_location_consistency(
+                pick.location_dest_id, pick.move_line_ids.location_dest_id
+            )
```
yostashiro:
```suggestion
            pick._check_location_consistency(pick.location_id, line_source_locations)
            pick._check_location_consistency(
                pick.location_dest_id, pick.move_line_ids.location_dest_id
            )
```

yostashiro (changes requested):

yostashiro (approved):

kanda999 (approved):
Functional review: LGTM

AungKoKoLin1997:
@oca/stock-logistics-warehouse-maintainers
Is is possible to take a look at this new module?

stock_picking_location_check/models/stock_picking.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _action_done(self):
+        for pick in self:
+            if pick.allow_location_inconsistency:
+                continue
+            line_source_locations = [
+                line.location_id
+                for line in pick.move_line_ids
+                if not getattr(line.move_id, "is_subcontract", False)
```
rousseldenis, line 42: Why is `is_subcontract` field here as there is no dependency on `mrp_subcontracting`. If you have problem with that, a glue module is necessary.

stock_picking_location_check/models/stock_picking.py
```diff
@@ -0,0 +1,51 @@
...
+            ]
+            pick._check_location_consistency(pick.location_id, line_source_locations)
+            if pick._check_immediate():
+                line_dest_locations = pick.move_line_ids.location_dest_id
+            else:
+                line_dest_locations = pick.move_line_ids.filtered(
+                    lambda line: line.qty_done > 0
```
rousseldenis: A `float_compare()`  is necessary here.

stock_picking_location_check/readme/CONTEXT.md
```diff
@@ -0,0 +1,4 @@
+Users sometimes change locations (location_id/location_dest_id) in the picking after
```
rousseldenis, line 1: In standard, it is not possible to select a location (dest  or source) that is not the same or child as the picking one (replicated on stock.move level) and then on move lines.

Do you have particular addons that change that behaviour ?

I've tested on runbot a case where I changed an internal transfer from Stock to Stock to WH to Stock (after reservation):

Move line:

image


Move:

image


That is coherent.

Could you explain your use case ?

stock_picking_location_check/models/stock_picking.py
AungKoKoLin1997: If `mrp_subtcontracting` was installed in the odoo environment, we want to filter out these subcontracted move lines because they are coming from `subcontracting location` and their location and picking's location will never be the same. Then, I think `getattr` is enough for this and we don't need to depend on `mrp_subcontracting`.

stock_picking_location_check/readme/CONTEXT.md
AungKoKoLin1997: This is the example use case we want to use.

https://github.com/user-attachments/assets/e4a9a25f-a002-4c39-805a-10113b88df1a

stock_picking_location_check/models/stock_picking.py
AungKoKoLin1997: Done!

rousseldenis (approved):
Ok, I see. That is fixed in v17

rousseldenis:
@sbidoul

sbidoul:
name reserved on pypi

PR #2042, opened 2024-05-21
[17.0][MIG] product_quantity_update_force_inventory

AnaisLForgeFlow:
Migration to v17.0:

The stock.inventory model and related inventory form views have been deprecated and are no longer available. This migration adapts the module by redirecting to the inventory adjustment view 'view_stock_quant_tree_inventory_editable' while filtering by the specific product.

@ForgeFlow


product_quantity_update_force_inventory/tests/test_stock_quant_editable_view_block.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2020-24 Forgeflow, S.L.
+#   (http://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestStockQuantEditableViewBlock(TransactionCase):
+    def setUp(self):
```
rousseldenis, line 9: Could you change this to class method?

product_quantity_update_force_inventory/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,11 @@
+With this module, when a user goes to a product and presses the button
+'Update Quantity' they will be directed to a new Inventory Adjustment
+for this product, instead of driving the user to update the quantities
+in the stock quants.
+
+By doing this users will be able to have full traceability on the
```
rousseldenis, line 6: I'm wondering what is the difference with standard flow as:

- If you update the quantity through quants, you'll get the move name with 'Quantity updated'-like message
- If you use the Action, you can add a reason:
image


image

rousseldenis (changes requested):

product_quantity_update_force_inventory/readme/DESCRIPTION.md
AnaisLForgeFlow: As it is stated in this pull request https://github.com/OCA/stock-logistics-warehouse/pull/846, this module was created to force users to create inventory adjustments instead of directly updating quantities on the product card but since the stock.inventory is no longer available, now it is not possible to start inventory adjustments without updating quantities. Given these changes, I adapted the module to redirect users to the inventory adjustment view filtered by the product when they initiate a quantity update. However, I recognize that this adaptation does not fully replicate the original functionality of the module.
Since the module has lost its original purpose, what do you think it would be appropiate to do with it?

rousseldenis: Yes, I know the drawback of not using inventory lines anymore.

But, IMHO, we should try not duplicating existing functionalities as it can add confusion for users, add an additional entry point to do the same thing.

I would say, something has to be done on the actions visible on stock quant view (in inventory mode):

image


For that 'Apply' action, something should be done to display the reason popup too if needed.

rousseldenis: In fact your functionality is not needed anymore. Odoo has managed that well. If you don't have 'Storage Locations' enabled on settings, the popup will be launched as there is only 1 location (Stock). That popup is launched if you click on the magic button 'On Hand' in the product form.

If you enable 'Storage Locations' setting, you are both redirected to the stock quant view.

AnaisLForgeFlow: Yes, i agree with you. I will be closing this PR as it is not needed anymore.
Thank you for your cooperation.

rousseldenis: @AnaisLForgeFlow The only missing feature is to be able to add a reason for one line (the Apply button on the line)

AnaisLForgeFlow: But the feature that you mean is what is being added in the stock_change_qty_reason module, right?

rousseldenis: Right

PR #2044, opened 2024-05-21
[16.0][FIX] stock_reserve: Set the correct _default_picking_type_id value

victoralmau:
FWP from 15.0: https://github.com/OCA/stock-logistics-warehouse/pull/2043

Set the correct `_default_picking_type_id` value

Do not use a fixed value because it will not be compatible with multi-company.

@Tecnativa TT49168


pedrobaeza (approved):
Let's finish the fw-port chain:

/ocabot merge patch

PR #2046, opened 2024-05-23
[ADD] stock_exception: when to run the exception (confirm/validate)

mav-adhoc:
When we were testing the model, in case we choose 'stock.picking' as the model where we wanted to apply the exception, we noticed that the exception also ran when we confirmed sale orders.
This happens because, by confirming sale orders, the action confirm creates the respective records in stock.picking (IN/OUT, depending on the configuration you have) and run the exception.
What we want is to have the decision when to run the exception, if at the creation of the stock.picking record or if at the validation of the stock.picking record.

Due to this desire, we created two new boolean fields in the stock.exception class (check_on_validate and check_on_confirm) in order to indicate when to run the exception. If you select check_on_confirm, it will run the exception when you create the stock.picking record. If you select check_on_validate, it will run the exception when you validate the stock.picking record.

### DIFF
--- a/stock_exception/__manifest__.py
+++ b/stock_exception/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock Exception",
     "summary": "Custom exceptions on stock picking",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
     "category": "Generic Modules/Warehouse Management",
     "author": "Ecosoft, Odoo Community Association (OCA)",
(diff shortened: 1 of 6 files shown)


mav-adhoc:
@yostashiro @rousseldenis Hi! is there a possibility we can add this improvement on the stock_exception module? Thanks

jcadhoc:
Hello!

@OCA/deployment-tools-maintainers merge please :)

PR #2048, opened 2024-05-23
[16.0][FIX] stock_inventory: prevent error when applying inventory without quantities

AaronHForgeFlow:
FW port of https://github.com/OCA/stock-logistics-warehouse/pull/2047/


LoisRForgeFlow (approved):
Ty!

/ocabot merge patch

PR #2049, opened 2024-05-24
[17.0][MIG] stock_orderpoint_generator

astirpe:


stock_orderpoint_generator/models/orderpoint_template.py
```diff
@@ -82,6 +82,20 @@ class OrderpointTemplate(models.Model):
     qty_forecast = fields.Float(compute=False, store=False)
     product_category_id = fields.Many2one(related=False, store=False)
 
+    allowed_warehouse_ids = fields.One2many(
```
rousseldenis, line 85: @astirpe Could you put those changes that are enhancement in a separate commit in order to be able to backport them (and separate from real migration) ? Thanks

astirpe: @rousseldenis They are part of the real migration, not enhancement. Odoo would raise an error without this part. I think this should be backported as well.

rousseldenis: That's why a separate commit is more convenient

astirpe: Not sure I get your point. When backporting from V17 to V16, other parts of the code should be changed, like the qty_done and the views, but not this part that could be kept as it is.

rousseldenis:
@astirpe In fact, I didn't realize but you should do the job here instead : https://github.com/OCA/stock-logistics-orderpoint/

astirpe:
@rousseldenis thank you! Will do that. Closing this one.

PR #2050, opened 2024-05-27
[ADD] stock_quant_product_packaging

kevinkhao:


stock_quant_product_packaging/models/stock_quant.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    packaging_info_ids = fields.One2many(
+        "stock.quant.packaging.info",
+        "quant_id",
+        string="Packaging additional information",
+    )
```
thomaspaulb, line 14: Is this persistent? My understanding is that Odoo can randomly decide to [merge](https://github.com/OCA/OCB/blob/16.0/addons/stock/models/stock_quant.py#L939) quants that look similar, and sum the quantities into one total, and then you will lose your packaging information.

If you don't want that, you have to make at least one of these fields unique: `GROUP BY product_id, company_id, location_id, lot_id, package_id, owner_id`.

Funnily, that actually contains `package_id`, so maybe you can make use of that. I don't know what it does, though.

Another problem might be that once your stock on hand drops to zero, your quant is [unlinked](https://github.com/OCA/OCB/blob/16.0/addons/stock/models/stock_quant.py#L929). If stock comes in again, a quant will be created, but you'll need to somehow add it to the right package by default. Did you add code for that?

kevinkhao: package_id refers to the stock.quant.package model which from my understanding is more oriented towards e.g making a parcel with 2 chairs and 1 table (different products). In this module I am using product.packaging, as in, one carton of 20 units (of the same product) on the quant in question

As for the merging, of the quants, I will look into it

As for unlinking the quant, the intention was to have the quants created by some other external code with the right packagings

Thanks for the fast review!

rousseldenis:
@kevinkhao I would have renamed this module to stock_quant_packaging_additional_information

rousseldenis:
@kevinkhao ave you ever looked at https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_product_qty_by_packaging (and https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_packaging_calculator) ?

yajo:
We've been thinking a lot about this subject too and I think implementing it in an extra model, as you are doing here, isn't maybe the best approach.

IIUC in Odoo, `stock.quant` records are essentially a sum of `stock.move.line`s grouped by certain criteria, such as location, lot, `stock.quant.package`, owner, etc.

In `stock_move_packaging_qty` we already have information such as what's the `product.packaging` and its qtys involved in one `stock.move.line`: https://github.com/OCA/stock-logistics-warehouse/blob/f84bce2bc63033731b948219ad42197acfbcb919/stock_move_packaging_qty/models/stock_move_line.py#L6-L19

Thus our hypothetical design for a module that tracks packagings by quant would just extend  `stock_move_packaging_qty`  and:
- Add a `product_packaging_id` to the `stock.quant` model.
- Modify the algorithms that group the `stock.move.line`s into `stock.quant`s to have that field into account.

I think that would simplify a lot the design and would integrate better into the rest of Odoo. WDYT?

cc @rafaelbn @EmilioPascual @Shide @moduon

rafaelbn:
Hello @kevinkhao ,

We are interested in this. We need to solve too @moduon

Ping @yajo @Shide @EmilioPascual , MT-5287 and MT-5287

Thank you! 😄 ❤️

kevinkhao:
> We've been thinking a lot about this subject too and I think implementing it in an extra model, as you are doing here, isn't maybe the best approach.
> 
> IIUC in Odoo, `stock.quant` records are essentially a sum of `stock.move.line`s grouped by certain criteria, such as location, lot, `stock.quant.package`, owner, etc.
> 
> In `stock_move_packaging_qty` we already have information such as what's the `product.packaging` and its qtys involved in one `stock.move.line`:
> 
> https://github.com/OCA/stock-logistics-warehouse/blob/f84bce2bc63033731b948219ad42197acfbcb919/stock_move_packaging_qty/models/stock_move_line.py#L6-L19
> 
> Thus our hypothetical design for a module that tracks packagings by quant would just extend `stock_move_packaging_qty` and:
> 
> * Add a `product_packaging_id` to the `stock.quant` model.
> * Modify the algorithms that group the `stock.move.line`s into `stock.quant`s to have that field into account.
> 
> I think that would simplify a lot the design and would integrate better into the rest of Odoo. WDYT?
> 
> cc @rafaelbn @EmilioPascual @Shide @moduon

This seems like a good approach. On our side, we decided to opt for another hacky solution that fit the need temporarily... We need to go for the "real" solution now, so I will look further into this path.

kevinkhao:
> @kevinkhao ave you ever looked at https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_product_qty_by_packaging (and https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_packaging_calculator) ?

@rousseldenis AFAIK these modules allow for grouping a quantity of products by optimized packaging. e.g: 
Product X has packagings by 50, 20, 10
I want 130 units -> I use these modules to calculate that to attain these quantities I need 2 packagings by 50, 1 by 20, 1 by 10

Mine was is an attempt to inform the number of each packaging **(not packages)** we have in stock. 
The logical way seems to put product packagings on quants, e.g:
Product X has packagings by 50, 20, 10. In my stock I have 200 units
Then my stock is represented by 3 quants as follows:
2 packagings of 50
4 packagings of 20
2 packagings of 10

With some investigation we need to take into account 1. the create/update/delete which seems doable, and 2. all the places that quants are read, take into account the packagings. This seems like an undertaking that seems risky/costly that I don't think I can afford.

@rafaelbn @yajo @Shide @EmilioPascual have you found a way to deal with this need without rewriting a sizeable part of the stock module ?

yajo:
Sorry, we don't use those modules and didn't have that specific need until now.

bosd:
> We are interested in this. We need to solve too @moduon

@rafaelbn Have you implemented a solution for this use case?

yajo:
@rafaelbn and me work together, so the answer is the same: [we still didn't have this particular need](https://github.com/OCA/stock-logistics-warehouse/pull/2050#issuecomment-2363101870), so we have not implemented any solution, but we think that the proper solution should follow [this design](https://github.com/OCA/stock-logistics-warehouse/pull/2050#issuecomment-2146813581).

rousseldenis:
@kevinkhao Any update on this ?

bosd:
@kevinkhao Can you re-trigger tests / Runboat?

kevinkhao:
@rousseldenis @bosd 
sorry for late reply guys - this one took some time to solve and then got lost, in the end we didn't go through this route. Instead bypassed the quants entirely and went through direct SQL requests to compile packaging info

PR #2051, opened 2024-05-29
[15.0][IMP] stock_inventory: adjust restriction

AaronHForgeFlow:
This change allows to create several inventory adjustments on the same location if the products involved are different

You may say that creating inventory adjustment in the same location by product category is not efficient and I agree with that. However, the possibility is there, and the constraint should not be raised in this case.

@ ForgeFlow

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -190,25 +190,49 @@ class InventoryAdjustmentsGroup(models.Model):
             ]
         )
 
+    def _get_products(self):
+        self.ensure_one()
+        if self.product_selection == "all":
+            return self.env["product.product"].search([])
+        elif self.product_selection == "category" and self.category_id:
+            return self.env["product.product"].search(
+                [
+                    "|",
+                    ("categ_id", "=", self.category_id.id),
+                    ("categ_id", "in", self.category_id.child_id.ids),
+                ]
+            )
+        else:
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        self.ensure_one()
+        inventories = self.search([("state", "=", "in_progress")])
+        if not inventories:
+            return self.env["stock.inventory"].browse()
+        inventory_products = self._get_products()
+        overlapping_inventories = inventories.filtered(
+            lambda x: (
+                (
+                    any(i in x.location_ids for i in self.location_ids)
+                    or (
+                        any(
+                            i in x.location_ids.child_internal_location_ids
+                            for i in self.location_ids
+                        )
+                        and not x.exclude_sublocation
+                    )
+                )
+                and any(product in inventory_products for product in x._get_products())
+            )
+        )
+        return overlapping_inventories
+
     def refresh_stock_quant_ids(self):
         for rec in self:
             rec.stock_quant_ids = rec._get_quants(rec.location_ids)
 
     def action_state_to_in_progress(self):
-        active_rec = self.env["stock.inventory"].search(
-            [
-                ("state", "=", "in_progress"),
-                ("location_ids", "child_of", self.location_ids.ids),
-            ],
-            limit=1,
-        )
-        if active_rec:
-            raise ValidationError(
-                _(
-                    "There's already an Adjustment in Process using one requested Location: %s"
-                )
-                % active_rec.name
-            )
         self.state = "in_progress"
         self.refresh_stock_quant_ids()
         self.stock_quant_ids.update(
@@ -265,28 +289,18 @@ class InventoryAdjustmentsGroup(models.Model):
         result["context"] = []
         return result
 
-    @api.constrains("state", "location_ids")
+    @api.constrains(
+        "state", "location_ids", "product_ids", "category_id", "product_selection"
+    )
     def _check_inventory_in_progress_not_override(self):
         inventories = self.search([("state", "=", "in_progress")])
         for rec in inventories:
-            inventory = inventories.filtered(
-                lambda x: x.id != rec.id
-                and (
-                    any(i in x.location_ids for i in rec.location_ids)
-                    or (
-                        any(
-                            i in x.location_ids.child_internal_location_ids
-                            for i in rec.location_ids
-                        )
-                        and not x.exclude_sublocation
-                    )
-                )
-            )
-            if len(inventory) > 0:
+            overlapping_inventories = rec._get_overlapping_inventories()
+            if len(overlapping_inventories) > 1:
                 raise ValidationError(
                     _(
                         "Cannot be more than one in progress inventory adjustment "
-                        "affecting the same location at the same time."
+                        "affecting the same location and product at the same time."
                     )
                 )
 
--- a/stock_cycle_count/tests/test_stock_cycle_count.py
+++ b/stock_cycle_count/tests/test_stock_cycle_count.py
@@ -26,10 +26,14 @@ class TestStockCycleCount(common.TransactionCase):
         cls.partner = cls.env.ref("base.res_partner_1")
         cls.g_stock_manager = cls.env.ref("stock.group_stock_manager")
         cls.g_stock_user = cls.env.ref("stock.group_stock_user")
+        cls.g_manager_settings = cls.env.ref("base.group_system")
 
         # Create users:
         cls.manager = cls._create_user("user_1", [cls.g_stock_manager], cls.company).id
         cls.user = cls._create_user("user_2", [cls.g_stock_user], cls.company).id
+        cls.admin_settings_user = cls._create_user(
+            "user_boss", [cls.g_stock_manager, cls.g_manager_settings], cls.company
+        ).id
 
         # Create warehouses:
         cls.big_wh = cls.stock_warehouse_model.create(
@@ -230,17 +234,21 @@ class TestStockCycleCount(common.TransactionCase):
         """Tests workflow."""
         with self.assertRaises(AccessError):
             self.cycle_count_1.action_create_inventory_adjustment()
+        # actually the inv adjustment was not created
+        # but test were not failing because the next serach was
+        # returning an empty recorset
+        self.cycle_count_1.with_user(
+            self.admin_settings_user
+        ).action_create_inventory_adjustment()
         inventory = self.inventory_model.search(
             [("cycle_count_id", "=", self.cycle_count_1.id)]
         )
-        with self.assertRaises(AssertionError):
-            self.assertTrue(inventory, "Inventory not created.")
+        self.assertEqual(len(inventory), 1)
         inventory.action_state_to_in_progress()
         inventory.action_state_to_done()
-        with self.assertRaises(AssertionError):
-            self.assertEqual(
-                self.cycle_count_1.state, "done", "Cycle count not set as done."
-            )
+        self.assertEqual(
+            self.cycle_count_1.state, "done", "Cycle count not set as done."
+        )
         self.cycle_count_1.do_cancel()
         self.assertEqual(
             self.cycle_count_1.state, "cancelled", "Cycle count not set as cancelled."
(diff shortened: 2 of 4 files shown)


stock_inventory/models/stock_inventory.py
```diff
@@ -196,26 +196,53 @@ def refresh_stock_quant_ids(self):
...
+                    if x.product_selection != "all"
+                    else self.env["product.product"].search([])
+                    or self.env["product.product"].search(
+                        [("categ_id", "=", x.category_id.id)]
+                        if x.product_selection == "category" and x.category_id
+                        else []
+                    )
+                )
+            )
+        )
```
JordiMForgeFlow: it could be good to move the logic to get an overlapping inventory in a separate method and use it also for the constrain

stock_inventory/models/stock_inventory.py
```diff
@@ -196,26 +196,53 @@ def refresh_stock_quant_ids(self):
...
+                    x.product_ids
+                    if x.product_selection != "all"
+                    else self.env["product.product"].search([])
+                    or self.env["product.product"].search(
+                        [("categ_id", "=", x.category_id.id)]
+                        if x.product_selection == "category" and x.category_id
+                        else []
+                    )
+                )
+            )
```
MiquelRForgeFlow: A bit confusing. Put more `(` and `)` to be sure about what you are doing is correct.

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: I agree, done that!

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: now it is more simple :)

AaronHForgeFlow:
CI fixed now :tada:

stock_inventory/models/stock_inventory.py
```diff
@@ -266,28 +297,16 @@ def action_view_stock_moves(self):
         result["context"] = []
         return result
 
-    @api.constrains("state", "location_ids")
+    @api.constrains("state", "location_ids", "product_ids", "category_id")
```
JordiMForgeFlow:
```suggestion
    @api.constrains("state", "location_ids", "product_ids", "category_id", "product_selection")
```

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,55 @@ def _get_domain_category_quants(self, base_domain):
...
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        for rec in self:
+            inventories = self.search([("state", "=", "in_progress")])
+            inventory_products = rec._get_products()
+
+            overlapping_inventories = inventories.filtered(
+                lambda x: (
+                    x.id != rec.id
```
JordiMForgeFlow:
```suggestion
                    x.id != rec.id
```

you could remove this if when doing the search you already do like: [("state", "=", "in_progress"), ("id", "!=", rec.id)]

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,56 @@ def _get_domain_category_quants(self, base_domain):
...
+        elif self.product_selection == "category" and self.category_id:
+            return self.env["product.product"].search(
+                [("categ_id", "=", self.category_id.id)]
+            )
+        else:
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        for rec in self:
+            inventories = self.search(
```
rousseldenis: This search should be done outside of the loop. Then, filter on id.

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,56 @@ def _get_domain_category_quants(self, base_domain):
...
+                [("state", "=", "in_progress"), ("id", "!=", rec.id)]
+            )
```
JordiMForgeFlow:
```suggestion
            inventories = self.search(
                [("state", "=", "in_progress"), ("id", "!=", rec.id)]
            )
            if not inventories:
                continue
```

just a small improvement, sorry for the nitpicking 😆 for the rest LGTM!

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,56 @@ def _get_domain_category_quants(self, base_domain):
...
+            return self.env["product.product"].search([])
+        elif self.product_selection == "category" and self.category_id:
+            return self.env["product.product"].search(
+                [("categ_id", "=", self.category_id.id)]
+            )
+        else:
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        for rec in self:
```
JordiMForgeFlow: here it should be only one record actually, right?

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: I did this change, but if then I move the search out of the loop, so now this filtered is needed again

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: yes, indeed, I had self.ensure_one() before but sometimes it was called with an empty recordset and that was causing tests to fail

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: the more nitpicking the more quality :)

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: in the normal case there is only 1 record, but yes, it is more efficient this way.

stock_inventory/models/stock_inventory.py
JordiMForgeFlow: yes I saw that, for some reason in the test of stock_cycle_count they are calling the method with empty self. I think the problem in this case is with the test, I don't see the reason why it should be done that way

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,56 @@ def _get_domain_category_quants(self, base_domain):
...
+            return self.env["product.product"].search(
+                [("categ_id", "=", self.category_id.id)]
+            )
+        else:
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        inventories = self.search([("state", "=", "in_progress")])
+        if not inventories:
+            return self.env["stock.inventory"]
```
```suggestion
            return self.env["stock.inventory"].browse()
```

is better

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,56 @@ def _get_domain_category_quants(self, base_domain):
...
+        overlapping_inventories = self._get_overlapping_inventories()
+        if overlapping_inventories:
             raise ValidationError(
                 _(
-                    "There's already an Adjustment in Process using one requested Location: %s"
+                    "There's already an Adjustment in Process using one requested Location"
+                    " and Product: %s"
                 )
-                % active_rec.name
+                % overlapping_inventories.name
```
rousseldenis: Using translation function arguments is better.

```suggestion
                _(
                    "There's already an Adjustment in Process using one requested Location"
                    " and Product: %(name)s", name=overlapping_inventories.name
                    )
```

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: perhaps that test is incorrect, but I am not sure if the method it can be called with empty recordset in any other way, and I thought that doing this change was safer

JordiMForgeFlow: I think that the correct thing is to put the self.ensure_one(), and then correct the test. Because it should give an error if you try to execute action on empty record I would say

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: removed the loop, and added this change again

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: Sure!, thanks

stock_inventory/models/stock_inventory.py
```diff
@@ -191,24 +191,55 @@ def _get_domain_category_quants(self, base_domain):
...
+            return self.env["product.product"].search([])
+        elif self.product_selection == "category" and self.category_id:
+            return self.env["product.product"].search(
+                [("categ_id", "=", self.category_id.id)]
+            )
+        else:
+            return self.product_ids
+
+    def _get_overlapping_inventories(self):
+        self.ensure_one()
```
rousseldenis, line 209: I'm just wondering why not using multi. As this function is very interesting. Maybe to build a report, or...

AaronHForgeFlow: Well, it is not needed for the purpose of the change, but yes, it may be useful for other purposes, so ok to me.

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: The tests for the stock_cycle_count are incorrect as far as I can see, but the fix is not totally clear to me, I will try to investigate a bit more

AaronHForgeFlow:
@rousseldenis , @JordiMForgeFlow  I dediced to fix the tests in the stock_cycle_count module, indeed the test that was failing was actually not testing anything as long the inventory adjustment was never created in that test.

for now I left the method _get_overlapping_inventories as self.ensure_one(), for the purpose of the PR that is sufficient. If this change is a blocking change then I can consider make it multi.

JordiMForgeFlow:
@AaronHForgeFlow could you add some test to validate the new changes for the constrain?

AaronHForgeFlow:
@JordiMForgeFlow Added the test. Also remove the constraint check in the action_in_progress method. It is not necessary as the constraint method depends in the state. Adapted a bit the code.

stock_inventory/models/stock_inventory.py
```diff
@@ -191,25 +191,45 @@ def _get_domain_category_quants(self, base_domain):
...
         )
 
+    def _get_products(self):
+        self.ensure_one()
+        if self.product_selection == "all":
+            return self.env["product.product"].search([])
+        elif self.product_selection == "category" and self.category_id:
+            return self.env["product.product"].search(
+                [("categ_id", "=", self.category_id.id)]
+            )
```
JordiMForgeFlow: @AaronHForgeFlow one last comment: should we consider also child categories here?

MiquelRForgeFlow:
conflicts

stock_inventory/models/stock_inventory.py
AaronHForgeFlow: good point!

AaronHForgeFlow:
> conflicts

rebased & solved

JordiMForgeFlow (approved):

JoanSForgeFlow:
Hi @AaronHForgeFlow 

I've noticed we have overlapping PRs (https://github.com/OCA/stock-logistics-warehouse/pull/2021) . Here are my observations and proposal:

1. **Review of this PR**:
   - This PR improves the system by allowing two stock inventories in the same location for different products.
   - However, when performing a general review of a product in a global location (like a warehouse), all interior sublocations are blocked.
   - In my opinion, only sublocations where the product being reviewed is located should be blocked.
   - Sublocations without any quant under review should remain accessible.

2. **My Proposal**:
   - In my PR I have done a small refactor of the `to_do`, where `to_do` represents the quant currently in a stock inventory in progress.
   - With my change, only the stock.inventory reviewing a quant already marked as `to_do = True` gets blocked in order to allow reviews in sublocations where this product is not in stock. (You can review test_09 of my PR, this is the one I think this PR is not covering https://github.com/OCA/stock-logistics-warehouse/pull/2021/files#diff-4307516705b49705a1076c0c27cb158684c4bde931c08199f7273efa32d1de7cR457 )

I hope this clarifies my proposal. Looking forward to your thoughts. 

Thanks.

AaronHForgeFlow:
I am closing this PR in favor of https://github.com/OCA/stock-logistics-warehouse/pull/2021/ Please do not delete the branch

The other PR covers more use cases than this one, and that is why it is ok to me to keep the other.

PR #2052, opened 2024-05-29
[14.0][ADD]new module stock_location_simple

clementmbr:
Add a new simplified menu for locations to avoid user to make annoying changes

cc @Kev-Roche @sebastienbeau


rousseldenis:
> Add a new simplified menu for locations to avoid user to make annoying changes
> 
> cc @Kev-Roche @sebastienbeau

@clementmbr It seems this does other things too. Maybe refine the module name to reflect actual functionalities.

clementmbr:
> @clementmbr It seems this does other things too. Maybe refine the module name to reflect actual functionalities.

Hi @rousseldenis, thanks for the comment.
Indeed, this module only adds a simplified Location menu. But in order to really be resistant to newbie users creativity, we need to make some Location's fields readonly, through this menu, if the location is linked to a warehouse.

I can imagine this boolean "is_linked_to_warehouse" on stock.location can be useful to some other future module and I confess I also first thought to create a distinct module for this "functionality".

But later I realized it would mean to create 3 modules instead of one (one for the simplified view, one for the boolean, and a third one to add the boolean on the simplified view), and it looked like a lot of structural complexity for a module which does something very simple, only to make this boolean available for a future hypothetical module which does not even exist yet.

What I can propose is to make this split in 3 modules only if someone shows up asking for this boolean outside of this module (I separated the code in a specific commit in order to ease the split if necessary). Like this, we will add the structural complexity only needed.

What do you think?

PS: I realize after writing this that you were only talking about the module's name. Something like "dumbie-proof location menu" would be more explicit for sure, but I didn't find a better name translating this idea. Do you think of something better?
Thanks!

rousseldenis:
A question comes through my mind, but why not using standard security features and using inventory user groups for the users that should not manage locations ?

clementmbr:
> A question comes through my mind, but why not using standard security features and using inventory user groups for the users that should not manage locations ?

The idea is that we do want basic users to be able to create and modify locations. But they should be able to create/modify only internal locations, child of an existing one, and not be able to create/modify the locations linked to a warehouse.
Does it make sense to you?

Kev-Roche (approved):
LGTM

PR #2057, opened 2024-05-31
[14.0][IMP] stock_location_package_restriction: add nopackage

jbaudoux:
* Allow to configure that you don't want a package on the location.
* Refactor code. Run check on quants after move is done. Drop the `read_group` on each `action_done`
* Allow to check a move line location before validation
* Show & search locations not respecting the restriction.
* Split tests in separate test files

cc @mt-software-de @TDu

### DIFF
--- a/stock_location_package_restriction/models/stock_location.py
+++ b/stock_location_package_restriction/models/stock_location.py
@@ -4,6 +4,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.osv.expression import NEGATIVE_TERM_OPERATORS
 
 NOPACKAGE = "nopackage"
 SINGLEPACKAGE = "singlepackage"
@@ -38,6 +39,67 @@ class StockLocation(models.Model):
             (MULTIPACKAGE, "Mandatory"),
             (SINGLEPACKAGE, "Mandatory and unique"),
         ]
+
+    has_package_restriction_violation = fields.Boolean(
+        compute="_compute_has_package_restriction_violation",
+        search="_search_has_package_restriction_violation",
+    )
+
+    def _has_package_restriction_violation_query(self):
+        self.flush()
+        query = """
+            SELECT stock_quant.location_id
+            FROM stock_quant
+            JOIN stock_location ON stock_location.id = stock_quant.location_id
+            WHERE
+                quantity != 0
+                AND stock_location.package_restriction IS NOT NULL
+            """
+        if self:
+            query += "AND stock_quant.location_id in %s"
+        query += f"""
+            GROUP BY
+                stock_quant.location_id, stock_location.package_restriction
+            HAVING
+                (
+                    stock_location.package_restriction = '{NOPACKAGE}'
+                    AND count(distinct(stock_quant.package_id)) > 0
+                ) OR (
+                    stock_location.package_restriction = '{SINGLEPACKAGE}'
+                    AND (
+                        count(distinct(stock_quant.package_id)) > 1
+                        OR count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0
+                    )
+                ) OR (
+                    stock_location.package_restriction = '{MULTIPACKAGE}'
+                    AND count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0
+                )
+        """
+        return query
+
+    @api.depends("package_restriction")
+    def _compute_has_package_restriction_violation(self):
+        self.env.cr.execute(
+            self._has_package_restriction_violation_query(), (tuple(self.ids),)
+        )
+        error_ids = [r[0] for r in self.env.cr.fetchall()]
+        for location in self:
+            location.has_package_restriction_violation = location.id in error_ids
+
+    def _search_has_package_restriction_violation(self, operator, value):
+        search_has_violation = (
+            # has_restriction_violation != False
+            (operator in NEGATIVE_TERM_OPERATORS and not value)
+            # has_restriction_violation = True
+            or (operator not in NEGATIVE_TERM_OPERATORS and value)
+        )
+        self.env.cr.execute(self._has_package_restriction_violation_query())
+        error_ids = [r[0] for r in self.env.cr.fetchall()]
+        if search_has_violation:
+            op = "in"
+        else:
+            op = "not in"
+        return [("id", op, error_ids)]
 
     def _check_package_restriction(self, move_lines=None):
         """Check if the location respect the package restrictions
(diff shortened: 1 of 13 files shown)


stock_location_package_restriction/models/stock_location.py
```diff
@@ -16,21 +19,68 @@ class StockLocation(models.Model):
...
     @api.model
     def _selection_package_restriction(self):
         return [
+            (NOPACKAGE, "Forbidden"),
             (MULTIPACKAGE, "Mandatory"),
             (SINGLEPACKAGE, "Mandatory and unique"),
         ]
+
+    def _check_package_restriction(self, move_lines):
+        if not self.package_restriction:
```
mt-software-de:
```suggestion
        self.ensure_one()
        if not self.package_restriction:
```

stock_location_package_restriction/models/stock_move.py
```diff
@@ -1,109 +1,29 @@
...
-                            location=location.display_name,
-                            packages=", ".join(new_packages.mapped("display_name")),
-                        )
-                    )
-                    continue
-                new_packages |= package
-
+            try:
+                with self.env.cr.savepoint():
+                    location._check_package_restriction(move_lines)
```
mt-software-de: Is the Savepoint needed?
There is a Union missing. Move lines is just a list.

```suggestion
                move_lines = self.env["srick.move.line"].union(*move_lines)
                location._check_package_restriction(move_lines)
```

mt-software-de (changes requested):
Looks good. Just two remarks

TDu (approved):

jbaudoux:
@mt-software-de It wasn't ready yet :) Please review again as I changed the approach

jbaudoux:
Ping @mt-software-de

stock_location_package_restriction/models/stock_move.py
```diff
@@ -1,112 +1,21 @@
...
-                    continue
-                new_packages |= package
-
-        if error_msgs:
-            raise ValidationError("\n".join(error_msgs))
+    def _check_location_package_restriction(self):
+        """Check if the moves can be done regarding potential package restrictions"""
+        self.filtered(
+            lambda m: m.state == "done"
+        ).location_dest_id._check_package_restriction()
```
mt-software-de, line 16: This is now always checking all quants on the location. Should it not only check the current move lines?
```suggestion
        ).location_dest_id._check_package_restriction(self.move_line_ids)
```

stock_location_package_restriction/models/stock_location.py
```diff
@@ -16,21 +20,158 @@ class StockLocation(models.Model):
...
+        )
+        self.env.cr.execute(self._has_package_restriction_violation_query())
+        error_ids = [r[0] for r in self.env.cr.fetchall()]
+        if search_has_violation:
+            op = "in"
+        else:
+            op = "not in"
+        return [("id", op, error_ids)]
+
+    def _check_package_restriction(self, move_lines=None):
```
mt-software-de, line 149: Is it really needed to make move_lines optional?
It should called with move_lines from stock.move action_done.

If it stays optional it would be make sense to call the has_package_violation instead of redoing the check here for a specific location?

stock_location_package_restriction/models/stock_location.py
```diff
@@ -16,21 +20,158 @@ class StockLocation(models.Model):
...
+                    AND (
+                        count(distinct(stock_quant.package_id)) > 1
+                        OR count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0
+                    )
+                ) OR (
+                    stock_location.package_restriction = '{MULTIPACKAGE}'
+                    AND count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0
+                )
+        """
+        return query
```
mt-software-de: Would it not make sense to return the violation in sql?
Then we could display as well a violation string on the location and it could be reused by the `_check_package_restriction` method. 

```suggestion
        self.flush()
        nopackage_violation = f"stock_location.package_restriction = '{NOPACKAGE}' AND count(distinct(stock_quant.package_id)) > 0"
        singlepackage_violation = f"stock_location.package_restriction = '{SINGLEPACKAGE}' AND (count(distinct(stock_quant.package_id)) > 1 OR OR count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0)"
        multipackage_violation = f"stock_location.package_restriction = '{MULTIPACKAGE}' AND count(*) FILTER (WHERE stock_quant.package_id IS NULL) > 0"
        
        query = f"""
            SELECT
                stock_quant.location_id,
                CASE WHEN {nopackage_violation} THEN 1 ELSE 0 END AS nopackage_violation,
                CASE WHEN {singlepackage_violation} THEN 1 ELSE 0 END AS singlepackage_violation,
                CASE WHEN {multipackage_violation} THEN 1 ELSE 0 END AS multipackage_violation
            FROM stock_quant
            JOIN stock_location ON stock_location.id = stock_quant.location_id
            WHERE
                quantity != 0
                AND stock_location.package_restriction IS NOT NULL
            """
        if self:
            query += "AND stock_quant.location_id in %s"
        query += f"""
            GROUP BY
                stock_quant.location_id, stock_location.package_restriction
            HAVING
                ({nopackage_violation}) OR ({singlepackage_violation}) OR ({multipackage_violation})
        """
        return query
```

mt-software-de (changes requested):

stock_location_package_restriction/models/stock_move.py
jbaudoux: If the location has a violation but the just moved move lines are correct then you would accept to put them on the location. I thought it is safer to check the entire location.

stock_location_package_restriction/models/stock_location.py
jbaudoux: I prefer to not make explicit sql query in a transaction but use data from the cache.
On a list of location, as there may be many data to process, I prefer the sql query.

stock_location_package_restriction/models/stock_location.py
jbaudoux: I added the violation message

stock_location_package_restriction/models/stock_move.py
mt-software-de: Yes it would be safer, but this would change the current implementation. 
At the moment it is to book a moves on a location, although the location has a violation.

stock_location_package_restriction/models/stock_location.py
mt-software-de: :-) Ok. I prefer to not duplicate functionality `_check_package_restriction` is not anymore called with move_lines.

jbaudoux: But you can call it (for example from shopfloor) to validate a move line destination location before posting the move to know if it is good or not

PR #2058, opened 2024-05-31
[16.0][FIX] stock_inventory: create missing quants if needed

benwillig:
Based on https://github.com/OCA/stock-logistics-warehouse/pull/2017

- Display quant smartbutton even when there is no existing quant to allow creating new ones
- Adds a boolean on the inventory to create missing quants with zero quantity

cc @DavidJForgeFlow @rousseldenis

### DIFF
--- a/stock_inventory/models/res_company.py
+++ b/stock_inventory/models/res_company.py
@@ -13,3 +13,7 @@ class ResCompany(models.Model):
         "are done, the adjustment is automatically set to done.",
         default=False,
     )
+    stock_inventory_auto_create_missing_quants = fields.Boolean(
+        help="If enabled, missing quants will be created on the inventory confirmation",
+        default=False,
+    )
--- a/stock_inventory/models/res_company.py
+++ b/stock_inventory/models/res_company.py
@@ -11,9 +11,7 @@ class ResCompany(models.Model):
     stock_inventory_auto_complete = fields.Boolean(
         help="If enabled, when all the quants prepared for the adjustment "
         "are done, the adjustment is automatically set to done.",
-        default=False,
     )
     stock_inventory_auto_create_missing_quants = fields.Boolean(
         help="If enabled, missing quants will be created on the inventory confirmation",
-        default=False,
     )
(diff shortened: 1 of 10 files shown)


stock_inventory/models/res_company.py
```diff
@@ -13,3 +13,7 @@ class ResCompany(models.Model):
         "are done, the adjustment is automatically set to done.",
         default=False,
     )
+    stock_inventory_auto_create_missing_quants = fields.Boolean(
+        help="If enabled, missing quants will be created on the inventory confirmation",
+        default=False,
```
rousseldenis: That is the default.

stock_inventory/models/stock_quant.py
```diff
@@ -1,70 +1,89 @@
...
+        comodel_name="stock.inventory",
+        string="Inventory Adjustment",
+        readonly=True,
+    )
     to_do = fields.Boolean(default=True)
 
+    def _compute_virtual_in_progress_inventory_id(self):
+        Inventory = self.env["stock.inventory"]
+        for rec in self:
+            rec.virtual_in_progress_inventory_id = Inventory.search(
```
rousseldenis: Prefer a search outside the loop and then filtered_domain on locations.

rousseldenis (changes requested):
Great. Code review.

stock_inventory/models/stock_quant.py
benwillig: Done

benwillig:
ping @DavidJForgeFlow @LoisRForgeFlow 

just a gently reminder -> https://github.com/OCA/stock-logistics-warehouse/pull/1935#issuecomment-2082849933

LoisRForgeFlow:
Hi @benwillig 

Thanks for the reminder and thanks for adding the optional stetting to enable that, when you do have a lot of bin locations and many products, creating all the quants for every inventory can become a little crazy, but I understand that in some situations is usefull.

Could you fix the conflicts that prevent merging?

PR #2075, opened 2024-06-14
[16.0][FIX] stock_location_product_restriction: Ensure restriction is alway…

lmignon:
…s checked

Transform check at move validation by a constraint on the stock.quant to ensure that the rule is checked in every cases


rousseldenis (approved):
Code and functional review

jbaudoux (approved):
Interesting improvement :)

We should do same improvement in `stock_location_package_restriction` https://github.com/OCA/stock-logistics-warehouse/pull/2068

PR #2076, opened 2024-06-14
WIP [ADD] stock_product_template_procurement

petrus-v:


petrus-v:
Closing this because is not wise and should be allowed by odoo core in future version. we will see if we have energy to contribute and discuss in this way

PR #2082, opened 2024-06-20
[ADD] stock_secondary_unit migrate to 17.0

fredericgrall:


pedrobaeza:
If it's a migration, you should preserve commit history following https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

fredericgrall:
@pedrobaeza,thank you, I will resume the migration following the procedure

Issue #2066, opened 2024-06-07
[14.0][BUG] stock_request_tier_validation: ValidationError dont check pending Tiers


kaynnan:
## Module
stock_request_tier_validation

## Describe the bug

The module in v14.0 and v15.0 does not have the correct behavior in tier.validation, it ignores pending reviews when only SRO Tier Validation is activated, in versions 12 and 13 the behavior displayed a Warning indicating that there are pending reviews.

## To Reproduce
**Affected versions**:

v14.0:

![image](https://github.com/OCA/stock-logistics-warehouse/assets/53870822/0e7e22b4-f918-4a73-a30e-b4946485d9bf)

1. Create new tier.definition for Stock Request Order (only active SRO)
2. Create new SRO
3. Save
4. Request for Validation
5. Confirm

**Expected behavior**

Warning: "A validation process is still open for at least."

**Additional context**

Runboat v12.0 (Work):

![Captura de imagem_20240607_190802](https://github.com/OCA/stock-logistics-warehouse/assets/53870822/2760c5c4-f404-484d-91f8-5daf2d8fe136)

Runboat v13.0 (Work):

![Captura de imagem_20240607_191340](https://github.com/OCA/stock-logistics-warehouse/assets/53870822/4d9de9f5-d1bb-4195-b580-6babbd28cf5c)

cc @douglascstd @marcelsavegnago

kaynnan:
I conducted some tests with the stock_request module and noticed that after implementing this change to obtain the status, which prevents stock.request.order from remaining in the open state when it has already been completed [[14.0][IMP] stock_request: Convert the state field of stock.request.order as compute store to complete all use cases](https://github.com/OCA/stock-logistics-warehouse/pull/1822/), the tier.validation of stock.request.order began to ignore pending reviews. This causes a significant issue where completed SROs or those needing to revert to draft status cannot perform this action if there are pending reviews.

Additionally, I performed tests using the previous commit [[14.0][FIX] stock_request: set stock request order state to done when partially delivered](https://github.com/OCA/stock-logistics-warehouse/pull/1741), and the tier.validation for stock.request.order functioned individually as expected.

kaynnan:
cc @victoralmau @pedrobaeza

pedrobaeza:
Sorry, but that module is not ours and we are not even using it.

kaynnan:
cc @HviorForgeFlow

kaynnan:
cc @ForgeFlow

PR #2095, opened 2024-07-02
[14.0][FIX] stock_request: write method to update state in _compute_state 

kaynnan:
cc @marcelsavegnago @ForgeFlow @victoralmau @pedrobaeza

As described in issue [#2066](https://github.com/OCA/stock-logistics-warehouse/issues/2066), after implementing the compute_state method in stock_request, a conflict arose with stock_request_tier_validation. When an SRO had tier validation active and pending reviews, it was confirmed without considering the tier validation status.

Upon analyzing the code, I opted to use item.write to set the status instead of simply assigning it with =. Following this change, the behavior aligned with expectations, where each SR and SRO operates individually according to their tier validation. Prior to this, if an SR was confirmed with an SRO having pending reviews, the action was executed while ignoring the pending reviews.

### DIFF
--- a/stock_request/models/stock_request_order.py
+++ b/stock_request/models/stock_request_order.py
@@ -151,13 +151,13 @@ class StockRequestOrder(models.Model):
         for item in self:
             states = item.stock_request_ids.mapped("state")
             if not item.stock_request_ids or all(x == "draft" for x in states):
-                item.state = "draft"
+                item.write({"state": "draft"})
             elif all(x == "cancel" for x in states):
-                item.state = "cancel"
+                item.write({"state": "cancel"})
             elif all(x in ("done", "cancel") for x in states):
-                item.state = "done"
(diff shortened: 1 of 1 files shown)


WesleyOliveira98 (approved):
LGTM

marcelsavegnago (approved):

kaynnan:
ping @victoralmau @pedrobaeza

pedrobaeza (changes requested):
This is incorrect. A compute method must never write nor create any data. You can use dot notation or `update` method, but never `write`.

kaynnan:
> This is incorrect. A compute method must never write nor create any data. You can use dot notation or `update` method, but never `write`.

For now I'll leave this pull request in draft while I search for solutions to resolve the compute method conflict with tier.validation, because it's ignoring validation based on the status of SRO when using update or dot notation.

PR #2097, opened 2024-07-04
[17.0][MIG] product route profile: Migration to 17.0

StefanRijnhart:
Based on unmerged https://github.com/OCA/stock-logistics-warehouse/pull/1775
Includes https://github.com/OCA/stock-logistics-warehouse/pull/1939


bosd (approved):
LGTM

jbaudoux (commented):
I would drop the binary files from the commit that introduced them

StefanRijnhart:
@jbaudoux That would be the best but it's not possible without a hard reset of the target branch, because the files were added in a commit related to another module. It was discussed in https://odoo-community.org/groups/contributors-15/contributors-1946820, and it was agreed that the cons would outweigh the pros.

rousseldenis:
> @jbaudoux That would be the best but it's not possible without a hard reset of the target branch, because the files were added in a commit related to another module. It was discussed in https://odoo-community.org/groups/contributors-15/contributors-1946820, and it was agreed that the cons would outweigh the pros.

In fact, in this case, base 17.0 branch does not contain those files. So, please remove them.

StefanRijnhart:
@rousseldenis In that case, the files have been removed in another PR. I rebased onto 17.0 so that the removal of the files is no longer visible in the diff.

jbaudoux (approved):

PR #2098, opened 2024-07-08
[FIX] stock_mts_mto_rule: multiple procs same product

ferran-S73:
Reopen of https://github.com/OCA/stock-logistics-warehouse/pull/1872

When launching 2 mts+mto procurements for the same product, if the first one has a greater o equal qty than the second, the second won't run it's mto action. This is because the replenishment move gets confirmed whereas the original moves are still in drat state.

Steps to reproduce the issue:
  - Configure a product with the mts+mto rule and some other rule (in my case it was a manufacturing rule)
  - Launch it's rule twice at the same time (i.e. confirming a sales order with two lines of the same product) without having any stock for the product. First line with 10units, second with 5.
  - Only the procurement for the first line will apply it's mto rule (in my case it only creates one manufacturing order with 10 units).

Debugging the code I found out that when the `_run_split_procurement` method calls `get_mto_qty_to_order` for the second procurement the product `virtual_available` includes the qty that from the manufacturing order created by the first proc (it's move is confirmed) but isn't taking into account the sale's move as it is still in draft. So I think we can take this into account by keeping track of what's being replenished for each product.

### DIFF
--- a/stock_mts_mto_rule/models/stock_rule.py
+++ b/stock_mts_mto_rule/models/stock_rule.py
@@ -36,14 +36,16 @@ class StockRule(models.Model):
                     ) % (rule.name,)
                     raise ValidationError(msg)
 
-    def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
+    def get_mto_qty_to_order(
+        self, product, product_qty, product_uom, values, done_proc_qty=0.0
+    ):
         self.ensure_one()
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
+        virtual_available = product_location.virtual_available - done_proc_qty
         qty_available = product.uom_id._compute_quantity(virtual_available, product_uom)
         if float_compare(qty_available, 0.0, precision_digits=precision) > 0:
             if (
@@ -59,6 +61,7 @@ class StockRule(models.Model):
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
+        done_procurements_by_product = {}
         for procurement, rule in procurements:
             domain = self.env["procurement.group"]._get_moves_to_assign_domain(
                 procurement.company_id.id
@@ -68,6 +71,9 @@ class StockRule(models.Model):
                 procurement.product_qty,
                 procurement.product_uom,
                 procurement.values,
+                done_proc_qty=done_procurements_by_product.get(
+                    procurement.product_id.id, 0.0
+                ),
             )
             if float_is_zero(needed_qty, precision_digits=precision):
                 getattr(self.env["stock.rule"], "_run_%s" % rule.mts_rule_id.action)(
@@ -104,4 +110,8 @@ class StockRule(models.Model):
                 getattr(self.env["stock.rule"], "_run_%s" % rule.mto_rule_id.action)(
                     [(mto_procurement, rule.mto_rule_id)]
                 )
+            done_procurements_by_product.setdefault(procurement.product_id.id, 0)
+            done_procurements_by_product[
+                procurement.product_id.id
+            ] += procurement.product_qty
         return True
--- a/stock_mts_mto_rule/tests/test_mto_mts_route.py
+++ b/stock_mts_mto_rule/tests/test_mto_mts_route.py
@@ -213,6 +213,50 @@ class TestMtoMtsRoute(TransactionCase):
         self.warehouse.name = new_warehouse_name
         self.assertEqual(new_rule_name, self.warehouse.mts_mto_rule_id.name)
 
+    def test_multiple_procurements(self):
+        mto_mts_route = self.env.ref("stock_mts_mto_rule.route_mto_mts")
+        self.product.route_ids = [(6, 0, [mto_mts_route.id])]
+        self._create_quant(1.0)
+        group2 = self.env["procurement.group"].create({"name": "test_2"})
+        proc_vals2 = {"warehouse_id": self.warehouse, "group_id": group2}
+        self.env["procurement.group"].run(
+            [
+                self.group.Procurement(
+                    self.product,
+                    2.0,
+                    self.uom,
+                    self.customer_loc,
+                    self.product.name,
+                    "test",
+                    self.warehouse.company_id,
+                    self.procurement_vals,
+                ),
+                group2.Procurement(
+                    self.product,
+                    2.0,
+                    self.uom,
+                    self.customer_loc,
+                    self.product.name,
+                    "test",
+                    self.warehouse.company_id,
+                    proc_vals2,
+                ),
+            ]
+        )
+        moves = self.env["stock.move"].search([("group_id", "=", self.group.id)])
+        self.assertEqual(3, len(moves))
+        moves2 = self.env["stock.move"].search([("group_id", "=", group2.id)])
+        self.assertEqual(2, len(moves2))
+        move_mto = self.env["stock.move"].search(
+            [
+                ("group_id", "=", group2.id),
+                ("location_dest_id", "=", self.customer_loc.id),
+                ("procure_method", "=", "make_to_order"),
+            ]
+        )
+        self.assertEqual(1, len(move_mto))
+        self.assertEqual(2.0, move_mto.product_uom_qty)
+
     def setUp(self):
         super().setUp()
         self.move_obj = self.env["stock.move"]


rousseldenis (approved):
Code review. Seems great

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -104,4 +112,8 @@ def _run_split_procurement(self, procurements):
                 getattr(self.env["stock.rule"], "_run_%s" % rule.mto_rule_id.action)(
                     [(mto_procurement, rule.mto_rule_id)]
                 )
+            done_procurements_by_product.setdefault(procurement.product_id.id, 0)
+            done_procurements_by_product[
+                procurement.product_id.id
+            ] += procurement.product_qty
```
jbaudoux, line 116: I would have removed only the mts quantity. If you have 7 in stock and you decide to route the need of 10 to mto, then you still have the 7 in stock to take the decision for routing the 5

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -68,6 +75,7 @@ def _run_split_procurement(self, procurements):
                 procurement.product_qty,
                 procurement.product_uom,
                 procurement.values,
+                done_procs=done_procurements_by_product,
```
jbaudoux: I would have passed the pre_consumed_qty as `done_procs.get(product.id, 0.0)` as this is what the method needs.

(minor remark)

jbaudoux (changes requested):

jbaudoux:
FYI @TDu @mt-software-de

stock_mts_mto_rule/models/stock_rule.py
TDu: Good point, this needs to be fixed.
@ferran-S73 Could you attend this comment ?

rousseldenis:
@ferran-S73

stock_mts_mto_rule/models/stock_rule.py
ferran-S73: @jbaudoux I don't think I fully understand what you're trying to say. I think the case you're talking about is already tested in the test code I wrote.
If we have 1 unit in stock and 2 procurements for 2 units each, we need to resupply a total of 3 units. If we only take into account mts quantity for the second procurement it will only resupply 1 unit, thus only making a total of 2 units.

jbaudoux: @ferran-S73 If you have 1 unit in stock, a procurement of 2 units and a procurement of 1 unit, and you don't want to split between mto & mts, then the first procurement of 2 will be routed to mto, and the used mts quantity is 0. What is still available for the next procurement is 1 - 0 and not 1 - 2.
This can happen if you enhance `get_mto_qty_to_order` to take more advanced decision. After the call to that method, the math should be correct.

ferran-S73: @jbaudoux maybe I'm not so used to this module but I don't know how you could decide to not split between mto & mts. As far as I'm concerned the expected behavior would always be for the first procurement to take the available mts quantity.

jbaudoux: @ferran-S73 A use case is when MTO is dropshipping and you prefer to dropship the total quantity if you don't have enough in stock

ferran-S73: @jbaudoux I understand that, buy I don't see how would you do it using this module since (correct me if I'm wrong) it doesn't allow you to choose if you wish to use the on hand quantity or not

ferran-S73:
@jbaudoux @rousseldenis is there anything else you think needs to be done here?

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -36,14 +36,16 @@ def _check_mts_mto_rule(self):
                     ) % (rule.name,)
                     raise ValidationError(msg)
 
-    def get_mto_qty_to_order(self, product, product_qty, product_uom, values):
+    def get_mto_qty_to_order(
+        self, product, product_qty, product_uom, values, done_proc_qty=0.0
```
jbaudoux, line 40:
```suggestion
        self, product, product_qty, product_uom, values, qty_unavailable=0.0
```

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -36,14 +36,16 @@ def _check_mts_mto_rule(self):
...
+    ):
         self.ensure_one()
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
+        virtual_available = product_location.virtual_available - done_proc_qty
```
jbaudoux, line 48:
```suggestion
        virtual_available = product_location.virtual_available - qty_unavailable
```

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -59,6 +61,7 @@ def _run_split_procurement(self, procurements):
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
+        done_procurements_by_product = {}
```
jbaudoux, line 64:
```suggestion
        qty_unavailable_by_product = {}
```

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -68,6 +71,9 @@ def _run_split_procurement(self, procurements):
                 procurement.product_qty,
                 procurement.product_uom,
                 procurement.values,
+                done_proc_qty=done_procurements_by_product.get(
+                    procurement.product_id.id, 0.0
+                ),
```
jbaudoux, line 76:
```suggestion
                qty_unavailable=qty_unavailable_by_product.get(
                    procurement.product_id.id, 0.0
                ),
```

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -104,4 +110,8 @@ def _run_split_procurement(self, procurements):
                 getattr(self.env["stock.rule"], "_run_%s" % rule.mto_rule_id.action)(
                     [(mto_procurement, rule.mto_rule_id)]
                 )
+            done_procurements_by_product.setdefault(procurement.product_id.id, 0)
+            done_procurements_by_product[
+                procurement.product_id.id
+            ] += procurement.product_qty
```
jbaudoux, line 116:
```suggestion
            if float_is_zero(needed_qty, precision_digits=precision):
                ...
                qty_unavailable_by_product.setdefault(procurement.product_id.id, 0)
                qty_unavailable_by_product[
                    procurement.product_id.id
                ] += procurement.product_qty
           elif:
              ...
              # here don't increase as it's mto
           else:  
               ... 
               qty_unavailable_by_product.setdefault(procurement.product_id.id, 0)
               qty_unavailable_by_product[
                    procurement.product_id.id
               ] += mts_qty
```

jbaudoux (changes requested):
You can fix it like that

stock_mts_mto_rule/models/stock_rule.py
ferran-S73: @jbaudoux this isn't ok, I think you're not getting the point of this PR. I'll try to explain the best I can step by step;

When two procurements are run for the same product, without any available stock, first procurement (proc A) creates a move replenishing the needed quantity.
When the second iteration is run with the second procurement (proc B), it finds the replenishment by created by proc A as an incoming move but does not find the original move that generated said procurement. Thus taking the replenished quantity as available.
That is why we need to take that quantity as unavailable even though it's mto.

Let me know if my explanation is understandable

jbaudoux: Hello @ferran-S73 
If there is no stock, both procurements will generate an mto move and this has always worked I think.
What you are fixing as I understand and according to your added test, if there is a mts quantity available and used by the first procurement, it cannot be available for the second procurement, that's why we must decrease the available qty by the mts qty used by previous procurements.

I just tried my suggestion and your test is successful

jbaudoux: looks like I can't make a pr on your repo but here is my fix https://github.com/OCA/stock-logistics-warehouse/compare/16.0...jbaudoux:stock-logistics-warehouse:16.0-fix-stock_mts_mto_rule2-fix

stock_mts_mto_rule/models/stock_rule.py
mt-software-de: I would also suggest to use @jbaudoux naming, because this more precise. 
But i would switch to a `defaultdict` this would ease the later calls. 
```suggestion
        done_procurements_by_product = defaultdict(float)
```

stock_mts_mto_rule/models/stock_rule.py
mt-software-de: with a use of defaultdict it would be just
```suggestion
                done_proc_qty=done_procurements_by_product[procurement.product_id.id],
```

stock_mts_mto_rule/models/stock_rule.py
mt-software-de: There is the method `get_mto_qty_to_order` which is returning the `needed_qty`.
the `needed_qty` represents the qty which you need to get by order. 
That means the following
`needed_qty == 0` you get all from stock
`needed_qty == proc.product_qty` you get all from order
in the other cases normally only `needed_qty < proc.product_qty`
You will get the `needed_qty` from order and the rest from stock.

So this means that the `proc.product_qty` isn't the right value to use.
You should only decrease the `virtual_available` by the qty which is taken from stock. 
And that is exactly what @jbaudoux  is saying.

stock_mts_mto_rule/models/stock_rule.py
mt-software-de, line 116: Here with defaultdict it would look like.
```suggestion
            done_procurements_by_product[
                procurement.product_id.id
            ] += procurement.product_qty
```

mt-software-de: This needs to be changed, like @jbaudoux said. I commented it on the other thread.

mt-software-de (changes requested):

GabbasovDinar:
@ferran-S73 Any news?

ferran-S73:
> @ferran-S73 Any news?

@GabbasovDinar not really, I still think my approach is what is needed in this case. We've been using these PR in ot clients for 2 years now without any issue. You can check the discussion in the [other pr](https://github.com/OCA/stock-logistics-warehouse/pull/2331) for more info as to why this may be the way to solve it.

Reyes4711-S73 (approved):
LGTM

Reyes4711-S73:
@rousseldenis @jbaudoux As @ferran-S73 said, we use this module in several customers without issues. Can you merge this PR?

ypapouin:
@ferran-S73, I was planning to add your fix to the `stock_mts_mto_rule` 18.0 PR but your test is not working:
`test_multiple_procurements` never FAIL with or without `qty_unavailable` FIX.
I would be nice to have a working test to ensure that the fix is working correctly.
Tested on 14.0, 16.0 and 18.0

ferran-S73:
> @ferran-S73, I was planning to add your fix to the `stock_mts_mto_rule` 18.0 PR but your test is not working: `test_multiple_procurements` never FAIL with or without `qty_unavailable` FIX. I would be nice to have a working test to ensure that the fix is working correctly. Tested on 14.0, 16.0 and 18.0

Hi @ypapouin thanks for pointing that out. There must be something off in the way I'm launching the procurement rules in test. Testing it functionally I found out that the problem was caused by one delivery move being still in draft state when confirming a sales order with 2 lines of the same product. However I haven't been able to get this case in my test as the moves are always confirmed one by one.. Any idea of how could I correctly reproduce the issue?

PR #2113, opened 2024-07-22
[14.0][FIX] stock_request_tier_validation: action_confirm validation after compute state in SRO

kaynnan:
cc @marcelsavegnago @douglascstd @WesleyOliveira98 @Matthwhy 

The applied fix addresses the issue where, when the action to confirm an SRO is performed, it is not being validated with its pending reviews. After the change in the method by which it obtains the computed status [[14.0][IMP] stock_request: Convert the state field of stock.request.order as compute store to complete all use cases](https://github.com/OCA/stock-logistics-warehouse/pull/1822/), it ignores the pending reviews. Therefore, the goal is to reinforce the validations of this action_confirm in the SR/SRO.


WesleyOliveira98 (approved):
LGTM

PR #2122, opened 2024-07-30
[14.0][IMP] stock_move_auto_assign_auto_release: Release release_ready moves together

mt-software-de:
A a transfer is waiting for more than product to release and the release policy is "one".
The whole transfer should be released, instead of only the last received one. 

Within this PR
Also releaseable siblings will be released.

### DIFF
--- a/stock_move_auto_assign_auto_release/models/product_product.py
+++ b/stock_move_auto_assign_auto_release/models/product_product.py
@@ -1,4 +1,5 @@
 # Copyright 2022 ACSONE SA/NV
+# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import models
@@ -14,16 +15,15 @@ class ProductProduct(models.Model):
             ("is_auto_release_allowed", "=", True),
         ]
 
-    def moves_auto_release(self):
-        """Job trying to auto release moves based on product
+    def pickings_auto_release(self):
+        """Job trying to auto release pickings based on product
 
-        It searches all* the moves auto releasable and trigger the release
-        available to promise process.
+        It searches all* the moves auto releasable
+        and triggers a delayed release available to promise for their pickings.
         """
         self.ensure_one()
         moves = self.env["stock.move"].search(self._moves_auto_release_domain())
         pickings = moves.picking_id
         if not pickings:
             return
-        self._lock_pickings_or_retry(pickings)
-        moves.release_available_to_promise()
+        pickings._delay_auto_release_available_to_promise()
--- a/stock_move_auto_assign_auto_release/tests/test_assign_auto_release.py
+++ b/stock_move_auto_assign_auto_release/tests/test_assign_auto_release.py
@@ -1,4 +1,5 @@
 # Copyright 2022 ACSONE SA/NV
+# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from datetime import datetime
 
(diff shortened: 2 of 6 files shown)


mt-software-de:
cc @lmignon @TDu @jbaudoux

lmignon (commented):
LGTM (Code review only)

jbaudoux (commented):
What about dropping this `moves_auto_release` method and in `_enqueue_auto_assign`, check that the `move.picking_id` is `is_auto_release_allowed`, and then use as delayable `_delay_auto_release_available_to_promise` (which is better placed in stock_move_auto_assign_auto_release than in stock_release_channel_auto_release)

jbaudoux:
> A a transfer is waiting for more than product to release and the release policy is "one". The whole transfer should be released, instead of only the last received one.

It would be worth adding a fix on `stock_available_to_promise_release` so that when `release_available_to_promise` is called on `stock.move` it raises an error. And on `stock.picking` `release_available_to_promise`, call `moves...._run_stock_rule` instead of `moves....release_available_to_promise`

mt-software-de:
> `stock_available_to_promise_release`

mt-software-de:
> > A a transfer is waiting for more than product to release and the release policy is "one". The whole transfer should be released, instead of only the last received one.
> 
> It would be worth adding a fix on `stock_available_to_promise_release` so that when `release_available_to_promise` is called on `stock.move` it raises an error. And on `stock.picking` `release_available_to_promise`, call `moves...._run_stock_rule` instead of `moves....release_available_to_promise`

So you mean a check if all moves of a picking are in the recordset if the policy is one?

mt-software-de:
> What about dropping this `moves_auto_release` method and in `_enqueue_auto_assign`, check that the `move.picking_id` is `is_auto_release_allowed`, and then use as delayable `_delay_auto_release_available_to_promise` (which is better placed in stock_move_auto_assign_auto_release than in stock_release_channel_auto_release)

Ok you mean moving `_delay_auto_release_available_to_promise` into stock_move_auto_assign_auto_release - yes this sounds good.
Needs to be merged before https://github.com/OCA/wms/pull/566

stock_move_auto_assign_auto_release/tests/test_assign_auto_release.py
```diff
@@ -166,3 +173,48 @@ def test_move_field_is_auto_release_allowed(self):
...
+            self.assertFalse(job)
+        with trap_jobs() as trap:
+            self._receive_product(self.product2, 100)
+            shipping.invalidate_cache()
+            shipping.move_lines.invalidate_cache()
+            trap.perform_enqueued_jobs()
+            job = self._get_job_for_method(
+                trap.enqueued_jobs, shipping.auto_release_available_to_promise
+            )
+            job.perform()
```
jbaudoux, line 218: This is failing
https://github.com/OCA/stock-logistics-warehouse/actions/runs/10195951314/job/28205684041?pr=2122#step:8:1118

mt-software-de: I don't know why yet, i will have a look.

mt-software-de: Locally it works. :-)

mt-software-de: Somehow the trap jobs wasn't working as expected with the current queue.job version. I fixed it

jbaudoux (approved):

stock_move_auto_assign_auto_release/models/product_product.py
```diff
@@ -14,16 +15,15 @@ def _moves_auto_release_domain(self):
...
 
-    def moves_auto_release(self):
-        """Job trying to auto release moves based on product
+    def pickings_auto_release(self):
+        """Job trying to auto release pickings based on product
 
-        It searches all* the moves auto releasable and trigger the release
-        available to promise process.
+        It searches all* the moves auto releasable
+        and trigger a delayed release available to promise for there pickings.
```
TDu:
```suggestion
        and trigger a delayed release available to promise for their pickings.
```

TDu (approved):
Maybe the readme could be updated also ?
LG otherwise.

stock_move_auto_assign_auto_release/models/product_product.py
jbaudoux: triggers

Issue #2125, opened 2024-08-01
[16.0] stock_mts_mto_rule: company-specific products crash upgrade of module


kevinkhao:
module affected: stock_mts_mto_rule

In an intercompany context, when a product is company-specific and uses mts+mto rule, upgrading the module raises this error:

> File "/opt/odoo/odoo/models.py", line 3464, in _check_company
>     raise UserError("\n".join(lines))
> odoo.exceptions.UserError: Incompatible companies on records:
> - 'Make To Order + Make To Stock' belongs to company False and 'Products' (product_ids: 'San Francisco company-specific product') belongs to another company.

**Steps to reproduce:**

1. Install this module
2. Use MTS+MTO rule on a company-specific product
3. Upgrade the module

**Expected behavior**
No exception on module upgrade

Temporary fix/workaround can be found here https://github.com/OCA/stock-logistics-warehouse/pull/2126

PR #2126, opened 2024-08-01
[16.0][FIX] stock_mts_mto_rule: make route noupdate

kevinkhao:
See https://github.com/OCA/stock-logistics-warehouse/issues/2125
This is just a quick workaround that keeps the initial rule, allows for persistent customization of it and prevents raise on module upgrade if the DB happens to have the wrong configuration case.


kevinkhao:
@florian-dacosta what do you think ?

jbaudoux (approved):

PR #2120, opened 2024-07-26
[15.0][FIX] stock_inventory: Resolve error when applying inventory with multiple adjustments in the same location for different products

JoanSForgeFlow:
[FIX] stock_inventory: Resolve error when applying inventory with multiple adjustments in the same location for different products

Error resolved:

```
  File "/home/joan/odoo/15.0/odoo/odoo/models.py", line 5243, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
Exception

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/joan/odoo/15.0/odoo/odoo/http.py", line 658, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/joan/odoo/15.0/odoo/odoo/http.py", line 301, in _handle_exception
    raise exception.with_traceback(None) from new_cause
ValueError: Expected singleton: stock.inventory(105, 104)
```

PR #2127, opened 2024-08-02
[16.0][FIX] stock_inventory: Expected singleton error when applying the inventory adjustments

AaronHForgeFlow:
Happens when you have more than one inventory adjustment in progress in the same location for different products

Create an inventory adjustment in stock and choose "Manual Selection" and a single product:

image

Put it in progress, and modify the counted quantity:
image

Create another inventory adjustment for another product in the same location and put it "In Progress" . Modify the counted quantity. Try to apply the inventory. Boom.

Solution is not doing a search by trusting in the stock_inventory_ids field filtering by the ones in progress and by location. Maybe we can use a sanity check, and if no adjustment is found show a user error or similar.

@ForgeFlow

### DIFF
--- a/stock_inventory/models/stock_quant.py
+++ b/stock_inventory/models/stock_quant.py
@@ -18,11 +18,10 @@ class StockQuant(models.Model):
         record_moves = self.env["stock.move.line"]
         adjustment = self.env["stock.inventory"].browse()
         for rec in self:
-            adjustment = (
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
Duplicated from https://github.com/OCA/stock-logistics-warehouse/pull/2120

PR #2131, opened 2024-08-05
[15.0][MIG] stock_inventory_verification_request

JoanSForgeFlow:
**Migration of the `stock_inventory_verification_request` to 15.0**

- **Main Model Introduced**: Slot Verification Request.
- **Threshold Handling**: When the threshold defined in inventory adjustments is exceeded, users cannot validate the applied quantity. (This is already addressed in `stock_inventory_discrepancy`.)
- **Validation Process**: The stock manager must handle the validation. A button will appear for the manager to create a Slot Verification Request, allowing the supervisor to review the issue later and ensure it is recorded.


stock_inventory_verification_request/models/stock_inventory.py
```diff
@@ -71,3 +63,20 @@ def action_open_svr(self):
...
+                    {
+                        "quant_id": quant.id,
+                        "location_id": quant.location_id.id,
+                        "state": "wait",
+                        "product_id": quant.product_id.id,
+                        "company_id": self.company_id.id,
+                        "lot_id": quant.lot_id.id if quant.lot_id else False,
+                    }
+                )
+                self.requested_verification = True
```
rousseldenis: `quant.requested_verification = True`

rousseldenis: And add : `return {"type": "ir.actions.act_window_close"}` which is more correct syntaxically.

rousseldenis (changes requested):
@JoanSForgeFlow No migration script for the `requested_verification` field ?

JoanSForgeFlow:
Hello @rousseldenis 

I think it doesn't make much sense to create a migration script for this, right? After all, it's just a field used to mark that a review is in progress.

But from 12.0 to 15.0, the adjustments flow have changed a lot technically. Previously, this field was in `stock.inventory`, and `stock.inventory `had `stock.inventory.line`s. Now, this field is in `stock.quant`. So, what would need to be done is to check which stock.inventory you have with requested_verification set to True in 12.0 and look at which lines you have with which products and where, and then make the equivalence to set those quants to True.

I think it's an effort that isn't worth it, right? Better before migrating closing the SVRs and start in 15.0 with all quants in requested=False.

Thank you

stock_inventory_verification_request/models/stock_inventory.py
ThiagoMForgeFlow: Thank you for yours comments, we have applied the changes.

rousseldenis (approved):
LGTM

stock_inventory_verification_request/security/stock_security.xml
```diff
@@ -7,7 +7,7 @@
             <field name="global" eval="True" />
             <field
             name="domain_force"
-        >['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
```
LoisRForgeFlow: ['|',('company_id','=',False),('company_id', 'in', company_ids)]

stock_inventory_verification_request/views/stock_slot_verification_request_view.xml
```diff
@@ -7,7 +7,7 @@
         <field name="name">stock.slot.verification.request.tree</field>
         <field name="model">stock.slot.verification.request</field>
         <field name="arch" type="xml">
-            <tree string="Stock Slot Verification Request">
```
LoisRForgeFlow: I think string is just deprecated in views, you can remove it

LoisRForgeFlow (approved):
LGTM

PR #2140, opened 2024-08-19
[16.0][FIX] stock_vlm_mgmt: location general inventory

chienandalu:
We were lacking the default location id so we could select the tray. It was also needed to avoid compting a pos when no tray is selected.

please review @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1680, opened 2023-03-29
[16.0] stock_picking_volume: Speedup install and avoid useless recompute

lmignon:

PR #2148, opened 2024-08-28
[14.0][BACKPORT] [16.0] stock_picking_volume: Speedup install and avoid useless recompute

sebalix:
Port of #1680 from 16.0 to 14.0.

SQL request in `pre_init_hook` has been adapted to 14.0 (see https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/stock/16.0.1.1/pre-migration.py#L17-L18 )


lmignon (approved):
LGTM (Code review only)

gbrito (approved):

PR #2149, opened 2024-08-28
[14.0][FIX] stock_picking_volume: pre_init_hook, replace column type from 'numeric' to 'double precision'

sebalix:
As no digits parameters are provided on `<stock.move>.volume` and `<stock.picking>.volume` fields, the 'double precision' type is used instead of 'numeric'.

This avoids time consuming SQL requests 'ALTER TABLE...' at module install.


sebalix:
Closing, including it there to avoid conflicts: https://github.com/OCA/stock-logistics-warehouse/pull/2148

PR #2152, opened 2024-09-02
[14.0][IMP] stock_request_purchase: Cancel stock request when unlinking related purchase order lines

carlos-lopez-tecnativa:
Before this commit, when a purchase order line was unlinked, the stock request remained open. After this commit, related stock requests are canceled.

@Tecnativa TT50615 

@victoralmau @pedrobaeza could you please take a look?

### DIFF
--- a/stock_request_purchase/models/purchase_order_line.py
+++ b/stock_request_purchase/models/purchase_order_line.py
@@ -12,6 +12,21 @@ class PurchaseOrderLine(models.Model):
         comodel_name="stock.request", string="Stock Requests", copy=False
     )
 
+    def unlink(self):
+        """
+        Cancel the stock.request
+        related to the purchase order line
+        because it does not occur automatically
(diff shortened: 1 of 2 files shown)


stock_request_purchase/models/purchase_order_line.py
```diff
@@ -12,6 +12,15 @@ class PurchaseOrderLine(models.Model):
         comodel_name="stock.request", string="Stock Requests", copy=False
     )
 
+    def unlink(self):
```
pedrobaeza, line 15: Include docstring about the general what, the whys, and so on.

carlos-lopez-tecnativa: done

victoralmau (approved):
Code and functional review OK.

IMO it's ready to merge.

pedrobaeza (approved):
/ocabot merge patch

To be fw-ported it to upper versions.

PR #2153, opened 2024-09-09
[14.0][ADD] stock_available_exclude_location: add new module

WesleyOliveira98:
Ticket: HT00837

This module allow defining excluded locations for product availability.

Select the excluded locations in Inventory Settings:
image

Now the quantity in these locations is excluded of product availability
image

Even when we have quantities of this product in the locations we marked to exclude
image

cc: @marcelsavegnago @kaynnan @douglascstd


douglascstd (commented):
[FUNCTIONAL REVIEW]

The test is conducted to check if the configuration is applied to the child locations; if so, those locations are excluded.

image
image
image

Is this the expected behavior? Wouldn't it be better to exclude all child locations from the selected location? Alternatively, could there be two fields:
-    Unique
-    Parent and Children

WesleyOliveira98:
> Wouldn't it be better to exclude all child locations from the selected location?

Makes sense, we add the module `stock_location_children` in depends and add all the children the excluded location have

kaynnan (approved):
LGTM

Code Review

stock_available_exclude_location/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+        string="Locations to Exclude as Available",
+        compute="_compute_excluded_location_ids",
+        readonly=True,
+        default=None,
+    )
+
+    @api.depends()
+    def _compute_excluded_location_ids(self):
+        for product in self:
+            stock_quant = self.env["stock.quant"].search(
```
rousseldenis: @WesleyOliveira98 This is very bulky as it will generate an sql query for each product in the loop.

Prefer:

```python
stock_quants = self.env["stock.quant"].search(
    [
                    ("product_id", "in", self.ids),
                    ("on_hand", "=", True),
                ]

```

before the loop and then filter the results in the loop.

stock_available_exclude_location/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+                warehouse_ids = location_ids.mapped(lambda x: x.warehouse_id)
+                excluded_ids = warehouse_ids.mapped(
+                    lambda x: x.stock_excluded_location_ids
+                )
+                if excluded_ids:
+                    for excluded_location in excluded_ids:
+                        excluded_ids |= excluded_location.children_ids
+                    product.excluded_location_ids = excluded_ids
+                else:
+                    product.excluded_location_ids = None
```
rousseldenis: To avoid a `set()` for each product in that case, in v14, prefer:

```python
self.update({"excluded_location_ids: False})
```

before the loop. You can then remove both `else`

rousseldenis:
This looks great.

@WesleyOliveira98 I'm just wondering if those excluded locations should be excluded at warehouse level, why not defining those locations outside the warehouse view location ?

WesleyOliveira98:
> I'm just wondering if those excluded locations should be excluded at warehouse level, why not defining those locations outside the warehouse view location ?

@rousseldenis We thought about defining it in the warehouse as it was the model that made the most sense for defining location rules, where do you suggest we place this configuration?

rousseldenis:
> > I'm just wondering if those excluded locations should be excluded at warehouse level, why not defining those locations outside the warehouse view location ?
> 
> @rousseldenis We thought about defining it in the warehouse as it was the model that made the most sense for defining location rules, where do you suggest we place this configuration?

No, I wanted to say, from a functional point of view about locations that should be excluded from available quantities.
If you define them outside the warehouse view location, they will be excluded from available quantities. But maybe you have a use case that does not correspond to that.

WesleyOliveira98:
@rousseldenis I change the configuration of the excluded locations to inventory settings based on the company:

image

Could you review it please?

cc: @marcelsavegnago @kaynnan @douglascstd

rousseldenis:
Ok, but you didn't answer my question :sweat_smile:

WesleyOliveira98:
> No, I wanted to say, from a functional point of view about locations that should be excluded from available quantities.
> If you define them outside the warehouse view location, they will be excluded from available quantities. But maybe you have a use case that does not correspond to that.

Sure, this way it is much easier to deal with this specific problem of not counting the quantities of a location, but we have some cases such as not counting a specific shelf in stock, or excluding pre-production and post -production when using MRP with 3 steps, and using the `stock_available_base_exclude_location` module we just need to define where it will be configured and pass this as a context to the product, that's why we are proposing this module, it will be very useful for some of our customers

WesleyOliveira98:
ping @OCA/stock-weighing-maintainers @rousseldenis @marcelsavegnago @kaynnan @douglascstd

stock_available_exclude_location/models/product_product.py
```diff
@@ -0,0 +1,45 @@
...
+                )
+                if excluded_ids:
+                    for excluded_location in excluded_ids:
+                        excluded_ids |= excluded_location.children_ids
+                    product.excluded_location_ids = excluded_ids
+
+    def _compute_quantities_dict(
+        self, lot_id, owner_id, package_id, from_date=False, to_date=False
+    ):
+        context = dict(self._context, excluded_location_ids=self.excluded_location_ids)
```
rousseldenis:
```suggestion
        context = dict(self.env.context, excluded_location_ids=self.excluded_location_ids)
```

WesleyOliveira98:
ping @OCA/stock-weighing-maintainers @rousseldenis @marcelsavegnago @kaynnan @douglascstd

kaynnan (approved):

douglascstd (approved):
[FUNCTIONAL REVIEW][APPROVED]

marcelsavegnago (approved):

PR #2156, opened 2024-09-18
[15.0] [MIG] stock_request_cancel_confirm: Migration to 15.0

bizzappdev:


rousseldenis:
@bizzappdev Thanks for this. Adding tests should be great!

stock_request_cancel_confirm/tests/test_stock_request.py
```diff
@@ -0,0 +1,116 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestStockRequest(TransactionCase):
+    def setUp(self):
```
rousseldenis: @bizzappdev Could you use classmethod instead ?

rousseldenis (approved):

PR #2172, opened 2024-10-08
[IMP] stock_exception: method exception

mav-adhoc:
This PR aims to add exceptions to the methods. Today,we can only apply exceptions by domain or by python code, but with his improve, also when running a method the exception can be validated, which helps users to make simpler and more precise exceptions in certain cases. 
Only two methods were added. action_confirm, which runs when a picking is created, and button_validate, which runs when a picking is validated. Both from stock.picking model


stock_exception/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Stock Exception",
     "summary": "Custom exceptions on stock picking",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
```
rousseldenis: don't include this change unless you have migration scripts. Oca bot will do it.

stock_exception/models/exception_rule.py
```diff
@@ -15,3 +15,10 @@ class ExceptionRule(models.Model):
         ],
         ondelete={"stock.picking": "cascade", "stock.move": "cascade"},
     )
+    method = fields.Selection(
+        selection_add=[
+            ("button_validate", "Validation"),
+            ("action_confirm", "Confirmation"),
+        ],
+        readonly=False,
```
rousseldenis, line 23: could you add ondelete ?

rousseldenis:
@mav-adhoc Could you describe a little what you try to fix here? Thanks

mav-adhoc:
@rousseldenis Done!

PR #1917, opened 2024-01-18
[14.0][IMP] stock_helper: Getting closest warehouse for record set

mt-software-de:
This PR extends the get closest warehouse logic,
now it is possible to get the closest warehouse for a record set.

@rousseldenis and @simahawk could you please review this PR.

PR #2176, opened 2024-10-15
[17.0][FW] stock_helper: multiple ports from 14.0

henrybackman:
Port of the following PRs from 14.0 to 17.0:
- #1466
- #1917

### DIFF
--- a/stock_helper/models/stock_location.py
+++ b/stock_helper/models/stock_location.py
@@ -55,7 +55,7 @@ class StockLocation(models.Model):
             .search([])
             .sorted(lambda w: w.view_location_id.parent_path, reverse=True)
         )
-        res = defaultdict(self.env["stock.warehouse"])
+        res = defaultdict(lambda: self.env["stock.warehouse"])
         for location in self.filtered("parent_path"):
             for warehouse in warehouses:
                 if location.parent_path.startswith(
--- a/stock_helper/models/stock_location.py
+++ b/stock_helper/models/stock_location.py
@@ -36,3 +36,26 @@ class StockLocation(models.Model):
             if rule.procure_method == procure_method:
                 return rule.location_src_id
             current_location = rule.location_src_id
+
+    def get_closest_warehouse(self):
+        """Returns closest warehouse for current location.
+
+        By default the get_warehouse (which is located in the odoo core module stock)
+        returns the warehouse via searching the view_location_id with parent_of.
+        If we have multiple warehouses
+        where a view_location_id is a parent of an other view_location_id
+        Then it depends on the sorting of the warehouses which this method returns
+
+        With this methods we will really get the closest warehouse of a location
+        """
+        self.ensure_one()
+        location_ids = [int(x) for x in self.parent_path.split("/") if x]
+        warehouses = (
+            self.env["stock.warehouse"]
+            .search([("view_location_id", "in", location_ids)])
+            .sorted(lambda w: w.view_location_id.parent_path, reverse=True)
+        )
+        for warehouse in warehouses:
+            if self.parent_path.startswith(warehouse.view_location_id.parent_path):
+                return warehouse
+        return warehouses.browse()
--- a/stock_helper/models/stock_location.py
+++ b/stock_helper/models/stock_location.py
@@ -2,6 +2,8 @@
 # Copyright 2022-2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
+from collections import defaultdict
+
 from odoo import models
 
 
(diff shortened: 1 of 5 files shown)


henrybackman:
The tests fail but I don't see how the failure is related to the changes done here

rousseldenis:
> The tests fail but I don't see how the failure is related to the changes done here

@henrybackman That's because in Odoo stock module, `get_warehouse`has been replaced by `warehouse_id`field.

stock_helper/models/stock_location.py
```diff
@@ -48,14 +48,44 @@ def get_closest_warehouse(self):
...
+                        break
+            res[location.id] = wh
+        return res
+
+    def get_closest_warehouse(self):
+        """Returns closest warehouse for current location."""
+        self.ensure_one()
+        location_and_warehouse = self._get_closest_warehouse()
+        warehouse = location_and_warehouse[self.id]
+        return warehouse or self.env["stock.warehouse"]
```
mmequignon:
```suggestion
        res = defaultdict(self.env["stock.warehouse"])
        for location in self.filtered("parent_path"):
            for warehouse in warehouses:
                if location.parent_path.startswith(
                    warehouse.view_location_id.parent_path
                ):
                    res[location.id] = warehouse
                    break
        return res

    def get_closest_warehouse(self):
        """Returns closest warehouse for current location."""
        self.ensure_one()
        return  self._get_closest_warehouse()[self.id]
```

Also, I kinda feel we can improve `_get_closest_warehouse` and drop the nested loop here.

stock_helper/tests/test_location_get_closest_warehouse.py
```diff
@@ -0,0 +1,33 @@
...
+
+class TestStockLocationGetClosestWarehouse(StockHelperCommonCase):
+    def test_get_closest_warehouse(self):
+        test_warehouse = self.env["stock.warehouse"].create(
+            {"name": "Test Warehouse", "code": "Test WH Code"}
+        )
+        test_warehouse.view_location_id.location_id = self.wh.lot_stock_id.id
+        location = test_warehouse.lot_stock_id
+
+        self.assertEqual(location.warehouse_id, self.wh)
```
henrybackman, line 16: This fails but I don't have enough context knowledge to say what it is testing and how to fix it.

When running the tests locally the `location.warehouse_id` ID increased after each run, so this way of testing doesn't seem stable.

First run: `AssertionError: stock.warehouse(4,) != stock.warehouse(1,)`
Second run: `AssertionError: stock.warehouse(5,) != stock.warehouse(1,)`
etc.

stock_helper/models/stock_location.py
henrybackman: Applied the recommended change, but didn't want to mess about with the function more as I would first need to get more familiar with the context

stock_helper/tests/test_location_get_closest_warehouse.py
rousseldenis: That's normal. The warehouse(1) is the one defined in demo data (from stock module). The warehouse `test_warehouse` is the one created in tests (and so the id is incrementing at each run).

That said, your test will never succeed as your structure looks like :

     [WAREHOUSE(1)]
                  ||
     [TEST WAREHOUSE]

As the parent of your view location of test warehouse is the stock location of the warehouse 1

henrybackman:
After looking into this more, it seems that the closest warehouse search has been implemented already in the `stock_location` `_compute_warehouse_id`. Would this make these PRs not worth porting? @mt-software-de @jbaudoux 

[https://github.com/odoo/odoo/blob/17.0/addons/stock/models/stock_location.py#L140](https://github.com/odoo/odoo/blob/17.0/addons/stock/models/stock_location.py#L140)
```
    @api.depends('warehouse_view_ids', 'location_id')
    def _compute_warehouse_id(self):
        warehouses = self.env['stock.warehouse'].search([('view_location_id', 'parent_of', self.ids)])
        warehouses = warehouses.sorted(lambda w: w.view_location_id.parent_path, reverse=True)
        view_by_wh = OrderedDict((wh.view_location_id.id, wh.id) for wh in warehouses)
        self.warehouse_id = False
        for loc in self:
            if not loc.parent_path:
                continue
            path = set(int(loc_id) for loc_id in loc.parent_path.split('/')[:-1])
            for view_location_id in view_by_wh:
                if view_location_id in path:
                    loc.warehouse_id = view_by_wh[view_location_id]
                    break
```

mt-software-de:
> After looking into this more, it seems that the closest warehouse search has been implemented already in the `stock_location` `_compute_warehouse_id`. Would this make these PRs not worth porting? @mt-software-de @jbaudoux
> 
> https://github.com/odoo/odoo/blob/17.0/addons/stock/models/stock_location.py#L140
> 
> ```
>     @api.depends('warehouse_view_ids', 'location_id')
>     def _compute_warehouse_id(self):
>         warehouses = self.env['stock.warehouse'].search([('view_location_id', 'parent_of', self.ids)])
>         warehouses = warehouses.sorted(lambda w: w.view_location_id.parent_path, reverse=True)
>         view_by_wh = OrderedDict((wh.view_location_id.id, wh.id) for wh in warehouses)
>         self.warehouse_id = False
>         for loc in self:
>             if not loc.parent_path:
>                 continue
>             path = set(int(loc_id) for loc_id in loc.parent_path.split('/')[:-1])
>             for view_location_id in view_by_wh:
>                 if view_location_id in path:
>                     loc.warehouse_id = view_by_wh[view_location_id]
>                     break
> ```

The PRs don't need a porting. It was fixed by this Odoo PR https://github.com/odoo/odoo/pull/118128

PR #2179, opened 2024-10-21
[14.0][FIX] stock_request: Remove route_ids fields from requests

victoralmau:
Remove `route_ids` field from requests

It is not correct to set the `route_ids` field with the product routes, you must be able to select any route (similar to what is shown in the product form view).

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT50610

### DIFF
--- a/stock_request/models/stock_request_abstract.py
+++ b/stock_request/models/stock_request_abstract.py
@@ -96,46 +96,13 @@ class StockRequest(models.AbstractModel):
     route_id = fields.Many2one(
         "stock.location.route",
         string="Route",
-        domain="[('id', 'in', route_ids)]",
+        domain=[("product_selectable", "=", True)],
         ondelete="restrict",
-    )
-
-    route_ids = fields.Many2many(
-        "stock.location.route",
-        string="Routes",
-        compute="_compute_route_ids",
-        readonly=True,
     )
 
     _sql_constraints = [
         ("name_uniq", "unique(name, company_id)", "Name must be unique")
     ]
-
-    @api.depends("product_id", "warehouse_id", "location_id")
-    def _compute_route_ids(self):
-        route_obj = self.env["stock.location.route"]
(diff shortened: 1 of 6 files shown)


carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge minor

rousseldenis:
Didn't see this. @JordiBForgeFlow @LoisRForgeFlow What do you think ? There is still  https://github.com/OCA/stock-logistics-warehouse/pull/2184

LoisRForgeFlow:
@rousseldenis Thanks for the pointer.

LoisRForgeFlow:
@victoralmau @pedrobaeza I do not agree at all with this change, you tag removing a field and changing functionality as a fix... I don't think this is correct. I have processes using routes in stock request that are not product selectable...

pedrobaeza:
Let's focus the discussion on the other opened PR.

hparfr:
So... this PR should be reverted ?

max3903:
Yes. How is the inventory user supposed to decide how to fulfill a stock request if there is no route?

Like the first comment says "if we must be able to select any route", the field should remain. Maybe with a different filter.

pedrobaeza:
The field is there, @max3903. What has been removed is the other supporting field with the "allowed routes". Now every route present in the product is able to be selected in the request.

LoisRForgeFlow:
> So... this PR should be reverted ?

From my PoV, yes, this one should be removed to keep things aligned. @pedrobaeza is it ok?

pedrobaeza:
OK if you want to revert it. We are already going to 16 and our patch has converted to custom.

LoisRForgeFlow:
Hi all, revert being done in https://github.com/OCA/stock-logistics-warehouse/pull/2198.

PR #2181, opened 2024-10-23
[16.0][ADD] stock_safe_scrap

rousseldenis:


rousseldenis:
@lmignon @phschmidt

stock_safe_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,70 @@
...
+            lambda a_scrap: dict(
+                {key: a_scrap[key] for key in self._scrap_origin_common_keys()}
+            )
+        ).items():
+            scrap.in_progress_picking_ids = move_lines.filtered(
+                lambda line: all(
+                    (line[move_key] == scrap.mapped(move_key))
+                    for move_key in self._scrap_origin_common_keys()
+                )
+            ).picking_id
```
lmignon, line 54: @rousseldenis Can you explain this code? It's difficult to understand it. A comment would be welcome.

lmignon (changes requested):

PR #2187, opened 2024-11-05
[15.0][IMP] stock_inventory_verification_request: compute involved lines and quant

JoanSForgeFlow:
If new moves are completed in that location for the specified product after confirming, they should automatically appear in the related lines.

 Otherwise, it breaks traceability and hinders easy understanding of the process

Extra commit: do not show related inventory adjustments button if created_inventory_count is zero
2nd Extra commit: show lot_id for stock.group_production_lot group

### DIFF
--- a/stock_inventory_verification_request/models/stock_slot_verification_request.py
+++ b/stock_inventory_verification_request/models/stock_slot_verification_request.py
@@ -100,6 +100,8 @@ class SlotVerificationRequest(models.Model):
         column1="slot_verification_request_id",
         column2="move_line_id",
         string="Involved Stock Moves",
+        compute="_compute_involved_move_lines",
+        store=False,
     )
     involved_move_line_count = fields.Integer(
         compute="_compute_involved_move_line_count"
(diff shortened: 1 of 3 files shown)


stock_inventory_verification_request/models/stock_slot_verification_request.py
```diff
@@ -120,6 +124,27 @@ def _compute_created_inventory_count(self):
     )
     created_inventory_count = fields.Integer(compute="_compute_created_inventory_count")
 
+    @api.depends("location_id", "product_id", "lot_id", "state")
+    def _compute_involved_move_lines(self):
+        for rec in self:
+            # Solo calcular si el estado es 'open'
```
AaronHForgeFlow: Comments better in English

AaronHForgeFlow (approved):
Code review

stock_inventory_verification_request/models/stock_slot_verification_request.py
JoanSForgeFlow: sorry I missed this personal comment. Petition attended.

Thank you @AaronHForgeFlow

JordiMForgeFlow (approved):
code review 👍🏼

stock_inventory_verification_request/models/stock_slot_verification_request.py
LoisRForgeFlow: I would go one step further, why do you need a comment that states what the code does just right into the next line? I'm not a big friend of comments that explain what code does, an useful comment would be one that explains **why** something is done. So in this case for me the comment can be removed.

JoanSForgeFlow: Comment attended

Thank you @LoisRForgeFlow

PR #2188, opened 2024-11-07
[16.0][ADD] stock_location_fill_state

rousseldenis:
This is different from https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_location_empty

As we don't compute stock amount.

Depends on:

- https://github.com/OCA/stock-logistics-warehouse/pull/2200


stock_location_occupancy/models/stock_location.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @api.depends("quant_ids.location_id")
+    def _compute_occupied(self):
+        """
+        Compute the 'occupied' field.
+        It is True when location does not contain any quant.
+        """
+        quants_result = self.env["stock.quant"].read_group(
+            [("location_id", "in", self.ids)], ["location_id"], ["location_id"]
+        )
```
jbaudoux: You can still have quants with 0 quantity. So that's not inline with the help `no product quantity`

jbaudoux (changes requested):
According to me, the right implementation is here 
https://github.com/OCA/wms/blob/16.0/stock_storage_type/models/stock_location.py#L319

I would have migrated the stock_location_empty and extracted the computation from the above code to compute `stock_amount` and `location_is_empty`. It's not the exact same computation as in 14.0 but it's the most correct one in my opinion and improving previous implementation. Then we don't end-up with 3 modules computing the same concept in 3 different ways. 

I prefer `empty` than `void`. And occupied state is like there is an hostile army ;)

Occupancy is also misleading as I thought initially it was about volume...

jbaudoux:
cc @p-tombez @TDu @sebalix

rousseldenis:
> According to me, the right implementation is here
> https://github.com/OCA/wms/blob/16.0/stock_storage_type/models/stock_location.py#L319

Yes, in fact, I had this in mind but wanted simpler implementation.

I would say, we should maybe transform the boolean state to a Selection one, with an 'Operation in Progress' (yellow color) there:

image

jbaudoux:
> I would say, we should maybe transform the boolean state to a Selection one, with an 'Operation in Progress' (yellow color)

* empty = green
* being filled = yellow
* filled = red
* being emptied = orange

rousseldenis:
> > I would say, we should maybe transform the boolean state to a Selection one, with an 'Operation in Progress' (yellow color)
> 
> * empty = green
> * being filled = yellow
> * filled = red
> * being emptied = orange

@jbaudoux 

`being emptied` and `being filled` : which one has priority on the other ?

That's why I proposed `operation in progress` as we can have both at the same time.

jbaudoux:
> `being emptied` and `being filled` : which one has priority on the other ?
> 

Being emptied is when there is stock but entirely reserved for move lines with qty done. Equals to current implementation of is empty in storage type module.

Being filled is when it is empty but there are move line with location as destination.

As one is when there is stock and the other when there is no stock, you cannot have both, it is exclusive

stock_location_fill_state/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Location Fill State",
+    "summary": """This module allows to identify the fill state of stock locations""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
jbaudoux:
```suggestion
    "author": "ACSONE SA/NV,BCIM,Odoo Community Association (OCA)",
```

stock_location_fill_state/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "depends": ["base_partition", "stock", "stock_location_pending_move"],
+    "maintainers": ["rousseldenis"],
```
jbaudoux:
```suggestion
    "maintainers": ["rousseldenis","jbaudoux"],
```

stock_location_fill_state/i18n/fr_BE.po
```diff
@@ -0,0 +1,72 @@
+# Translation of Odoo Server.
```
jbaudoux: name file `fr.po` instead of `fr_BE.po`

stock_location_fill_state/readme/CONTEXT.md
```diff
@@ -0,0 +1,3 @@
+Stock logistic managers need to know the void stock locations in order to
```
jbaudoux:
```suggestion
Stock logistic managers need to know the empty stock locations in order to
```

stock_location_fill_state/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Denis Roussel <denis.roussel@acsone.eu>
```
jbaudoux, line 1: And Guewen Baconnier for extracted code from `stock_storage_type`
```suggestion
- Denis Roussel <denis.roussel@acsone.eu>
- Guewen Baconnier

Design
~~~~~~
- Denis Roussel <denis.roussel@acsone.eu>
- Jacques-Etienne Baudoux <je@bcim.be>
```

stock_location_fill_state/models/stock_location.py
```diff
@@ -0,0 +1,77 @@
...
+        ):
+            location_id = group["location_id"][0]
+            qty_by_location[location_id] = group["quantity"]
+        records_by_state = defaultdict(lambda: self.browse())
+        for rec in self:
+            qty_in_location = qty_by_location.get(rec.id, 0.0)
+            out_by_location = out_qty_by_location.get(rec.id, 0.0)
+            if qty_in_location > 0.0:
+                records_by_state["filled"] |= rec
+            if qty_in_location <= 0.0:
```
jbaudoux:
```suggestion
            else:
```

stock_location_fill_state/models/stock_location.py
```diff
@@ -0,0 +1,77 @@
...
+                records_by_state["empty"] |= rec
+            if out_by_location and (qty_in_location - out_by_location <= 0):
+                records_by_state["being_emptied"] |= rec
+            if (qty_in_location <= 0) and (
+                rec.pending_in_move_ids or rec.pending_in_move_line_ids
+            ):
+                records_by_state["being_filled"] |= rec
```
jbaudoux: i would do those tests first and use `if elif elif else`, otherwise a record can be in multiple dicts

stock_location_fill_state/models/stock_location.py
```diff
@@ -0,0 +1,77 @@
...
+            groupby=["location_id"],
+        ):
+            location_id = group["location_id"][0]
+            out_qty_by_location[location_id] = group["qty_done"]
+        for group in self.env["stock.quant"].read_group(
+            location_domain, fields=["quantity:sum"], groupby=["location_id"]
+        ):
+            location_id = group["location_id"][0]
+            qty_by_location[location_id] = group["quantity"]
+        records_by_state = defaultdict(lambda: self.browse())
```
jbaudoux, line 84: What about making a list of ids and browsing only once instead of at each concat?

stock_location_fill_state/models/stock_location.py
```diff
@@ -0,0 +1,77 @@
...
+                records_by_state["empty"] |= rec
+            if out_by_location and (qty_in_location - out_by_location <= 0):
+                records_by_state["being_emptied"] |= rec
+            if (qty_in_location <= 0) and (
+                rec.pending_in_move_ids or rec.pending_in_move_line_ids
+            ):
+                records_by_state["being_filled"] |= rec
+        for state, records in records_by_state.items():
+            # Don't update if value is already set
+            records.filtered(
```
jbaudoux, line 102: And browse here
```suggestion
        for state, location_ids in records_by_state.items():
            # Don't update if value is already set
            self.browse(location_ids).filtered(
```

stock_location_fill_state/readme/USAGE.md
```diff
@@ -0,0 +1,3 @@
+- Go to Inventory > Configuration > Warehouse Management > Locations
+- In the list of locations, enable the column visibility 'Fill State'
+- You can filter the locations per 'Empty' or 'Filled' property.
```
jbaudoux:
```suggestion
- You can filter the locations per 'Empty', 'Filled', 'Being filled' or 'Being emptied' property.
```

stock_location_fill_state/i18n/fr_BE.po
```diff
@@ -0,0 +1,72 @@
...
+"        [filled] Filled location\n"
+"        [being_filled] The location is empty and an incoming move is in progress\n"
+"        [being_emptied] The location is filled and the outgoing move(s) will empty the location\n"
+"        "
+msgstr ""
+"\n"
+"        Ceci montre le statut de remplissage de l'emplacement.\n"
+"        Valeurs possibles:\n"
+"        [empty] Emplacement vide\n"
+"        [filled] Emplacement remplu\n"
```
jbaudoux: :)
```suggestion
"        [filled] Emplacement rempli\n"
```

stock_location_fill_state/models/stock_location.py
```diff
@@ -0,0 +1,77 @@
...
+        "pending_in_move_ids",
+        "pending_in_move_line_ids",
+    )
+    def _compute_fill_state(self):
+        """ """
+        location_domain = [("location_id", "in", self.ids)]
+        out_qty_by_location = {}
+        qty_by_location = {}
+        for group in self.env["stock.move.line"].read_group(
+            PENDING_MOVE_DOMAIN + location_domain,
```
jbaudoux: Better use `expression.AND`

stock_location_fill_state/models/stock_location.py
rousseldenis: It was non performant in v14, but in v16, it's equivalent

stock_location_fill_state/models/stock_location.py
rousseldenis: Yes, that was the last part I needed to refactor from `stock_storage_type`

rousseldenis:
@jbaudoux Comments attended

lmignon (approved):

jbaudoux (approved):

PR #2189, opened 2024-11-07
[16.0] [ADD] stock_inventory_security

ivantodorovich:
In standard, the inventory adjustments can only be applied by **Inventory / Manager** users.

This module introduces a new security group named **Stock: Inventory Adjustments**, which
grants regular stock users the ability to apply inventory adjustments.


stock_inventory_security/demo/demo.xml
Camille0907, line 1: No need to add it to a "demo" category in the manifest ?

yankinmax (approved):
Very nice and well tested, thanks @ivantodorovich

rousseldenis:
> In standard, the inventory adjustments can only be applied by **Inventory / Manager** users.
> 
> This module introduces a new security group named **Stock: Inventory Adjustments**, which grants regular stock users the ability to apply inventory adjustments.

@ivantodorovich 

I suppose you are aware the Odoo standard use case is:

- The inventory users can count the stock levels.
- Then, the manager confirm the counted quantity.

stock_inventory_security/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+        # the Inventory Manager group.
+        # OVERRIDE: Hijack the check to replace it with our own group.
+        if groups == "stock.group_stock_manager" and self.env.context.get(
+            "_stock_inventory_security"
+        ):
+            groups = "stock_inventory_security.group_inventory_adjustment"
+        return super().user_has_groups(groups)
+
+    def action_open_quants(self):
+        if self.user_has_groups("stock_inventory_security.group_inventory_adjustment"):
```
Camille0907, line 21: Why do we need to test if the user has the group before adding the context ? 
In any case, it will be tested in the `user_has_groups` method, no ? 
(Same question in the other files)

stock_inventory_security/tests/test_stock_inventory_security.py
```diff
@@ -0,0 +1,156 @@
...
+    @users("stock")
+    def test_stock_user_quant_action(self):
+        """Test that the stock user does not get into inventory mode from quants"""
+        res = self.product_quants.with_user(self.env.user).action_view_inventory()
+        self.assertTrue(res["context"].get("search_default_my_count"))
+
+    @users("inventory", "admin")
+    def test_inventory_user_lot_action(self):
+        """Test that the inventory user gets into inventory mode from lots"""
+        res = self.product_lot.with_user(self.env.user).action_lot_open_quants()
```
Camille0907, line 82: Why do you need `search_default_my_count` here and above, whereas you don't need it in the tests below ? I don't see the difference

Camille0907 (approved):
Technically LGTM, I just have a few questions

stock_inventory_security/models/product_product.py
ivantodorovich: It's to avoid bypassing the groups if it's not needed

ivantodorovich:
> > In standard, the inventory adjustments can only be applied by **Inventory / Manager** users.
> > This module introduces a new security group named **Stock: Inventory Adjustments**, which grants regular stock users the ability to apply inventory adjustments.
> 
> @ivantodorovich
> 
> I suppose you are aware the Odoo standard use case is:
> 
> * The inventory users can count the stock levels.
> * Then, the manager confirm the counted quantity.

Hello @rousseldenis ,

Indeed. However here we need to grant non-manager users the ability to apply inventories

stock_inventory_security/tests/test_stock_inventory_security.py
ivantodorovich: We're essentially testing this:
https://github.com/odoo/odoo/blob/534220ee90da3e213ea99ec5da7ce6cc9eff4203/addons/stock/models/stock_quant.py#L400-L401

stock_inventory_security/demo/demo.xml
ivantodorovich: Indeed, fixed!

stock_inventory_security/tests/test_stock_inventory_security.py
Camille0907: Sorry, I've just realized that there were a copy/paste issue in my comment : my question was not about `search_default_my_count` but about `with_user(self.env.user)` -> So, when do you need the `with_user`, when don't you ?

ivantodorovich: The `@users` decorator will set the user in `self.env`.

However, recordsets that were already initialized during `setUpClass`, referenced in variables like `self.product_lot`, have their own `env` set, which isn't modified by the decorator.

So, we need to set the user when interacting with them

Camille0907: Okayy thank you for the explanation :+1:

PR #2193, opened 2024-11-12
[16][FIX] stock_secondary_unit: Error when user try to select a secondary uom without product in picking operations

sergio-teruel:
cc @Tecnativa  TT51683

Commit from 15 (bf39020eeb4fcc7b02f4be0cae0bffe238417f61)


pedrobaeza (approved):
Bugfix cherry-picking, so fast-tracking:

/ocabot merge patch

PR #2205, opened 2024-11-28
[16.0][ADD] stock_quant_task_deferred

rousseldenis:
@jbaudoux @grindtildeath @simahawk


jbaudoux:
Cc @sebalix

stock_quant_task_deferred/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,2 @@
+- Denis Roussel <denis.roussel@acsone.eu>
+- Jacques-Étienne Baudoux <je@bcim.be>
```
sebalix, line 2: Not sure if I should be listed as well? As I gave the idea to J-E (I implemented that for one of our customer).
EDIT: if not I don't really care, happy to see this consolidated in a module :)

sebalix (commented):
Question: why not using `stock.skip_quant_tasks` config parameter? (I guess there is a reason)

To not pollute the list of quants for users, we also added a default filter to display only those having quantity != 0. Should it be part of this module as well?

rousseldenis:
> we also added a default filter to display only those having quantity != 0

I hesitated about that, but could be a good idea

sebalix:
@rousseldenis any news about this module? Would be nice to have it migrated in 18.0 (can take care of that once you're done with 16.0)

PR #2207, opened 2024-11-28
[16.0][IMP] procurement_auto_create_group: create a procurement.group for each stock.picking instead of each stock.move

ghost:
This update ensures that procurement groups are created at the `stock.picking` level, rather than at the level of individual `stock.move` records. 

This adjustment improves the grouping logic by associating all stock moves within a single picking under the same procurement group. It prevents unnecessary fragmentation of procurement groups and simplifies the management of pickings and related operations.


aitoreizmendi (approved):
LGTM

rousseldenis:
@jbaudoux

LuisNovalio (approved):
Functional Review OK.

jbaudoux (changes requested):
You cannot do that. The purpose is to change the procurement group of the move instead of propagating any existing. It's done at move level to allow to create for example a picking per product to perform pick by product instead of pick by order

PR #2214, opened 2024-12-05
[18.0][MIG] stock_exception

mav-adhoc:


rousseldenis (changes requested):
@mav-adhoc It seems you took some commits from `stock_quant_manual_assign` module. Please update this.

PR #2027, opened 2024-05-10
[14.0][ADD]stock_package_autoload: new module

PicchiSeba:
This module allows to automatically add the content of a package to the transfer lines

PR #2218, opened 2024-12-11
[16.0][MIG] stock_package_autoload: Migration to 16.0

anusriNPS:
Migrating stock_package_autoload to 16.0
This module is migrated in 16.0 as this module is introduced as part of this PR in 14.0 :  https://github.com/OCA/stock-logistics-warehouse/pull/2027


stock_package_autoload/models/stock_move.py
```diff
@@ -0,0 +1,80 @@
...
+            [
+                (
+                    "id",
+                    "in",
+                    self.env["stock.quant.package"]
+                    .search([("quant_ids.product_id", "=", self.product_id.id)])
+                    .ids,
+                )
+            ]
+            if self.product_id
```
HekkiMelody: **question**:Would it be possible to avoid the search (line 28) and return

```py
[("quant_ids.product.id", "=", self.product_id.id ])
```
?

HekkiMelody (approved):
Code review, LGTM, one possible suggestion

stock_package_autoload/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "summary": """
+        Select a package to add its content to a transfer line
+    """,
+    "version": "16.0.1.0.0",
+    "author": "PyTech SRL, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "category": "Warehouse Management",
+    "depends": [
+        "stock",
+        "web_domain_field",
```
HekkiMelody: **issue**: this module is deprecated, the same can be achieved in base Odoo by using a Binary field. See: https://github.com/OCA/web/blob/16.0/web_domain_field/README.rst

HekkiMelody (changes requested):

stock_package_autoload/__manifest__.py
anusriNPS: Sorry on missing to notice the deprecation. Thank you for noticing it.

stock_package_autoload/models/stock_move.py
anusriNPS: Updated without search

stock_package_autoload/models/stock_move.py
```diff
@@ -0,0 +1,69 @@
...
+    def _compute_package_domain(self):
+        """
+        There's no need to compute this field if the current user doesn't have the
+        necessary group
+        """
+        if not self.env.user.has_group("stock.group_tracking_lot"):
+            self.write({"package_domain": "[]"})
+            return
+        for sm in self:
+            sm.package_domain = sm._package_domain()
```
HekkiMelody, line 36: **issue**: please keep the `json.dumps()` https://github.com/OCA/stock-logistics-warehouse/compare/edb1d646519c4e674a32693ca26c3dc85cf83d0f..15738ffc2951eba89528320cc84f8ff18a3e00c6#diff-166f2ecf2683b4ba59287e869e301a4d7bbb12e6a0ac826345a1addd6a11688bL38-L39

it's a best practice when passing data from the python backend to the JavaScript frontend to ensure that there are no issues with encoding, character escaping etc. (Most of the times it works fine even without it, but it's better to keep it, in case)

HekkiMelody (changes requested):

anusriNPS:
While trying to use json.dumps() on domain value, I was observing below frontend error where argument expects to be array:

image

   I thought due to json encoding the domain is converted to string and hence it is not required to use json.dumps() here. I also tried to check on other modules for similar implementation such as sale_purchase_force_vendor module under "purchase-workflow" which uses similar binary field and does not encode domain using json.(https://github.com/OCA/purchase-workflow/blob/16.0/sale_purchase_force_vendor/models/sale_order_line.py#L36 )

   As you have mentioned, it is best pratice to encode while passing to frontend, I am not sure whether my understanding regarding above error is right. I am not able to figure out other way to overcome it. Please let me know your suggestions on it.

HekkiMelody:
@anusriNPS You're probably correct that with the Binary field, it's not necessary to use `json.dumps`, I'm seeing several other implementations across OCA all omit it as well. It's probably fine as it is. Thank you for checking!

HekkiMelody (approved):
Code and functional review, LGTM

PR #2220, opened 2024-12-12
[17.0][FIX] stock_helper: fix tests based on modules installed

LoisRForgeFlow:
Since this change in v17
https://github.com/odoo/odoo/commit/439ca89a68fe957adbe0f7a6147047593155aa9f _get_rule fallbacks to location's warehouse to check for routes. This makes that the receipt route is found by _get_source_location_from_route. However, this is the case only with `stock` module installed, because as soon as `mrp` or `purchase` is installed that first step of the receipt route get removed in favor of manufacture or purchase routes. See https://github.com/odoo/odoo/commit/d25d320e0e870980971de945719f7b7d5deadd3c

cc @MeritxellAForgeFlow

### DIFF
--- a/stock_helper/tests/test_location_source_from_route.py
+++ b/stock_helper/tests/test_location_source_from_route.py
@@ -37,4 +37,18 @@ class TestStockLocationSoruceFromRoute(StockHelperCommonCase):
         source_location = location._get_source_location_from_route(
             route, "make_to_stock"
         )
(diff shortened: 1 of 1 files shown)


CarlosVForgeFlow (approved):
LGTM!

MeritxellAForgeFlow (approved):
Thank you Lois! LGTM:)

PR #2222, opened 2024-12-17
[14.0][IMP] stock_location_last_inventory_date: Make last_inventory_date storeable

mt-software-de:
I wanted to make this field searchable. At first i tried to create a search method for it, but it was not nice to implement, since the location of `stock.inventory` is used. Storing the date is easier. Therefor i changed the computation to use `stock.inventory.line` instead of `stock.inventory` , because this record contains a direct link to a location.

### DIFF
--- a/stock_location_last_inventory_date/__manifest__.py
+++ b/stock_location_last_inventory_date/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Stock Location Last Inventory Date",
     "summary": "Show the last inventory date for a leaf location",
-    "version": "14.0.1.0.0",
+    "version": "14.0.2.0.0",
     "development_status": "Alpha",
     "category": "Warehouse",
     "website": "https://github.com/OCA/stock-logistics-warehouse",
@@ -13,4 +13,5 @@
     "installable": True,
     "depends": ["product", "stock"],
     "data": ["views/stock_location_views.xml"],
+    "post_init_hook": "post_init_hook",
 }
--- a/stock_location_last_inventory_date/hooks.py
+++ b/stock_location_last_inventory_date/hooks.py
@@ -1,0 +1,36 @@
+# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def set_initial_last_inventory_date(cr):
+    cr.execute(
+        """
+    Update
+        stock_location
+    set
+        last_inventory_date = sub.date
+    from (
+        Select
+            line.location_id,
+            max(inventory.date) as date
+        from
+            stock_inventory_line as line
+        join
+            stock_inventory as inventory
+        on
+            inventory.id = line.inventory_id
+        group by
+            line.location_id
+    ) as sub
+    where
+        stock_location.id = sub.location_id;
+    """
+    )
+
+
+def post_init_hook(cr, registry):
+    logger.info("Calculate last inventory date for locations")
+    set_initial_last_inventory_date(cr)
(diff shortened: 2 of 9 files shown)


mt-software-de:
@hailangvn @sebalix @simahawk 
Could you maybe review this one, since you did the migration for it?
https://github.com/OCA/stock-logistics-warehouse/pull/1261

mt-software-de:
cc @i-vyshnevska @jbaudoux

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -9,19 +10,23 @@ class StockLocation(models.Model):
...
         for location in self:
-            location_ids = [
-                int(location_id)
-                for location_id in location.parent_path.rstrip("/").split("/")
+            last_inventory_date = False
+            last_inventory_dates = [
+                line.inventory_id.date
+                for line in location.inventory_line_ids
+                if line.state == "done"
             ]
```
vuwnevska: search method was more efficient

stock_location_last_inventory_date/models/stock_location.py
```diff
@@ -9,19 +10,23 @@ class StockLocation(models.Model):
...
     def _compute_last_inventory_date(self):
         for location in self:
-            location_ids = [
-                int(location_id)
-                for location_id in location.parent_path.rstrip("/").split("/")
+            last_inventory_date = False
+            last_inventory_dates = [
+                line.inventory_id.date
+                for line in location.inventory_line_ids
+                if line.state == "done"
```
jbaudoux: can move to fields definition `done_inventory_line_ids` as a domain

jbaudoux (changes requested):
You can make an inventory of a location without making any inventory line. I would have not changed that concept.
In newer versions, it's updating the date when the inventory is posted. You could go that way and make a migration script of the initialization of the field.

stock_location_last_inventory_date/models/stock_location.py
mt-software-de: The last_inventory_date will be now set on action_done.
But still i am using the locations set on `stock.inventory.line`.
You can also create a `stock.inventory` without setting location_ids. 
The lines should be created by `_action_start` 
(https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/stock/models/stock_inventory.py#L170).
If the lines aren't created, then they should be created manually.

This is then now the same as it is in newer versions. `stock.inventory.line` is removed and just `stock.quant` is used. 
In v14 the method `_get_inventory_lines_values`
(https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/stock/models/stock_inventory.py#L297)
is getting the data from `stock.quant`.

stock_location_last_inventory_date/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Stock Location Last Inventory Date",
     "summary": "Show the last inventory date for a leaf location",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
```
vuwnevska:
```suggestion
    "version": "14.0.1.1.0",
```

vuwnevska (approved):

stock_location_last_inventory_date/hooks.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo import SUPERUSER_ID, api
+from odoo.tools import groupby
+
+logger = logging.getLogger(__name__)
+
+
+def set_initial_last_inventory_date(env):
+    inventory_lines = env["stock.inventory.line"].search(
+        [("inventory_id.state", "=", "done")]
+    )
```
jbaudoux: This can load millions of lines. I think you better do it with one SQL query doing the update without loading anything in odoo cache

stock_location_last_inventory_date/__manifest__.py
jbaudoux: I would even put it as a major change
```suggestion
    "version": "14.0.2.0.0",
```

stock_location_last_inventory_date/hooks.py
mt-software-de: Yes of course you are right.

stock_location_last_inventory_date/__manifest__.py
mt-software-de: done

jbaudoux (approved):

mt-software-de:
can this be merged?

PR #2232, opened 2025-01-10
[16.0][IMP] stock_inventory_discrepancy: better inheritance

thomaspaulb:
During the migration to 13.0 of this module it was decided to introduce a hook.

Refactoring to a regular override plays better with inheritance.


thomaspaulb:
@HviorForgeFlow Any reason not to revert back to this more common way?

stock_inventory_discrepancy/models/stock_quant.py
```diff
@@ -90,3 +90,39 @@ def action_apply_inventory(self):
...
+            and not self.user_has_groups(
+                "stock_inventory_discrepancy.group_stock_inventory_validation_always"
+            )
+        ):
+            raise UserError(
+                _("Only a stock manager can validate an inventory adjustment.")
+            )
+        # Allow to write last_inventory_date on stock.location
+        self = self.sudo().with_context(from_apply_inventory=True)
+        return super(StockQuant, self)._apply_inventory()
```
rvalyi, line 133: this should be super()._apply_inventory()

rvalyi (dismissed):

stock_inventory_discrepancy/models/stock_quant.py
thomaspaulb: Would that use the new self?

HviorForgeFlow:
> @HviorForgeFlow Any reason not to revert back to this more common way?

No, I guess it was a decision without taking a count a possible inheritance. It makes more sense move it away from a hook.

stock_inventory_discrepancy/models/stock_quant.py
rvalyi: ah sorry, I didn't pay attention to this, you are right indeed.

thomaspaulb:
Now green

HviorForgeFlow (approved):

stock_inventory_discrepancy/models/stock_quant.py
yostashiro: I thought `super()._apply_inventory()` would be valid. It should use the new self if I'm not mistaken.

rousseldenis (approved):
Code review.

Yes, indeed, no need of patch in this case.

thomaspaulb:
@OCA/logistics-maintainers merge?

hparfr (approved):
/ocabot merge major

PR #2236, opened 2025-01-15
[17.0][PERF] stock_inventory: reduce amount of write operations

anikeenko-viktor:
Speed up `_apply_inventory` by writing all new moves at once to inventory. Almost 10 times faster for my case on average.

Before:
image

After (there's no bunch of writes anymore):
image

### DIFF
--- a/stock_inventory/models/stock_quant.py
+++ b/stock_inventory/models/stock_quant.py
@@ -1,4 +1,5 @@
 from odoo import _, api, fields, models
+from odoo.tools import groupby
 
 
 class StockQuant(models.Model):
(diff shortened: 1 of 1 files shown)


stock_inventory/models/stock_quant.py
```diff
@@ -22,39 +23,44 @@ def _apply_inventory(self):
...
-            )
-            rec.to_do = False
-            rec.current_inventory_id = False
+
+        for adjustment, quant_list in groupby(
+            self, key=lambda x: x.current_inventory_id
+        ):
+            new_moves = self.env["stock.move.line"]
+            for quant in quant_list:
+                moves = record_moves.search(
```
rousseldenis, line 32: IMHO, this should not be done in the loop.

Do it before (even before the groupby(). Then use the filter in the loop. That can increase performances massively

stock_inventory/models/stock_quant.py
```diff
@@ -22,39 +23,44 @@ def _apply_inventory(self):
...
+                    ],
+                    order="create_date asc",
+                ).filtered(
+                    lambda x, quant=quant: not x.company_id.id
+                    or not quant.company_id.id
+                    or quant.company_id.id == x.company_id.id
+                )
+                if len(moves) == 0:
+                    raise ValueError(_("No move lines have been created"))
+                move = moves[len(moves) - 1]
```
rousseldenis, line 48: this is more correct to get the last one:

```suggestion
                move = moves[-1]
```

stock_inventory/models/stock_quant.py
```diff
@@ -22,39 +23,44 @@ def _apply_inventory(self):
...
+                    reference = adjustment.name + ": " + move.reference
+                elif adjustment.name:
+                    reference = adjustment.name
+                move.write(
+                    {
+                        "inventory_adjustment_id": adjustment.id,
+                        "reference": reference,
+                    }
+                )
+                quant.to_do = False
```
rousseldenis, line 60:
```suggestion
                quant.update({
                    "to_do": False,
                    "current_inventory": False
                })
```

stock_inventory/models/stock_quant.py
```diff
@@ -22,39 +23,44 @@ def _apply_inventory(self):
...
-                    "reference": reference,
-                }
-            )
-            rec.to_do = False
-            rec.current_inventory_id = False
+
+        for adjustment, quant_list in groupby(
+            self, key=lambda x: x.current_inventory_id
+        ):
+            new_moves = self.env["stock.move.line"]
```
rousseldenis, line 30: This is more correct syntaxically:

```suggestion
            new_moves = self.env["stock.move.line"].browse()
```

rousseldenis (changes requested):

PR #2246, opened 2025-01-23
[18.0][MIG] stock_quant_cost_info: Migration to 18.0

juancarlosonate-tecnativa:
Migration to 18.0

@Tecnativa TT54352

@CarlosRoca13 @sergio-teruel can you check? thanks!


stock_quant_cost_info/hooks.py
CarlosRoca13, line 1: This is still necessary because if there were millions of records, the installation would take a long time since the field's value would have to be computed for each record. This way, we are pre-creating the column in the database with a value set to 0 for the quants already existing in the system.

CarlosRoca13 (changes requested):

stock_quant_cost_info/readme/INSTALL.md
CarlosRoca13, line 1: Keep this file

CarlosRoca13 (approved):
Minor comment

stock_quant_cost_info/readme/INSTALL.md
juancarlosonate-tecnativa: True, sorry..

stock_quant_cost_info/hooks.py
juancarlosonate-tecnativa: Done

sergio-teruel (approved):

PR #2255, opened 2025-02-04
[18.0][MIG] stock_vertical_lift: Migration to 18.0

LucasTran380381:
## Depend
- https://github.com/OCA/stock-logistics-warehouse/pull/2257
### Put from receipt
* create receipt to shuttle 
image
* Switch shuttle to put mode -> Insert barcode product and tray code -> Save to tray location
image
* Receipt change state to done
image
### Pick from Delivery
* Create Delivery
image
* Switch Shuttle to Pick mode -> Insert barcode destination -> Save
image
* Delivery change state to done
image
### Adjust Inventory
* Set stock.quant allow to adjust
image
* Input real quantity
image
* Inventory stock.quant has adjusted
image


stock_vertical_lift/views/shuttle_screen_templates.xml
```diff
@@ -0,0 +1,23 @@
+<odoo>
+    <template id="packagings">
+        <ul class="o_vlift_packaging list-unstyled">
+            <t t-foreach="product.packaging_ids" t-as="packaging">
+                <li>
+                    <span>1</span>
+                    <span
+                        class="packaging_name"
+                        itemprop="name"
+                        t-esc="packaging.name"
```
henrybackman: `t-esc` directives are deprecated
- https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

stock_vertical_lift/models/stock_location.py
```diff
@@ -0,0 +1,167 @@
...
+        recursive=True,
+        store=True,
+    )
+
+    # This is a one2one in practice, but this one is not really interesting.
+    # It's there only to be in the depends of 'vertical_lift_shuttle_id', which
+    # give the unique shuttle for any location in the tree (whether it's a
+    # shuttle, a tray or a cell)
+    inverse_vertical_lift_shuttle_ids = fields.One2many(
+        comodel_name="vertical.lift.shuttle", inverse_name="location_id", readonly=True
```
henrybackman: If possible avoid `readonly` fields
- https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

henrybackman (commented):
Looks good to me, just a couple of comments

stock_vertical_lift/models/vertical_lift_operation_inventory.py
```diff
@@ -0,0 +1,250 @@
...
+    tray_name = fields.Char(compute="_compute_tray_data")
+    tray_type_id = fields.Many2one(
+        comodel_name="stock.location.tray.type",
+        compute="_compute_tray_data",
+        string="Tray Type",
+    )
+    tray_type_code = fields.Char(compute="_compute_tray_data", string="Tray Code")
+    tray_x = fields.Integer(string="X", compute="_compute_tray_data")
+    tray_y = fields.Integer(string="Y", compute="_compute_tray_data")
+    tray_matrix = Serialized(string="Cells", compute="_compute_tray_data")
```
ivantodorovich: IMO we can just use the `Json` field and drop the `base_sparse_field` dependency

stock_vertical_lift/controllers/main.py
```diff
@@ -0,0 +1,23 @@
...
+from odoo import http
+from odoo.http import request
+
+_logger = logging.getLogger(__name__)
+
+
+class VerticalLiftController(http.Controller):
+    @http.route(["/vertical-lift"], type="http", auth="public", csrf=False)
+    def vertical_lift(self, answer, secret):
+        if secret == self._get_env_secret():
```
ivantodorovich, line 15: 🔴 This seems unsafe.

I know this change isn't introduced by the migration, but we could take the opportunity to improve this

I don't see this `VERTICAL_LIFT_SECRET` variable documented anywhere, and the default is just an unsecured endpoint

IMO we should check this secret is set, and fail with a message explaining how to set it.
If an unprotected (no secret) option wants to be given, IMO it should be explicit (e.g.: by setting the secret to `False`, for example)

Additionally, it'd be better to not rely on an environment variable for this secret. Some environments like Odoo.sh don't allow customizing these variables. Instead, we should read this from an `ir.config_parameter` instead. The config parameter itself can optionally be configured through environment with the usage of https://github.com/OCA/server-env/tree/18.0/server_environment_ir_config_parameter (no direct dependency needed)

stock_vertical_lift/models/stock_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+        "was being processed in the Vertical Lift, the operator decided to "
+        "skip its processing.",
+    )
+
+    def fetch_vertical_lift_tray_source(self):
+        self.ensure_one()
+        self.location_id.fetch_vertical_lift_tray()
+        # We reload mainly because otherwise, it would close
+        # the popup. This action is provided by the OCA module
+        # web_ir_actions_act_view_reload
```
ivantodorovich:
```suggestion
```

misleading/outdated comment. In fact `soft_reload` is a core feature now, that module is deprecated

stock_vertical_lift/controllers/main.py
LucasTran380381: improved to read secret by config_param

stock_vertical_lift/controllers/main.py
```diff
@@ -0,0 +1,35 @@
...
+        secret = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("stock_vertical_lift.secret", None)
+        )
+        if not secret:
+            _logger.error("secret is not set")
+            msg = self.env._(
+                "Vertical Lift secret not set. "
+                "Pls set secret in Inventory/Settings/Vertical Lift"
```
ivantodorovich:
```suggestion
                "Please set it in Inventory/Settings/Vertical Lift"
```

stock_vertical_lift/controllers/main.py
```diff
@@ -0,0 +1,35 @@
...
+            return Unauthorized()
+
+    def _get_env_secret(self):
+        secret = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("stock_vertical_lift.secret", None)
+        )
+        if not secret:
+            _logger.error("secret is not set")
```
ivantodorovich: Please a little bit more verbose with this message 🙏🏻 
Mention also how to configure it.
You can probably use the same message than the exception

ivantodorovich (approved):
Thanks, minor comments

stock_vertical_lift/controllers/main.py
LucasTran380381: I use the same message than the exception to make it more verbose. Thanks for suggestion

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -53,26 +51,19 @@ def extract_and_action_done(move):
...
             new_picking = picking.copy(
                 {
                     "name": "/",
-                    "move_lines": [],
+                    "move_ids": [],
                     "move_line_ids": [],
                     "backorder_id": picking.id,
                 }
             )
-            new_picking.message_post(
```
grindtildeath: Why removing this message?

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -351,7 +343,9 @@ def _send_notification_refresh(self):
             "action": "refresh",
             "params": self._get_user_notification_params(),
         }
-        self.env["bus.bus"].sendone(channel, bus_message)
+        self.env["bus.listener.mixin"]._bus_send(
```
grindtildeath: `self.env.user._bus_send(` :question: 

Looks weird to call the function on the mixin

stock_vertical_lift/models/vertical_lift_operation_inventory.py
```diff
@@ -142,44 +124,44 @@ def _compute_tray_data(self):
...
-            product = record.current_inventory_line_id.product_id
+            product = record.quant_id.product_id
             location = record.tray_location_id
             record.tray_qty = self._get_tray_qty(product, location)
 
     def _compute_number_of_ops(self):
         for record in self:
-            line_model = self.env["stock.inventory.line"]
+            line_model = self.env["stock.quant"]
             record.number_of_ops = line_model.search_count(
```
grindtildeath: I would suggest to rename to `quant_model` if it's not an inventory line anymore.

stock_vertical_lift/models/vertical_lift_operation_inventory.py
```diff
@@ -189,21 +171,21 @@ def _domain_inventory_lines_to_do_all(self):
...
+        stock_quant = self.quant_id
         return (
             float_compare(
-                line.theoretical_qty,
+                stock_quant.quantity,
                 self.quantity_input,
-                precision_rounding=line.product_uom_id.rounding,
+                precision_rounding=stock_quant.product_uom_id.rounding,
             )
             == 0
```
grindtildeath: You could use `float_is_zero` there

stock_vertical_lift/models/vertical_lift_operation_inventory.py
```diff
@@ -223,36 +205,34 @@ def clear_current_inventory_line(self):
...
-            line.vertical_lift_done = True
-            if self.quantity_input != line.product_qty:
-                line.product_qty = self.quantity_input
-            inventory = line.inventory_id
-            if all(line.vertical_lift_done for line in inventory.line_ids):
-                inventory.action_validate()
+        stock_quant = self.quant_id
+        if not stock_quant.vertical_lift_done:
+            stock_quant.vertical_lift_done = True
+            if self.quantity_input != stock_quant.inventory_quantity:
```
grindtildeath: Use `float_compare`

grindtildeath (approved):
Looks very good, thanks :+1: 

Couple small things to check, not blocking.

stock_vertical_lift/models/vertical_lift_operation_base.py
Kimkhoi3010: You can take a look [bus.py #Line102](https://github.com/odoo/odoo/blob/f60188fc9d2efc9e961fb4fecad800e529bee041/addons/bus/models/bus.py#L102)
This docstring explains a low-level method used to send notification_type and message to a specific target. It recommends using _bus_send() from bus.listener.mixin for better simplicity and security. When using _sendone() directly, it's important to ensure that the target (if it's a string) is not guessable by an attacker to prevent unauthorized access.

stock_vertical_lift/models/vertical_lift_operation_base.py
Kimkhoi3010: Because from Odoo 18.0, there is a message notifying user about creating a backorder, you see it [here](https://github.com/odoo/odoo/blob/2b109261bc6e7550b23a0d444a4e7c63496e4c55/addons/stock/models/stock_picking.py#L1582)

stock_vertical_lift/models/vertical_lift_operation_inventory.py
Kimkhoi3010: Hello @grindtildeath, I updated by your comment!

simahawk:
@Kimkhoi3010 can you drop the pending merge and rebase pls?

Kimkhoi3010:
> @Kimkhoi3010 can you drop the pending merge and rebase pls?

Hello @simahawk, I updated by your comment. Thanks for your time.

stock_vertical_lift/controllers/main.py
simahawk: FTR There's no need to translate a log message. Not blocking.

stock_vertical_lift/controllers/main.py
simahawk: FTR Even if it was switched to a config param, security is not improved at all. Plus, the use of sys param implies one step more for the configuration. No need to change it now but I'll have a look on how to improve this.

jbaudoux:
@LucasTran380381 @simahawk please next time, follow repository guidelines and put on the right repository

PR #2257, opened 2025-02-06
[18.0][MIG] stock_location_tray: Migration to 18.0

xaviedoanhduy:
1. **changes from version 14.0 to 18.0**

- remove `_get_id` and `_get_object_reference` that were one liner to
`_xmlid_lookup` in commit https://github.com/odoo/odoo/pull/69120/commits/15692f394.
- change the content of `UserError` in stock location in commit https://github.com/odoo/odoo/pull/139605/commits/2b738dbe. 
- change the .js file extension to .esm.js to avoid the pre-commit error `Parsing error: 'import' and 'export' may appear only with 'sourceType: module'`.
- move static files from `/static/js/` to `/static/js/backend/` to match odoo convention (see an example in: https://github.com/odoo/odoo/tree/18.0/addons/web_editor/static/src/js/backend).
- re-style, re-class some fields displayed on the backend form view to match changes from the current version.
- The button to view **detailed operation** has been changed to use js code instead of python so the context value (`show_source_location` and `show_destination_location`) will never be passed through the `action_show_details` function as before. This change fixes the issue that opens the replacement window we were talking about (video demonstration below), [ref](https://github.com/odoo/odoo/commit/4da8c6ebca024b31278a946aef55cc37f0210b33#diff-516cd63c49bc728d6348266542e792102aa1da763c5eab089350fd27a1ac7605R1-R15)
- Similarly the detailed operation view button has also been removed from the stock move view in the commit above, [ref](https://github.com/odoo/odoo/commit/4da8c6ebca024b31278a946aef55cc37f0210b33#diff-3bad372563263ef117de6ed9e222008e2a27b2ea212df89f72f269066a8eacacL440) 
- `location_id` is now hidden from list view and replaced with `stock quant`, see: https://github.com/odoo/odoo/pull/117513/commits/19f07f05fcc1ac46b2f37ab7b2414860170b2736

2. **from the above changes, I chose the following approach** 

- `show_source_location` and `show_destination_location` calculation approach (instead of passing context via the outdated action).
- change the widget appearance position to always be behind the `quant_id` and `location_dest_id` fields.
- and for `ROADMAP.md` seems to be fixed since the above commit (no more RPC call will not replace the current window).

3.  **video comparing 2 versions 14.0 (the issue that `ROADMAP.md` mentioned) and it seems to have disappeared since version 17.0:**

- [version_14.0.webm](https://github.com/user-attachments/assets/2c5cb373-2920-4013-ae92-9532eec59c80)

- [version_18.0.webm](https://github.com/user-attachments/assets/9d33943a-5ce9-46d7-802d-103b22d0456b)


henrybackman (approved):

stock_location_tray/models/stock_location.py
```diff
@@ -121,7 +125,7 @@ def write(self, vals):
             if not trays_to_update and "cell_name_format" in vals:
                 new_format = vals.get("cell_name_format")
                 trays_to_update = location.cell_name_format != new_format
-            super(StockLocation, location).write(vals)
+            super().write(vals)
```
ivantodorovich: This change is not correct.

Since this is called within a loop, the previous code was actually calling super only for the iterated record.

With this change, we'd be now calling super on the whole recordset, multiple times.

stock_location_tray/models/stock_location.py
```diff
@@ -228,9 +232,8 @@ def has_ref(xmlid):
             namespace, tray_name = tray_external_id.split(".")
             if module != namespace:
                 continue
-            tray_external = self.env["ir.model.data"].browse(
-                self.env["ir.model.data"]._get_id(module, tray_name)
-            )
+            xmlid = f"{module}.{tray_name}"
+            tray_external = ModelData.browse(ModelData._xmlid_lookup(xmlid)[1])
```
ivantodorovich: This is not the same than before.

`ModelData._xmlid_lookup(xmlid)[1]` here would return the `res_id` of the related model. Not the `id` of the `ir.model.data` record.

So we'd be mixing apples and oranges here.

IMO you can solve this with a simple search

```py
tray_external = self.env["ir.model.data"].search([("module", "=", module), ("name", "=", tray_name)], limit=1)
```

stock_location_tray/tests/test_stock_move_line.py
```diff
@@ -27,16 +27,16 @@ def create_stock_move(self):
...
                 "location_id": move.location_id.id,
                 "location_dest_id": move.location_dest_id.id,
             }
         )
-        self.move = move
+        return move
 
     def test_compute_tray_matrix(self):
-        self.create_stock_move()
+        self.move = self.create_stock_move()
```
ivantodorovich: You're on the right path here.
But we should also try to avoid setting `self.move` here, too

Instead use a local variable if possible

`move = self.create_stock_move()`

ivantodorovich (changes requested):
Thanks for the work done!

stock_location_tray/models/stock_location.py
xaviedoanhduy: hi @ivantodorovich, thanks for your suggestion. actually i was a bit confused about [this docstring](https://github.com/odoo/odoo/blob/2e9c546748fa9ff7e20abb1ec03ccb5e4d5a0616/odoo/addons/base/models/ir_model.py#L2231) which made me assume that this function still returns values ​​including `id`. but i updated it according to your suggestion and have prepared a PR on odoo side to update this `docstring`

stock_location_tray/models/stock_location.py
xaviedoanhduy: thanks, I fixed it

ivantodorovich (approved):
Thanks!

twalter-c2c (approved):

PR #2260, opened 2025-02-11
[14.0][FIX] stock_vertical_lift: put package multiple lines

TDu:
The workflow of this module moves one move line at at time even if the user scanned a package.

When the package is spread on multiple stock.move.line Odoo core will raise an error.

To avoid a refactor of the workflow, this fix tries to merge the move lines that are part of the same package


twalter-c2c (approved):
LGTM

simahawk (approved):

marcelsavegnago (approved):
LGTM

PR #2263, opened 2025-02-12
[18.0][MIG] account_move_line_product

StefanRijnhart:
* Ensure product_id is only in the list view once.
* Allow the user to hide the fields in the various views.


PieterPaulussen (approved):

jbaudoux (commented):
What has this module to do with stock?

StefanRijnhart:
@jbaudoux according to the original submitter (@JordiBForgeFlow): https://github.com/OCA/stock-logistics-warehouse/pull/231

jbaudoux (changes requested):
Let's now move this module to https://github.com/OCA/account-financial-reporting or https://github.com/OCA/account-financial-tools

StefanRijnhart:
@jbaudoux thanks for the suggestions. @JordiBForgeFlow do you have a preference?

PR #2268, opened 2025-02-20
[14.0][FIX] stock_reserve_rule: favor largest package when reserving full packaging

henrybackman:
Fixes issue where a smaller packaging is selected for picking even if another bin would have a larger one available.

Switched the inner and outer loop in `apply_strategy_packaging` to check the largest packaging size for all locations first before considering smaller sizes.

### DIFF
--- a/stock_reserve_rule/models/stock_reserve_rule.py
+++ b/stock_reserve_rule/models/stock_reserve_rule.py
@@ -284,19 +284,22 @@ class StockReserveRuleRemoval(models.Model):
         def is_greater_eq(value, other):
             return float_compare(value, other, precision_rounding=rounding) >= 0
 
-        for location, location_quants in quants_per_bin:
-            location_quantity = sum(location_quants.mapped("quantity")) - sum(
-                location_quants.mapped("reserved_quantity")
-            )
-            if location_quantity <= 0:
+        # first search for the largest packaging in all locations
+        for packaging_quantity in packaging_quantities:
+            # check if asked at least packaging quantity
+            if not is_greater_eq(need, packaging_quantity):
                 continue
+            for location, location_quants in quants_per_bin:
+                location_quantity = sum(location_quants.mapped("quantity")) - sum(
+                    location_quants.mapped("reserved_quantity")
+                )
+                if location_quantity <= 0:
+                    continue
 
-            for pack_quantity in packaging_quantities:
-                enough_for_packaging = is_greater_eq(location_quantity, pack_quantity)
-                asked_at_least_packaging_qty = is_greater_eq(need, pack_quantity)
-                if enough_for_packaging and asked_at_least_packaging_qty:
+                # check if quantity is enough for packaging
+                if is_greater_eq(location_quantity, packaging_quantity):
                     # compute how much packaging we can get
-                    take = (need // pack_quantity) * pack_quantity
+                    take = (need // packaging_quantity) * packaging_quantity
                     need = yield location, location_quantity, take, None, None
 
     def _apply_strategy_full_bin(self, quants):
--- a/stock_reserve_rule/models/stock_reserve_rule.py
+++ b/stock_reserve_rule/models/stock_reserve_rule.py
@@ -298,8 +298,12 @@ class StockReserveRuleRemoval(models.Model):
 
                 # check if quantity is enough for packaging
                 if is_greater_eq(location_quantity, packaging_quantity):
-                    # compute how much packaging we can get
-                    take = (need // packaging_quantity) * packaging_quantity
+                    # compute how many packagings we can get
+                    available_packagings = location_quantity // packaging_quantity
+                    need_packagings = need // packaging_quantity
+                    take = (
+                        min(available_packagings, need_packagings) * packaging_quantity
+                    )
                     need = yield location, location_quantity, take, None, None
 
     def _apply_strategy_full_bin(self, quants):
--- a/stock_reserve_rule/tests/test_reserve_rule.py
+++ b/stock_reserve_rule/tests/test_reserve_rule.py
@@ -826,6 +826,69 @@ class TestReserveRule(common.SavepointCase):
         )
         self.assertEqual(move.state, "assigned")
 
+    def test_rule_packaging_does_not_exceed_location_qty(self):
+        self._setup_packagings(
+            self.product1,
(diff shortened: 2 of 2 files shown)


stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -284,17 +284,20 @@ def _apply_strategy_packaging(self, quants):
         def is_greater_eq(value, other):
             return float_compare(value, other, precision_rounding=rounding) >= 0
 
-        for location, location_quants in quants_per_bin:
-            location_quantity = sum(location_quants.mapped("quantity")) - sum(
-                location_quants.mapped("reserved_quantity")
-            )
-            if location_quantity <= 0:
+        # first search for the largest packaging in all locations
+        for pack_quantity in packaging_quantities:
```
jbaudoux: to remove any confusion, can you rename to packaging as it's not a pack
```suggestion
        for packaging_quantity in packaging_quantities:
```

jbaudoux (approved):

stock_reserve_rule/models/stock_reserve_rule.py
henrybackman: Renamed

jbaudoux (approved):

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -284,19 +284,22 @@ def _apply_strategy_packaging(self, quants):
...
-            location_quantity = sum(location_quants.mapped("quantity")) - sum(
-                location_quants.mapped("reserved_quantity")
-            )
-            if location_quantity <= 0:
+        # first search for the largest packaging in all locations
+        for packaging_quantity in packaging_quantities:
+            # check if asked at least packaging quantity
+            if not is_greater_eq(need, packaging_quantity):
                 continue
+            for location, location_quants in quants_per_bin:
```
mmequignon, line 292: This is a little bit worrying:
Let's say need is 27, available qty in location is 15, packaging qty is 5.
Here:
```python
                    take = (need // packaging_quantity) * packaging_quantity
                    take = (27 // 5) * 5 = 25
```
Are you yielding that take is 25 while available qty in location is 15?
Might've been a little fast in my assumption though

mmequignon (changes requested):

stock_reserve_rule/models/stock_reserve_rule.py
mmequignon: Instead of need here, I would use `min(need, location_qty)`.
```
avail_for_need = min(need, location_qty) # 15
take = (avail_for_need // packaging_quantity) * packaging_qty
take == 15
```
But again, I might miss some context.
Just it feels weird to use `need`, that might not be in stock.

mmequignon: @jbaudoux any insight here?

henrybackman: The `location_quantity` is also returned, which is later used as the `available_quantity`. If this returns `take` of 25, the `taken_quantity` in `_update_reserved_quantity` of `stock.move` is going to be just 15.
- https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L1253

The other rules have the same pattern and they only check for `location_quantity` below zero, but not if the `take` is larger than `location_quantity`

jbaudoux: And if the location quantity is 16 ? You should still take a multiple of 5. Will it take 15 or 16 ?

henrybackman: That's a good point. It looks to me that it would then take 16, which would be incorrect. I guess makes sense to fix it in this same PR instead of separately? I can write a test to confirm the wrong behavior and fix the logic so that the `take` doesn't exceed the `location_quantity`.

jbaudoux: In this PR. Yes best it to write a unit test for that

henrybackman: Unit test added to confirm the issue and the fix to resolve it

stock_reserve_rule/models/stock_reserve_rule.py
```diff
@@ -284,19 +284,22 @@ def _apply_strategy_packaging(self, quants):
...
+                location_quantity = sum(location_quants.mapped("quantity")) - sum(
```
TDu, line 293: Because the loop on the `quants_per_bin` is integrated in another loop are we not risking reserving multiple time from the same bin ?
And over reserving

henrybackman: Yes, for example if we have packaging quantities 5 and 50 and the need 60, this rule would first yield take of 50 and then later take of 10. After doing a quick test, these seem to be merged into one move line of quantity 60 for this location, so it seems to be working fine and as expected. However I'm not aware if there are unexpected side-effects from having multiple reservations from the same location.

Earlier the multiple reservations would not have happened because the packaging quantities were checked in ascending order, so if there are side-effects from making multiple reservations then that is a concern.

jbaudoux: I think @TDu is right. If the location quantity is 51, it should not take 50 and then 1 but only 50. As 51 >= 5, I think it will consider the location as valid for the packaging of 5 which is wrong as there is only 1 left

henrybackman: @jbaudoux  @TDu The `location_quantity` in the `_apply_strategy_packaging` considers also the `reserved_quantity`of the `quant`. This is updated after each take by the main loop in `stock_reserve_rule/models/stock_move.py:_update_reserved_quantity` by calling the `super()`. 

In the case of location quantity of 51, packaging quantities [50,5] and need of 55, the process first reserves the 50 and that updates the location quant `reserved_quantity`. After that it tries to get more from the location but there's not enough unreserved quantity for the smallest packaging size and the end result is "partially_available" move line with quantity 50.

I've added a test "test_rule_packaging_multiple_reservations_from_same_location" for this case to confirm

jbaudoux (approved):

jbaudoux:
@mmequignon ok for you now ?

PR #2269, opened 2025-02-20
[18.0] [MIG] stock_location_bin_name

rlimaeco:


rlimaeco:
@grindtildeath and @mmequignon could please review this PR

TDu (approved):

stock_location_bin_name/models/stock_location.py
```diff
@@ -37,14 +38,25 @@ class StockLocation(models.Model):
...
+    @api.constrains("location_name_format")
+    def _check_location_name_format(self):
+        for loc in self:
+            if loc.location_kind == "bin" and loc.location_name_format:
+                raise ValidationError(
+                    self.env._(
+                        "The area of the location: '%(location)s' "
+                        "should have the name format, not in a Bin location",
+                        location=loc.name,
+                    )
```
mmequignon: This looks a little bit too specific and at the same time a little vague.
Location Name Format could be set on the Area or on the Zone, as long as it's not a bin.
I would be more direct about what the issue is.
```suggestion
                    self.env._(
                        "Location Name Format cannot be set on 'bin' type locations."
                    )
```

stock_location_bin_name/tests/test_stock_location_bin_name.py
```diff
@@ -0,0 +1,137 @@
...
+            location_form.posy = 2
+            location_form.posz = 3
+            location_form.level = "2"
+
+        self.assertEqual(self.location_bin.name, "Zone Location-Z-02-123.005")
+
+    def test_location_bin_with_name_format(self):
+        """Test location constraint with location name format in a bin location"""
+
+        with self.assertRaises(Exception) as context:
```
mmequignon: a little tip here:
When testing an exception, we want to ensure that the raised exception is actually the one we introduced.
For that, we should use `assertRaisesRegex` instead:
```suggestion
regex = "Location Name Format cannot be set on 'bin' type locations."
with self.assertRaisesRegex(ValidationError, regex):
```
Then you don't need to check the context :blush:

mmequignon (changes requested):
Looks good, only minor remarks :+1:

mmequignon (approved):

jbaudoux (approved):

PR #2274, opened 2025-02-26
[17.0][FIX] stock_quant_manual_assign: fix calculation in the quant manual reservation for MO

MeritxellAForgeFlow:
When manually assigning the quants for a MO, in case the previously reserved quantity was higher than the on-hand quantity, the line quantity was negative. This PR fixes the problem so the amount can only be 0 or positive.

@DavidJForgeFlow


DavidJForgeFlow (approved):
Makes sense!

LoisRForgeFlow (changes requested):
I would like to have steps to reproduce and ideally a tests case.

LoisRForgeFlow:
Just rebased for now.

PR #2223, opened 2024-12-18
[18.0][MIG] stock_helper (renamed to stock_location_is_sublocation)

JasminSForgeFlow:
Standard Migration

Rename module to stock_location_is_sublocation and include check stock location is sublocation functionality

will add _get_source_location_from_route to different module

@ForgeFlow

PR #2283, opened 2025-03-07
[18.0][MIG] stock_helper: Migration to 18.0. (split feature to stock_route_location_source)

mt-software-de:
To complete the Migration of `stock_helper` started with https://github.com/OCA/stock-logistics-warehouse/pull/2223
This PR split the method for getting the source location of a route into a own addon. 
The initial method `_get_source_location_from_route` was moved to `stock.route` and renamed to `_get_source_location`. 

cc @ThiagoMForgeFlow @jbaudoux @LoisRForgeFlow @JasminSForgeFlow


LoisRForgeFlow (commented):
Thanks! looks like the tests are failing, could you check?

stock_route_location_source/models/stock_route.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+class StockRoute(models.Model):
+    _inherit = "stock.route"
+
+    def _get_source_location(self, location, procure_method):
```
jbaudoux: What about having this as a computed field?

stock_route_location_source/i18n/hr.po
mt-software-de, line 1: Needs to be removed

stock_route_location_source/readme/CONTRIBUTORS.md
mt-software-de, line 1: Needs to be changed.

stock_route_location_source/readme/DESCRIPTION.md
mt-software-de, line 1: Needs to be updated

stock_route_location_source/models/stock_route.py
mt-software-de: At the moment it is just used here. And here it is just quite specific. 
https://github.com/OCA/stock-logistics-orderpoint/blob/31cfed56eebf1cef7422f2adda4b40b876832554/stock_location_orderpoint/models/stock_location_orderpoint.py#L128

But it is not possible to implement it as computed method, because you need a least a location as argument otherwise you cannot find a src location. 

But after thinking about your question, i don't know it this method is needed at all. 
Because it is only once used and very specific. And for `stock_location_orderpoint` it also easier to figure out what the src location is. 
For example in `stock_location_orderpoint` we could just filter the rules of the route.  
`location_src = orderpoint.route_id.rule_ids.filtered(lambda rule: rule.location_id == orderpoint.location_id and rule.procure_method == 'make_to_stock').location_src_id`

mt-software-de: Also this is not correct that it can be simplified. Because the main issue is here that i'm just talking from a very easy route rule configuration. So something a need is created in Stock take it from stock2. 
But if it it would be a more complex configuration this is not working it all. 
If it is a more complex route we need to set from the last make_to_stock rule the location src.
This also would mean that we need to remove the route from the call.

mt-software-de: I will update this feature and propose a change.

mt-software-de: Could you please check the last commit.

jbaudoux (approved):
LG. Just write a nice commit msg :)

LoisRForgeFlow (approved):
Good enough :+1:

PR #2289, opened 2025-03-13
[17.0][FIX] stock_quant_manual_assign: manual assign option was enabled in 'done' transfers

MeritxellAForgeFlow:
Previous to this fix, the option of manually assigning quants was shown in 'done' transfers.

### DIFF
--- a/stock_quant_manual_assign/views/stock_move_view.xml
+++ b/stock_quant_manual_assign/views/stock_move_view.xml
@@ -12,7 +12,7 @@
                     icon="fa-tags"
                     title="Manual Quants"
                     options='{"warn": true}'
-                    attrs="{'invisible':['|',('picking_code','=','incoming'),('state','not in',('confirmed','assigned','partially_available'))]}"
+                    invisible="picking_code == 'incoming' or state not in ('confirmed', 'assigned', 'partially_available')"
                 />
             </button>
         </field>


stock_quant_manual_assign/views/stock_move_view.xml
```diff
@@ -12,7 +12,7 @@
                     icon="fa-tags"
                     title="Manual Quants"
                     options='{"warn": true}'
-                    attrs="{'invisible':['|',('picking_code','=','incoming'),('state','not in',('confirmed','assigned','partially_available'))]}"
+                    invisible="picking_code == 'incoming' or state not in ['confirmed', 'assigned', 'partially_available']"
```
Shide: **nitpick:** Avoid uneeded negations (not in)

```suggestion
                    invisible="picking_code == 'incoming' or state in ['draft', 'waiting', 'done', 'cancel']"
```

MeritxellAForgeFlow: Right, thanks!

LoisRForgeFlow: @Shide Why should they be avoided? For me it is better to control in which statuses you actually allow this feature to be accesible which are ('confirmed','assigned','partially_available'). I find it more easy to understand and less prone to unexpected errors if more statuses are added to the move in the future or by other module.

ThiagoMForgeFlow (approved):

LoisRForgeFlow (approved):
/ocabot merge minor

PR #2299, opened 2025-03-19
[16.0][FIX] stock_move_auto_assign: Use order to assign moves

rousseldenis:
Use same fields as Odoo does to sort stock moves at assignation. This in order to be sure more priority moves are assigned before other ones.

### DIFF
--- a/stock_move_auto_assign/models/product_product.py
+++ b/stock_move_auto_assign/models/product_product.py
@@ -31,6 +31,12 @@ class ProductProduct(models.Model):
             # out anyway.
         ]
 
+    def _get_move_auto_assign_order(self):
+        # See _run_scheduler() method in stock module without the
+        # reservation_date as we are in real time, we should use
+        # date field instead
+        return "priority desc, date asc, id asc"
+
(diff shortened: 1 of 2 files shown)


stock_move_auto_assign/models/product_product.py
```diff
@@ -31,6 +31,10 @@ def _moves_auto_assign_domain(self, locations):
             # out anyway.
         ]
 
+    def _get_move_auto_assign_order(self):
+        # See _run_scheduler() method in stock module
+        return "reservation_date, priority desc, date asc, id asc"
```
jbaudoux: As we do real time re-reservation, I would drop the reservation_date

```suggestion
        return "priority desc, date asc, id asc"
```

jbaudoux (approved):

jbaudoux:
@rousseldenis Tests are failing

rousseldenis:
> @rousseldenis Tests are failing

Yes. Other module. I'll check

jbaudoux (approved):

PR #2303, opened 2025-03-28
[IMP] [16.0] stock_inventory_discrepancy: UI improvements

Shide:
Changes:
- optional="hide" on list view fields
- Discrepancy view group with stock.group_stock_manager
- Company related icon on Settings

image
image

MT-9768 @moduon @rafaelbn @yajo @EmilioPascual @LoisRForgeFlow please review if you want 😄

### DIFF
--- a/stock_inventory_discrepancy/views/stock_location_view.xml
+++ b/stock_inventory_discrepancy/views/stock_location_view.xml
@@ -7,13 +7,19 @@
         <field name="model">stock.location</field>
         <field name="inherit_id" ref="stock.view_location_form" />
         <field name="arch" type="xml">
-            <field name="return_location" position="after">
-                <label for="discrepancy_threshold" />
-                <div>
-                    <field name="discrepancy_threshold" class="oe_inline" /> %
-                </div>
-                <field name="propagate_discrepancy_threshold" />
-            </field>
+            <xpath expr="//form/sheet/group" position="inside">
+                <group
+                    name="grp_discrepancy_threshold"
+                    string="Inventory Discrepancy"
+                    groups="stock.group_stock_manager"
+                >
+                    <label for="discrepancy_threshold" />
+                    <div>
+                        <field name="discrepancy_threshold" class="oe_inline" /> %
+                    </div>
+                    <field name="propagate_discrepancy_threshold" />
+                </group>
+            </xpath>
         </field>
     </record>
 </odoo>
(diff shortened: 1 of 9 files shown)


rafaelbn:
Hello @LoisRForgeFlow @AaronHForgeFlow @JoanSForgeFlow @JasminSForgeFlow what do you think about this UX improvements? 😄

stock_inventory_discrepancy/views/stock_location_view.xml
```diff
@@ -7,13 +7,19 @@
...
-                <div>
-                    <field name="discrepancy_threshold" class="oe_inline" /> %
-                </div>
-                <field name="propagate_discrepancy_threshold" />
-            </field>
+            <xpath expr="//form/sheet/group" position="inside">
+                <group
+                    name="grp_discrepancy_threshold"
+                    string="Inventory Discrepancy"
+                    groups="base.group_no_one"
```
LoisRForgeFlow: I agree in general with the improvements. Just a bit hesitant on making them only displayed in debug mode, any specific reason for this?

Shide: There is a general threshold configured in settings, so... 99% of the time, this information is redundant (most of the time) and doesn't change too often.
And other "configurable" fields in stock location, which follows the same kind of usage are displayed on debug mode too.

LoisRForgeFlow: I'm not sure if what you state is correct:

There is no discrepancy threshold in the settings
image

And there is not a single field with group_no_one in standard location form view:
image


I think I would keep standard visibility for these fields, you can apply the group "group_stock_manager" if you want to restrict it a little bit.

Regarding the WH view, I would move the new group to the wh config tab

image


Is this fine for you?

Shide: Totally 😄 
But checking it again, I've respected the original position (added manager group) because this pages are not shown if you don't activate certain settings.

LoisRForgeFlow (approved):
Thanks! LGTM now :+1:

AaronHForgeFlow (approved):
LGTM :+1:

rafaelbn (approved):
/ocabot merge minor

LoisRForgeFlow:
@Shide could you take care of fwporting these change to next versions?

LoisRForgeFlow:
@Shide could you fwport this?

Shide:
@LoisRForgeFlow doing in https://github.com/OCA/stock-logistics-warehouse/pull/2346, but it will take me more time
We don't have customers in v17, so I've ported all to v18 (all of this PRs must be ported together)

PR #2219, opened 2024-12-11
[18.0][MIG] stock_picking_show_linked: Migration to 18.0

fd-oerp:

PR #2307, opened 2025-04-01
[18.0][MIG] stock_picking_show_linked

ThiagoMForgeFlow:
Supersede:

* https://github.com/OCA/stock-logistics-warehouse/pull/2219

@ForgeFlow


LoisRForgeFlow (changes requested):
I have tested in runboat, it looks like Odoo added a similar button for the destination transfer:

image

Could you check if there is any differnce between "Next Transfer" and "Destination Transfers" (maybe the former does not support m2m?) ?

Maybe we should use this module just to add the origin transfer button that Odoo did not add. I we go that way, aligning icon and wording with the ones used by Odoo I think  makes sense.

ThiagoMForgeFlow:
> I have tested in runboat, it looks like Odoo added a similar button for the destination transfer:
> 
> image
> 
> Could you check if there is any differnce between "Next Transfer" and "Destination Transfers" (maybe the former does not support m2m?) ?
> 
> Maybe we should use this module just to add the origin transfer button that Odoo did not add. I we go that way, aligning icon and wording with the ones used by Odoo I think makes sense.

@LoisRForgeFlow You are right, the Next Transfer smart button does the same functionality as Dest. Transfers of this module. I just adapted the code to only add Origin Transfers. I also change the name of the button to Previous Transfer  and icon to the same as next Transfer.

rousseldenis:
@ThiagoMForgeFlow Could you improve commit message ? Thanks

image

ThiagoMForgeFlow:
> @ThiagoMForgeFlow Could you improve commit message ? Thanks
> 
> image

Commit message changed!

rousseldenis (approved):
LGTM

LoisRForgeFlow (approved):
Looks good now, thanks for sorting things out :+1:

PR #2309, opened 2025-04-02
[14.0][fix] stock_request: use deadline date in procurement and other fixes

petrus-v:
replace #1637 

In this PR I've multiple changes; let me now It you prefer distincts PRs:

* improves unit test to make sure `assetRaises` realy test what it expected to test
* makes api.constraints working with recordset
* do not force default expected date on stock.request if user provide one at creation without stock.request.order
* use the expected date as deadline_date on procurement creation to avoid merging `stock.move` from different stock.request with different expected date. I wonder the impact on running prod env and wonder if I should do something less impacting ? 
* refactor code to replace api.onchange per api.depends: which helps to reuse code while using inherits on stock.request.order

### DIFF
--- a/stock_request/models/stock_request.py
+++ b/stock_request/models/stock_request.py
@@ -40,6 +40,9 @@ class StockRequest(models.Model):
     requested_by = fields.Many2one(
         "res.users",
         "Requested by",
+        compute="_compute_requested_by",
+        store=True,
+        readonly=False,
         required=True,
         tracking=True,
         default=lambda s: s._get_default_requested_by(),
@@ -48,6 +51,8 @@ class StockRequest(models.Model):
         "Expected Date",
         index=True,
         default=_get_expected_date,
+        compute="_compute_expected_date",
+        store=True,
         required=True,
         readonly=True,
         states={"draft": [("readonly", False)]},
@@ -59,6 +64,8 @@ class StockRequest(models.Model):
             ("one", "Receive all products at once"),
         ],
         string="Shipping Policy",
+        compute="_compute_picking_policy",
+        store=True,
         required=True,
         readonly=True,
         states={"draft": [("readonly", False)]},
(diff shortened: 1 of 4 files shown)


stock_request/models/stock_request.py
```diff
@@ -149,6 +170,62 @@ def _check_qty(self):
                     _("Stock Request product quantity cannot be negative.")
                 )
 
+    @api.depends("order_id.warehouse_id")
+    def _compute_warehouse_id(self):
+        for request in self:
+            if request.order_id.warehouse_id:
+                request.warehouse_id = request.order_id.warehouse_id
+            else:
+                request.warehouse_id = request.warehouse_id
```
alexandregaldeano, line 179: **question:** is the `else` statement really necessary? Same everywhere
```suggestion
        for request in self:
            if request.order_id.warehouse_id:
                request.warehouse_id = request.order_id.warehouse_id
```

alexandregaldeano (approved):
LGTM! And thanks for adding recordset supports! 👏

stock_request/models/stock_request.py
petrus-v: when field is not store, I'm certain that would raise caching error, I don't remember exactly when the field is store

damdam-s (approved):

rousseldenis:
> In this PR I've multiple changes; let me now It you prefer distincts PRs:

As you did them in different commits, not necessary :smiley:

stock_request/models/stock_request.py
```diff
@@ -149,6 +170,62 @@ def _check_qty(self):
...
+    def _compute_procurement_group_id(self):
+        for request in self:
+            if request.order_id.procurement_group_id:
+                request.procurement_group_id = request.order_id.procurement_group_id
+            else:
+                request.procurement_group_id = request.procurement_group_id
+
+    @api.depends("order_id.company_id")
+    def _compute_company_id(self):
+        for request in self:
```
rousseldenis, line 223: @petrus-v Could `request` and `request order` be in different companies????

petrus-v: I don't think so because of the existing constraint: https://github.com/OCA/stock-logistics-warehouse/pull/2309/files#diff-7da7c1c93876e1d2eeebd0827ded3675f1d2603621937dfc1ea551847d8672b2L259

a `stock.request` can leave alone without `stock.request.order`.


I'm not sure what's the point?

PR #2312, opened 2025-04-07
[16.0][IMP] stock_quant_manual_assign: Improve assign_quants() performance

AungKoKoLin1997:
Previously, when there were hundreds of quants assigned to a stock move, it would take a very long time for assign_quants() to complete, as the system would try to reassign all selected quants.

This PR mitigates the issue by identifying only the necessary changes — specifically, the move lines to be unlinked and the quant lines to be assigned — so that updates are made only to those records.

@qrtl QT5295

### DIFF
--- a/stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
+++ b/stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
@@ -172,3 +172,23 @@ class TestStockQuantManualAssign(TransactionCase):
             sum(line.qty for line in wizard.quants_lines),
             self.move.reserved_availability,
         )
+
+    def test_quant_assign_get_discrepancies(self):
+        self.move._action_assign()
+        # There should be no move lines to unlink and no quant lines to assign
+        # if there is no discrepancy between move lines and quant lines
+        wizard = self._create_wizard()
+        self.assertEqual(
+            len(wizard.quants_lines.ids),
+            3,
+            "Three quants created, three quants got by default",
+        )
+        move_lines_to_unlink, quant_lines_to_assign = wizard._get_discrepancies()
+        self.assertFalse(move_lines_to_unlink)
+        self.assertFalse(quant_lines_to_assign)
+        # There should be a move line to unlink
+        # and a quant line to assisgn (due to qty discrepancy)
+        wizard.quants_lines[2].write({"selected": True, "qty": 50.0})
+        move_lines_to_unlink, quant_lines_to_assign = wizard._get_discrepancies()
+        self.assertEqual(move_lines_to_unlink, self.move.move_line_ids[2])
+        self.assertEqual(quant_lines_to_assign, wizard.quants_lines[2])
--- a/stock_quant_manual_assign/wizard/assign_manual_quants.py
+++ b/stock_quant_manual_assign/wizard/assign_manual_quants.py
@@ -58,11 +58,54 @@ class AssignManualQuants(models.TransientModel):
         string="Move",
     )
 
+    def _get_discrepancies(self):
+        """Get discrepancies between quant lines and move lines.
+
+        This is so that we can work on the minimum sets of records to unlink or assign.
+
+        :return: two recordsets, one with move lines to unlink and the other with quant
+            lines to assign.
+        """
+        self.ensure_one()
+        ml_to_unlink = self.env["stock.move.line"].browse()
+        ql_to_assign = self.env["assign.manual.quants.lines"].browse()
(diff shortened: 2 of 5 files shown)


stock_quant_manual_assign/wizard/assign_manual_quants.py
```diff
@@ -58,11 +58,55 @@ def _compute_qties(self):
         string="Move",
     )
 
+    def _get_discrepancies(self):
+        """Get discrepancies between quant lines and move lines.
+
+        This is to so that we can work on the minimum sets of records to unlink or
+        assign.
```
yostashiro:
```suggestion
        This is so that we can work on the minimum sets of records to unlink or assign.
```

yostashiro (approved):
LGTM.

rousseldenis (commented):
@AungKoKoLin1997 Thanks for this. Could you add tests that cover the move lines to unlink case ?

stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
```diff
@@ -172,3 +172,27 @@ def test_quant_assign_wizard_after_availability_check(self):
...
+                "qty_done": 1.0,
+                "location_id": self.location1.id,
+                "location_dest_id": self.location_dst.id,
+                "lot_id": lot1.id,
+            }
+        )
+        self.assertIn(move_line, self.move.move_line_ids)
+        wizard = self._create_wizard()
+        move_lines_to_unlink, _ = wizard._get_discrepancies()
+        self.assertEqual(move_lines_to_unlink, move_line)
```
yostashiro: This is because there is no stock for `lot1`, correct? In a way, having such a move line record like the one in this test case is unnatural. Would be nice to add some comments to make the intention easier to follow.

stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
```diff
@@ -172,3 +174,23 @@ def test_quant_assign_wizard_after_availability_check(self):
...
+        move_lines_to_unlink, quant_lines_to_assign = wizard._get_discrepancies()
+        self.assertFalse(move_lines_to_unlink)
+        self.assertFalse(quant_lines_to_assign)
+        # There should be a move line to unlink (due to owner mismatch)
+        # and a quant line to assisgn (due to qty discrepancy)
+        owner = self.env["res.partner"].create({"name": "Owner"})
+        self.move.move_line_ids[0].owner_id = owner.id
+        wizard = self._create_wizard()
+        wizard.quants_lines[0].write({"selected": True, "qty": 100.0})
+        wizard.quants_lines[1].write({"selected": True, "qty": 100.0})
```
yostashiro: The scenario is still hard to follow. IIUC, `self._create_wizard()` creates three selected quant line records (with no owner, since the value is taken from the quant), which creates a discrepancy for the first quant line, since the corresponding move line has an owner assigned in the previous step.

If this is the case, we shouldn't need the steps of selecting quants. Instead, we should just assert the contents of `self.move.move_line_ids` and `wizard.quants_lines`. Am I missing something?

stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
```diff
@@ -172,3 +174,23 @@ def test_quant_assign_wizard_after_availability_check(self):
...
+        move_lines_to_unlink, quant_lines_to_assign = wizard._get_discrepancies()
+        self.assertEqual(move_lines_to_unlink, self.move.move_line_ids[0])
+        self.assertTrue(quant_lines_to_assign, wizard.quants_lines[2])
```
yostashiro:
```suggestion
        self.assertEqual(quant_lines_to_assign, wizard.quants_lines[2])
```

stock_quant_manual_assign/tests/test_stock_quant_manual_assign.py
AungKoKoLin1997: Please review my latest changes.

yostashiro (approved):
LGTM.

AungKoKoLin1997:
@rousseldenis Could you please review the PR?

rousseldenis (approved):
LGTM

kanda999 (approved):
Functional review: This change is running in the production environment.

AungKoKoLin1997:
@rousseldenis Can we get this PR merged?

yostashiro:
@AungKoKoLin1997 Please plan to forward port this.

PR #2321, opened 2025-04-29
[18.0][MIG] stock_location_pending_move

sebalix:


sebalix:
Stock moves are now open in pivot view by default, should we force it to list view instead?
image

jbaudoux:
> Stock moves are now open in pivot view by default, should we force it to list view instead? 

@sebalix I think it's fine with default pivot. What about `search_default_todo` that I don't see anymore ?

sebalix:
@jbaudoux I compared between v16/v18 runbot: they are identical on this point

1. `search_default_todo` set for `stock.move` is useless, there is no such filter available on `stock.move` (both v16 and v18, even on v14)
2. it is however available (and working) for `stock.move.line` action

EDIT: this filter is in fact named `future` for `stock.move`, starting from v14
I made a fix in a separate commit that could be backported later in v16

image

jbaudoux (approved):

rousseldenis (approved):

PR #2331, opened 2025-05-09
[16.0][FIX] stock_mts_mto_rule: multiple procs same product

jbaudoux:
Replaces:
* https://github.com/OCA/stock-logistics-warehouse/pull/2098

cc @ferran-S73 @mt-software-de

### DIFF
--- a/stock_mts_mto_rule/models/stock_rule.py
+++ b/stock_mts_mto_rule/models/stock_rule.py
@@ -1,4 +1,7 @@
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from collections import defaultdict
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


ferran-S73 (approved):

ferran-S73:
@rousseldenis @mt-software-de do you think we can fast-track this PR so that I can close mine?

stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -36,14 +39,16 @@ def _check_mts_mto_rule(self):
...
+        self, product, product_qty, product_uom, values, qty_unavailable=0.0
+    ):
         self.ensure_one()
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
         )
         src_location_id = self.mts_rule_id.location_src_id.id
         product_location = product.with_context(location=src_location_id)
-        virtual_available = product_location.virtual_available
+        virtual_available = product_location.virtual_available - qty_unavailable
```
mt-software-de, line 51: I thought about this pull request again, and I noticed potential sources of error.

 First, we can't simply subtract the unavailable quantity from the virtual quantity. Since the source location can change with each rule, so we might subtract quantities from other locations.

And secondly, it is problematic to first calculate the quantity for each procurement and then execute it. Since, of course, each execution can lead to an increase in the virtual available quantity, I think that was @ferran-S73 initial problem.

mt-software-de (changes requested):

stock_mts_mto_rule/models/stock_rule.py
ferran-S73: @jbaudoux any news on this?

PR #2342, opened 2025-05-15
[IMP] stock_location_product_restriction: Allow to put products in location if being emptied

rousseldenis:
Depends on :

- https://github.com/OCA/stock-logistics-warehouse/pull/2188

In day-to-day stock operations, operators will take products from a location and often empty that location.

But, if the movement quantity is already changed ('qty_done'), the picking is not yet validated.

In parallel, some operators want to fill in the location with another product as it is already physically void.

NOTE: This is a major change as a dependency is added

- Adds also an helper function to check if a product can enter a location


lmignon (approved):
LGTM (code review)

lmignon (approved):
Works in production for months

rousseldenis:
@jbaudoux

PR #2353, opened 2025-05-21
[18.0][ADD] stock_warehouse_out_pull: restore pull rules for multi-steps delivery route

sebalix:
This module restores the pull rules configuration on warehouse delivery route as it was done prior to Odoo 18.0.

image


rousseldenis (approved):
Great!

@sebalix Doesn't this missing something if we only check the `delivery_pull` value ?

sebalix:
@rousseldenis sorry I'm not sure to understand?
If we set `delivery_pull = True` by code while the `delivery_steps` stays to `ship_only`, this won't change anything as this route is not impacted in `get_rules_dict`.

rousseldenis:
> @rousseldenis sorry I'm not sure to understand? If we set `delivery_pull = True` by code while the `delivery_steps` stays to `ship_only`, this won't change anything as this route is not impacted in `get_rules_dict`.

What I wanted to say, is this 

https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_warehouse.py#L804

called when we change to 'delivery_pull': True only?


Imagine we pick pack ship with standard Odoo behavior. But we change our mind and activate the `delivery_pull`. We should trigger the recomputation of rules. I think there is something with the `depends` key

sebalix:
@rousseldenis you're right, I added a test for that and it fails, so the `depends` key has to be updated as well :+1:

sebalix:
@rousseldenis force-pushed with a fix and a new test, thank you

jbaudoux (commented):
isn't this module about workflow ?

rousseldenis:
> isn't this module about workflow ?

@jbaudoux It's more about warehouse configuration. So, IMHO, it can stay here

jbaudoux:
> > isn't this module about workflow ?
> 
> @jbaudoux It's more about warehouse configuration. So, IMHO, it can stay here

## repo README:
Extend the stock related models (warehouse, location, picking, move...) but without impact flows and processes. It's mainly adding fields or buttons.

sebalix:
@jbaudoux it doesn't override business workflow methods, but configuration ones (that aren't call very often in a project life), it is really about config there that is then used by workflows. I put it there for that reason.

ivantodorovich (approved):

jbaudoux (approved):

PR #2354, opened 2025-05-22
[15.0][IMP] stock_inventory_lockdown: add hook to decide what locations are excluded from the lockdown

AaronHForgeFlow:
Some companies may don't want to block all inventory loss locations and some others will want to block other types of locations.

This is just a hook, no change by itself :pray:

### DIFF
--- a/stock_inventory_lockdown/models/stock_move_line.py
+++ b/stock_inventory_lockdown/models/stock_move_line.py
@@ -18,6 +18,13 @@ class StockMoveLine(models.Model):
         self.ensure_one()
         return self.move_line_ids.mapped("location_dest_id")
 
+    def _location_usage_without_restriction(self):
+        self.ensure_one()
+        return (
+            self.location_dest_id.usage == "inventory"
+            or self.location_id.usage == "inventory"
+        )
+
     @api.constrains("location_dest_id", "location_id", "state")
     def _check_locked_location(self):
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
~~Failing tests already exists in main Odoo15 branch: https://github.com/OCA/stock-logistics-warehouse/actions/runs/15157529327/job/42615744592 I will try to fix it in a separate PR~~

Edit: It seems a rebase did the thing :confused:

stock_inventory_lockdown/models/stock_move_line.py
```diff
@@ -18,6 +18,12 @@ def _get_dest_locations(self):
         self.ensure_one()
         return self.move_line_ids.mapped("location_dest_id")
 
+    def _location_usage_without_restriction(self, move_line):
+        return (
+            move_line.location_dest_id.usage == "inventory"
+            or move_line.location_id.usage == "inventory"
+        )
```
JordiMForgeFlow: wouldn't it be better to pass move_line as self, and then in this method you evaluate self

AaronHForgeFlow: Good point! I just did it! :rocket:

MiquelRForgeFlow (approved):

JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
:+1: 

/ocabot merge patch

PR #2357, opened 2025-05-27
[15.0][ADD] stock_inventory_lockdown_product in v15

AaronHForgeFlow:
By default, stock_inventory_lockdown blocks any stock movement to or from locations involved in an in-progress inventory adjustment.

With this module installed and enabled, the restriction becomes product-aware if you activate a setting.

Depends on:

- [x] https://github.com/OCA/stock-logistics-warehouse/pull/2354

It adds a minor improvement in stock_inventory module, in order to be able to inherit the view properly.


LoisRForgeFlow (approved):
Nothing against, it could have been a improvement to the base module (because you added a setting anyway) but in any case not blocking for me.

PR #2361, opened 2025-05-28
[18.0][FIX] procurement_auto_create_group:Fixed the view of stock rule

bizzappdev:
In Odoo v18, the 'Auto-create Procurement Group' field is included by default. 
However, in v17, it was not present, so this module added it to the view. It is no longer needed as it is available by default.

stock-rule-view


simahawk:
@bizzappdev could you pls drop the odoo version from the commit msg?
Also, the msg should talk about a fix... Eg:
`[FIX] procurement_auto_create_group: fix stock rule view duplicated field`

bizzappdev:
@simahawk Done.

PR #2364, opened 2025-06-05
[18.0][MIG] stock_orderpoint_manual_procurement: Migration to 18.0

henrybackman:


stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
```diff
@@ -71,9 +71,7 @@ def make_procurement(self):
             if not item.orderpoint_id:
                 raise ValidationError(_("No reordering rule found!"))
             values = item.orderpoint_id._prepare_procurement_values()
-            values["date_planned"] = fields.Datetime.to_string(
-                fields.Date.from_string(item.date_planned)
-            )
+            values["date_planned"] = fields.Date.from_string(item.date_planned)
```
ivantodorovich: This is probably not needed. `item.date_planned` should already be a date, no?

```suggestion
            values["date_planned"] = item.date_planned
```

ivantodorovich (approved):

stock_orderpoint_manual_procurement/wizards/make_procurement_orderpoint.py
henrybackman: Good catch, applied suggested :+1:

twalter-c2c (approved):
LGTM

jbaudoux (changes requested):
There is already a replenish action in standard.
image

It should integrate with that, like add a new option in the dropdown to launch the manual orderpoint wizard where you can change quantities.

Also it needs to move to https://github.com/OCA/stock-logistics-orderpoint/

henrybackman:
@jbaudoux I moved the PR here:
- https://github.com/OCA/stock-logistics-orderpoint/pull/46
but I would need help making the requested changes as I'm not familiar enough with this part

PR #2365, opened 2025-06-06
[18.0][ADD] stock_move_reset_quantity: Easily reset quantities on all moves / move lines

jbaudoux:
When you receive one product out of a reception, it's convenient to set to 0 all the lines and just put a quantity on the received line

Hotkey: SHIFT-R

On the move lines:

On the picking:


rousseldenis (commented):
LGTM. Functionally and technically reviewed

rousseldenis (approved):

sebalix (approved):

jbaudoux:
I removed the highlight of the button on the move lines. It now appears as secondary like on the picking

stock_move_reset_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Camptocamp SA
```
twalter-c2c: Is it correct, or should it be 2025 and your reference (nitpicking)?

twalter-c2c (approved):
Nice one, LGTM.

stock_move_reset_quantity/models/stock_move_line.py
jbaudoux: good catch, fixed

PR #2367, opened 2025-06-16
[FIX] stock_location_fill_state: Consider a being emptied location as being filled

rousseldenis:
If the 'being emptied' location has an operation that points finally to, it should be considered as 'being filled'.

### DIFF
--- a/stock_location_fill_state/models/stock_location.py
+++ b/stock_location_fill_state/models/stock_location.py
@@ -86,9 +86,13 @@ class StockLocation(models.Model):
         for rec in locations_to_compute:
             qty_in_location = qty_by_location.get(rec.id, 0.0)
             out_by_location = out_qty_by_location.get(rec.id, 0.0)
-            if out_by_location and (qty_in_location - out_by_location <= 0):
+            if (
+                out_by_location
+                and (qty_in_location - out_by_location <= 0)
+                and not (rec.pending_in_move_ids or rec.pending_in_move_line_ids)
+            ):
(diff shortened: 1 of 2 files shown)


stock_location_fill_state/models/stock_location.py
```diff
@@ -86,11 +86,17 @@ def _compute_fill_state(self):
...
+            if (
+                out_by_location
+                and (qty_in_location - out_by_location <= 0)
+                and not (rec.pending_in_move_ids or rec.pending_in_move_line_ids)
             ):
+                records_by_state["being_emptied"] |= rec
+            elif (
+                out_by_location
+                and (qty_in_location - out_by_location <= 0)
+                or (qty_in_location <= 0)
```
jbaudoux:
```suggestion
```

jbaudoux (changes requested):
add missing parenthesis or remove useless `or` condition

stock_location_fill_state/models/stock_location.py
rousseldenis: No. The case is : when you have outgoing quantities OR no quantity

jbaudoux: Instead of 
`out_by_location and (qty_in_location - out_by_location <= 0) or (qty_in_location <= 0)`
you can just do
`(qty_in_location - out_by_location <= 0)`

jbaudoux (approved):
After second read, what you did is right but it can be simplified :)

PR #2384, opened 2025-07-08
[FIX] stock_picking_supplier_ref: clarify label and help text for supplier_reference field

ppyczko:
This change updates the label and help text of the supplier_reference field on the stock picking model to better reflect its actual use. Previously, the field’s description implied that the supplier reference from the related purchase order would be automatically copied to the picking, which is not the case. This improves user understanding and reduces confusion about the purpose and behavior of this field.

cc https://github.com/APSL 16755

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @BernatObrador please review

### DIFF
--- a/stock_picking_supplier_ref/i18n/es.po
+++ b/stock_picking_supplier_ref/i18n/es.po
@@ -4,41 +4,34 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 14.0\n"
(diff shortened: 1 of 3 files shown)


peluko00 (approved):
LGTM!

mpascuall (approved):
LGTM!

lbarry-apsl (approved):
LGTM

rousseldenis (approved):

PR #2390, opened 2025-07-18
[18.0][IMP] stock_picking_volume: Add hooks

mmequignon:
Add hooks allowing to use a different quantity I.E. stock_available_to_promise_release.
Also make the code more robust, for instance by checking that product is not a service.


simahawk (approved):
LG
the fix in the picking should go to its own commit tho

PR #2394, opened 2025-07-28
[16.0][FIX] stock_cycle_count: filter stock.inventory is not inherited yet

xaviedoanhduy:
fix `My Adjustments` filter is now missing because the view has inherited it incorrectly

result: 
image


LoisRForgeFlow (approved):
Indeed, thank you!

PR #2395, opened 2025-07-28
[16.0][FIX] stock_cycle_count: fix multiple issues in location accuracy report

xaviedoanhduy:
Fixes:
- Incorrect field used in domain (`location_id` → `location_ids`)
- Wrong report template name in `ir.actions.report`
- Missing `data` in QWeb rendering context due to incorrect report model name; renamed `report.stock_location_accuracy` → `report.stock_cycle_count.stock_location_accuracy`
- Replaced Bootstrap 4 classes with Bootstrap 5 equivalents in report template
- Removed deprecated `render_html`, replaced with `_get_report_values`, see: https://github.com/odoo/odoo/pull/39527

Also improved the display a bit on this occasion:
- Added percentage units to Accuracy values
- left aligned the status and date columns (right aligned for numeric data only)

result: [Accuracy Report (4).pdf](https://github.com/user-attachments/files/21463311/Accuracy.Report.4.pdf)

### DIFF
--- a/stock_cycle_count/reports/report_stock_location_accuracy.py
+++ b/stock_cycle_count/reports/report_stock_location_accuracy.py
@@ -6,13 +6,13 @@ from odoo import api, models
 
 
 class LocationAccuracyReport(models.AbstractModel):
-    _name = "report.stock_location_accuracy"
+    _name = "report.stock_cycle_count.stock_location_accuracy"
     _description = "Location Accuracy Report"
 
     @api.model
     def _get_inventory_domain(self, loc_id, exclude_sublocation=True):
         return [
-            ("location_id", "=", loc_id),
+            ("location_ids", "in", [loc_id]),
             ("exclude_sublocation", "=", exclude_sublocation),
             ("filter", "=", "none"),
             ("state", "=", "done"),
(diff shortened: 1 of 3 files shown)


xaviedoanhduy:
Hi @LoisRForgeFlow, can you take a look?

LoisRForgeFlow:
@xaviedoanhduy What am I missing? I don't see the option to print in the ruboat:

image

xaviedoanhduy:
Since this report record does not have the `binding_model_id` field, we have to manually add it in Settings / Technical / Reporting / Reports

But I recommend you create some sample data (Inventory Adjustments) to make the report more complete.

image

xaviedoanhduy:
FYI @LoisRForgeFlow, I got this error when I tried to access this report before. So this patch is to fix it

image

LoisRForgeFlow:
> Since this report record does not have the `binding_model_id` field, we have to manually add it in Settings / Technical / Reporting / Reports
> 
> But I recommend you create some sample data (Inventory Adjustments) to make the report more complete.
> 
> image

I think this should be done via data directly in the module. Could you add that in this PR?

xaviedoanhduy:
> I think this should be done via data directly in the module. Could you add that in this PR?

@LoisRForgeFlow ok, done.
also took the opportunity to fix the display of the Cycle Count detail table (the columns displaying text should be right-aligned instead of centered)

[Cycle Count (2).pdf](https://github.com/user-attachments/files/21465038/Cycle.Count.2.pdf)
[Accuracy Report (8).pdf](https://github.com/user-attachments/files/21465042/Accuracy.Report.8.pdf)

LoisRForgeFlow (approved):
Looks great now, thanks for cleaning all this :+1:

PR #2403, opened 2025-07-30
[18.0][MIG] stock_cycle_count: Migration to 18.0

xaviedoanhduy:
pick some commits into commit history like:

- https://github.com/OCA/stock-logistics-warehouse/pull/2394
- https://github.com/OCA/stock-logistics-warehouse/pull/2395

reports:

- [Cycle Count.pdf](https://github.com/user-attachments/files/21461078/Cycle.Count.pdf)
- [Accuracy Report.pdf](https://github.com/user-attachments/files/21461244/Accuracy.Report.pdf)

note: 

- Because the Accuracy Report uses `web.external_layout` and introduce the change from https://github.com/odoo/odoo/pull/169512, change the table showing the location details in this report, based on another report, the Purchase Order report, which also `uses web.external_layout`

old table:
image

new table:
image


JoanSForgeFlow (approved):
Functional review, works as expected!

stock_cycle_count/models/stock_quant.py
```diff
@@ -7,38 +7,38 @@
 class StockQuant(models.Model):
     _inherit = "stock.quant"
 
+    # pylint: disable=W8110
```
LoisRForgeFlow, line 10: Why? It was capturing and returning super correctly before.

LoisRForgeFlow (commented):
Hi @xaviedoanhduy 

Functional and code review looks good, just a couple of comments.

Could you move the menu under "adjustments"?

image

The other comment below inline.

stock_cycle_count/models/stock_quant.py
xaviedoanhduy: the original function is a void function so it doesn't return anything. ([ref](https://github.com/odoo/odoo/blob/b9d6747abc913c41e936b29e0890f83133941497/addons/stock/models/stock_quant.py#L1013-L1034))
maybe in previous versions they returned to bypass this check (`W8110`). 
I removed this return and bypass check so that readers would avoid misunderstanding or expecting a return value from `super()`

xaviedoanhduy:
> Hi @xaviedoanhduy
> 
> Functional and code review looks good, just a couple of comments.
> 
> Could you move the menu under "adjustments"?
> 

done!

image

stock_cycle_count/models/stock_quant.py
LoisRForgeFlow: OK, good enough. It is common in OCA to do that even though there is nothing to return actually, so most people will already consider that, but you have a point and I think it doesn't really affect end result so let's keep it like this. 

Thanks for the quick response!

LoisRForgeFlow (approved):
/ocabot merge nobump

PR #2231, opened 2025-01-10
[16.0][IMP] stock_inventory: forward port feature, add index

thomaspaulb:
- [x] Migrate over https://github.com/OCA/stock-logistics-warehouse/pull/1781
- [x] Add index on `inventory_adjustment_id` on stock.move.line as described below

On a system with many move lines, opening an inventory adjustment view can result in the following slow query:

```
SELECT
    min(l.id) AS id,
    count(l.id),
    l.inventory_adjustment_id
FROM stock_move_line l
LEFT JOIN stock_inventory i ON l.inventory_adjustment_id = i.id
WHERE l.inventory_adjustment_id = 17
AND ((l.company_id = 1) OR l.company_id IS NULL)
GROUP BY l.inventory_adjustment_id,i.date,i.id;
ORDER BY  "stock_move_line__inventory_adjustment_id"."date" DESC,"stock_move_line__inventory_adjustment_id"."id" DESC;
```

The system will have to churn through all the millions of move lines to find the matching ones. This index makes that search quicker.

PR #2408, opened 2025-08-07
[17.0][IMP] stock_inventory: do not delete inv adjustment groups if in process or done + index

LoisRForgeFlow:
Fwport of https://github.com/OCA/stock-logistics-warehouse/pull/1781 and https://github.com/OCA/stock-logistics-warehouse/pull/2231


AaronHForgeFlow (approved):
:+1:  makes sense

Issue #2389, opened 2025-07-17
[17.0] Modification of blocking rule for inventrory per location  (stock_inventory)


AndrzejGerasimukARCHIMEDES:
I've just installed the `stock_inventory` module for Odoo version 17 on our test environment. This module provides an excellent backport of inventory functionality similar to R14.

There is an option to limit products to specific ones or to a product category. What concerns me is that the system currently blocks multiple inventories to be run in the same `location_ids`. I'm curious about the rationale behind this.

My assumption is that the product limiting options were developed later, and initially, inventory was only managed per location. This presents a challenge for us, as we need to be able to run several inventories within one location, each for different product categories.

Could you elaborate on why this is currently restricted?

In the `action_state_to_in_progress()` method, `quants = self._get_quants(self.location_ids)` is used to find quants, and separate methods then limit the results. Wouldn't it be sufficient to calculate all `quants` that have `current_inventory_id` set and compare this set with the one obtained from the `_get_quants()` method? 
If any records overlap, then a block could be implemented (or a wizard with warning about creating an inventory without products already assigned to another inventory)?

Furthermore, would it be feasible to overload the `action_apply()` method from `odoo/addons/stock/wizard/stock_inventory_adjustment_name.py`? Before calling `super()`, it could be filtered to include only those quants that have a matching `current_inventory_id` set.

P.S. The `_check_inventory_in_progress_not_override` method appears to no longer be in use anywhere.

AndrzejGerasimukARCHIMEDES:
@DavidJForgeFlow @LoisRForgeFlow 

can You take a look on my proposal of modification and share Your opinion?

LoisRForgeFlow:
@AndrzejGerasimukARCHIMEDES this was solved in a previous version, see https://github.com/OCA/stock-logistics-warehouse/pull/2120. But apparently it did not reach the newer versions.

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow 
So in that case can You implement it to 17.0?

It is very important for us to have abillity to run several inventories at the same time on the same location for different product categories

LoisRForgeFlow:
@AndrzejGerasimukARCHIMEDES Hi again,

I have reviewed the commit history and the mentioned PR did actually reach v17 and v18, so I was too quick. I have now tested in runboat and I cannot reproduce the problem you describe, I have created 2 inventory adjustments and put them in progress for the same location an different products:

[...]

PR #2411, opened 2025-08-07
[17.0] Issue of the blocking rule for inventory per location, stock_inventory, v17.0.1.4.0, #2389

AndrzejGerasimukARCHIMEDES:


AndrzejGerasimukARCHIMEDES:
related with https://github.com/OCA/stock-logistics-warehouse/issues/2389

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow Can You review it please?

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow 
Hello.
I would like to kindly remind myself about this PR

LoisRForgeFlow:
Hi @AndrzejGerasimukARCHIMEDES 

Thanks for your contribution and sorry for the slow reply. There are a few things that you need to take care of here before this is ready to review.

- You should make sure that test are green.
image

- You should not bump the version manually in the manifiest if you are not adding a migration script. This is done automatically by ocabot at the time of merging.
- You should add a new test case to the module that tests the new feature.
- You should propose the change in a single commit following the [guidelines](https://www.odoo.com/documentation/12.0/developer/reference/guidelines.html#commit-message-structure). In short: 

```
[IMP] module_name: short title 

Long description
```


You might want to review the full how-to guide at OCA: https://odoo-community.org/resources/code.

Reviewing how other PRs are done can also help you to better understand the process.

Thanks again for steping in and trying to contribute!

AndrzejGerasimukARCHIMEDES:
Hello @LoisRForgeFlow,

Thanks for your feedback.

When I uploaded my code, I took a look at the errors in the test and they weren't related to this module.
The error is in the database on the `product_rel` table, and it looks like a problem generated by the `product_merge` app.
Other warnings are generated from the `stock_demand_estimate` module.

I don't know how I can resolve them.

I also see a problem in `test_09_product_inventory_global_and_sublocations_review` on `stock_inventory`.
I suppose that's why you wrote,

>   You should add a new test case to the module that tests the new feature.

I will try to do this (I have never done it and I need to check how it should be done).

Regarding the version bump, I'm sorry. I wasn't aware of it. On our customization project, I always do it manually. If I push a commit with the old version, will that be enough?

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow 
I've modified the code according to the feedback (thank you for that).
I've prepared a test for my modification.
I've squashed all the commits and uploaded them to this PR in a single commit.
All tests have passed successfully and are now green.

DavidJForgeFlow (approved):

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow Great. Thank You for approval.
Can You tell me what now?

You will merge it?
Or i should pass it to some one else for approval?

DavidJForgeFlow:
@AndrzejGerasimukARCHIMEDES I review it and seems okay to me, that's why I approved, however we need here another approve and as @LoisRForgeFlow was already checking this he should review it before going to next step!

AndrzejGerasimukARCHIMEDES:
@DavidJForgeFlow Thanks for feedback 

certainly - i didn't notice that approval was not done by @LoisRForgeFlow but by You - sorry for that.

LoisRForgeFlow (changes requested):
@AndrzejGerasimukARCHIMEDES

First, sorry for the late review. Regardings the changes proposed, I agree with @JoanSForgeFlow analysis but not in the conclusion (EDIT: ups he has deleted his response, but anyway, the problem is described in the next sentences here). For me it should not be possible to count 2 quants in 2 inventory adjusments. Imagine this siutation, you have 2 inventories:

1. Inventory for Product category All.
2. Inventory for Stock location WH/Stock

The overlap in quants will be massive. This was not possible prior to this PR and now it is.

Thefore I agree with the issue you reported, but not with the fix. We have to completely restrict any overlap.

JoanSForgeFlow:
Hi @AndrzejGerasimukARCHIMEDES 

Sorry, I had deleted my previous comment, but it was the following:

> I noticed a potential issue with the logic
>
> Example:
>
> - Start an inventory by category
> - Then start another one by location (cycle count) on a location that contains some of those products.
>
> Those quants will already be marked as `to_do=True` for the first inventory, so they’ll be excluded from the second one.
> The new inventory will start fine, but those products won’t appear, which can be confusing, because it looks like they’re missing from that location.
>
> Maybe it would help to show a link or smart button to view the quants that would have been included but are already being counted in another in-progress inventory.

I deleted it because after reviewing the code more carefully, I wanted to be more specific.

The real issue comes from this line with `to_do = False`:  
https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R217

That condition allows you to create inventory adjustments that overlap, but Odoo won’t show the quants that should appear in the new inventory.

The problem you described is real for **categories**, but not for **products**, and I think this new condition you are adding here already fixes the categories the same way the previous condition handled products.  https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R299

So maybe it’s not necessary to change the `_get_base_domain`, right?

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow 

Hi and thank you for your feedback.

Regarding the overlapping issue: the reason I included domain = [("to_do", "=", False)] in the _get_base_domain(): method is to prevent the possibility of counting the same quantities in different inventory adjustments.

I'm not sure how inventories are organized in your company (or in others using this model). In our experience (during the annual main inventory), all counting teams use the same method. Two years ago, we assigned inventories based on stock locations. This year, we plan to split them by product categories (without any exceptions).

This same method was also successfully implemented in two other companies where I worked as an ERP Administrator.

I also believe that a permanent block on running inventory because single products are currently being counted in another inventory could potentially cause more harm (from a company or process perspective) than allowing it to run without those specific products.

Blocking one team's work and making it dependent on another team's inventory completion, while simultaneously forcing the same stock to be counted multiple times, would likely lead to a less efficient use of human resources.

Regarding @JoanSForgeFlow comment (which I saw in an email, but not here): Yes, it's possible that one user might not see products that are in another inventory. This approach is generally safer than unlocking multiple inventories for a single quant.

> Maybe it would help to show a link or smart button to view the quants that would have been included but are already being counted in another in-progress inventory.

That's a great idea! I can certainly prepare that. However, I'd first like to hear @LoisRForgeFlow  opinion on my response and this suggestion.

AndrzejGerasimukARCHIMEDES:
@JoanSForgeFlow 

> Hi @AndrzejGerasimukARCHIMEDES
> 
> Sorry, I had deleted my previous comment, but it was the following:
> 
> I deleted it because after reviewing the code more carefully, I wanted to be more specific.

OK. I understand.

> The real issue comes from this line with `to_do = False`: https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R217
> 
> That condition allows you to create inventory adjustments that overlap, but Odoo won’t show the quants that should appear in the new inventory.
> 
> The problem you described is real for **categories**, but not for **products**, 

Yes, and No.
When You run inventory for **product** and second one for **location** You wil have the same problem (case that @LoisRForgeFlow introduce in https://github.com/OCA/stock-logistics-warehouse/pull/2411#pullrequestreview-3330426300 )

> and I think this new condition you are adding here already fixes the categories the same way the previous condition handled products. https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R299

Yes - i think it will work without it.

> So maybe it’s not necessary to change the `_get_base_domain`, right?

I have add this as a fallback option if in some way the `action_state_to_in_progress` did not work.
But theoretically it should work without this modification.

LoisRForgeFlow:
@AndrzejGerasimukARCHIMEDES Avoiding duplicated quants being counted in multiple inventory adjustments is a current feature of this module and one that should be kept, so if you really need to have overlapping inventories, you could propose another module that is optional to skip the constraint (something like stock_inventory_allow_overlap or similar). EDIT: or directly in this module as a setting that is disabled by default to not modify the existing behavior for current installations.

Speaking of business process, I generally think that counting by product or product category is ineficient because you make the worker travel across multiple locations, so it is surprising to me that after already completed a count strategy based on locations you are now willing to count by products. Pure curiosity, you don't have to answer if you don't want, what made you make the change?

> Blocking one team's work and making it dependent on another team's inventory completion, while simultaneously forcing the same stock to be counted multiple times, would likely lead to a less efficient use of human resources.

We are blocking it at inventory confirmation time so you realize that there is an overlap, therefore the system is actually preventing you  from counting twice the same inventory position. In any case if you are doing a strategy by products or product categories, they should not overlap, they will only overlap if you mix inventories for products and for locations.

AndrzejGerasimukARCHIMEDES:
@LoisRForgeFlow 
OK. I understand Your point of view - thank you for the detailed explanation of the requirements. You are an author of the module and I wil adjust to the main basic features and functionality.

So, as @JoanSForgeFlow already noticed:
> The real issue comes from this line with `to_do = False`: https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R217

> The problem you described is real for **categories**, but not for **products**, and I think this new condition you are adding here already fixes the categories the same way the previous condition handled products. https://github.com/OCA/stock-logistics-warehouse/pull/2411/files#diff-5482ee4f2814dcfe860064db41234e2fe26681bfc92e377671c89fba90a14c60R299
> 
> So maybe it’s not necessary to change the `_get_base_domain`, right?

Would it be appropriate if I restore `_get_base_domain` method to the original, will it meet your requirements? Or is there something else that needs to be improved?

JoanSForgeFlow:
Hi @AndrzejGerasimukARCHIMEDES

Yes, I agree — if you keep the `_get_base_domain` method as it was originally, and just keep your additional condition to fix the issue with product categories, everything should work fine and solve the problem with the categories.

Once you push that change, I’ll test it functionally and approve the PR if works as expected.

thank you very much

AndrzejGerasimukARCHIMEDES:
@JoanSForgeFlow @LoisRForgeFlow 
I have commit fix,

Test errors is related with problem of Docker image postgres, not with code

>   Error response from daemon: Head "https://registry-1.docker.io/v2/library/postgres/manifests/12.0": received unexpected HTTP status: 503 Service Unavailable

PR #2415, opened 2025-08-21
[16.0][FIX] stock_inventory: Memory exhaust issue when selected locations have huge number of child locations

BhaveshHeliconia:
Steps To Reproduce:
- Open this 'Inventory Adjustment' record: http://oca-stock-logistics-warehouse-16-0-3e646a776d23.runboat.odoo-community.org/web?debug=1#cids=1&menu_id=179&action=434&model=stock.inventory&view_type=form&id=1

- Click on 'Begin Adjustments' button.

- Following Error occurs:
image

- FYI: I've imported 30k 'stock.location' records to reproduce this issue.

### DIFF
--- a/stock_inventory/models/stock_inventory.py
+++ b/stock_inventory/models/stock_inventory.py
@@ -1,6 +1,7 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 from odoo.osv import expression
+from odoo.tools import split_every
 
 READONLY_STATES = {
     "draft": [("readonly", False)],
@@ -229,11 +230,11 @@ class InventoryAdjustmentsGroup(models.Model):
     def _get_base_domain(self, locations):
         return (
             [
-                ("location_id", "in", locations.mapped("id")),
+                ("location_id", "in", locations.ids),
             ]
             if self.exclude_sublocation
             else [
-                ("location_id", "child_of", locations.child_internal_location_ids.ids),
+                ("location_id", "child_of", locations.ids),
             ]
         )
 
@@ -334,29 +335,34 @@ class InventoryAdjustmentsGroup(models.Model):
                 "stock_quant_ids": [(6, 0, quants.ids)],
             }
         )
-        quants.write(
-            {
-                "to_do": True,
-                "user_id": self.responsible_id,
-                "inventory_date": self.date,
-                "current_inventory_id": self.id,
-            }
-        )
+        batch_size = self.company_id.stock_inventory_batch_size or 1000
+        for quant_batch in split_every(batch_size, quants):
+            self.env["stock.quant"].concat(*quant_batch).write(
+                {
+                    "to_do": True,
+                    "user_id": self.responsible_id,
+                    "inventory_date": self.date,
+                    "current_inventory_id": self.id,
+                }
+            )
         return
 
     def action_state_to_done(self):
         self.ensure_one()
         self.state = "done"
-        self.stock_quant_ids.filtered(
+        quants = self.stock_quant_ids.filtered(
             lambda q: q.current_inventory_id.id == self.id
-        ).update(
-            {
-                "to_do": False,
-                "user_id": False,
-                "inventory_date": False,
-                "current_inventory_id": False,
-            }
         )
+        batch_size = self.company_id.stock_inventory_batch_size or 1000
+        for quant_batch in split_every(batch_size, quants):
+            self.env["stock.quant"].concat(*quant_batch).write(
+                {
+                    "to_do": False,
+                    "user_id": False,
+                    "inventory_date": False,
+                    "current_inventory_id": False,
+                }
+            )
         return
 
     def action_auto_state_to_done(self):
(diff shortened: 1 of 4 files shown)


BhaveshHeliconia:
Hey @OCA/logistics-maintainers could you please review this PR as soon as possible? This issue has occurred for our client in production.

hparfr (commented):
Hello,

I don't get precisely why this memory exaution happens during the search.

Was it possible to just iterate over the `search`  with `asset` and `limit` ?

BhaveshHeliconia:
@hparfr : In such case, extracting location_ids from domain would be tricky as we've other method calls before this one, which might modify domain when we've other values of 'product_selection' field and I think main problem is here in this part: https://github.com/OCA/stock-logistics-warehouse/blob/16.0/stock_inventory/models/stock_inventory.py#L236

hparfr:
Ok now I get it. Thanks.

May be add a small comment explaining the orm has issues to build a query with huge number of locations.

hparfr (approved):

BhaveshHeliconia:
Hello @OCA/logistics-maintainers , kindly look into this PR in Priority; reviews are welcome. As urgent as this module is, we are using it in production, and once the client had so many child locations and its start breaking this like mentioned, at the moment they are not able to do these functions.

BhaveshHeliconia:
Hello @OCA/logistics-maintainers, could you please review this when you get a chance? We need this module urgently.

BhaveshHeliconia:
@jbaudoux @simahawk @gurneyalex @rafaelbn @LoisRForgeFlow could you please review this when you get a chance?

stock_inventory/models/stock_inventory.py
```diff
@@ -213,29 +217,57 @@ def _compute_action_state_to_cancel_allowed(self):
...
+                domain = self._get_domain_all_quants(base_domain)
+            elif self.product_selection == "manual":
+                domain = self._get_domain_manual_quants(base_domain)
+            elif self.product_selection == "one":
+                domain = self._get_domain_one_quant(base_domain)
+            elif self.product_selection == "lot":
+                domain = self._get_domain_lot_quants(base_domain)
+            elif self.product_selection == "category":
+                domain = self._get_domain_category_quants(base_domain)
+            stock_quant_ids |= stock_quant_env.search(domain)
```
baguenth: @BhaveshHeliconia Instead of actually batching the domain itself, I think the more typical approach would be to do something like: 

```python
return self.env["stock.quant"].search(domain).ids
```

This will only return the ids without actually loading the objects entirely to memory.

Then in the subsequent processing you can process batch by batch: 

```python
from odoo.tools import float_compare, split_every

for stock_quants in split_every(1000, all_stock_quant_ids):
    stock_quant_ids = self.env["stock.quant"].browse(stock_quants)
    ...
```

This will allow for future adaptions in the domain without breaking the batching.

Additional approaches could be to additionally use 

`.with_context(prefetch_fields=False)`

or to use pagination (limit, offset) to limit the results per search.

baguenth (changes requested):

BhaveshHeliconia:
@baguenth I’ve addressed your suggestions. Could you please review the changes?

baguenth:
@BhaveshHeliconia Please check out the failing tests

BhaveshHeliconia:
> @BhaveshHeliconia Please check out the failing tests

The tests are not failing from this module; actually, those failing from the following modules :

- base_product_merge
- stock_package_type_category
- stock_inventory_preparation_filter

It's ready for the review.

BhaveshHeliconia:
Hello @OCA/logistics-maintainers, could you please review this fix? It has been open for a long time, and it’s important that it gets merged. Thank you!

BhaveshHeliconia:
Hello @OCA/logistics-maintainers, it would be great if someone could have a look at these changes.

stock_inventory/models/stock_inventory.py
```diff
@@ -327,36 +329,41 @@ def action_state_to_in_progress(self):
...
-        quants.write(
-            {
-                "to_do": True,
-                "user_id": self.responsible_id,
-                "inventory_date": self.date,
-                "current_inventory_id": self.id,
+                "stock_quant_ids": [(6, 0, quant_ids)],
             }
         )
+        for batch_ids in split_every(1000, quant_ids):
```
mmequignon: I wouldn't hardcode such value, and would create an option for it.
That way, if for some reason you have an issue and 1000 is too high, you can reduce that value without touching the code.

stock_inventory/models/stock_inventory.py
```diff
@@ -327,36 +329,41 @@ def action_state_to_in_progress(self):
                     error_message % {"names": names, "blocking_names": blocking_names}
                 )
 
-        quants = self._get_quants(self.location_ids)
+        quant_ids = self._get_quants(self.location_ids)
```
mmequignon: ⚠️ I Wouldn't change the return value of `_get_quants`. If you want the ids, just do `_get_quants().ids`.
This will also lower the qty of lines modified, which will ease the review and make this pr more likely to be merged.
Here's the issue, if that `_get_quants` method is used elsewhere, and this might break thing.

stock_inventory/models/stock_inventory.py
```diff
@@ -327,36 +329,41 @@ def action_state_to_in_progress(self):
...
-                "to_do": False,
-                "user_id": False,
-                "inventory_date": False,
-                "current_inventory_id": False,
-            }
-        )
+        ).ids
+        for batch_ids in split_every(1000, quant_ids):
+            quants = self.env["stock.quant"].browse(batch_ids)
+            quants.write(
```
mmequignon: You can use `split_every` on recordsets.
No need for getting ids from a recordset, and then browse them again.

stock_inventory/models/stock_inventory.py
```diff
@@ -368,16 +375,19 @@ def action_auto_state_to_done(self):
...
-            {
-                "to_do": False,
-                "user_id": False,
-                "inventory_date": False,
-                "current_inventory_id": False,
-            }
-        )
+        ).ids
+        for batch_ids in split_every(1000, quant_ids):
+            quants = self.env["stock.quant"].browse(batch_ids)
```
mmequignon: Same

mmequignon (changes requested):

BhaveshHeliconia:
> @BhaveshHeliconia Please check out the failing tests

@baguenth now all checks are passing.

BhaveshHeliconia:
ping @mmequignon @baguenth

mmequignon (approved):
Looks much better, thanks 👍

BhaveshHeliconia:
@pedrobaeza can we merge this?

hparfr (approved):

hparfr (approved):
/ocabot merge major

PR #2422, opened 2025-09-08
[18.0][ADD] stock_maturity: new module

chienandalu:
https://www.loom.com/share/d1d9294df8ad46f48c7db63069d62e36?sid=d2f235ab-3bac-43b9-a13f-09f03a81cb83

cc @moduon MT-9625

please review @Shide @EmilioPascual @Gelojr @rafaelbn


Gelojr (changes requested):
Congratulations on the work @chienandalu but some changes are needed before merge.
Reason: Test 7 reveals an issue: with traceability set to “By quantities” and “Maturity date” unchecked (is not possible), the “Maturity date” column still appears in quants via “View Stock”; it should be hidden.
image
image
image

We executed the following tests in Odoo; here are the results.

- Test 1: OK — Enabled maturity control on a lot/serial–tracked storable product; option available under Traceability.
- Test 2: OK — Created two lots (past/future maturity); “Not mature yet” badge and kanban icon shown only for the future-maturity lot.
- Test 3: OK — With scheduled date before maturity, only mature quantities were reserved (partial reservation).
- Test 4: OK — Moving the scheduled date after maturity allowed full reservation.
- Test 5: OK — Lots without a maturity_date remain reservable as expected.
- Test 6: OK — “Not mature products” filter lists quants with future maturity; “Maturity date (remaining days)” column visible.
- Test 7: NOK — For a product with traceability set to “By quantities” (not per lot/serial) and “Maturity date” unchecked, “View Stock” still shows a “Maturity date” column in quants; expected: column hidden.
- Test 8: OK — With serial tracking, only the serial with past maturity was reserved; after moving the date past the second serial’s maturity, both were reserved.
- Test 9: OK — Expiration compatibility verified: only lots that are not expired and already mature are reservable.
- Test 10: OK — Two-step receipts capture maturity on lots. Note: in Odoo you can only set the maturity date after validating the first step of the reception; there’s no place to add it before validation.
- Test 11: OK — If maturity_date is not set on receipt, the lot is reservable; this is an intentional warning case.
- Test 12: OK — Pick reserves only usable stock (not expired and already mature). Note: the availability/forecast popover on the sales line shows total on-hand (not filtered by maturity/expiration), which matches Odoo standard for expiration but may mislead about effective availability.

stock_maturity/models/product_template.py
EmilioPascual, line 1: IMHO: Maybe you might add a new behavior to calculate dates and alerts such as expiration behavior but for madurity

Shide: The behavior is the opposite.
We only need to reserve products when this maturity date is reached. (it's in the past or today).

Expiration date alert is done to sold those products as fast as possible or scrap them.

stock_maturity/models/stock_lot.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import api, fields, models
+
+
+class StockLot(models.Model):
+    _inherit = "stock.lot"
+
+    use_maturity_date = fields.Boolean(related="product_id.use_maturity_date")
+    maturity_date = fields.Datetime()
```
Shide, line 10: Can you add a help to this field?

stock_maturity/views/stock_lot_views.xml
```diff
@@ -0,0 +1,60 @@
...
+        <field name="arch" type="xml">
+            <field name="ref" position="after">
+                <field name="maturity_date" />
+            </field>
+            <xpath expr="//div[hasclass('oe_title')]" position="inside">
+                <span
+                    class="badge text-bg-warning"
+                    invisible="not product_maturity_alert"
+                >Not mature yet</span>
+            </xpath>
```
Shide, line 16: Will be easy to show an alert (alert-warning) on the top of the form.
Otherwise, fix bottom margin (I think would be too much space to show this kind of badge)

Image

stock_maturity/views/stock_lot_views.xml
```diff
@@ -0,0 +1,60 @@
...
+        </field>
+    </record>
+
+    <record id="view_production_lot_tree" model="ir.ui.view">
```
Shide, line 20: Would be great to add a search view that adds "Already Matured / Not Matured yet" to filters.

stock_maturity/views/stock_quant_views.xml
```diff
@@ -0,0 +1,64 @@
...
+                    maturity_date &gt; current_date or quantity &lt; 0
+                </attribute>
+            </xpath>
+            <xpath expr="//field[@name='quantity']" position="after">
+                <field name="maturity_date" widget="remaining_days" />
+            </xpath>
+        </field>
+    </record>
+
+    <record id="view_stock_quant_tree_editable" model="ir.ui.view">
```
Shide, line 18: Search view also on Quants if possible.

Shide (commented):
I like this approach! Great job!

PR #2424, opened 2025-09-09
[18.0][IMP] stock_picking_volume: Add volume digits to the volume fields

ferran-S73:
I think that this improves flexibility in the module as it allows the user to set custom digits number via the decimal precision. It also adds more consistency with the digits throughout all volume fields in the database.

If this change is approved I'll gladly backport it to previous versions.

### DIFF
--- a/stock_picking_volume/hooks.py
+++ b/stock_picking_volume/hooks.py
@@ -11,7 +11,7 @@ _logger = logging.getLogger(__name__)
 def pre_init_hook(env):
     """Pre init create volume column on stock.picking and stock.move"""
     if not column_exists(env.cr, "stock_move", "volume"):
-        create_column(env.cr, "stock_move", "volume", "double precision")
+        create_column(env.cr, "stock_move", "volume", "numeric")
         # First we compute the reserved qty by move_id
         # the reserved qty is the sum of the reserved qty of the move lines
         # linked to the move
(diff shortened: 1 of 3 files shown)


ferran-S73:
Hi @sebalix @henrybackman @mmequignon can I ask you your opinion about this improvement?

lmignon:
@ferran-S73 AFAIK we don't want to round the volume fields. Otherwise we loose precision when we sum volumes. Furthermore, if you are handling very small parts, there is a risk of having zero volumes. However, I did not check what's the defined precision for Volume in Odoo.

ferran-S73:
> @ferran-S73 AFAIK we don't want to round the volume fields. Otherwise we loose precision when we sum volumes. Furthermore, if you are handling very small parts, there is a risk of having zero volumes. However, I did not check what's the defined precision for Volume in Odoo.

@lmignon the default value is 2 digits but since this allows the user to set the desired value that should cover for every use case

lmignon:
I'm still convinced it's a bad idea. The sum of the rounded values is not equal to the rounded sum of the values. And this difference could matter.

ferran-S73:
> I'm still convinced it's a bad idea. The sum of the rounded values is not equal to the rounded sum of the values. And this difference could matter.

Alright @lmignon I get what you're saying. I'll abandon this idea for now and will explore other alternatives. Thanks so much for your quick review!

lmignon:
> > I'm still convinced it's a bad idea. The sum of the rounded values is not equal to the rounded sum of the values. And this difference could matter.
> 
> Alright @lmignon I get what you're saying. I'll abandon this idea for now and will explore other alternatives. Thanks so much for your quick review!

I could also be wrong. :smirk:

PR #2427, opened 2025-09-22
[18.0][FIX] stock_vertical_lift: Ensure that button_save keep the destination

florentx:
The destination location was not saved correctly when the stock move was split.

This change fixes the issue.

Also the readability of field was not optimal: black on purple.  Now it is white on purple like similar fields.

### DIFF
--- a/stock_vertical_lift/models/vertical_lift_operation_base.py
+++ b/stock_vertical_lift/models/vertical_lift_operation_base.py
@@ -48,20 +48,19 @@ def extract_and_action_done(move):
     to first extract some move lines in a separate move, then validate it
     with this method.
     """
-    # Put remaining qty to process from partially available moves
-    # in their own move (which will be then 'confirmed')
-    partial_moves = move.filtered(lambda m: m.state == "partially_available")
-    for partial_move in partial_moves:
-        split_other_move_lines(partial_move, partial_move.move_line_ids)
     # Process assigned moves
     moves = move.filtered(lambda m: m.state == "assigned")
     if not moves:
         return False
     for picking in moves.picking_id:
         moves_todo = picking.move_ids & moves
+        # No need to create a new transfer if we are processing all moves
         if moves_todo == picking.move_ids:
-            # No need to create a new transfer if we are processing all moves
             new_picking = picking
+        # We process some available moves of the picking, but there are still
+        # some other moves to process, then we put the moves to process in
+        # a new transfer to validate. All remaining moves stay in the
+        # current transfer.
         else:
             new_picking = picking.copy(
                 {
@@ -71,13 +70,12 @@ def extract_and_action_done(move):
                     "backorder_id": picking.id,
                 }
             )
-            moves_todo.write({"picking_id": new_picking.id})
-            moves_todo.package_level_id.write({"picking_id": new_picking.id})
-            moves_todo.move_line_ids.write({"picking_id": new_picking.id})
-            moves_todo.move_line_ids.package_level_id.write(
-                {"picking_id": new_picking.id}
-            )
-            new_picking.action_assign()
+            moves_todo.picking_id = new_picking.id
+            moves_todo.package_level_id.picking_id = new_picking.id
+            moves_todo.move_line_ids.picking_id = new_picking.id
+            moves_todo.move_line_ids.package_level_id.picking_id = new_picking.id
+            if moves_todo.state != "assigned":
+                moves_todo._action_assign()
             assert new_picking.state == "assigned"
         new_picking.button_validate()
     return True
--- a/stock_vertical_lift/models/vertical_lift_shuttle.py
+++ b/stock_vertical_lift/models/vertical_lift_shuttle.py
@@ -63,10 +63,10 @@ class VerticalLiftShuttle(models.Model):
     @property
     def _screen_view_for_mode(self):
         return {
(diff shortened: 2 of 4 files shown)


stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -34,6 +34,15 @@ def split_other_move_lines(move, move_lines):
     return False
 
 
+def set_picking_location_dest(picking, location):
+    """Set location_dest_id on the picking and related moves."""
+    if picking.location_dest_id != location:
+        picking.location_dest_id = location.id
```
sebalix:
```suggestion
        picking.location_dest_id = location
```
So you'll benefit of the type checking from Odoo (ensuring that `location` is a `stock.location` recordset)

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -34,6 +34,15 @@ def split_other_move_lines(move, move_lines):
...
+    moves = picking.move_ids.filtered(lambda mv: mv.location_dest_id != location)
+    if moves:
+        moves["location_dest_id"] = location.id
```
sebalix: Not sure it's needed? There is an `api.depends('picking_id.location_dest_id')` on the compute method of `<stock.move>.location_dest_id` that should align the destination with the one from the picking, is it not triggered?

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -72,6 +82,7 @@ def extract_and_action_done(move):
             )
             new_picking.action_assign()
             assert new_picking.state == "assigned"
+        set_picking_location_dest(new_picking, location_dest)
```
sebalix: It's weird we have to do this, this `extract_and_action_done` method is a copy of the one implemented in `shopfloor` module, and it should work as expected there (or we also have the same issue in `shopfloor`?)

cc @jbaudoux

stock_vertical_lift/models/vertical_lift_operation_base.py
florentx: Right, this 2nd part is not needed, thanks to the `@api.depends`.

Then the function has only 2 lines ... maybe I could inline it into `extract_and_action_done` instead of having a separate function.

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -72,6 +72,8 @@ def extract_and_action_done(move):
             )
             new_picking.action_assign()
             assert new_picking.state == "assigned"
+        # Set the destination on the Picking too
+        new_picking.location_dest_id = moves_todo.move_line_ids[0].location_dest_id
```
jbaudoux: can you apply same fix than here https://github.com/OCA/stock-logistics-shopfloor/pull/51/files

stock_vertical_lift/models/vertical_lift_operation_base.py
jbaudoux: was fixed here in shopfloor https://github.com/OCA/stock-logistics-shopfloor/pull/51
odoo now recomputes put-away on action assign. So we should no call it if the move lines are moved

stock_vertical_lift/models/vertical_lift_operation_base.py
florentx: It was done in addition with:
- #2429

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -71,14 +70,14 @@ def extract_and_action_done(move):
...
-            moves_todo.move_line_ids.write({"picking_id": new_picking.id})
-            moves_todo.move_line_ids.package_level_id.write(
-                {"picking_id": new_picking.id}
-            )
-            new_picking.action_assign()
+            moves_todo.picking_id = new_picking.id
+            moves_todo.package_level_id.picking_id = new_picking.id
+            moves_todo.move_line_ids.picking_id = new_picking.id
+            moves_todo.move_line_ids.package_level_id.picking_id = new_picking.id
+            moves_todo._action_assign()
```
jbaudoux: I meant here, the move is already assigned, you just move it to another picking for validation. Calling this will reset the destination location.
If you skip this line, you should not need to put back the destination location.

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -71,14 +70,14 @@ def extract_and_action_done(move):
...
             assert new_picking.state == "assigned"
+        # Set the destination on the Picking too
+        new_picking.location_dest_id = moves_todo.move_line_ids[0].location_dest_id
```
grindtildeath: :warning: 
This might break chained move in case the `location_dest_id` is a sublocation of the original picking's location, because odoo will call `stock.move._break_mto_link` in case the source location of the destination move is not a child of the destination location of the original move: https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_move.py#L1093

IMO we need instead to keep the location_dest_id of the original picking here.

grindtildeath (changes requested):

jbaudoux (approved):
You squash the commits?

grindtildeath (approved):

PR #2436, opened 2025-10-08
[18.0][MIG] product_route_profile: Migration to 18.0

StefanRijnhart:
Standard migration (applying Fields.Command where possible), but with an additional commit to improve backwards compatibility.


product_route_profile/models/stock_route.py
```diff
@@ -0,0 +1,54 @@
...
+    _inherit = "stock.route"
+
+    product_ids = fields.Many2many(
+        compute="_compute_product_ids",
+        inverse="_inverse_product_ids",
+        search="_search_product_ids",
+        store=False,
+    )
+
+    def _compute_product_ids(self):
```
jbaudoux, line 22: missing depends on route_profile_id.route_ids

stock_warehouse_resupply_route_push/tests/test_stock_warehouse_resupply_route_push.py
```diff
@@ -257,6 +257,7 @@ def test_05_recreate_routes_assignation(self):
         wh_4.route_ids += r_wh_3_to_wh4
         # Change to push and recreate routes:
         wh_4.write({"resupply_wh_push": True})
+        self.env.invalidate_all()  # product_route_profile compatibility
```
jbaudoux: This should not occur here. There is a missing depend for cache invalidation or the invalidation should occur in the module that requires it, but not here.

product_route_profile/models/stock_route.py
StefanRijnhart: Thanks! By adding a many2many inverse I could easily define the `api.depends`.

stock_warehouse_resupply_route_push/tests/test_stock_warehouse_resupply_route_push.py
StefanRijnhart: Thanks, also solved by adding the compute method depends.

product_route_profile/models/stock_route.py
jbaudoux: Great. You remove the know issue note then ?
:)

StefanRijnhart: Done!

jbaudoux (approved):

PR #2443, opened 2025-10-23
[18.0][FIX] stock_putaway_product_template: fix test

jbaudoux:
Fix singleton error

The method `_get_putaway_strategy` is not meant to be called on a recordset but on a single record

See https://github.com/odoo/odoo/blob/91133cf41bb5e75c1aa9b185233700924f107c45/addons/stock/models/stock_location.py#L371

cc @pedrobaeza @victoralmau


jbaudoux:
I fixed the tests creating the putaway from a main location to a new destination. Computation of the putaway must occur from that main location

pedrobaeza (approved):

PR #2444, opened 2025-10-29
[18.0][MIG] stock_warehouse_relationship: Migration to 18.0

christian-ramos-tecnativa:
Removed the `warehouse_id` field from the `stock.quant` model, as it is already included in this version.

@Tecnativa TT58957


christian-ramos-tecnativa:
Finally we are not gonna continue using this module. 
Feel free to fork it and propose it again.

PR #1925, opened 2024-01-29
[16][MIG] stock_multi_warehouse_security

florian-dacosta:
Start from https://github.com/OCA/stock-logistics-warehouse/pull/1543

@petrus-v 
@damdam-s 
@bguillot

PR #2445, opened 2025-10-29
[18.0][MIG] stock_warehouse_security: Migration to 18.0

christian-ramos-tecnativa:
Migrated from https://github.com/OCA/stock-logistics-warehouse/pull/1925.

- Module renamed from `stock_multi_warehouse_security` to `stock_warehouse_security` as the tag `multi` wasn't adding extra information about the module.
- Removed dependency to stock_warehouse_relationship

@Tecnativa TT58828


christian-ramos-tecnativa:
After some research, I will re-propose this module without the stock_warehouse_relationship dependency as I think it may not be needed anymore

pedrobaeza (approved):

gurneyalex (approved):

carlos-lopez-tecnativa (changes requested):
I think it might be necessary to rewrite the entire Git history, or is it not needed, @pedrobaeza? Something like:
```
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
git rebase origin/branch
```

Also, I think this module might need to invalidate the cache when the warehouse is assigned to the user. I tested it with the demo user, but on another page I can still see all records (pickings, operation types, etc.).

image

image

image

I restarted the Runboat instance to see the changes:
image

pedrobaeza:
Yes, it would be good to rewrite whole history.

christian-ramos-tecnativa:
I found that when a warehouse is selected, if any product has stock in a location belonging to a different warehouse, you can no longer access the product view.
This issue seems to be related to how `qty_available` is computed and will most likely require some adjustments.
I’ll provide an update as soon as I check it.

christian-ramos-tecnativa:
Thanks for the review @carlos-lopez-tecnativa, last changes done:

- Reword commit history renaming the module
- Added the field  `warehouse_ids` to the method `_get_invalidation_fields` to invalidate the cache 
- Added the user warehouse filtered to avoid the error in products view

carlos-lopez-tecnativa (changes requested):
> I think it might be necessary to rewrite the entire Git history, or is it not needed, @pedrobaeza? Something like:
> 
> ```
> git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
> git rebase origin/branch
> ```

This is still pending. If you review the first commit, the module name is the old one. Did you apply the command I suggested

image

christian-ramos-tecnativa:
Oh sorry @carlos-lopez-tecnativa, I thought the command was just for commit renaming, but now I see is way more complex than that
Thanks!

carlos-lopez-tecnativa (approved):
LGTM

PR #2446, opened 2025-11-02
[18.0][FIX] stock_vertical_lift put

jbaudoux:
* Add missing current_move_line_id to the view and force save as it is readonly
   The missing line was automatically added by odoo in the view but as it is marked as readonly in the model, it was not transmitted on save
* Validate the picking even if the lot is expired

cc @simahawk @sebalix

### DIFF
--- a/stock_vertical_lift/models/vertical_lift_operation_base.py
+++ b/stock_vertical_lift/models/vertical_lift_operation_base.py
@@ -486,6 +486,7 @@ class VerticalLiftOperationTransfer(models.AbstractModel):
 
     def process_current(self):
         line = self.current_move_line_id
+        line.ensure_one()
         if line.state in ("assigned", "partially_available"):
             # if the move has other move lines, it is split to have only this move line
             split_other_move_lines(line.move_id, line)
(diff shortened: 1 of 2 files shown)


stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -77,7 +77,9 @@ def extract_and_action_done(move):
             if moves_todo.state != "assigned":
                 moves_todo._action_assign()
             assert new_picking.state == "assigned"
-        new_picking.button_validate()
+        new_picking.with_context(
+            skip_backorder=True, skip_expired=True, skip_sms=True
+        ).button_validate()
```
sebalix, line 82: Any reason to use `button_validate()` instead of `_action_done()`?

simahawk: indeed :thinking:

jbaudoux: yes, it's doing all what is needed and that's how a picking is processed. It's a choice made long time ago and it's like that in all oca modules

simahawk (approved):

PR #2448, opened 2025-11-04
[18.0][ADD] stock_lot_condition: New module

victoralmau:
New module

Please @christian-ramos-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT57112


stock_lot_condition/demo/stock_lot_condition_demo.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+    <record id="condition_good" model="stock.lot.condition">
```
pedrobaeza: Please check if there's any standard classification for this, and at least I'm thinking this should have 4 levels.

victoralmau: I don't know of any existing standard that could be useful.

What do you mean with 4 levels? (Perhaps creating 4 conditions in demo data?)

It is important to remember that the objective of this module is to create the conditions that each person considers/needs.

pedrobaeza: Yes, I mean four records. And I would like to have them as data if we start from a more or less standarized classification of the condition.

stock_lot_condition/models/stock_lot_condition.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class StockLotCondition(models.Model):
+    _name = "stock.lot.condition"
+    _description = "Stock Lot Condition"
+    _order = "sequence, id"
+
+    sequence = fields.Integer()
```
pedrobaeza, line 12: Add `active` field for archiving conditions.

stock_lot_condition/demo/stock_lot_condition_demo.xml
pedrobaeza: Check https://share.google/aimode/y4CCbOYSGjgC4iRAV for possible classifications.

stock_lot_condition/models/stock_lot_condition.py
```diff
@@ -0,0 +1,13 @@
...
+    name = fields.Char(required=True, translate=True)
```
pedrobaeza, line 13: I would also add a `description` field for adding details about the condition.

pedrobaeza (approved):

stock_lot_condition/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_stock_lot_condition_user,access_stock_lot_condition_user,model_stock_lot_condition,base.group_user,1,0,0,0
+access_stock_lot_condition_admin,access_stock_lot_condition_admin,model_stock_lot_condition,stock.group_stock_user,1,1,1,1
```
christian-ramos-tecnativa: Im not sure about allowing regular stock users to create these conditions since IMO they shouldn't be created on the fly. WDYT 
```suggestion
access_stock_lot_condition_user,access_stock_lot_condition_user,model_stock_lot_condition,base.group_user,1,0,0,0
access_stock_lot_condition_admin,access_stock_lot_condition_admin,model_stock_lot_condition,stock.group_stock_user,1,0,0,0
access_stock_lot_condition_admin,access_stock_lot_condition_admin,model_stock_lot_condition,stock.group_stock_manager1,1,1,1
```

christian-ramos-tecnativa (approved):
LGTM

PR #2454, opened 2025-11-06
[12.0][FIX] stock_available_unreserved: handle merge of quants

MiquelRForgeFlow:
When you change `quantity` by query, the `unreserved_quantity` is not updated.


rousseldenis:
@MiquelRForgeFlow Ugh. IMHO, this code has no place here as it is normal. You should update the quantity through ORM

PR #2460, opened 2025-11-13
[18.0][MIG] stock_picking_batch_packaging_qty: Migration to 18.0

Xino61122:
[18.0][MIG]stock_picking_batch_packaging_qty: Migration to 18.0

https://www.loom.com/share/fcba7b35f4a54df4b8878be9c83a6f42

@moduon MT-11845

PR #2465, opened 2025-11-18
[IMP][18.0]stock_move_packaging_qty: move columnt packaging

Xino61122:
In version 16, the packaging column is in a different position.
In Odoo 18, it also appears as it did in v16, so to improve the
adaptation of the migration to v18 and make it less confusing,
the column is placed in the same location.

Before > 

image

After > 

image

@moduon MT-11845

### DIFF
--- a/stock_lot_image/i18n/it.po
+++ b/stock_lot_image/i18n/it.po
@@ -6,129 +6,131 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 18.0\n"
 "Report-Msgid-Bugs-To: \n"
-"Last-Translator: Automatically generated\n"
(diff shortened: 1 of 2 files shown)


yajo (changes requested):
Please don't merge this improvement with https://github.com/OCA/stock-logistics-warehouse/pull/2460. You have to remove those commits from this PR.

Thanks!

yajo:
If this is ready to review, please take it out of the draft status.

stock_lot_image/i18n/it.po
yajo, line 10: Something wrong is happening here. This file should not be included in the PR...

yajo (changes requested):
It looks to me like the order is not still the same: 

Image

PR #2472, opened 2025-11-26
[19.0][MIG] stock_picking_supplier_ref: Migration to 19.0

sergioM-S73:


ioans73 (approved):
Functional review 👍🏻

stock_picking_supplier_ref/tests/test_supplier_ref.py
```diff
@@ -0,0 +1,72 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestStockPickingSupplierReference(TransactionCase):
```
BhaveshHeliconia:
```suggestion
class TestStockPickingSupplierReference(Basecommon):
```

stock_picking_supplier_ref/tests/test_supplier_ref.py
```diff
@@ -0,0 +1,72 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        # Create a partner
+        cls.partner = cls.env["res.partner"].create(
```
BhaveshHeliconia:
```suggestion
```

BhaveshHeliconia (changes requested):
Consider using BaseCommon instead of TransactionCase. In this case, there’s no need to redefine res.partner is already available via BaseCommon. This would help simplify and optimize the test code.

stock_picking_supplier_ref/tests/test_supplier_ref.py
sergioM-S73: @BhaveshHeliconia done!

BhaveshHeliconia (approved):

Reyes4711-S73 (approved):
LGTM

PR #2474, opened 2025-11-27
[18.0][MIG] stock_vlm_mgmt: Migration to version 18.0

christian-ramos-tecnativa:
@Tecnativa TT57303
 
A couple of changes were needed in the python to handle the new `picked` and `quantity` fields in `stock.move.lines` and some owl reformat for 18.0 version.

Functionaly the module remains exactly the same as in 16.0 version.

@pedrobaeza


stock_vlm_mgmt/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Vertical Lift Module management",
     "summary": "Light self contained alternative for VLM integrations",
-    "version": "16.0.1.0.2",
+    "version": "18.0.1.0.2",
```
pedrobaeza: 18.0.1.0.0

pedrobaeza (approved):
Code review

stock_vlm_mgmt/__manifest__.py
christian-ramos-tecnativa: Nice catch

stock_vlm_mgmt/models/stock_location.py
```diff
@@ -0,0 +1,145 @@
...
+            password=self.vlm_password,
+            **options,
+        )
+        response = vlm_connector.request_operation(data)
+        response_code = response.get("code", "")
+        # These negative codes are issued by the connector and they're common connection
+        # problems.
+        if response_code == "-1":
+            raise UserError(
+                _("The connection was refused by the VLM and couldn't be stablished.")
```
carlos-lopez-tecnativa: And could you change all occurrences to use the new translation method?

```suggestion
                self.env._("The connection was refused by the VLM and couldn't be stablished.")
```

stock_vlm_mgmt/models/stock_location_vlm_tray_type.py
```diff
@@ -0,0 +1,53 @@
...
+    tray_matrix = fields.Serialized(compute="_compute_tray_matrix")
+
+    @api.depends("width", "cols")
+    def _compute_width_per_cell(self):
+        for record in self:
+            width = record.width
+            if not width:
+                record.width_per_cell = 0.0
+                continue
+            record.width_per_cell = width / record.cols
```
carlos-lopez-tecnativa, line 31: Please prevent division by zero.
```suggestion
            record.width_per_cell = record.cols and (width / record.cols) or 0
```
image

stock_vlm_mgmt/models/stock_location_vlm_tray_type.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @api.depends("depth", "rows")
+    def _compute_depth_per_cell(self):
+        for record in self:
+            depth = record.depth
+            if not depth:
+                record.depth_per_cell = 0.0
+                continue
+            record.depth_per_cell = depth / record.rows
```
carlos-lopez-tecnativa, line 40:
```suggestion
            record.depth_per_cell = record.rows and (depth / record.rows) or 0
```
image

stock_vlm_mgmt/wizards/stock_vlm_task_action.py
```diff
@@ -0,0 +1,186 @@
...
+                "pos_y": task.pos_y,
+                "quantity_done": task.quantity_done,
+                "quantity_pending": task.quantity_pending,
+                "tray_id": task.tray_id.id,
+                "location_id": task.location_id.id,
+                "tray_type_id": task.tray_id.tray_type_id.id,
+            }
+        )
+        if not self.env.context.get("default_vlm_task_ids"):
+            vals.update({"vlm_task_ids": [(6, 0, task.ids)]})
```
carlos-lopez-tecnativa: Could you change it to use the [Command](https://github.com/odoo/odoo/blob/0622c210450a69e4331ba38bc1c9b7f389566ff3/odoo/fields.py#L4240
) class? It is more readable.
```suggestion
            vals.update({"vlm_task_ids": [Command.set(task.ids)]})
```

stock_vlm_mgmt/wizards/stock_vlm_task_action_views.xml
```diff
@@ -0,0 +1,149 @@
...
+<odoo>
+    <record id="vlm_task_action_form_view" model="ir.ui.view">
+        <field name="model">stock.vlm.task.action</field>
+        <field name="arch" type="xml">
+            <form create="false" string="VLM Operation action needed">
+                <field name="state" invisible="1" />
+                <sheet>
+                    <field name="next_vlm_task_id" invisible="1" />
+                    <field name="previous_vlm_task_id" invisible="1" />
+                    <field name="splitted_vlm_task_id" invisible="1" />
```
carlos-lopez-tecnativa: If these fields are used only in readonly, invisible, or domain attributes, they can be removed from the view. Odoo adds the fields automatically, as indicated in the migration guides. This keeps the view as simple as possible.

image

stock_vlm_mgmt/wizards/stock_vlm_task_action_views.xml
```diff
@@ -0,0 +1,149 @@
...
+                            type="object"
+                            class="btn btn-secondary"
+                            invisible="not next_vlm_task_id or state in ('done', 'skipped')"
+                        />
+                        <button
+                            string="Next 🡺"
+                            disabled="1"
+                            class="btn btn-secondary"
+                            invisible="next_vlm_task_id or state != 'done'"
+                        />
```
carlos-lopez-tecnativa, line 129: Is this button working?
I don’t see the `name` used to call a Python function or action. Additionally, the invisible attribute was changed during the migration, but it is not equivalent to the previous version. Why?
`state != 'done'` is not equal to `['state', '=', 'done']`.

image

image

carlos-lopez-tecnativa (changes requested):

stock_vlm_mgmt/models/stock_location.py
christian-ramos-tecnativa: Absolutely, done

stock_vlm_mgmt/models/stock_location_vlm_tray_type.py
christian-ramos-tecnativa: I added the condition to the `if` above, IMO it looks cleaner

stock_vlm_mgmt/wizards/stock_vlm_task_action.py
christian-ramos-tecnativa: Totally agree, done

stock_vlm_mgmt/wizards/stock_vlm_task_action_views.xml
christian-ramos-tecnativa: Didn't find any js that make them visible, IMO thats the only rational reason to keep this perma-invisible fields in views. I removed every `invisible="1"` field from views.

stock_vlm_mgmt/wizards/stock_vlm_task_action_views.xml
christian-ramos-tecnativa: I misswrite the domain but now it should be correct. 
Regarding the button itself, I believe its purpose is purely visual since it is `diabled="1"`

christian-ramos-tecnativa:
@carlos-lopez-tecnativa all of your comments have been resolved
Thanks for your review!

carlos-lopez-tecnativa (approved):
Code review LGTM.
Non-blocking comment: I think some administrative commits can be squashed.

image

christian-ramos-tecnativa:
Squash done! I agree is better keeping the history as small as posible

pedrobaeza:
@christian-ramos-tecnativa on runboat, I have now this JS error:

image

Please check

PR #2478, opened 2025-11-28
[18.0][FIX] stock_inventory: auto done not work with more than 1 adjustment

ArnauCForgeFlow:
When using barcode to do more than 1 adjustment at the same time, the auto completion to done is not working properly, and only sets to done the last adjustment:

As you can see we have 2 adjustments:

We complete the 2 adjustments using the barcode app:

Whenever we apply, only 1 is marked as done, when both should be done:

As you can see the adjustment that is in progress, is done:


JoanSForgeFlow (approved):
LGTM!

LauraCForgeFlow (approved):

JoanSForgeFlow:
Hi @JordiBForgeFlow,

When you have a moment, could you please merge this?

PR #2482, opened 2025-12-03
[18.0][IMP] stock_secondary_unit: add secondary unit in stock quant

AungKoKoLin1997:
This PR adds secondary qty and secondary uom in the stock quant.

@qrtl QT6192 QT6876


yostashiro:
Depends on https://github.com/OCA/product-attribute/pull/2166.

AungKoKoLin1997:
I changed `secondary_uom_id` to a stored related field to avoid `secondary_uom_qty` being computed with an empty `secondary_uom_id` during `stock.quant` creation, which caused incorrect values to be stored.

nobuQuartile (approved):
LGTM: Code review and functional review

AungKoKoLin1997:
Added a new commit for allowing to enter the counted quantity of an inventory adjustment in the secondary unit.

stock_secondary_unit/models/stock_move.py
```diff
@@ -43,6 +42,33 @@ def _recompute_state(self):
...
+            return vals_list
+        for vals in vals_list:
+            vals["secondary_uom_qty"] = move._convert_qty_to_secondary_uom(
+                vals["quantity"]
+            )
+            vals["secondary_uom_id"] = {
+                "id": move.secondary_uom_id.id,
+                "display_name": move.secondary_uom_id.display_name,
+            }
+        return vals_list
```
nobuQuartile: You should comment on why we need this code.

stock_secondary_unit/models/stock_quant.py
```diff
@@ -0,0 +1,87 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if "secondary_uom_inventory_quantity" not in vals:
+                continue
+            product = self.env["product.product"].browse(vals.get("product_id"))
+            secondary_uom = product.stock_secondary_uom_id
+            if not self._is_convertible_secondary_uom(secondary_uom):
+                continue
+            vals["inventory_quantity"] = self._convert_secondary_uom_qty_to_qty(
```
nobuQuartile, line 79: If both inventory_quantity and secondary_uom_inventory_quantity are provided, does this mean that secondary_uom_inventory_quantity takes precedence?

Shouldn't inventory_quantity take precedence instead?

stock_secondary_unit/models/stock_move.py
```diff
@@ -43,6 +42,33 @@ def _recompute_state(self):
...
+                ("picking_id", "=", picking_id),
+                ("product_id", "=", product_id),
+                ("secondary_uom_id", "!=", False),
+            ],
+            limit=1,
+        )
+        if not move or move.secondary_uom_id.dependency_type == "independent":
+            return vals_list
+        for vals in vals_list:
+            vals["secondary_uom_qty"] = move._convert_qty_to_secondary_uom(
```
nobuQuartile:
```suggestion
            vals["secondary_uom_qty"] = move.secondary_uom_id._get_secondary_qty(vals["quantity"], product.uom_id)
```

nobuQuartile (commented):
Partial review

stock_secondary_unit/models/stock_move.py
AungKoKoLin1997: I don't think so since it easy to understand by looking at the code.

stock_secondary_unit/models/stock_move.py
AungKoKoLin1997: We have a line record with secondary_uom_id already populated. So, we should use `_convert_qty_to_secondary_uom`.

stock_secondary_unit/models/stock_quant.py
AungKoKoLin1997: I think it is kind of over engineering. I mean there is no chance to fill in both field when you create new record in the real operation even we can fill in both field as technically. Then, we can't say inventory_quantity should always take precedence. The rule may different depends on the users. So, I will keep current design.

stock_secondary_unit/models/stock_move.py
nobuQuartile: The point isn't which record holds `secondary_uom_id` — it's which UoM the input quantity is expressed in.

`_convert_qty_to_secondary_uom()` treats `qty` as being in the record's own `uom_field`, which for `stock.move` is `product_uom`. But the vals built by core are in the **product's** UoM: `action_generate_lot_line_vals()` sets `product_uom_id = product.uom_id`, and `default_quantity` comes from `move.product_qty`, whose help reads "Quantity in the default UoM of the product".

So when the move's UoM differs from the product's, the quantity is converted a second time. Reproduced on 18.0 — product in Units, move in Dozens, secondary unit with factor 2:

    quantity = 1 (Unit)  ->  secondary_uom_qty = 6.0   (expected 0.5)

`_convert_qty_to_secondary_uom(1)` reads the 1 as 1 Dozen, converts it to 12 Units, then divides by 2. If the client sends `secondary_uom_qty` on save, that 6.0 is stored against a "Pack of 2" unit — a line whose unit and quantity disagree.

Passing the source UoM explicitly avoids this (corrected from my earlier suggestion, which referenced an undefined `product`):

```python
vals["secondary_uom_qty"] = move.secondary_uom_id._get_secondary_qty(
    vals["quantity"], move.product_id.uom_id
)
_get_secondary_qty() takes the source UoM as an argument, so it cannot be misread. The existing test only covers a product and move both in Units, which is exactly the case that hides this.

stock_secondary_unit/models/stock_quant.py
nobuQuartile: Fair enough that there's no universally right answer here — but the same question is answered the other way in `mrp_secondary_unit` (OCA/manufacture#1855), where `_is_driven_by_secondary_uom()` ignores the secondary quantity whenever the primary one is given. Measured on 18.0 with both fields set (primary 99, secondary 3):

    stock.quant.inventory_quantity = 6.0    (secondary wins)
    mrp.production.product_qty     = 99.0   (primary wins)

I'm not asking for a branch — a one-line comment saying which one this module takes as authoritative would be enough. Two modules from the same series resolving the same conflict in opposite directions is what I'd like to avoid.

stock_secondary_unit/models/stock_move.py
nobuQuartile: When the move has a unit, this method is redundant on save: it returns "Pack of 10" and the stored line comes out "Pack of 2" — the parent move's unit. Its only durable job is what the dialog shows before save, since the client builds rows from these vals without a round trip (`// create records directly from values to bypass onchanges`, `generate_serial.js`). Fair reason to have it, but not recoverable from the code — hence the comment request.

stock_secondary_unit/models/stock_quant.py
AungKoKoLin1997: I will make the same approach as `mrp_secondary_unit`.

stock_secondary_unit/models/stock_move.py
AungKoKoLin1997: You are right. I misunderstood.

nobuQuartile (approved):
LGTM: Code review and functional review

PR #2486, opened 2025-12-04
[18.0][FIX] stock_inventory: prevent auto_state_to_done to empty recordsets

ArnauCForgeFlow:
Fixed an issue where action_auto_state_to_done() was being executed on an empty stock.inventory() recordset, caused by adding undefined adjustments to the `adjustments` list. Now only valid inventory records are added, preventing errors during auto-complete.

### DIFF
--- a/stock_inventory/models/stock_quant.py
+++ b/stock_inventory/models/stock_quant.py
@@ -21,8 +21,9 @@ class StockQuant(models.Model):
     def _apply_inventory(self):
         res = super()._apply_inventory()
         record_moves = self.env["stock.move.line"]
-        adjustments = []
-        for rec in self:
+        adjustments_to_process = self.env["stock.inventory"]
+
+        for rec in self.filtered("current_inventory_id"):
             adjustment = rec.current_inventory_id
             moves = record_moves.search(
                 [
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):
Thanks! For more modern Odoo-esque code, maybe change line 25 to say `for rec in self.filtered('current_inventory_id')`. You can also skip keeping track of unique inventories by making `action_auto_state_to_done` iterate over its `self` and then do `self.current_inventory_id.action_auto_state_to_done()`.

But this will work as well, of course.

DavidJForgeFlow:
@ArnauCForgeFlow could you take a look at the feedback?

ArnauCForgeFlow:
@StefanRijnhart Thanks for your review, could you check if now it looks better?

stock_inventory/models/stock_quant.py
```diff
@@ -55,10 +54,7 @@ def _apply_inventory(self):
             )
             rec.to_do = False
             rec.current_inventory_id = False
-            if adjustment not in adjustments:
-                adjustments.append(adjustment)
-        if adjustments and self.env.company.stock_inventory_auto_complete:
-            for adjustment in adjustments:
+            if self.env.company.stock_inventory_auto_complete:
```
StefanRijnhart: Not sure this is correct if `self` consists of multiple quants linked to the same inventory adjustment. The adjustment will be set to done after the first quant.

ArnauCForgeFlow: It won't, since the method action_auto_state_to_done method explicitly checks that all quants are processed (to_do = False) before setting the inventory to done.

https://github.com/OCA/stock-logistics-warehouse/blob/078fdd6ae92d3efac66e6bb465682bd218e9153f/stock_inventory/models/stock_inventory.py#L349

StefanRijnhart: OK, but it does not look right, so it may confuse generations of developers to come. Would very much prefer to have this called outside the loop.

ArnauCForgeFlow: Done

StefanRijnhart (approved):

LoisRForgeFlow (approved):
/ocabot merge patch

PR #2425, opened 2025-09-12
[18.0] [MIG] stock_mts_mto_rule : migrate to 18.0

apik-mgu:

PR #2489, opened 2025-12-18
[18.0][MIG] stock_mts_mto_rule: Migration to 18.0

ypapouin:
New PR written from scratch since https://github.com/OCA/stock-logistics-warehouse/pull/2425 made the tests work with the code instead of the other way around, there were too many things to re-fix.

- The dummy route from tests is now created with `location_dest_from_rule=True` to ensure that stock move's `location_dest_id` is properly computed.
- The dummy route picking type set to "IN" since its way is supplier -> stock (longstanding test bug)
- Small code and test refactoring for easier readability

Note that I made a port of https://github.com/OCA/stock-logistics-warehouse/pull/2098 + https://github.com/OCA/stock-logistics-warehouse/pull/2331 that could be added in this PR if necessary (waiting for a feedback from original author)

Another thing that made me scratch my head is the introduction of the `location_final_id` on the `stock.move` model that I though was a way to know the final product location when there were chained moves, but that's not the case since openupgrade just rename `location_dest_id` to `location_final_id` and let the `location_dest_id` be auto-computed.


stock_mts_mto_rule/models/stock_rule.py
```diff
@@ -0,0 +1,118 @@
...
+        """Shortcut to run the MTO rule linked to this MTS+MTO rule."""
+        self.mto_rule_id.ensure_one()
+        return self._run_subrule_action(procurement, self.mto_rule_id)
+
+    @api.constrains("action", "mts_rule_id", "mto_rule_id")
+    def _check_mts_mto_rule(self):
+        for rule in self:
+            if rule.action == "split_procurement":
+                if not rule.mts_rule_id or not rule.mto_rule_id:
+                    msg = _(
```
arielbarreiros96: It's recommended to use self.env._ instead, I've seen the _() in some other places

arielbarreiros96:
I've been testing the module and it seems to be working fine when combined with purchases and manufacturing, stock movements makes sense, from what I can see

arielbarreiros96 (changes requested):

stock_mts_mto_rule/models/stock_rule.py
ypapouin: Thanks for the tip, I didn't know that.

Saurat:
@ypapouin 
Is the addon migrated to V18? i wanted to provide some help for migrating to V19.

HectorMerazChavez (approved):
Tested locally on Odoo 18.0, all 13 tests pass successfully. Code changes look clean and well structured.

alan196 (approved):
Functional review

aleboldro02:
Can anyone merge please?

rousseldenis:
@arielbarreiros96 Could you update your review ?

rousseldenis (approved):

rousseldenis:
/ocabot merge nobump

Moving forward on this as all comments have been solved.

PR #2490, opened 2025-12-19
[15.0][FIX] stock_secondary_unit: Secondary uom qty is accumulated when only product_uom_qty is changed in PO

sergio-teruel:
cc @Tecnativa TT59838

ping @CarlosRoca13 @carlosdauden

### DIFF
--- a/stock_secondary_unit/models/stock_move.py
+++ b/stock_secondary_unit/models/stock_move.py
@@ -51,12 +51,6 @@ class StockMove(models.Model):
             )
         return vals
 
-    def _merge_moves_fields(self):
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

pedrobaeza (commented):
Shouldn't this sum both secondary quantities?

sergio-teruel:
Hi, no, because both stock move have the same second qty data. So thia is the actual issue , we sum the secondary qty and is not ok..

pedrobaeza (approved):
OK, I see that in `_merge_moves` you are already adding secondary quantities.

/ocabot merge patch

PR #2491, opened 2025-12-23
[18.0][MIG] stock_lot_filter_available: Migration to 18.0

Tisho99:


Tisho99:
Closed because module is no longer needed: https://github.com/OCA/stock-logistics-warehouse/issues/1883#issuecomment-3700024778

PR #2495, opened 2026-01-02
[18.0][MIG] stock_package_autoload: migration to 18.0

quirino95:
This module allows to automatically add the content of a package to each Transfer line.

This module was introduced in 14.0 : https://github.com/OCA/stock-logistics-warehouse/pull/2027
Then a migration (with some improvements) was done for 16.0: https://github.com/OCA/stock-logistics-warehouse/pull/2218


stock_package_autoload/models/stock_move.py
```diff
@@ -0,0 +1,69 @@
...
+            data = common_line_data.copy()
+            data.update(
+                {
+                    "quantity": quant.quantity,
+                    "product_uom_id": quant.product_uom_id.id,
+                    "lot_id": quant.lot_id.id,
+                }
+            )
+            data_list.append(data)
+        self.move_line_ids = [fields.Command.create(data) for data in data_list]
```
HekkiMelody: **chore**: This change fixes an actual issue that was present but was not picked up by tests. I think it'd be best to refactor the existing test to use the Form class, which should hopefully help reproduce the error.

**suggestion**: this can be refactored slightly to use a single list instead of creating a list and then unpacking it into a second list


```suggestion
        command_list = []
        for quant in product_quants:
            data = common_line_data.copy()
            data.update(
                {
                    "quantity": quant.quantity,
                    "product_uom_id": quant.product_uom_id.id,
                    "lot_id": quant.lot_id.id,
                }
            )
            command_list.append(fields.Command.create(data))
        self.move_line_ids = command_list
```

HekkiMelody (commented):
Generally looks good, and seems to work as expected

stock_package_autoload/tests/test_stock_package_autoload.py
```diff
@@ -57,36 +57,29 @@ def setUpClass(cls):
...
         self.assertEqual(package_domain, self.package)
-        move.load_products_from_package_id = self.package
-        move._onchange_load_products_from_package_id()
+
+        with Form(
+            move[0],
+            view=self.env.ref("stock_package_autoload.view_stock_move_operations"),
+        ) as move_f:
+            move_f.load_products_from_package_id = self.package
+        move = move_f.save()
```
HekkiMelody: **chore**: if you use `Form` as a context manager, you don't need to explicitly save, it already saves automatically

https://github.com/OCA/OCB/blob/18.0/odoo/tests/form.py#L50-L64

stock_package_autoload/tests/test_stock_package_autoload.py
```diff
@@ -100,8 +93,13 @@ def test_autoload_package(self):
         self.assertLess(len(move.move_line_ids), n_move_lines)
         # by selecting the same package again, only the missing serial will be added to
         # the move lines
-        move.load_products_from_package_id = self.package
-        move._onchange_load_products_from_package_id()
+        with Form(
+            move[0],
+            view=self.env.ref("stock_package_autoload.view_stock_move_operations"),
+        ) as move_f:
+            move_f.load_products_from_package_id = self.package
```
HekkiMelody, line 99: **chore**: if you use `Form` as a context manager, you don't need to explicitly save, it already saves automatically

https://github.com/OCA/OCB/blob/18.0/odoo/tests/form.py#L50-L64

HekkiMelody (approved):
Code review, LGTM

setup/stock_package_autoload/odoo/addons/stock_package_autoload
SirPyTech, line 1: **chore**: Please remove the created folder in `setup`: this is the only module in `18.0` having it so it probably isn't needed.

stock_package_autoload/__manifest__.py
SirPyTech, line 1: **chore**: This not a migration because this module was never merged, please clarify it in the PR title/description.

As a reviewer, this is an important clarification: in a migration I only check the commit history and the migration commits because the module had been already approved and merged in a previous version; here this module had never been approved so I have to review all of it.

It is fine too keep the history from previous PRs to track the module's evolution though.

stock_package_autoload/views/stock_move_view.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_stock_move_operations" model="ir.ui.view">
+        <field name="name">stock.move.operations.form</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_stock_move_operations" />
+        <field name="arch" type="xml">
+            <xpath expr="//group/group/.." position="inside">
```
SirPyTech, line 8: **suggestion**: Can we use a more specific `xpath`? If some other module adds a `group` it's not clear where these new fields will go.

For instance, if you want this to go in the same group where `product_id` is, you could use something like `//fields[@name='product_id']/..`.

stock_package_autoload/views/stock_move_view.xml
```diff
@@ -0,0 +1,21 @@
...
+                <group>
+                    <field name="package_domain" invisible="1" />
+                    <field
+                        name="load_products_from_package_id"
+                        domain="package_domain"
+                        groups="stock.group_tracking_lot"
```
SirPyTech, line 14: **suggestion**: when the user does not have the security group `stock.group_tracking_lot` we are adding an empty `group` tag and this could mess up the UI, please move the `groups` attribute to the `group` tag so the whole group is hidden when needed.

This could also save us the trouble of needing this piece of code in the domain computation:
```py
        """
        There's no need to compute this field if the current user doesn't have the
        necessary group
        """
        if not self.env.user.has_group("stock.group_tracking_lot"):
            self.write({"package_domain": "[]"})
            return
```

stock_package_autoload/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,3 @@
+- [PyTech-SRL](info@pytech.it):
+
+  > - Sebastiano Picchi \<sebastiano.picchi@pytech.it\>
```
SirPyTech, line 3: **suggestion**: This is currently displayed as a quote: Image
I think it could be fixed by removing the `>`:
```suggestion
  - Sebastiano Picchi \<sebastiano.picchi@pytech.it\>
```

SirPyTech (changes requested):
Thanks for the PR!
I have reviewed the code and it looks good overall, just a few suggestions.

I also tried it and I have found the following use-case: when a move line is selected with the usual "Add line", it cannot be added again; on the other hand, adding all the quants of a package adds it again, see the following video:

https://github.com/user-attachments/assets/3f5ba5cc-7b40-4ead-805d-f56de7f5b914

I think this should not happen, could you please have a look?

SirPyTech:
This can be closed because superseded by #2621.

PR #2507, opened 2026-01-22
[18.0][IMP] stock_move_packaging_qty: move columnt packaging

Andrii9090-tecnativa:
In version 16, the packaging column is in a different position. In Odoo 18, it also appears as it did in v16, so to improve the adaptation of the migration to v18 and make it less confusing, the column is placed in the same location.
v16
image
v18 now
image

MT-11845
@moduon

### DIFF
--- a/stock_move_packaging_qty/views/stock_picking_view.xml
+++ b/stock_move_packaging_qty/views/stock_picking_view.xml
@@ -7,13 +7,9 @@
         <field name="arch" type="xml">
             <!-- Packaging quantity -->
             <xpath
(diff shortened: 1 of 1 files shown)


yajo (approved):
Code review OK.

Please @Gelojr could you do the functional review?

chienandalu (approved):

Shide (approved):
Review "columnt" word on commit and PR

Gelojr (approved):
Nice job @Andrii9090

PR #2509, opened 2026-01-29
[18.0][IMP] stock_vertical_lift: shared stock location for shuttles

vvrossem:


stock_vertical_lift/models/stock_location.py
```diff
@@ -28,15 +28,15 @@ class StockLocation(models.Model):
...
+        comodel_name="vertical.lift.shuttle", inverse_name="shared_storage_location_id"
     )
     # compute the unique shuttle for any shuttle, tray or cell location, by
     # going through the parents
     vertical_lift_shuttle_id = fields.Many2one(
         comodel_name="vertical.lift.shuttle",
         compute="_compute_vertical_lift_shuttle_id",
         recursive=True,
-        store=True,
+        store=False,
```
SilvioC2C, line 39: 2 quick (non blocking) comments:
- the `compute` method doesn't depend anymore on the location's parent value `location_id.vertical_lift_shuttle_id`, but `recursive=True` will still force Odoo to recompute this field on single records instead of a whole recordset (see https://github.com/odoo/odoo/blob/18.0/odoo/fields.py#L1455-L1463 and https://github.com/odoo/odoo/blob/18.0/odoo/fields.py#L1303-L1309), leading to worse performances overall => we can remove `recursive=True`
- since we're moving from a stored to a non-stored field, could be nice to have a `search` method implemented here

SilvioC2C (approved):
Nice work, LGTM overall!

gurneyalex:
superseeded by https://github.com/OCA/stock-logistics-warehouse/pull/2646

PR #2513, opened 2026-01-30
[19.0][ADD] stock_lot_warranty

MarinaAForgeFlow:
This module aims to provide a foundation for warranty management at the lot level.

There is an existing OCA module in the rma repository for version 15 that also addresses warranties. However, it is quite simplistic and does not fully align with common warranty processes. Building over that existing module would likely involve working against its design, which is why this new module was created from scratch.

By separating the logic into multiple methods, this module is highly extensible, allowing easy customization for RMA or repair additional workflows.

Depends on https://github.com/OCA/rma/pull/506


stock_lot_warranty/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,20 @@
...
+the product is sold again.
+
+This module also allows users to manually edit warranty dates directly on the lot/serial form,
+with all changes logged in the chatter for full traceability.
+
+The logic is the most basic logic for warranty management, and has been implemented in different hook methods,
+allowing other modules to easily override or extend warranty management, such as managing returns in RMA or
+repair workflows.
+
+
```
MiquelRForgeFlow, line 20: Too much blank space.

stock_lot_warranty/models/stock_lot.py
```diff
@@ -0,0 +1,102 @@
...
+    customer_warranty_end_date = fields.Date(
+        string="Customer Warranty End", tracking=True
+    )
+
+    vendor_warranty_start_date = fields.Date(
+        string="Vendor Warranty Start", tracking=True
+    )
+    vendor_warranty_end_date = fields.Date(string="Vendor Warranty End", tracking=True)
+
+    is_serial_tracked = fields.Boolean(compute="_compute_is_serial_tracked")
```
DavidJForgeFlow: Why not making it related to product_id.tracking?

stock_lot_warranty/models/stock_move_line.py
```diff
@@ -0,0 +1,29 @@
...
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _action_done(self):
+        res = super()._action_done()
+
+        for line in self:
+            lot = line.lot_id
+            if not lot or lot.product_id.tracking != "serial":
```
DavidJForgeFlow: here you can use your new field, or the related one!

stock_lot_warranty/models/stock_move_line.py
```diff
@@ -0,0 +1,29 @@
...
+                continue
+
+            src_usage = line.location_id.usage
+            dest_usage = line.location_dest_id.usage
+
+            if dest_usage == "customer":
+                lot._update_customer_warranty_on_delivery(line)
+            elif src_usage == "customer":
```
DavidJForgeFlow: Maybe here you should use plain ifs, as you could be facing some dropshipping.

DavidJForgeFlow (changes requested):

stock_lot_warranty/models/stock_move_line.py
MarinaAForgeFlow: You are right, didn't thought about dropshipping. Thanks!

MarcGForgeFlow:
Two commits added on top of this PR  with the reasoning behind them:

**1. `[IMP] show warranty dates for any tracking type`**

The dates were hidden unless the product was tracked by serial number, so a lot or an
untracked product could not have its dates filled in by hand, even though the dates
themselves are stored on the lot and can be read by other modules without looking at the
tracking. That leaves the lot form showing nothing while the warranty is in fact set.

The fields are now shown on every lot. The automatic computation is untouched and still
applies to serial numbers only, for the reason already stated in the module: a lot can be
received or delivered on several dates and there is no single date the whole lot could
derive its warranty from.

**2. `[IMP] do not compute when the product is not configured`**

`_compute_warranty_end_date` returns nothing when the product, or the supplier info, has no
warranty duration, and that empty value was written to the lot regardless. A delivery or a
receipt therefore replaced whatever the lot held with an empty end date, and left a start
date behind with no end date to go with it.

The rule is about the computation, not about precedence: when a duration is configured the
behaviour is exactly the same as before and computed values overwrite manual ones as usual.
When it is not configured nothing is computed and nothing is written, so a value entered by
hand is preserved.

Both commits come with their test.

PR #2515, opened 2026-02-05
[14.0][UPD] stock_warehouse_calendar_orderpoint: method belongs to the Warehouse

florentx:
Method is related to the Warehouse, not the Replenishment line.

It is private method, we don't have to preserve the old method.

### DIFF
--- a/stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
+++ b/stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
@@ -57,6 +57,14 @@ class StockWarehouse(models.Model):
             self = self.with_context(force_wh_company=vals["company_id"])
         return super().create(vals)
 
+    def _get_next_reordering_date(self):
+        self.ensure_one()
+        now = fields.Datetime.now()
+        calendar = self.orderpoint_calendar_id
+        # TODO: should we take into account days off of the reordering calendar with
+        #  'compute_leaves=True' here?
+        return calendar and calendar.plan_hours(0, now) or now
+
     def _get_lead_date(self, date_order, lead_days):
         self.ensure_one()
         # Get the WH calendar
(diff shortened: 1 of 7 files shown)


twalter-c2c (approved):
Looks good to me.

stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
```diff
@@ -57,6 +57,14 @@ def create(self, vals):
             self = self.with_context(force_wh_company=vals["company_id"])
         return super().create(vals)
 
+    def _get_next_reordering_date(self):
+        self.ensure_one()
+        now = fields.Datetime.now()
+        calendar = self.orderpoint_calendar_id
+        # TODO: should we take into account days off of the reordering calendar with
+        #  'compute_leaves=True' here?
```
mmequignon, line 65: yes.

mmequignon (changes requested):
Also, this is a breaking change.
You're dropping a method, any module depending on this one might be broken because a method disappeared.
Please restore that method, and eventually make it deprecated, and write smth in the roadmap.

florentx:
> Also, this is a breaking change. You're dropping a method, any module depending on this one might be broken because a method disappeared. Please restore that method, and eventually make it deprecated, and write smth in the roadmap.

Because this method was private, I thought it could be moved without a deprecation.
I will restore the method.

stock_warehouse_calendar_orderpoint/models/stock_warehouse.py
florentx: `# TODO`

twalter-c2c:
Only a small warning for squashing the fixup commit - I think you targeted the wrong commit with the fixup, so please make sure you squash to the correct one.

rousseldenis (approved):
LGTM

PR #2516, opened 2026-02-06
[18.0][FIX] stock_move_purchase_uom: also round if there is the same uom

JoanSForgeFlow:
if po and stock uom are the same we should also appy the rounding method

### DIFF
--- a/stock_move_purchase_uom/models/stock_move.py
+++ b/stock_move_purchase_uom/models/stock_move.py
@@ -16,7 +16,6 @@ class StockMove(models.Model):
                 and move.picking_type_id
                 and move.picking_type_id.use_purchase_uom
                 and purchase_uom
-                and move.product_uom != purchase_uom
                 and not move.origin_returned_move_id
             ):
                 updated_product_uom_qty = move.product_uom._compute_quantity(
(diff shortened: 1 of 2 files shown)


stock_move_purchase_uom/models/stock_move.py
```diff
@@ -16,16 +17,23 @@ def create(self, vals_list):
...
+                    )
+                    move.product_uom = move.product_id.uom_po_id
+                    move.product_uom_qty = updated_product_uom_qty
+                else:
+                    # Same UoM but still apply rounding with the configured method
+                    move.product_uom_qty = float_round(
+                        move.product_uom_qty,
+                        precision_rounding=purchase_uom.rounding,
+                        rounding_method=move.picking_type_id.purchase_uom_rounding_method,
+                    )
```
sebalix: Question: is the "same UoM" check mandatory? Does `_compute_quantity` not able to return the value rounded as expected even with the same UoM, so we don't need to call `float_round` here?
```suggestion
                updated_product_uom_qty = move.product_uom._compute_quantity(
                    move.product_uom_qty,
                    purchase_uom,
                    rounding_method=move.picking_type_id.purchase_uom_rounding_method,
                )
                move.product_uom = move.product_id.uom_po_id
                move.product_uom_qty = updated_product_uom_qty
```

JoanSForgeFlow:
Hi @sebalix,

You're right! Removing the `move.product_uom != purchase_uom` check from the condition was sufficient.

I've applied the changes and simplified the code accordingly.

Thank you

LauraCForgeFlow (approved):
Code LGTM, thanks!

dreispt (approved):

sebalix (approved):

PR #2517, opened 2026-02-06
[16.0][FIX] stock_picking_volume: volume not computed on record creation

sergiobstoj:
- Volume field was not computed when creating pickings/moves from the UI
- Root cause: `readonly=False` on a stored computed field causes the ORM to treat the explicit `0` sent by the web client as a manual override, skipping recompute
- Removed `readonly=False` and `states` from Python field definitions
- Moved readonly logic to XML views via `attrs` (editable in draft, readonly in done/cancel)

### DIFF
--- a/stock_picking_volume/models/stock_picking.py
+++ b/stock_picking_volume/models/stock_picking.py
@@ -10,10 +10,8 @@ class StockPicking(models.Model):
 
     volume = fields.Float(
         compute="_compute_volume",
-        readonly=False,
         store=True,
         compute_sudo=True,
-        states={"done": [("readonly", True)], "cancel": [("readonly", True)]},
     )
     volume_uom_name = fields.Char(
         string="Volume unit of measure label", compute="_compute_volume_uom_name"
--- a/stock_picking_volume/models/stock_move.py
+++ b/stock_picking_volume/models/stock_move.py
@@ -10,10 +10,8 @@ class StockMove(models.Model):
 
     volume = fields.Float(
         compute="_compute_volume",
-        readonly=False,
         store=True,
         compute_sudo=True,
-        states={"done": [("readonly", True)], "cancel": [("readonly", True)]},
     )
 
     volume_uom_name = fields.Char(
--- a/stock_picking_volume/views/stock_move.xml
+++ b/stock_picking_volume/views/stock_move.xml
@@ -11,7 +11,11 @@
                 <field name="location_dest_id" position="after">
                     <label for="volume" />
                     <div>
-                        <field name="volume" class="oe_inline" />
+                        <field
+                        name="volume"
+                        class="oe_inline"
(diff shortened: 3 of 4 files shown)


alvaro-gmz:
LGTM

dreispt (approved):

stock_picking_volume/models/stock_picking.py
```diff
@@ -10,10 +10,8 @@ class StockPicking(models.Model):
 
     volume = fields.Float(
         compute="_compute_volume",
-        readonly=False,
```
lmignon, line 13: This one should remain readonly=False also on the python side otherwise it's no more possible to manually assign a value.

lmignon (changes requested):

stock_picking_volume/models/stock_picking.py
lmignon: I wonder if the fix to the problem would be to simply declare the field as precompute=True :thinking:

sergiobstoj: Thanks for the feedback @lmignon. I tested both suggestions on a Odoo 16 instance.

About `readonly=False`, it causes two problems:
1. On `create()` from UI: The web client sends volume=0.0 in the vals because the field is editable. The ORM treats it as a manual value and skips the compute. Result: volume stays 0 until you edit the record.
2. On duplicate (`copy()`): `readonly=False` on a computed stored field makes `copy=True` implicit ([fields.py:L449](https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L449)). So duplicated pickings get the old volume instead of recomputing.

`precompute=True` doesn't help in either case. When volume is already in the vals (from UI or from copy), the precompute is skipped ([models.py:L4176](https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L4148) if fname not in vals).

Removing `readonly=False` fixes both: the web client stops sending the field, and copy goes back to False (the default for computed fields).

Also, this is the way it's currently being handled in [v18.0](https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_picking_volume/models/stock_picking.py)

If keeping manual override is important, it would need `copy=False` + a `create()` override to strip volume from vals. I think the simpler approach is better here. What do you think?

sergiobstoj: Hi, @lmignon 

Did you have a chance to review my response?

lmignon: I would prefer to add one additional fields 'manual_volume' only visible in edit mode. If set (>0) the computed volume use-it, otherwise it use the computed value.

lmignon: My feeling is that it's always a bad practice to express business constrain into a view...

PR #2519, opened 2026-02-11
[14.0][IMP] stock_location_last_inventory_date: propagate data on empty loc…

devRaumschmiede:
…ation

when inventory has no products, or no differences at the location is added the inventory data is not set on location

### DIFF
--- a/stock_location_last_inventory_date/migrations/14.0.2.1.0/post-migration.py
+++ b/stock_location_last_inventory_date/migrations/14.0.2.1.0/post-migration.py
@@ -1,0 +1,10 @@
+# Copyright 2026 Raumschmiede GmbH
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.stock_location_last_inventory_date.hooks import (
+    set_inventory_date_on_child_location,
+)
+
+
+def migrate(cr, version):
+    set_inventory_date_on_child_location(cr)
--- a/stock_location_last_inventory_date/models/stock_inventory.py
+++ b/stock_location_last_inventory_date/models/stock_inventory.py
@@ -9,6 +9,20 @@ class StockInventory(models.Model):
     def _action_done(self):
         super()._action_done()
         for inventory in self:
+            done_locations = inventory._get_all_inventory_locations()
             last_inventory_date = inventory.date
-            done_locations = inventory.line_ids.location_id
             done_locations.write({"last_inventory_date": last_inventory_date})
+
+    def _get_all_inventory_locations(self):
+        """Return all locations that should be updated with the inventory date."""
+        line_locations = self.line_ids.location_id
+        if self.product_ids:
+            return line_locations
+        LOCATION = self.env["stock.location"]
+        locations = LOCATION.search(
+            [
+                ("location_id", "child_of", self.location_ids.ids),
+                ("usage", "in", ("internal", "transit")),
+            ]
+        )
+        return LOCATION.browse(list(set(locations.ids) | set(line_locations.ids)))
--- a/stock_location_last_inventory_date/hooks.py
+++ b/stock_location_last_inventory_date/hooks.py
@@ -2,6 +2,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 import logging
 
+from odoo import SUPERUSER_ID, api
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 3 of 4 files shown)


mt-software-de:
Since it was changed to a stored filed within https://github.com/OCA/stock-logistics-warehouse/pull/2222
It didn't behaved as before, because only locations are used which are used in a inventory line. 
By adding all locations (+ their children) configured on an inventory it would be fixed. 

cc @jbaudoux

mt-software-de (approved):
Thx. LGTM

jbaudoux (changes requested):
So if you have a quant on `Stock` and want to remove it with an inventory, you will update the last inventory date of all locations under `Stock`. That doesn't look right to me.

mt-software-de:
> So if you have a quant on `Stock` and want to remove it with an inventory, you will update the last inventory date of all locations under `Stock`. That doesn't look right to me.

But then you wouldn't create an inventory adjustment for Stock. Because this will create inventory lines for all quants on stock and its children. You would need to delete a lot of lines if you simply want to remove just one quant.

stock_location_last_inventory_date/models/stock_inventory.py
```diff
@@ -9,6 +9,15 @@ class StockInventory(models.Model):
...
             done_locations.write({"last_inventory_date": last_inventory_date})
+
+    def _get_all_inventory_locations(self):
+        """Return all locations that should be updated with the inventory date."""
+        line_locations = self.line_ids.location_id
+        LOCATION = self.env["stock.location"]
+        locations = LOCATION.search(
+            [("location_id", "child_of", self.location_ids.ids)]
+        )
+        return LOCATION.browse(list(set(locations.ids) | set(line_locations.ids)))
```
mt-software-de, line 25: What we should add a check on product_ids configuration. To ensure not all other locations without this product are getting updated. cc @jbaudoux 


```suggestion
        if self.product_ids:
            return line_locations
        LOCATION = self.env["stock.location"]
        locations = LOCATION.search(
            [("location_id", "child_of", self.location_ids.ids)]
        )
        return LOCATION.browse(list(set(locations.ids) | set(line_locations.ids)))
```

jbaudoux:
> > So if you have a quant on `Stock` and want to remove it with an inventory, you will update the last inventory date of all locations under `Stock`. That doesn't look right to me.
> 
> But then you wouldn't create an inventory adjustment for Stock. Because this will create inventory lines for all quants on stock and its children. You would need to delete a lot of lines if you simply want to remove just one quant.

My point is that inventory lines are telling which locations are inventoried. You can add lines or remove lines. The location in the inventory is in my opinion only a helper to preload lines.

mt-software-de:
> > > So if you have a quant on `Stock` and want to remove it with an inventory, you will update the last inventory date of all locations under `Stock`. That doesn't look right to me.
> > 
> > But then you wouldn't create an inventory adjustment for Stock. Because this will create inventory lines for all quants on stock and its children. You would need to delete a lot of lines if you simply want to remove just one quant.
> 
> My point is that inventory lines are telling which locations are inventoried. You can add lines or remove lines. The location in the inventory is in my opinion only a helper to preload lines.

Yes i understand your point. But in fact for empty Locations before and after the inventory it will be then never set. Also If you checking the Code before my change it was exaclty computed as it would be with this PR. 

We could add new field in the inventory adjustment which inticates If preloaded lines we're removed. If the field indicates that nothing was removed. Then set the date as it is done here  otherwise set it only on Locations which are listed in the lines. @jbaudoux what do you think?

siyabongasilindokuhle22-ux (approved):

stock_location_last_inventory_date/models/stock_inventory.py
mt-software-de: ping @jbaudoux

mt-software-de:
> > > > So if you have a quant on `Stock` and want to remove it with an inventory, you will update the last inventory date of all locations under `Stock`. That doesn't look right to me.
> > > 
> > > 
> > > But then you wouldn't create an inventory adjustment for Stock. Because this will create inventory lines for all quants on stock and its children. You would need to delete a lot of lines if you simply want to remove just one quant.
> > 
> > 
> > My point is that inventory lines are telling which locations are inventoried. You can add lines or remove lines. The location in the inventory is in my opinion only a helper to preload lines.
> 
> Yes i understand your point. But in fact for empty Locations before and after the inventory it will be then never set. Also If you checking the Code before my change it was exaclty computed as it would be with this PR.
> 
> We could add new field in the inventory adjustment which inticates If preloaded lines we're removed. If the field indicates that nothing was removed. Then set the date as it is done here otherwise set it only on Locations which are listed in the lines. @jbaudoux what do you think?

ping @jbaudoux

stock_location_last_inventory_date/migrations/14.0.2.1.0/post-migration.py
mt-software-de, line 1: Please update the post_init_hook and use it then here.

mt-software-de (changes requested):

mt-software-de (approved):

mt-software-de:
@jbaudoux can we finish this?

stock_location_last_inventory_date/models/stock_inventory.py
```diff
@@ -9,6 +9,17 @@ class StockInventory(models.Model):
...
             done_locations.write({"last_inventory_date": last_inventory_date})
+
+    def _get_all_inventory_locations(self):
+        """Return all locations that should be updated with the inventory date."""
+        line_locations = self.line_ids.location_id
+        if self.product_ids:
+            return line_locations
+        LOCATION = self.env["stock.location"]
+        locations = LOCATION.search(
+            [("location_id", "child_of", self.location_ids.ids)]
```
jbaudoux: What about adding
```suggestion
            [("location_id", "child_of", self.location_ids.ids), ("usage", "in", ("internal", "transit"))]
```

jbaudoux (approved):
Fine for me :)

Can you squash the commits? A single commit for this change should be enough

PR #2520, opened 2026-02-12
[17.0][ADD] stock_lot_product_category: product category available on Lots

dreispt:


stock_lot_product_category/models/stock_lot.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2026 Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class StockLot(models.Model):
+    _inherit = "stock.lot"
+
+    product_category_id = fields.Many2one(
```
rousseldenis, line 10: @dreispt How does this behave on big databases ? Shouldn't we add a pre_init hook ?

rousseldenis (approved):
One question

carlossainznetkia (commented):
Looks good overall.
I agree with the comment: adding a stored related Many2one (product_category_id) can be expensive on large DBs at install/upgrade due to the backfill/recompute.
It’d be safer to add a pre_init_hook to populate it efficiently and avoid potential long installs/locks.

PR #2521, opened 2026-02-13
[IMP] stock_location_release_channel_restriction: Check a zone of locations

rousseldenis:
When several orders are grouped into several sub-locations of a parent, check that all pending outgoing moves are concerning the same release channel.

Use case: You have a chariot with several sub-locations. That chariot (parent location) should be dedicated to a same release channel.

### DIFF
--- a/stock_location_release_channel_restriction/models/stock_location.py
+++ b/stock_location_release_channel_restriction/models/stock_location.py
@@ -88,7 +88,10 @@ class StockLocation(models.Model):
         """
         This will compute the current release channel that
         """
-        for location in self:
+        locations_with_restriction = self.filtered(
+            lambda location: location.release_channel_restriction == "same"
+        )
+        for location in locations_with_restriction:
             # Get all locations related to this one
             # (same parent with "same" restriction)
             parent = self._get_first_ancestor_with_same_restriction()
@@ -98,6 +101,9 @@ class StockLocation(models.Model):
             )
 
             location.current_release_channel_restriction_id = release_channel_id
+        (
+            self - locations_with_restriction
+        ).current_release_channel_restriction_id = False
 
     @api.depends(
         "specific_release_channel_restriction", "parent_release_channel_restriction"
--- a/stock_location_release_channel_restriction/models/stock_location.py
+++ b/stock_location_release_channel_restriction/models/stock_location.py
@@ -147,7 +147,8 @@ class StockLocation(models.Model):
                 family_locations = self.browse()
             locations = family_locations | location
             if not (locations.pending_out_move_line_ids - lines) or force:
-                locations.current_release_channel_restriction_id = False
+                # users may not have write access on locations
+                locations.sudo().current_release_channel_restriction_id = False
 
     def action_reset_release_channel(self):
         view_id = self.env.ref(
--- a/stock_location_release_channel_restriction/models/stock_location.py
+++ b/stock_location_release_channel_restriction/models/stock_location.py
@@ -62,13 +62,16 @@ class StockLocation(models.Model):
         This will retrieve all first parent with "same" restriction
         on the recordset.
         """
-        parents = self.search(
-            [
-                ("id", "parent_of", self.ids),
-                ("location_id.release_channel_restriction", "!=", "same"),
-            ]
-        )
-        return parents
+        self.ensure_one()
+        location = self
+        while parent := location.location_id:
+            if (
+                parent.release_channel_restriction == "same"
+                and parent.location_id.release_channel_restriction != "same"
+            ):
+                return parent
+            location = parent
+        return self.browse()
 
     @api.depends(
         "release_channel_restriction",
@@ -79,16 +82,13 @@ class StockLocation(models.Model):
         """
         This will compute the current release channel that
         """
-        parents = self._get_first_ancestor_with_same_restriction()
         for location in self:
-            # Get all locations related to this one (same parent with "same" restriction)
-            family_location_ids = parents.filtered_domain(
-                [("id", "parent_of", location.id)]
-            ).child_ids.filtered(
-                lambda record: record.release_channel_restriction == "same"
-            )
+            # Get all locations related to this one
+            # (same parent with "same" restriction)
+            parent = self._get_first_ancestor_with_same_restriction()
+            family_location_ids = parent.child_ids
             release_channel_id = first(
-                family_location_ids.pending_out_move_line_ids.picking_id.release_channel_id
+                family_location_ids.pending_out_move_line_ids.picking_id.ship_picking_id.release_channel_id
             )
 
             location.current_release_channel_restriction_id = release_channel_id
--- a/stock_location_release_channel_restriction/models/stock_location.py
+++ b/stock_location_release_channel_restriction/models/stock_location.py
@@ -23,12 +23,8 @@ class StockLocation(models.Model):
         compute="_compute_current_release_channel_restriction_id",
         recursive=True,
         store=True,
-    )
-
-    has_release_channel_restriction = fields.Boolean(
-        compute="_compute_has_release_channel_restriction",
-        recursive=True,
-        store=True,
+        help="This is the current release channel that restrict this "
+        "location for future incoming movements.",
     )
 
     release_channel_restriction = fields.Selection(
@@ -46,6 +42,8 @@ class StockLocation(models.Model):
         readonly=True,
         related="location_id.release_channel_restriction",
         recursive=True,
+        help="This field is used to compute recursively the restriction parameter"
+        " from parent hierarchy.",
     )
 
     specific_release_channel_restriction = fields.Selection(
@@ -100,13 +98,6 @@ class StockLocation(models.Model):
             )
 
             location.current_release_channel_restriction_id = release_channel_id
-
-    @api.depends("current_release_channel_restriction_id")
-    def _compute_has_release_channel_restriction(self):
-        for location in self:
-            location.has_release_channel_restriction = bool(
-                location.current_release_channel_restriction_id
-            )
 
     @api.model
     def _selection_release_channel_restriction(self):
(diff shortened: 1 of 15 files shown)


rousseldenis:
@jbaudoux @phschmidt

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -49,6 +55,42 @@ class StockLocation(models.Model):
...
+        parents = self._get_first_ancestor_with_same_restriction()
+        for location in self:
+            # Get all locations related to this one (same parent with "same" restriction)
+            family_location_ids = parents.filtered_domain(
+                [("id", "parent_of", location.id)]
+            ).child_ids.filtered(
+                lambda record: record.release_channel_restriction == "same"
+            )
+            release_channel_id = first(
+                family_location_ids.pending_out_move_line_ids.picking_id.release_channel_id
```
jbaudoux: Normally the release channel is set on the delivery, not on internal transfer

```suggestion
                (pick := family_location_ids.pending_out_move_line_ids.picking_id)
                and pick.release_channel_id or pick.ship_picking_id.release_channel_id
```

or just

```suggestion
                family_location_ids.pending_out_move_line_ids.picking_id.ship_picking_id.release_channel_id
```

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -49,6 +55,42 @@ class StockLocation(models.Model):
...
+        "release_channel_restriction",
+        "pending_in_move_line_ids",
+        "pending_out_move_line_ids",
+    )
+    def _compute_current_release_channel_restriction_id(self):
+        """
+        This will compute the current release channel that
+        """
+        parents = self._get_first_ancestor_with_same_restriction()
+        for location in self:
```
jbaudoux: You should not make such searches for locations that do not have a release_channel_restriction

stock_location_release_channel_restriction/models/stock_location.py
jbaudoux:
```suggestion
```
I don't think the index is necessary

jbaudoux, line 56: Can't you simplify by just a single Boolean to enable the feature and an inverse method to propagate the value on the children.

jbaudoux: You even already have it `release_channel_restriction_in_move`.

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -49,6 +55,42 @@ class StockLocation(models.Model):
...
+            # Get all locations related to this one (same parent with "same" restriction)
+            family_location_ids = parents.filtered_domain(
+                [("id", "parent_of", location.id)]
+            ).child_ids.filtered(
```
jbaudoux: In the family, you need self and the children of the children.
It looks over complicated and `filtered_domain` inside a for loop is not ideal.
I think you could just find the top location having `release_channel_restriction_in_move` by looping over `location_id` and then collect all the children locations (incl self) having `release_channel_restriction_in_move` without making any call to `search`

stock_location_release_channel_restriction/tests/test_release_channel_restriction.py
```diff
@@ -297,3 +297,50 @@ def test_release_channel_restriction_pending_incoming_done(self):
         )
         self.picking_2.release_channel_id = self.default_channel
         self.picking_1._action_done()
+
+    def test_release_channel_restriction_children(self):
+        """
+        Check that a restriction
```
jbaudoux: Give a proper description

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -49,6 +55,42 @@ class StockLocation(models.Model):
...
 
+    def _get_first_ancestor_with_same_restriction(self):
+        """
+        This will retrieve all first parent with "same" restriction
+        on the recordset.
+        """
+        parents = self.search(
+            [
+                ("id", "parent_of", self.ids),
+                ("location_id.release_channel_restriction", "!=", "same"),
```
jbaudoux:
```suggestion
                ("location_id.release_channel_restriction", "!=", "same"),
                ("release_channel_restriction", "=", "same"),
```
but I would drop that method

jbaudoux (commented):
I suggest less and more readable code

lmignon:
ping @rousseldenis

stock_location_release_channel_restriction/models/stock_location.py
rousseldenis: The advantage with the search() will generate one query than accessing `location_id` field in each iteration would trigger a query in each level.

stock_location_release_channel_restriction/models/stock_location.py
rousseldenis: We should as we need to get the first parent that does not have a parent with a restriction in order to find the `same` family.

stock_location_release_channel_restriction/models/stock_location.py
rousseldenis: So, we introduce a limitation for that ?

stock_location_release_channel_restriction/models/stock_location.py
rousseldenis: If we use the `search()`, we do

rousseldenis:
@jbaudoux

rousseldenis:
@jbaudoux

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -51,24 +61,49 @@ class StockLocation(models.Model):
...
+        """
+        This will retrieve all first parent with "same" restriction
+        on the recordset.
+        """
+        self.ensure_one()
+        location = self
+        while parent := location.location_id:
+            if (
+                parent.release_channel_restriction == "same"
+                and parent.location_id.release_channel_restriction != "same"
```
jbaudoux, line 89: As you have it stored, you can use it :)
```suggestion
                and parent.parent_release_channel_restriction != "same"
```

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -51,24 +61,49 @@ class StockLocation(models.Model):
...
+        """
+        This will compute the current release channel that
+        """
+        locations_with_restriction = self.filtered(
+            lambda location: location.release_channel_restriction == "same"
+        )
+        for location in locations_with_restriction:
+            # Get all locations related to this one
+            # (same parent with "same" restriction)
+            parent = self._get_first_ancestor_with_same_restriction()
```
jbaudoux: This won't work if the restriction is on `location` and not on the parent

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -51,24 +61,49 @@ class StockLocation(models.Model):
...
+                return parent
+            location = parent
+        return self.browse()
+
+    @api.depends(
+        "location_id.release_channel_restriction",
+        "release_channel_restriction",
+        "pending_in_move_line_ids",
+        "pending_out_move_line_ids",
+        "location_id.current_release_channel_restriction_id",
```
jbaudoux: I still don't think it will properly trigger the recompute.

You have a restriction on `Output / Trolley A`.
You have a move line to `Output / Trolley A / Level X` for channel C.
It will compute `current_release_channel_restriction_id` in that location `Output / Trolley A / Level X` but not on `Output / Trolley A` or `Output / Trolley A / Level Y`.
You assume someone will change the location from the parent to a child but if there is a putaway, it won't work.

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -51,24 +61,49 @@ class StockLocation(models.Model):
...
+        This will compute the current release channel that
+        """
+        locations_with_restriction = self.filtered(
+            lambda location: location.release_channel_restriction == "same"
+        )
+        for location in locations_with_restriction:
+            # Get all locations related to this one
+            # (same parent with "same" restriction)
+            parent = self._get_first_ancestor_with_same_restriction()
+            family_location_ids = parent.child_ids
```
jbaudoux: Is this taking also the children of the children?

stock_location_release_channel_restriction/models/stock_move_line.py
```diff
@@ -29,13 +28,11 @@ def _compute_for_restriction_incoming_location_channel_ids(self):
         for line in self:
             line.for_restriction_incoming_location_channel_ids = line.location_dest_id.pending_in_move_line_ids.picking_id.release_channel_id  # noqa: E501
```
jbaudoux:
```suggestion
            line.for_restriction_incoming_location_channel_ids = line.location_dest_id.pending_in_move_line_ids.picking_id.ship_picking_id.release_channel_id  # noqa: E501
```

stock_location_release_channel_restriction/models/stock_location.py
```diff
@@ -51,24 +61,49 @@ class StockLocation(models.Model):
...
+            ):
+                return parent
+            location = parent
+        return self.browse()
+
+    @api.depends(
+        "location_id.release_channel_restriction",
+        "release_channel_restriction",
+        "pending_in_move_line_ids",
+        "pending_out_move_line_ids",
```
jbaudoux: You don't use `pending_in_move_line_ids` here, it's used in `_compute_for_restriction_incoming_location_channel_ids`, so you should not need it here

jbaudoux (changes requested):
I'm still not convinced (╥_╥)

Relying on `pending_out_move_line_ids` is a bit fragile as you rely on the fact that the chained move has a reservation policy on confirm so that you have a move line to read the channel from.

What about storing the channel restriction on the family of locations on action done on the move to the location and only when there is no value yet (you compute it only with first incoming done move and not at each new move line).
Then for emptying the value, you can rely on `pending_out_move_line_ids` and only if the location do not have an outgoing move line (pending_out_move_line_ids is False), then you check the family to see if there are still quants on the locations. If empty, remove the value.

jbaudoux (approved):

PR #2525, opened 2026-02-26
[19.0][ADD] stock_location_code

MiquelRForgeFlow:
Stock Location Code
===================

This module adds a code field in the location.


AaronHForgeFlow (commented):
LGTM! some small suggestions:

while searching I would like to see the code as well:

Image

Maybe we could add constraint unique code per company too.

MiquelRForgeFlow:
> while searching I would like to see the code as well:

Done.

> Maybe we could add constraint unique code per company too.

Sublocations could have same code, so I don't recommend this.

AaronHForgeFlow (approved):
LGTM

stock_location_code/models/stock_location.py
```diff
@@ -0,0 +1,24 @@
...
+        copy=False,
+    )
+
+    @api.depends("name", "location_id.complete_name", "usage", "code")
+    @api.depends_context("formatted_display_name")
+    def _compute_display_name(self):
+        res = super()._compute_display_name()
+        for location in self:
+            if location.code:
+                location.display_name = f"{location.code} {location.display_name}"
```
DavidJForgeFlow: We should use the [] to encapsulate the code. As in products.

Also, can you add some small tests?

stock_location_code/views/stock_location_view.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="name">stock.location.form - code</field>
+        <field name="model">stock.location</field>
+        <field name="inherit_id" ref="stock.view_location_form" />
+        <field name="arch" type="xml">
+            <field name="active" position="after">
```
DavidJForgeFlow, line 8: We should put it before the name, again to match product behavior.

Image

DavidJForgeFlow (changes requested):

stock_location_code/views/stock_location_view.xml
DavidJForgeFlow: I refernced the wrong view, you need to change the list view, not the form one.

DavidJForgeFlow (approved):

MiquelRForgeFlow:
Added tests ;)

BhaveshHeliconia (approved):
Code review LGTM!

WesleyOliveira98 (approved):
LGTM

PR #2530, opened 2026-02-28
[19.0][MIG] stock_putaway_product_template: Migration to 19.0

gdgellatly:
## Summary

Migration of `stock_putaway_product_template` module to 19.0.

- Version bump to 19.0.1.0.0
- Replaced the `filtered()` override on `stock.putaway.rule` with a cleaner `_get_putaway_location` override that filters out template rules not matching the product before calling super
- Removed `stock_location.py` entirely (the `filter_putaway_rule` context hack is no longer needed)
- Created an explicit `product.category` in tests since `product_category_all` was removed in v19, which caused `categ_id` to be `False` and the category domain to match all rules without a category

## Test plan

- [x] Module installs successfully
- [x] Tests pass with `--test-enable` on database with demo data
- [x] Pre-commit passes


dannyadair (approved):
I've reviewed the code and functionality - looks good.
The .pot file is still referencing 18.0

rousseldenis (approved):

gdgellatly:
@pedrobaeza Before this gets merged I want your opinion. I did change the overrride point because the filtered override did not work in v19 and it was confusing, however as I did it I was concerned that the original implementation was required for performance and whether that was the reason such efforts were gone to originally.

pedrobaeza:
In fact it was @sergio-teruel the author of the module. Sergio, can you please check about the new method?

stock_putaway_product_template/views/stock_putaway_rule_views.xml
```diff
@@ -0,0 +1,25 @@
...
+            <field name="product_id" position="attributes">
+                <attribute name="readonly">category_id or product_tmpl_id</attribute>
+            </field>
+            <field name="category_id" position="attributes">
+                <attribute name="readonly">product_id or product_tmpl_id</attribute>
+            </field>
+            <field name="product_id" position="before">
+                <field
+                    name="product_tmpl_id"
+                    readonly="category_id"
```
BhaveshHeliconia, line 18:
```suggestion
                    readonly="category_id or product_id"
```

BhaveshHeliconia (approved):
Code review LGTM!

minor suggestion -

WesleyOliveira98 (approved):

carlos-lopez-tecnativa (changes requested):
I think the priority should be:

1. Rules by product variant
2. Rules by product template
3. Rules by category
4. Empty rules

Right?

If this is correct, the rules by category currently take priority over the product template rules. In this example, I expected the `TEST S/M` variant to be assigned to the `WH/Stock/Shelf 2/Small Refrigerator` location instead of `WH/Stock/SELF 3`.
Image

Image

carlos-lopez-tecnativa:
ping @gdgellatly

gdgellatly:
> I think the priority should be:
> 
> 1. Rules by product variant
> 2. Rules by product template
> 3. Rules by category
> 4. Empty rules
> 
> Right?
> 
> If this is correct, the rules by category currently take priority over the product template rules. In this example, I expected the `TEST S/M` variant to be assigned to the `WH/Stock/Shelf 2/Small Refrigerator` location instead of `WH/Stock/SELF 3`. Image
> 
> Image

i don't really understand what you are saying. For me this is just migrating the existing behavior, but are you saying I changed it?

carlos-lopez-tecnativa (approved):
> i don't really understand what you are saying. For me this is just migrating the existing behavior, but are you saying I changed it?

Sorry, I simplified my example. What I meant is that if I have two putaway rules, one by product template and another by product category, I would expect the rule for the product template to be used. However, currently, the rule for the product category is used, which seems incorrect to me.

Image

Image

But anyway, I tested this in V18, and the behavior is the same, so this could be a general issue with the module and not something related to the migration.
LGTM 

TT64147
ping @sergio-teruel @pedrobaeza

PR #2533, opened 2026-03-01
[12.0][NEW] Product Stock availability

mymage:
Utility to show where a product is available


product_stock_available/models/stock_move.py
```diff
@@ -0,0 +1,25 @@
...
+            products.sudo()._update_available_locations()
+        return res
+
+    def _update_all_products(self):
+        prod_s = self.env['product.product'].search([])
+        for prod in prod_s:
+            # We use sudo() because an internal transfer might update quantities
+            # but the user running it might not have rights to update product.product
+            prod.sudo()._update_available_warehouses()
+            prod.sudo()._update_available_locations()
```
eLBati, line 25: This iterates over every product in the database, one by one, and for each product iterates over all warehouses/locations checking qty_available. On any "big" database (thousands of products, dozens of warehouses) this will timeout or lock tables for extended periods.

product_stock_available/views/stock_views.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <record id="action_update_all_products" model="ir.actions.server">
+        <field name="name">Update Products Availability</field>
+        <field name="model_id" ref="stock.model_stock_move"/>
+        <field name="binding_model_id" ref="stock.model_stock_move"/>
```
eLBati: AFAIK binding_model_id makes this action appear in the "Action" dropdown on stock move list/form views, which is confusing. Since it's triggered from a menu item, binding_model_id should be removed.

product_stock_available/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+Stefano Consolaro <stefano.consolaro@integralsolutions.it>
```
eLBati, line 1: should be readme/CONTRIBUTORS.rst

product_stock_available/models/product.py
```diff
@@ -0,0 +1,73 @@
...
+        store=True,
+        string='Available in Warehouses'
+    )
+
+    available_location_ids = fields.Many2many(
+        'stock.location',
+        compute='_compute_available_location_ids',
+        store=True,
+        string='Available in Locations'
+    )
```
eLBati, line 62: since this data is already present in the system, in the quants table, we should not add new fields and write them in a certain event (what would happen if the quants were created/modified/deleted outside of _action_done?) but add calculated fields based on the quants

eLBati (changes requested):

product_stock_available/models/stock_move.py
mymage: This a exceptional activity to do only once the application is installed and in a way or another you have to pass all product and all location to update data.
I changed a bit the code to limit going to all products/locations.

product_stock_available/readme/CONTRIBUTORS.rst
mymage: Right

product_stock_available/models/product.py
mymage: Mh, I suppose that you can't do operation on quants without doing a proper operation on stock (stock.move) because this will result in a discrepancy in the stock data.

product_stock_available/views/stock_views.xml
mymage: OK, fixed.

PR #2535, opened 2026-03-03
[18.0][IMP] stock_warehouse_resupply_route_push: include partner_address_id and push_domain from supplied wh

AlexPForgeFlow:
Include partner_address_id from supplied wh in pull rules and push_domain from supplied wh in push rules.
For cases where multiple warehouses are supplied from a central warehouses we automatically set up those fields on creation rules, for a proper routes resolution.

### DIFF
--- a/stock_warehouse_resupply_route_push/models/stock_warehouse.py
+++ b/stock_warehouse_resupply_route_push/models/stock_warehouse.py
@@ -2,6 +2,7 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
 
 from odoo import Command, fields, models
+from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
 
 
@@ -113,6 +114,14 @@ class StockWarehouse(models.Model):
     def create_resupply_routes(self, supplier_warehouses):
         if not self.resupply_wh_push:
             return super().create_resupply_routes(supplier_warehouses)
+        if not self.partner_id:
+            raise ValidationError(
+                _(
+                    "Address is required in warehouse %s to use "
+                    "Inter-warehouse push routes generation."
+                )
+                % self.name
+            )
         # For now, it is still not possible to use push for inter-warehouses
         # between 2 companies, because we cannot create a stock rule with
         # locations of different companies.
@@ -167,6 +176,7 @@ class StockWarehouse(models.Model):
                     "route_id": inter_wh_route.id,
                     "location_dest_from_rule": False,
                     "warehouse_id": self.id,
+                    "partner_address_id": self.partner_id.id,
                 },
             )
             if supplier_wh.delivery_steps != "ship_only":
@@ -198,8 +208,10 @@ class StockWarehouse(models.Model):
                     )
                 ],
                 values={
+                    "warehouse_id": False,
                     "route_id": inter_wh_route.id,
                     "propagate_warehouse_id": supplier_wh.id,
+                    "push_domain": "[('partner_id', '=', %d)]" % self.partner_id.id,
                 },
             )
             for rule_vals in rules_list:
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (commented):
@AlexPForgeFlow This makes defectively `partner_id` in warehouse required to use this feature, could you add a ValidationError if not set when using the "regenerate routes" action?

AlexPForgeFlow:
@LoisRForgeFlow added validation error in create_resupply_routes as when activating resupply_wh_push it will already create rules

stock_warehouse_resupply_route_push/models/stock_warehouse.py
```diff
@@ -113,6 +114,14 @@ def _is_internal_inter_wh(self, supplier_wh):
     def create_resupply_routes(self, supplier_warehouses):
         if not self.resupply_wh_push:
             return super().create_resupply_routes(supplier_warehouses)
+        if not self.partner_id:
```
LoisRForgeFlow, line 117: @AlexPForgeFlow It is only required if self.resupply_wh_push is true.

stock_warehouse_resupply_route_push/models/stock_warehouse.py
```diff
@@ -113,6 +114,14 @@ def _is_internal_inter_wh(self, supplier_wh):
...
+            raise ValidationError(
+                _(
+                    "Address is required in warehouse %s to use "
+                    "Inter-warehouse routes generation."
```
LoisRForgeFlow: "Inter-warehouse **push** routes"

stock_warehouse_resupply_route_push/models/stock_warehouse.py
AlexPForgeFlow: @LoisRForgeFlow if resupply_wh_push is false super call will be returned before reaching that point

stock_warehouse_resupply_route_push/tests/test_stock_warehouse_resupply_route_push.py
```diff
@@ -308,3 +324,62 @@ def test_07_apply_to_external_wh(self):
...
+        )
+        self.assertEqual(pull_rule_wh5.partner_address_id, self.wh_5.partner_id)
+        # Check push rule has push_domain has partner_id from wh_5
+        push_rule_wh5 = self.route_wh_1_to_wh_5.rule_ids.filtered(
+            lambda x: x.action == "push"
+        )
+        self.assertEqual(
+            push_rule_wh5.push_domain,
+            "[('partner_id', '=', %d)]" % self.wh_5.partner_id.id,
+        )
```
LoisRForgeFlow, line 394: @AlexPForgeFlow If I comment this line and the patches in this PR, this test passes, which means that the test is not really capturing the issue being fixed.

stock_warehouse_resupply_route_push/models/stock_warehouse.py
LoisRForgeFlow: @AlexPForgeFlow but that is fine, if we dont use push rules, we can let the system create rules in the standard way, and for that, the address is not needed.

LoisRForgeFlow: @AlexPForgeFlow wait, I understood you know, the if is just in the line before, so you are right. Thanks for clarifying!

stock_warehouse_resupply_route_push/tests/test_stock_warehouse_resupply_route_push.py
AlexPForgeFlow: @LoisRForgeFlow I will double check :+1:

LoisRForgeFlow (approved):
It looks perfect now, functional and code review. Thanks @AlexPForgeFlow 👍

PR #2543, opened 2026-03-19
[18.0][FIX] stock_cycle_count: use sudo() in check_zero_confirmation call

ThiagoMForgeFlow:
Use `sudo()` when calling `check_zero_confirmation` from `_action_done`, as it is an internal system operation that should not depend on the user's access rights. This follows the same pattern used in stock_account (odoo/addons/stock_account/models/stock_move.py) where `sudo()` is used to create stock.valuation.layer records and update product standard prices within `_action_done`.

@ForgeFlow


JoanSForgeFlow (approved):
LGTM!

LoisRForgeFlow (approved):
I think it is reasonable

PR #2544, opened 2026-03-20
[18.0][FIX] stock_move_purchase_uom: prevent unreservation in create

LauraCForgeFlow:
The `create` override writes `product_uom_qty` on the created move to convert the quantity into the purchase UoM. Because this is a `write()` on an existing record, it triggers the unreservation logic in `stock.move.write()`, which deletes `stock.move.line` records when `product_uom_qty` changes on a non-draft move.

Odoo interprets this numerical change as a demand reduction, which is incorrect: the write is not a demand reduction, it's the same physical quantity in a different UoM as part of the creation. The existing reservation (if any SML were linked during create via Command.link) remains valid and should not be discarded. Add `do_not_unreserve=True` to the context of the `product_uom_qty` write, to avoid triggering unreservation. This is consistent with how Odoo itself handles equivalent situations in the stock module.

### DIFF
--- a/stock_move_purchase_uom/models/stock_move.py
+++ b/stock_move_purchase_uom/models/stock_move.py
@@ -24,7 +24,9 @@ class StockMove(models.Model):
                     rounding_method=move.picking_type_id.purchase_uom_rounding_method,
                 )
                 move.product_uom = move.product_id.uom_po_id
-                move.product_uom_qty = updated_product_uom_qty
+                move.with_context(
+                    do_not_unreserve=True
+                ).product_uom_qty = updated_product_uom_qty
         return moves
 
     @api.onchange("product_id", "picking_type_id")
--- a/stock_move_purchase_uom/tests/test_stock_move.py
+++ b/stock_move_purchase_uom/tests/test_stock_move.py
@@ -1,5 +1,6 @@
 # Copyright 2024 ForgeFlow S.L. (https://www.forgeflow.com)
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo import Command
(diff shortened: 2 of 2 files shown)


JordiMForgeFlow (approved):
makes sense 👍🏼

LoisRForgeFlow (commented):
@LauraCForgeFlow could you add a test case depicting the issue that is being fixed?

LauraCForgeFlow:
@LoisRForgeFlow Sure, done! :)

stock_move_purchase_uom/models/stock_move.py
```diff
@@ -24,7 +24,9 @@ def create(self, vals_list):
                     rounding_method=move.picking_type_id.purchase_uom_rounding_method,
                 )
                 move.product_uom = move.product_id.uom_po_id
-                move.product_uom_qty = updated_product_uom_qty
+                move.with_context(
+                    do_not_unreserve=True
+                ).product_uom_qty = updated_product_uom_qty
```
LoisRForgeFlow, line 29: @LauraCForgeFlow I have commented this change in local and test test_create_move_linked_sml_not_unreserved did not fail

LoisRForgeFlow: Perfect, I tested now and the test is relevant with the latest pushes!

LoisRForgeFlow (approved):
Thanks for fixing!

/ocabot merge patch

stock_move_purchase_uom/models/stock_move.py
LauraCForgeFlow: Good catch! The data in the test was incorrect. It should be fixed now!

PR #2547, opened 2026-03-26
[18.0][IMP] stock_vertical_lift: revamp UI

divad1196:
The UI can overflow the screen vertically when we have too many data.
It has a lot of space lost that we can retrieve so that it can hold more data.

The current version of the UI:
- Maximize the use of the space
- Ensure compatibility with known modules that inherit the view.

The gained space is now more than enough to prevent overflow.

## New Visual

image

### DIFF
--- a/stock_vertical_lift/views/vertical_lift_operation_base_views.xml
+++ b/stock_vertical_lift/views/vertical_lift_operation_base_views.xml
@@ -118,55 +118,51 @@
             <xpath expr="//div[hasclass('o_shuttle_data')]" position="inside">
                 <field name="current_move_line_id" invisible="True" force_save="1" />
                 <!-- on the left of the screen -->
-                <div class="o_shuttle_data_content o_shuttle_move">
-                    <div>
+                <div class="o_shuttle_data_content">
+                    <group>
+                        <label for="picking_id" />
+                        <div>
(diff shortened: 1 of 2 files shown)


stock_vertical_lift/views/vertical_lift_operation_base_views.xml
```diff
@@ -64,6 +64,18 @@
                                 title="Barcode Input"
                             />
                         </div>
+                        <div
```
TDu, line 67: This moves the operation name/state that was spanning the whole screen horizontally in the middle of a flex display with multiple column ?
I struggle visualizing it, could you share a screen shot of the results, please (maybe blur any sensitive content)
Thanks

jbaudoux (approved):
LG, not tested

stock_vertical_lift/views/vertical_lift_operation_base_views.xml
divad1196: @TDu you are right, I added an image to the PR's description

TDu (approved):
From the screenshot, vertical space is optimized 👍

divad1196:
@jbaudoux can this be merged? It has been open for a while now

PR #2558, opened 2026-04-16
[18.0][FIX] stock_move_packaging_qty: ml pack columns order

chienandalu:
After the migration, users find annoying the change of column order as they expect it to be to the left of quantity instead of to the right.

image
image

cc @moduon MT-14533

please review @shide @EmilioPascual

### DIFF
--- a/stock_move_packaging_qty/i18n/es.po
+++ b/stock_move_packaging_qty/i18n/es.po
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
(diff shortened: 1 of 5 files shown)


EmilioPascual (approved):
LGTM

chienandalu:
I'm going to improve the labels as well

chienandalu:
@Shide @yajo please take a look

Shide (approved):

PR #2559, opened 2026-04-21
[18.0][ADD] stock_lot_list_price: New module

victoralmau:
New module (separated from `sale_order_lot_selection_price` to remove the dependency on `sale`)

@Tecnativa TT61752


pedrobaeza:
According the description, not sure about the module name and to decouple this from `sale`. At the end, if you are defining a sales price, is for selling it... In case you think it should be separated, it should be at least `stock_lot_sale_price` for being distinguishable from the lot cost price.

victoralmau:
I've separated the `sale_order_lot_selection_price` code to create a module that doesn't depend on `sale_order_lot_selection`.

In my opinion, the module should NOT have the `sale` prefix; it's simply adding a price field to the lots—nothing more—so it shouldn't introduce a dependency on `sale`. The exact same example exists in `product.template` with the `list_price` field, which adds it directly. [product](https://github.com/odoo/odoo/blob/73d73c5c6606e0b34c754bfc4de035840951dd3b/addons/product/models/product_template.py#L94).

pedrobaeza:
I'm not talking about a prefix, but `stock_lot_sale_price`, which puts `sale_price` as the part of the feature:

`stock` (area) + `lot` (document) + `sale_price` (feature)

victoralmau:
But if you add “sale” (whether as a prefix or anywhere in the name), wouldn't that make ‘sale’ a dependency? That's exactly what I want to avoid (the dependency on “sale”).

pedrobaeza:
Not really in this case, as you are saying that this is the "sales price". It's the same as if you do `stock_lot_cost_price`, you don't depend on a theoretical `cost` module. It's just the field you are creating. But if you prefer, call it `stock_lot_list_price`.

pedrobaeza (approved):

BhaveshHeliconia:
@victoralmau Could you please rebase the PR to make Runbot available?

BhaveshHeliconia:
@victoralmau On Runbot, after logging in, only a white screen is displayed. Could you please check?

victoralmau:
> @victoralmau On Runbot, after logging in, only a white screen is displayed. Could you please check?

Do you know  `sale_order_lot_selection_price` module? 
The purpose of this module is to allow you to set a `lst_price` for lots.

pilarvargas-tecnativa:
> @victoralmau Could you please rebase the PR to make Runbot available?

Try it on a Runbot with no modules installed – just the base – and install the necessary modules

BhaveshHeliconia (commented):
It works when using either module individually, but if both modules are installed together, there will be a conflict.

This module is not compatible with `sale_order_lot_selection_price`, as `lst_price` is already used by that module.

eduezerouali-tecnativa (approved):

victoralmau:
Can you try the process again?

WesleyOliveira98 (approved):

PR #2560, opened 2026-04-21
[18.0][ADD] stock_lot_warehouse: New module

victoralmau:
New module

@Tecnativa TT61752


stock_lot_warehouse/models/stock_lot.py
```diff
@@ -0,0 +1,21 @@
...
+        compute="_compute_warehouse_id",
+        store=True,
+    )
+
+    @api.depends("location_id", "location_id.warehouse_id")
+    def _compute_warehouse_id(self):
+        for lot in self:
+            lot.warehouse_id = (
+                lot.location_id.warehouse_id if lot.location_id else False
+            )
```
pedrobaeza:
```suggestion
            lot.warehouse_id = lot.location_id.warehouse_id
```

pedrobaeza (changes requested):
On installation, if you have a lot of lots, this will take a lot of time. Another thing: why it can be a related field? And it that case, search works over related fields with good performance, so why do you need to store it?

victoralmau:
Changed to a related field + search

pedrobaeza:
But the search is done automatically by ORM AFAIK, isn't it?

stock_lot_warehouse/views/stock_lot_views.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_production_lot_tree" model="ir.ui.view">
+        <field name="model">stock.lot</field>
+        <field name="inherit_id" ref="stock.view_production_lot_tree" />
+        <field name="arch" type="xml">
+            <field name="last_delivery_partner_id" position="after">
+                <field
+                    name="warehouse_id"
+                    invisible="not location_id"
```
pedrobaeza: Is this `invisible` really working?

stock_lot_warehouse/views/stock_lot_views.xml
```diff
@@ -0,0 +1,44 @@
...
+        <field name="model">stock.lot</field>
+        <field name="inherit_id" ref="stock.search_product_lot_filter" />
+        <field name="arch" type="xml">
+            <filter name="group_by_Company" position="before">
+                <filter
+                    name="group_by_warehouse"
+                    string="Warehouse"
+                    domain="[]"
+                    context="{'group_by': 'warehouse_id'}"
+                    groups="stock.group_stock_multi_warehouses"
```
pedrobaeza, line 40: The group is not correct for any of these views. It should be the "Lot" one, not the multiple locations/warehouses one.

pedrobaeza: Oh, wait, now I get it, but the group is for warehouse locations, not warehouses per se. I was confused when not seeing the group by warehouse here, and being able to do it through custom group by, but in the rest of the views, it may be reasonable to keep the group, as you won't see the location field either.

stock_lot_warehouse/views/stock_lot_views.xml
victoralmau: Yes, but I'm going to change it to use the `warehouse_id` field to avoid confusion.

stock_lot_warehouse/views/stock_lot_views.xml
victoralmau: Now that group is defined in all views, do we need to remove it from anywhere?

pedrobaeza: OK, let's keep it that way.

pedrobaeza (approved):

victoralmau:
Ping @CarlosRoca13

CarlosRoca13 (approved):

PR #2561, opened 2026-04-22
[ADD] New module stock_quant_package_archive

divad1196:
The module adds archiving capabilities on stock.quant.package model


divad1196:
wrong repository

PR #2563, opened 2026-04-27
[18.0][ADD] stock_lot_catalog: New module

victoralmau:
New module

ejemplo

@Tecnativa TT61752


stock_lot_catalog/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+The lot catalog option has been added.
```
pedrobaeza: Please explain a bit more the aim of this module. As I understand it right now, it's like a base one, but it's not doing anything by itself, isn't it?

CarlosRoca13 (approved):
LGTM 😄

pedrobaeza (approved):
/ocabot merge nobump

pedrobaeza:
Uhm, it seems to be a test failure in `stock_cycle_count`. Maybe some new code upstream that has only got to vanilla Odoo yet?

victoralmau:
Can you try again? (It looks like it only failed when committing the .po file.)

victoralmau:
The test for the `stock_cycle_count` module keeps failing randomly, and I haven't been able to reproduce the error locally. How would you like to proceed in this case? @pedrobaeza

pedrobaeza:
I think we should disable such test

PR #2565, opened 2026-04-27
[18.0][ADD] stock_lot_catalog_condition: New module

victoralmau:
New module

ejemplo

@Tecnativa TT61752


pedrobaeza (commented):
I think the condition should have a linked color, and show the tag with such color for helping users to easily identify it.

victoralmau:
> I think the condition should have a linked color, and show the tag with such color for helping users to easily identify it.

Do you have a visual example from somewhere else that shows how to do something similar?

victoralmau:
And another important question: should the color be added to the conditions in this module or in `stock_lot_condition`? Let's say, for example, that there is a module called `website_stock_lot_condition` (or something similar) to display that information on the website.

pedrobaeza:
Uhm, not right now, but I'm talking about putting the background of the pill in that color. You didn't include a screenshot for imaging how this module is going to work. Seeing the code, I see you render a string "Condition: <condition>`, so putting the background of that may work, or maybe putting an extra class in the div for giving the pill aspect.

pedrobaeza:
About where to include it, for me it's clear that it should be in `stock_lot_condition`.

victoralmau:
> Uhm, not right now, but I'm talking about putting the background of the pill in that color. You didn't include a screenshot for imaging how this module is going to work. Seeing the code, I see you render a string "Condition: `, so putting the background of that may work, or maybe putting an extra class in the div for giving the pill aspect.

Should we change the background color of the pill to a different color? How would we then tell the difference between when something has been added and when it hasn't?
I added an image to the PR description, even though I thought there wouldn't be any confusion from including informational text about the status of each lot, but it seems there is.

Before making any changes, it should be very clear exactly what needs to be done and how it should be displayed (of course, that’s just my opinion).

pedrobaeza:
I was talking about highlighting this something similar to this way:

image

and you even may not need the text "Condition:". This is for improving the user experience.

victoralmau:
> I was talking about highlighting this something similar to this way:
> image
> 
> and you even may not need the text "Condition:". This is for improving the user experience.

I don't see any examples (nor does it seem to exist in Core) where a many-to-one relationship can be displayed in the Kanban view with a defined color; that is, there is `<field name="condition_id" widget="many2many_tags" options="{‘color_field’: ‘color’}"/>`, but it's only for a many-to-many relationship.

pedrobaeza:
I think it can be done through divs and `t-attf-class`. Maybe @pilarvargas-tecnativa can tell, as she has done some of these things.

pedrobaeza (approved):
/ocabot merge minor

PR #2570, opened 2026-04-29
[FIX][IMP] stock_vertical_lift: Fix miscelleanous bugs and automate PUT steps

divad1196:
1 PR for multiple small changes instead of multiple PR with 1 commit.  
This is why the commits are not squashed.

## Fixes

Odoo changed the trigger prefix for `barcode_trigger` action:
- In V16: `O-BTN.{action}`
- In V18: `OBT{action}`

https://github.com/odoo/odoo/blob/16.0/addons/barcodes/static/src/barcode_handlers.js#L64
https://github.com/odoo/odoo/blob/18.0/addons/barcodes/static/src/barcode_handlers.js#L68

---

onchange method receives a list and not a string anymore

### Switch mode using barcodes

We can now use the scanner with barcodes to switch between PICK, PUT and Inventory modes

### Save & Release Button

Added a single button to do both `save` and `release`.  
The button can be triggered by scanning a barcode.

### Skip "Scan Tray Type" automatically

The tray type can be deduced from the destination location, which removes 1 step for the user.  
It's still possible to change the tray type later.

### Manual Barcode works like actual barcode

Manual Barcode was working in the backend only instead of the backend.  
This was causing differences with an actual scanner and not allowing proper testing.

The manual barcode feature now emit a proper `barcode_scanned` event.

### Confirm Tray Empty / Not Empty using barcodes

In PICK mode, when the last element of the tray cell has been picked, Odoo will ask you to confirm whether or not the cell is empty using buttons. This would interrupts a worker's flow.

We can now use a barcode scanner to trigger these buttons.


divad1196:
CI fails due to another module broken by a recent change made by Odoo:
https://github.com/OCA/stock-logistics-warehouse/issues/2573

Please, ignore the CI issue

stock_vertical_lift/static/src/js/vertical_lift.esm.js
```diff
@@ -21,7 +21,8 @@ patch(BarcodeHandlerField.prototype, {
         this.actionService = useService("action");
     },
     async onBarcodeScanned(event) {
-        const method = SWITCH_BARCODE_METHODS[event.detail.barcode];
+        const barcode = event.detail.barcode;
```
TDu, line 28: Nothing is gained with this change ?

stock_vertical_lift/static/src/js/vertical_lift.esm.js
```diff
@@ -82,3 +86,59 @@ patch(FormController.prototype, {
         }
     },
 });
+
+// Manual barcode popup: emits a barcode_scanned event on the global barcode
+// service bus, exactly like a real scanner, so the standard handlers and the
+// patched BarcodeHandlerField above pick it up.
+class ManualBarcodeDialog extends Component {
```
TDu, line 93: With replacing the backend wizard with a js client action, we loose some test coverage, what is the gain from it ? Except less back and forth exchange with the backend ?
Should we add some unit test for the js code ?
Do both rendering look the same ?

TDu (commented):
Whoooa, many changes.
Quick tip, when moving/renaming files uses `git mv ...`

stock_vertical_lift/static/src/js/vertical_lift.esm.js
divad1196: @TDu no behavior change, just a bit more readable to only have `barcode` instead of `event.detail.barcode`.

I had done it initially when trying to fix another issue. I ended up fixing the issue differently but kept this change.

stock_vertical_lift/static/src/js/vertical_lift.esm.js
divad1196: @TDu The `manual barcode` was not really a `manual barcode`. It would send a string to the backend and the backend decided what to run.

## Issues of old way
- Only what you implement in the backend of "manual barcode" can be triggered
- "manual barcode" does not guarantee the same behavior as actually scanning the barcode
- Not all barcode behaviors can be implemented on the backend
- Each module adding a barcode feature needed to also implement "manual barcode" which duplicates the logic

## Fix

The idea here is to emit a real barcode_scanned event and have the exact same behavior as when scanning a barcode. This way, all actions can be reproduce exactly.

## About tests

"Manual barcode" was really a workaround to glue together the pop-up in the frontend with the backend. It was just taking the barcode and calling the corresponding action.  
We don't loose tests here as long as the actions are still tested.

Testing the frontend could be an improvment, but we were not testing "manual barcode" from the web ui before.

stock_vertical_lift/static/src/js/vertical_lift.esm.js
TDu: Then it should be reverted because it adds to the diff for no reason and the one liner was find in my opinion.

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -303,6 +312,18 @@ def button_release(self):
...
+    def button_save_and_release(self):
+        """Confirm the operation (set move to done, ...)"""
+        self.ensure_one()
+        # Copy the code instead of calling button_save/release
+        #   because we still need to check the current step
+        if not self.step() == "save":
+            return
+        self.next_step()
+        if not self.step() == "release":
+            return
```
TDu, line 324: If the previous call to `next_step` move to another step that is not `release` or `save`  are we not going to be out of sync between the frontend and the backend ?
Could it happen ?

divad1196: @TDu The frontend is driven by the backend, which a good thing to avoid the kind of issue you mentioned.

1. Odoo changes the step in the backend
2. A refresh happens in the frontend

Also, I wrote the code this way because, if someone adds a new step between `save` and `release`, then we will correctly land on this intermediate step.

stock_vertical_lift/static/src/js/vertical_lift.esm.js
TDu: I thought there was some removed unit tests in the commit removing the wizard, maybe not :thinking: 

I don't understand what you mean with: 

> Not all barcode behaviors can be implemented on the backend

divad1196: @TDu `onchange` cannot trigger an action like a state refresh in V18. So you need a button for it like `button_save`.

Odoo has a special code `OBT{button-trigger-name}` that will trigger the button if the button is visible.
You cannot trigger it with `manual barcode`. There are also the special commands `OCD{action-name}`.

One issue I had was that the barcode was sent to the backend, mostly executed, but the screen would not update.

divad1196:
A bit confused by the project code coverage: 
https://app.codecov.io/gh/OCA/stock-logistics-warehouse/pull/2570

I rebased my branch and got `stock_lot_catalog_sale` which drops the test coverage. How did it got merged without passing the coverage ?

stock_vertical_lift/models/vertical_lift_operation_base.py
```diff
@@ -411,6 +432,10 @@ class VerticalLiftOperationTransfer(models.AbstractModel):
     )
     # TODO add a glue addon with product_expiry to add the field
 
+    def is_special_barcode(self, barcode):
+        # Ignore prefixed reserved by Odoo's barcode module
```
jbaudoux:
```suggestion
    def is_action_barcode(self, barcode):
        """ Detect if the barcode is a button or command barcode """
```

stock_vertical_lift/models/vertical_lift_operation_put.py
```diff
@@ -65,40 +68,48 @@ def on_barcode_scanned(self, barcode):
...
                 params=self._get_user_notification_params(),
             )
+            return
+        self.current_move_line_id = line
+        self.next_step()
+
+        dest = self.location_dest_id
+        tray_type = dest.tray_type_id or dest.cell_in_tray_type_id
+        if tray_type:
+            return self._scan_tray_type(tray_type)
```
jbaudoux: Makes sense but it could be that the matched move line has a destination that is not part of the current shuttle. In this case, you should not fetch the tray, i.e. call _scan_tray_type. Can you add a check that self.location_dest_id is a child of shuttle location?

jbaudoux: Also add a comment that if the destination location already corresponds to a tray type, you skip the scan tray type step

jbaudoux:
Regarding commit messages:
* Automatically pass tray type scan step > Automatically skip tray type scan step
* You have one commit missing the module name as prefix

stock_vertical_lift/models/vertical_lift_operation_base.py
divad1196: @jbaudoux Good proposal for the docstring

I did a comment because my wording wasn't suited for a docstring, but I should have instead changed my wording. Agree to make it a docstring.

But I wouldn't limit it to only the action barcodes. Today it's true that it's only the actions, but not necessarily tomorrow.

stock_vertical_lift/models/vertical_lift_operation_put.py
divad1196: I don't get the second message. Maybe it's a confusion coming from the `_scan_tray_type` function's name.

## Explanation

This code already checks whether the destination has a tray type.
```python
tray_type = dest.tray_type_id or dest.cell_in_tray_type_id
        if tray_type:
```

The reason I call `_scan_tray_type` afterward is because it will:
- Check if the tray type has free space
- Fetch the tray
- Go to the next step

Basically, `_scan_tray_type` does everything that needs to be done except... scanning the barcode. Scanning the barcode is done by `_scan_tray_type_action` which handle the barcode then call `_scan_tray_type`.

divad1196: For the first message: My understanding is:
- The delivery limits the location of the moves to be child location of the picking's location.
- And when the shuttle search for moves, it checks for moves that are in its location

Is that not correct?

jbaudoux: In the way we use it, I'm quite sure the move hasn't a destination location in the current shuttle. Either you scan the tray type and it searches a destination in the current shuttle. Or the tray type is already set on the move package and it searches a destination and skip scan tray type (module glue with storage type).
Purpose is to be able to match any move with the right picking type and find a destination in the current shuttle.
Do at this stage, imo, you have no warranty the current destination (as it's not yet recomputed) is for this shuttle. You shouldn't launch an action fetch tray on another shuttle.

divad1196:
@jbaudoux Here is what we discussed:

## Reason for the code in _scan_source_action

This code was added because the current flow was:
1. Scan package/product/... 
2. Scan tray type

The tray type can be deducted in most cases after step 1.
Therefore, at the end of the `_scan_source_action`, I added the code to:
1. Find the tray type
2. Call the same code that is actually called on step `Scan tray type`
```python
        dest = self.location_dest_id
        tray_type = dest.tray_type_id or dest.cell_in_tray_type_id
        if tray_type:
            return self._scan_tray_type(tray_type)
```

#### Note
The following code was already there, I just used a condition-guard pattern to removing the `if-else` nesting:
```python
# before
if cond:
    self.current_move_line_id = line
    self.next_step()
else:
    notify()
    return

# after
if not cond:
    notify()
self.current_move_line_id = line
self.next_step()
```

## Why this is strange

### Ensure matching destination

It's possible that the move's destination does not match the shuttle destination.

The module [stock_vertical_lift_storage_type](https://github.com/OCA/stock-logistics-warehouse/tree/18.0/stock_vertical_lift_storage_type) defines how the matchin is done.

I must add a test to reproduce this scenario.
The easiest fix would be to just add a check

### Risk of multiple fetches

Still because of [stock_vertical_lift_storage_type](https://github.com/OCA/stock-logistics-warehouse/tree/18.0/stock_vertical_lift_storage_type): we expected this module to:
- Skip the "Scan Tray Type" already
- Fetch the tray

In particular, I must check existing `Transition` objects and the `putaway_with_storage_type` method.
I will check this and add related tests.

divad1196:
@jbaudoux update on the situation

```python
            self.Transition(
                "scan_tray_type",
                "save",
                lambda self: self._has_storage_type()
                and self._putaway_with_storage_type(),
                # this is the trick that makes the transition applies
                # its function and directly jumps to save
                direct_eval=True,
            ),
            self.Transition(
                "scan_tray_type",
                "scan_source",
                # the transition above returned False because it could
                # not find a free space, in that case, abort the
                # put-away for this line in this shuttle
                lambda self: self._has_storage_type()
                and self._put_away_with_storage_type_failed()
                and self.clear_current_move_line(),
                # this is the trick that makes the transition applies
                # its function and directly jumps to save
                direct_eval=True,
            ),
```
The glue module `stock_vertical_lift_storage_type` is indeed trying to automatically skip `scan_tray_type` but:
- If it can't, the second `Transition` revert back to `scan_source`. That's not the case that we were facing but could still be annoying
- `_has_storage_type` is common to both `Transition` -> if it returns `False`, then it default back to the default behavior from `stock_vertical_lift`

I am exploring if we are indeed in the second case or not.

divad1196:
@jbaudoux you were right to push on this issue !

The cause is that the `stock_vertical_lift_storage_type` module only checked the `package_id`, but it's sometimes not defined. 
Fix: fallaback on `result_package_id` to find the storage classes

stock_vertical_lift/models/vertical_lift_operation_base.py
divad1196: @jbaudoux done

stock_vertical_lift/models/vertical_lift_operation_put.py
divad1196: I have added a check on the locations

stock_vertical_lift/models/vertical_lift_operation_put.py
```diff
@@ -65,40 +73,68 @@ def on_barcode_scanned(self, barcode):
...
+        is_shuttle_sublocation = (
+            self.env["stock.location"].search_count(
+                [
+                    ("id", "=", dest.id),
+                    ("id", "child_of", self.shuttle_id.location_id.id),
+                ],
+                limit=1,
+            )
+            > 0
+        )
```
jbaudoux: You can use
```suggestion
        is_shuttle_sublocation = dest._child_of(self.shuttle_id.location_id)
```

stock_vertical_lift_storage_type/models/vertical_lift_operation_put.py
```diff
@@ -17,50 +17,58 @@ class VerticalLiftOperationPut(models.Model):
...
+        # `scan_tray_type`, for the base transition doesn't have
+        # `direct_eval=True`
 
         return tuple(updated_transitions)
 
-    def _has_storage_type_domain(self):
+    def _get_current_package(self):
         move_line = self.current_move_line_id
-        package_type = move_line.package_id.package_type_id
+        package_id = move_line.package_id or move_line.result_package_id
```
jbaudoux: Only the source package matters. Nothing is creating a destination package. So either the destination package is equal to the source package or the destination package is empty.
If this solves a use case, then there is not right happening.

stock_vertical_lift_empty_tray_check/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 <meta name="generator" content="Docutils: https://docutils.sourceforge.io/" />
-<title>README.rst</title>
+<title>Vertical Lift Empty Tray Check</title>
```
jbaudoux, line 6: Do not update auto-generated files in imp/fix commits

stock_vertical_lift/models/vertical_lift_shuttle.py
```diff
@@ -148,6 +154,46 @@ def _hardware_update_tls_context(self, context):
...
+    def _compute_number_of_ops(self):
+        for record in self:
+            operation = record._operation_for_mode_no_create()
+            if record.mode == "inventory":
+                record.number_of_ops = self.env["stock.quant"].search_count(
+                    operation._domain_stock_quant_to_do()
+                )
+            else:
+                record.number_of_ops = self.env["stock.move.line"].search_count(
+                    operation._domain_move_lines_to_do()
```
jbaudoux, line 174: for put mode it is using _domain_move_lines_to_do_all() as operations are not targetting the shuttle yet

stock_vertical_lift/models/vertical_lift_shuttle.py
```diff
@@ -148,6 +154,46 @@ def _hardware_update_tls_context(self, context):
...
+                "res_model": "stock.quant",
+                "view_mode": "list,form",
+                "domain": operation._domain_stock_quant_to_do(),
+            }
+        return {
+            "type": "ir.actions.act_window",
+            "name": self.env._("Move Lines"),
+            "res_model": "stock.move.line",
+            "view_mode": "list,form",
+            "domain": operation._domain_move_lines_to_do(),
```
jbaudoux: For put mode, it is using _domain_move_lines_to_do_all
Shouldn't this be implemented in the operation?

stock_vertical_lift/models/vertical_lift_operation_put.py
```diff
@@ -29,7 +29,12 @@ def _transitions(self):
                 # (by on_barcode_scanner)
                 lambda self: self.current_move_line_id,
             ),
-            self.Transition("scan_tray_type", "save"),
+            self.Transition(
+                "scan_tray_type",
+                "save",
+                lambda self: self._skip_scan_tray_type(),
+                direct_eval=True,
```
henrybackman, line 36: Why is this `direct_eval=True` while the other transitions are not?

henrybackman (approved):

stock_vertical_lift_storage_type/models/vertical_lift_operation_put.py
divad1196: For tracking purposes: we discovered during our discussion that Odoo actually uses `result_package_id`.
To match Odoo behavior, we will completely remove `package_id` and use `result_pacakge_id` only instead

stock_vertical_lift/models/vertical_lift_operation_put.py
divad1196: `direct_eval` means that, when you lend on the source state (here `scan_tray_type`), it immediately check all Transition with `direct_eval=True` to see if one can be done immediately.

For each of the other Transition:
- `scan_source` -> `scan_tray_type`: `scan_source` is the first one, that's when the operator says what he wants to do. There is nothing before this step
- `save` -> `release`: Another development asked to add an additional `Save & Release` button but keep the possibility to do `save` then `release` if needed. 
- `release` -> `scan_source`: Automatically running this step means sending back the tray which can be dangerous. It's trigger by a button

That's the full explanation, but it's also that it was like that before and not needed to be changed here :smile:

stock_vertical_lift/models/vertical_lift_shuttle.py
divad1196: Change done

stock_vertical_lift/models/vertical_lift_shuttle.py
divad1196: Also did the change here.
And you are right -> I used the polymorphism approach: each operation model has its own method.

vvrossem (approved):
`[IMP] stock_vertical_life: Use polymorphism on operation for the action_open_operations` has a typo ➡️ `stock_vertical_lift` instead of `stock_vertical_life`

divad1196:
@vvrossem 
> `stock_vertical_lift` instead of `stock_vertical_life`

I would never have seen it. thank you!

stock_vertical_lift_storage_type/models/vertical_lift_operation_put.py
```diff
@@ -17,50 +17,58 @@ class VerticalLiftOperationPut(models.Model):
...
 
-    def _has_storage_type_domain(self):
+    def _get_current_package(self):
         move_line = self.current_move_line_id
-        package_type = move_line.package_id.package_type_id
+        package_id = move_line.result_package_id
+        return package_id
+
+    def _has_storage_type_domain(self):
+        package_id = self._get_current_package()
```
simahawk: nitpicking: it returns a record not a an id 
```suggestion
        package = self._get_current_package()
```

simahawk (approved):
LGTM overall. Just code review.

stock_vertical_lift_storage_type/models/vertical_lift_operation_put.py
divad1196: @simahawk I agree with the remark.
The reason I use `_id` suffix is because Odoo ~~does it~~ used to do it.

I checked the [Coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions)
- fields are suffixed `_id` (because it reflects the database) ([source](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/07_relations.html))
  > By convention, many2one fields have the _id suffix.
- code doesn't (because we do get records)

Will adapt it

TDu (approved):

divad1196:
@jbaudoux @lmignon Can this be merged ?

lmignon:
@divad1196 Unfortunately, I don't use this module and have no knowledge of how it works. I will therefore refrain from commenting on it :smirk:

divad1196:
@lmignon wise decision. thank you for notifying

jbaudoux (approved):

divad1196:
@jbaudoux thank you a lot !

PR #2575, opened 2026-05-05
[18.0][MIG] stock_inventory_location_state

sebalix:
Migration from 14.0 PR:
- https://github.com/OCA/stock-logistics-warehouse/pull/1461

What changed:
- depends on `stock_inventory` now
- intermediate locations (having children) are inventoried (it we do not want this behavior, it could deserve an option, but by default we should not skip such locations as Odoo allows to put goods there)

Improvement that would be nice: display counters on `stock.inventory.location` list view regarding quants processed to get an idea where we stand without going into the details

cc @bguillot


bguillot (approved):
Code review and test. LGTM Thanks

stock_inventory_location_state/models/stock_inventory.py
```diff
@@ -30,39 +30,22 @@ def _compute_location_count(self):
...
-        sub_locations = sub_locations - existing_locations
-        for location in sub_locations:
+        domain_loc = self._get_base_domain(self.location_ids)
+        locations = self.env["stock.location"].search(domain_loc)
+        # Retrieve locations from quants too to not forget any
+        quants = self._get_quants(self.location_ids)
+        locations |= quants.location_id
+        locations -= existing_locations
+        for location in locations:
             self.env["stock.inventory.location"].create(
```
rousseldenis: @sebalix You should use multi create here

stock_inventory_location_state/models/stock_inventory_location.py
```diff
@@ -0,0 +1,69 @@
...
+
+from odoo import fields, models
+from odoo.osv import expression
+
+
+class StockInventoryLocation(models.Model):
+    _name = "stock.inventory.location"
+    _description = "Stock Inventory Location"
+
+    _rec_name = "location_id"
```
rousseldenis, line 12: Maybe too much spaces. But nitpicking

stock_inventory_location_state/models/stock_inventory_location.py
```diff
@@ -0,0 +1,69 @@
...
+        (
+            "inventory_location_unique",
+            "UNIQUE(inventory_id, location_id)",
+            "Inventory location must be unique per inventory.",
+        )
+    ]
+
+    def action_start(self):
+        self.ensure_one()
+        assert self.state in ("pending", "started")
```
rousseldenis: We are in an Odoo context, you should use a derivative of `UserError` (or itself) instead.

rousseldenis: Moreover, that could be great to add better message for user.

stock_inventory_location_state/tests/test_inventory_location.py
```diff
@@ -0,0 +1,74 @@
...
+from odoo import Command
+from odoo.exceptions import UserError
+from odoo.tests.common import TransactionCase
+
+
+class TestStockInventoryLocationState(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis, line 13: This can be avoided inheriting from `odoo.addons.base.tests.common.BaseCommon`

stock_inventory_location_state/models/stock_inventory.py
```diff
@@ -0,0 +1,93 @@
...
+                }
+            )
+        return res
+
+    def action_state_to_done(self):
+        self.ensure_one()
+        if any(loc.state != "done" for loc in self.sub_location_ids):
+            raise UserError(
+                _(
+                    "The following locations have not been inventoried yet:\n%s\n"
```
rousseldenis: You should use translation function arguments instead.

rousseldenis (changes requested):
Some remarks

stock_inventory_location_state/tests/test_inventory_location.py
sebalix: We did some tests in the past, and I remember that `BaseCommon` was introducing some overhead, so its benefit heayvily depend on the number of data you'll generate in the tests.
Still, I do not disagree to depend on it for the sake of simplicity.

sebalix: Just tested locally: in average, without `BaseCommon` running these tests takes ~1.2s, while with `BaseCommon` it takes ~1.85s.
Not sure of the benefit for the CI if all tests of all repos are based on `BaseCommon` ?

For now I'll keep the old fashioned `tracking_disable=True`.

stock_inventory_location_state/models/stock_inventory_location.py
sebalix: Indeed, done

sebalix:
@rousseldenis added all required changes in a fixup, and rebased the PR on top of 18.0

stock_inventory_location_state/models/stock_inventory_location.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+    ]
+
+    def action_start(self):
+        self.ensure_one()
+        if self.state not in ("pending", "started"):
+            raise UserError(
+                self.env._(
+                    "Unable to start inventory %(name)s: it isn't pending.",
+                    name=self.name,
```
metaminux: `stock.inventory.location` has no "name" field...

```suggestion
                    name=self.inventory_id.name,
```

stock_inventory_location_state/models/stock_inventory_location.py
```diff
@@ -0,0 +1,87 @@
...
+        self.write({"state": "started"})
+        # TODO create missing inventory line
+
+    def action_done(self):
+        self.ensure_one()
+        if self.state != "started":
+            raise UserError(
+                self.env._(
+                    "Unable to validate inventory %(name)s: it hasn't been started.",
+                    name=self.name,
```
metaminux: `stock.inventory.location` has no "name" field...

```suggestion
                    name=self.inventory_id.name,
```

stock_inventory_location_state/models/stock_inventory_location.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+        self.write({"state": "done"})
+
+    def action_reopen(self):
+        self.ensure_one()
+        if self.state == "done":
+            raise UserError(
+                self.env._(
+                    "Unable to re-open inventory %(name)s: it has been validated.",
+                    name=self.name,
```
metaminux: `stock.inventory.location` has no "name" field...

```suggestion
                    name=self.inventory_id.name,
```

stock_inventory_location_state/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Inventory Location State",
+    "summary": "Verify that all locations are counted.",
+    "version": "18.0.1.0.0",
+    "development_status": "Beta",
```
metaminux, line 7: I can see this module now depends on `stock_inventory` so this migration is also a refactor, but isn't "Beta" a bit **strong** for the migration of a "Mature" module

metaminux (changes requested):

stock_inventory_location_state/models/stock_inventory_location.py
sebalix: Thanks, I added a test for this :+1:

metaminux (approved):
LGTM 👌🏼

sebalix:
Sorry @metaminux , re-reading and testing what I have done in these actions methods and there were some issues:
- we want to report the location name we are interacting with in the error message, not the inventory name
- except for the error message raised by `action_reopen`, as soon as the main inventory is done, we cannot re-open inventory on locations

Fixup pushed.

metaminux:
> Sorry @metaminux , re-reading and testing what I have done in these actions methods and there were some issues:
> 
>     * we want to report the location name we are interacting with in the error message, not the inventory name
> 
>     * except for the error message raised by `action_reopen`, as soon as the main inventory is done, we cannot re-open inventory on locations
> 
> 
> Fixup pushed.

No problem, last commit LGTM too 👍🏼

metaminux:
Hello @rousseldenis 

You're the last one having requested changes.
Only the use of BaseCommon and the development status in manifest file remain in the discussion

Regards,

rousseldenis:
@sebalix Could you squash your commits ?

sebalix:
@rousseldenis done

stock_inventory_location_state/__manifest__.py
rousseldenis: @metaminux In fact, this has been enforced several years ago about development status dependencies between modules.

So, as `stock_inventory` is in `Beta`, the dependent modules should be maximum `Beta` (they can be `Alpha` also).

You need to promote the whole chain in order to promote this one.

rousseldenis: https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_inventory/__manifest__.py#L6

metaminux: OK, thanks a lot for this explanation.

metaminux (approved):

metaminux:
Hello @rousseldenis 

Can you approve and merge this PR ?

Thanks.

stock_inventory_location_state/tests/test_inventory_location.py
rousseldenis: In fact, the little difference you observed here won't be exponential as the concerned code is in class method.

https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/tests/common.py#L24

The only thing created is a partner (which is convenient in most of use cases - except here).

So, maybe it is not really necessary here. Even if I'd prefer not to have such line of code. :-)

rousseldenis (approved):

PR #2576, opened 2026-05-05
[18.0][MIG] stock_removal_location_by_priority: Migration to 18.0

Andrii9090-tecnativa:
In odoo 18 the method name is `_get_removal_strategy_order`
[Commit Odoo](https://github.com/odoo/odoo/commit/bbbc1f1ff22c1e1c5fdec46f02d760af1485f5be#diff-9a22420c4a72cc9b6d558b3d832846c2e44a4f3731c6df9579d66c6279c4804e)

@sergio-teruel @CarlosRoca13 ping

@Tecnativa
TT62242


BhaveshHeliconia (commented):
@Andrii9090-tecnativa could we keep translation-related changes (such as _ → self.env._) and pre-commit fixes like lazy formatting in “pre-commit auto fixes” commit instead of including them in the migration commit?

We usually get those pre-commit issues before the migrated module is added, so those changes fit better in a dedicated pre-commit fixes commit.

Andrii9090-tecnativa:
> @Andrii9090-tecnativa could we keep translation-related changes (such as _ → self.env._) and pre-commit fixes like lazy formatting in “pre-commit auto fixes” commit instead of including them in the migration commit?
> 
> We usually get those pre-commit issues before the migrated module is added, so those changes fit better in a dedicated pre-commit fixes commit.

Hi @BhaveshHeliconia! 
I don’t think this is fully correct. The `_ → self.env._ ` changes are related to the migration. Moreover, the OCA 18.0 migration guide explicitly mentions this change as part of the migration tasks. The pre-commit commit should probably contain only automatic linter/formatter fixes.

BhaveshHeliconia:
@Andrii9090-tecnativa Agreed for 18.0, since the migration guide explicitly mentions the _ → self.env._ changes as part of the migration tasks.

My point was more about 19.0 migrations, where these changes are usually introduced while fixing pre-commit issues, so keeping them in the pre-commit fixes commit could make sense there.

BhaveshHeliconia (approved):
Code review LGTM!

eduezerouali-tecnativa (approved):
Code review

PR #2579, opened 2026-05-21
[18.0][IMP] stock_lot_catalog_sale: Exclude serials used in pending orders

victoralmau:
Exclude serials used in pending orders

Similar to https://github.com/OCA/sale-workflow/pull/4336

Locked by:
- [x] `sale_order_lot_selection`: https://github.com/OCA/sale-workflow/pull/4354

@Tecnativa TT61756

### DIFF
--- a/stock_lot_catalog_sale/models/sale_order.py
+++ b/stock_lot_catalog_sale/models/sale_order.py
@@ -24,10 +24,27 @@ class SaleOrder(models.Model):
             ),
         }
 
+    def _get_stock_lot_catalog_not_available_sol_domain(self):
+        self.ensure_one()
+        return [
+            ("state", "in", ("draft", "sent")),
+            ("lot_id", "!=", False),
+            ("product_uom_qty", ">", 0),
+            ("product_id.tracking", "=", "serial"),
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
This should apply only to products with serial.

victoralmau:
> This should apply only to products with serial.

It is important to remember that this is a domain used to display lots; for this reason, those currently in use for other pending orders are excluded (https://github.com/OCA/stock-logistics-warehouse/pull/2579/changes#diff-a8f31bd5711338a637d0c182e1198acd36aea0b32da96697f906776080fed776R33) by filtering on tracking=serial

pedrobaeza (approved):

victoralmau:
Related to https://github.com/OCA/sale-workflow/pull/4354, I think it would be useful to make this option configurable, so I'm going to add it.

pedrobaeza:
As we are not finally use it, we can simply wait till someone needs it.

pilarvargas-tecnativa (approved):

WesleyOliveira98 (approved):
LGTM

PR #2580, opened 2026-05-21
[18.0][IMP] stock_lot_catalog_sale: Add a method for get locations used in _get_stock_lot_catalog_order_domain()

victoralmau:
Add a method for get locations used in `_get_stock_lot_catalog_order_domain()`

Similar to https://github.com/OCA/sale-workflow/pull/4338

@Tecnativa TT61756

### DIFF
--- a/stock_lot_catalog_sale/models/sale_order.py
+++ b/stock_lot_catalog_sale/models/sale_order.py
@@ -24,10 +24,20 @@ class SaleOrder(models.Model):
             ),
         }
 
+    def _get_stock_lot_catalog_order_locations(self):
+        """This method retrieves the appropriate locations to use in the
+        _get_stock_lot_catalog_order_domain() method and is useful for extending
+        functionality in other modules (for example, if you do not want to use
+        warehouse_id.lot_stock_id because you want to use all of the
+        company's warehouses).
+        """
+        self.ensure_one()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why not overriding `_get_stock_lot_catalog_order_domain`? It's easy to decouple the rest needed domain there, as having 2 methods for the same method seems too much.

victoralmau:
> Why not overriding `_get_stock_lot_catalog_order_domain`? It's easy to decouple the rest needed domain there, as having 2 methods for the same method seems too much.

Although you can do this by overriding the method, it’s confusing (at least in my opinion) to have:

```
domain = [("location_id", "child_of", location_1.ids)] 
domain += [("location_id", "child_of", locations.ids)]
```

(locations contains location_1)

For that reason, I suggest this change, so that the locations used are more readable and easily definable.

pedrobaeza:
You can remove elements from the domain, just doing:

```
domain.remove(("location_id", "child_of", location_1.ids))
```

so it's not a problem to override the existing method.

victoralmau:
> You can remove elements from the domain, just doing:
> 
> ```
> domain.remove(("location_id", "child_of", location_1.ids))
> ```
> 
> so it's not a problem to override the existing method.

Okay, I'll use that method to remove that part of the domain.

PR #2589, opened 2026-06-05
[18.0][IMP] stock_location_lockdown: add outbound lock

divad1196:
# Context

`stock_location_lockdown` only lock inbound stock, not outbound.
Furthermore, it can only apply on location that already has stock.

The module as it existed was in fact just adding a way to prevent a location from having stock.

## Retro-compatibility

The changes made to this module are retro-compatible by default with how it previously behaved.
It's possible to allow lockdown on a location with stock. The setting is on the company level.

# Features

It adds 3 options to a location:
- **Lock inbound**: prevent stock from moving **in** the location. Propagates to child locations.
- **Lock outbound**: prevent stock from moving **out** of the location. Propagates to child locations.
- **Location without physical stock**: that's what the module was previously doing. 
  This one does not propagate to children location

When we lock for `outbound`, the quantities of the locked location appear as reserved.
This way,  `Check Availability` and forecasting can correctly take it into account.

# NOTE
## `stock_inventory_lockdown`

I think the existing module `stock_inventory_lockdown` should rely on this one or even be merged.

### DIFF
--- a/stock_location_lockdown/__manifest__.py
+++ b/stock_location_lockdown/__manifest__.py
@@ -4,13 +4,17 @@
 {
     "name": "Stock Location Lockdown",
     "summary": "Prevent to add stock on locked locations",
-    "author": "Akretion, Odoo Community Association (OCA)",
+    "author": "Akretion, Camptocamp, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-warehouse",
     "category": "Warehouse",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["stock"],
-    "data": ["views/stock_location.xml"],
+    "pre_init_hook": "pre_init_hook",
+    "depends": ["stock", "stock_inventory"],
+    "data": [
+        "views/res_config_settings_views.xml",
+        "views/stock_location.xml",
+    ],
 }
--- a/stock_location_lockdown/models/res_company.py
+++ b/stock_location_lockdown/models/res_company.py
@@ -1,4 +1,3 @@
-# Copyright 2026 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
--- a/stock_location_lockdown/models/res_company.py
+++ b/stock_location_lockdown/models/res_company.py
@@ -1,0 +1,36 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    allow_lockdown_on_stocked_location = fields.Boolean(
+        string="Allow Inbound Lockdown on Stocked Locations",
+        help="When enabled, an internal location can be marked as "
+        "'Block Stock Entrance' even if it already contains stock.",
+    )
+    block_location_on_inventory = fields.Boolean(
+        string="Block Location on Inventory",
+        help="When enabled, a location (and its descendants) is treated as "
+        "blocked for both inbound and outbound while an inventory is "
+        "in progress on it.",
+    )
+
+    def _refresh_inventory_blocked(self, company_ids):
+        locations = (
+            self.env["stock.location"]
+            .sudo()
+            .search([("company_id", "in", company_ids)])
+        )
+        if locations:
+            locations._compute_is_inventory_blocked()
+
+    def write(self, vals):
+        refresh_inventory = "block_location_on_inventory" in vals
+        res = super().write(vals)
+        if refresh_inventory:
+            self._refresh_inventory_blocked(self.ids + [False])
+        return res
--- a/stock_location_lockdown/models/stock_location.py
+++ b/stock_location_lockdown/models/stock_location.py
@@ -14,7 +14,7 @@ class StockLocation(models.Model):
         "aggregation."
     )
     block_stock_exit = fields.Boolean(
-        help="If checked, taking stock out of this location won't be allowed. "
+        help="If checked, moving goods out of this location won't be allowed. "
         "Quants on this location are treated as if reserved and won't be "
         "available for picking. Propagates to descendant locations through "
         "the outbound block aggregation."
--- a/stock_location_lockdown/models/stock_location.py
+++ b/stock_location_lockdown/models/stock_location.py
@@ -1,7 +1,7 @@
 # Copyright 2019 Akretion
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 from odoo.exceptions import UserError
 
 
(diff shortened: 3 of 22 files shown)


stock_location_lockdown/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+            .search([("company_id", "in", company_ids)])
+        )
+        if locations:
+            locations._compute_is_inventory_blocked()
+
+    def write(self, vals):
+        refresh_inventory = "block_location_on_inventory" in vals
+        res = super().write(vals)
+        if refresh_inventory:
+            self._refresh_inventory_blocked(self.ids + [False])
```
grindtildeath, line 34: What the goal with using `[False]` here?

stock_location_lockdown/models/res_company.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    allow_lockdown_on_stocked_location = fields.Boolean(
```
grindtildeath, line 9: I would use the term "not empty" or "non empty" instead of "stocked" or "contains stock", since it's usually what is used in Odoo for this concept, and does not rely on verb conjugation.

stock_location_lockdown/models/stock_location.py
```diff
@@ -1,33 +1,231 @@
...
-                        "It is impossible to prohibit this location from\
-                    receiving products as it already contains some."
-                    )
-                )
+        help="If checked, putting stock on this location won't be allowed. "
+        "Propagates to descendant locations through the inbound block "
+        "aggregation."
+    )
+    block_stock_exit = fields.Boolean(
+        help="If checked, taking stock out of this location won't be allowed. "
```
grindtildeath:
```suggestion
        help="If checked, moving goods out of this location won't be allowed. "
```

stock_location_lockdown/models/stock_location.py
```diff
@@ -1,33 +1,231 @@
...
+            "no_physical_stock",
+            "is_ancestor_outbound_blocked",
+            "is_inventory_blocked",
+        )
+
+    # ------------------------------------------------------------------
+    # Compute methods
+    # ------------------------------------------------------------------
+
+    @api.depends(lambda self: self._is_ancestor_inbound_blocked_depends())
```
grindtildeath, line 108: Why moving depends condition into a function if this function is not used anywhere else? When I look at the compute function, there's some value in having its dependencies listed without having to scroll to another function.

stock_location_lockdown/models/stock_location.py
```diff
@@ -1,33 +1,231 @@
...
+    is_outbound_blocked = fields.Boolean(
+        compute="_compute_is_outbound_blocked",
+        store=True,
+        help="Aggregated outbound block flag (self / ancestor / inventory). "
+        "Use this for condition evaluation, not 'Block Stock Exit' directly.",
+    )
+
+    # ------------------------------------------------------------------
+    # Depends declarations as overridable model methods so child modules
+    # can extend the dependency graph without rewriting the decorator.
```
grindtildeath, line 65: Actually there's no need for this, as Odoo will consider extra dependencies from child modules automatically...

stock_location_lockdown/models/stock_location.py
```diff
@@ -1,33 +1,231 @@
...
+    # Helpers
+    # ------------------------------------------------------------------
+
+    def _ancestor_location_ids(self, include_self=True):
+        """Return ids on the materialized parent_path."""
+        self.ensure_one()
+        ids = [int(x) for x in (self.parent_path or "").split("/") if x]
+        if not include_self:
+            ids = [i for i in ids if i != self.id]
+        return ids
```
grindtildeath, line 161: Why not using `parent_of` in your domains instead of this?

stock_location_lockdown/models/stock_location.py
```diff
@@ -1,33 +1,231 @@
...
+
+    def _check_lockdown_against_existing_stock(self, label):
+        """Raise if a location to be blocked already contains stock and the
+        company has not opted in to allowing that."""
+        if not self or self.env.company.allow_lockdown_on_stocked_location:
+            return
+        self.env["stock.quant"]._unlink_zero_quants()
+        stocked = self.filtered("quant_ids")
+        if stocked:
+            raise UserError(
```
grindtildeath, line 224:
```suggestion
            raise ValidationError(
```
https://github.com/odoo/odoo/blob/18.0/odoo/exceptions.py#L95

stock_location_lockdown/models/stock_move_line.py
```diff
@@ -0,0 +1,29 @@
...
+    _inherit = "stock.move.line"
+
+    @api.constrains("location_id", "location_dest_id", "state")
+    def _check_blocked_location(self):
+        for line in self.filtered(lambda m: m.state == "done"):
+            if line.location_id.is_outbound_blocked:
+                raise ValidationError(
+                    self.env._(
+                        "The location %(location)s is blocked for outbound and "
+                        "stock cannot be moved out of it.",
```
grindtildeath:
```suggestion
                        "goods cannot be moved out of it.",
```
IMO it would be better to state we are "moving goods" instead of "moving stock"

stock_location_lockdown/models/stock_move_line.py
```diff
@@ -0,0 +1,29 @@
...
+                        location=line.location_id.display_name,
+                    )
+                )
+            if line.location_dest_id.is_inbound_blocked:
+                raise ValidationError(
+                    self.env._(
+                        "The location %(location)s is blocked for inbound and "
+                        "stock cannot be moved into it.",
```
grindtildeath:
```suggestion
                        "goods cannot be moved into it.",
```

stock_location_lockdown/models/stock_route.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2026 Camptocamp SA
```
grindtildeath: To remove

stock_location_lockdown/models/stock_quant.py
```diff
@@ -1,19 +1,54 @@
...
 
 class StockQuant(models.Model):
     _inherit = "stock.quant"
 
+    # Outbound reservation lock: for a quant in an outbound-blocked location we
+    # hold ``reserved_quantity == quantity`` so the stock is not available for
+    # picking (available = quantity - reserved = 0) and is removed from the
+    # forecast figures natively. The genuine reservation is parked here so it
+    # can be restored when the location is unblocked.
+    lock_real_reserved = fields.Float(default=0.0)
```
grindtildeath, line 16: I do not get what's the goal here. If you want to block moves on a quant having reservations, IMO such reservations must be lifted or they would be incoherent while the location is blocked. Move lines must then be reassigned and the situation cannot be the same than it used to be before.

Afterwards, it's the reservation mechanism that must reset the reserved quantity according to the existing reservation.

Instead of all of these sketchy overrides, I'd suggest to add a confirmation wizard whenever the "outbound blocking" field is marked on the location while we have existing reservations, to warn the user that existing reservation will be lifted and stock moves will be reassigned.

stock_location_lockdown/models/stock_quant.py
```diff
@@ -24,3 +59,111 @@ def check_location_blocked(self):
...
+        return self.is_outbound_reservation_locked
+
+    def _pin_outbound_reservation(self):
+        """Hold reserved_quantity at quantity, parking the genuine reservation
+        in lock_real_reserved, so the stock is unavailable for picking."""
+        for quant in self:
+            if quant._is_outbound_reservation_pinned():
+                continue
+            quant.lock_real_reserved = quant.reserved_quantity
+            quant.reserved_quantity = quant.quantity
```
grindtildeath, line 81: ⚠️ 
This is wrong to me. By doing this you introduce a discrepancy between existing reservations (move lines) and reserved quantity on the quant.

grindtildeath (changes requested):
IMO the design is not right here.

We should never have `stock.quant.reserved_quantity` that does not reflect the existing reservations.

I would suggest the following:
- Add a confirmation wizard when a location is marked to restrict outgoing moves. Such wizard should warn the user that existing reservation will be lifted and reassigned.
- Materialize the new reserved qty through a stock move to a dedicated virtual location for location blocking.
- Make whenever goods enter a location that is "outbound blocked" are then reserved through such a stock move (it could be the same one)

stock_location_lockdown/models/res_company.py
divad1196: @grindtildeath common pattern to say "within this list or undefined"

stock_location_lockdown/models/stock_location.py
divad1196: @grindtildeath to extend the behavior without overriding.
The module `stock_inventory_lockdown` does a similar job and could depend on this module later.

I found this solution in the source code of Odoo

stock_location_lockdown/models/stock_location.py
divad1196: @grindtildeath AFAIK, defining `api.depends` on the override function replace the one from other modules.
That's why Odoo also uses `depends(lambda self: ...)` in some parts of its code

stock_location_lockdown/models/stock_location.py
divad1196: @grindtildeath First reason: I often forgot about `child_of` and even more about `parent_of`.

But 2 practical reasons:
- parent_path is pre-computed in BaseModel. 
  Using it can remove an IO
- `parent_of` is a lot heavier than searching for specific ids

stock_location_lockdown/models/res_company.py
grindtildeath: Which means if you write on CompanyA, a stock location that would be shared between CompanyA and CompanyB would then be impacted.

It's a choice you're doing, that as I user of CompanyB I would not expect if a setting is changed only in CompanyA...

stock_location_lockdown/models/stock_location.py
divad1196: > UserError(Exception):
>    [...] when the user tries to do something that has no sense given the current state of a record. 

It's not a data error, it's a functional error. The user should not have tried to use stock from a locked location.

stock_location_lockdown/models/stock_move_line.py
divad1196: completely agree

stock_location_lockdown/models/stock_location.py
grindtildeath: That's wrong, you can find examples in odoo, eg:
 - https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_move.py#L279
 - https://github.com/odoo/odoo/blob/18.0/addons/repair/models/stock_move.py#L32
 - https://github.com/odoo/odoo/blob/18.0/addons/mrp/models/stock_move.py#L320

stock_location_lockdown/models/stock_location.py
grindtildeath: cf my comment above

stock_location_lockdown/models/stock_quant.py
divad1196: We don't want to block quant with reservation. The "has reservation" is not a critieria. All `stock.quant` of the locked locations get reserved.
So, no need for a wisard either.

`lock_real_reserved` is the real reserved quantity when the lock is set.
When we lock a location, the reserved quantity is set to the quantity on hand so that the whole stock appear as reserved.
When the lock is turned off, the real reserved quantity is restored.

stock_location_lockdown/models/stock_route.py
divad1196: @grindtildeath okay. I saw that other companies left it in some modules, so I didn't know what is the norm

stock_location_lockdown/models/stock_quant.py
divad1196: move lines are not reservation and are not checked for that even in submodules. There are a whole bunch of module playing with reservation in https://github.com/OCA/stock-logistics-reservation

stock_location_lockdown/models/res_company.py
divad1196: @grindtildeath it is a choice yes.

Just for clarity: This function just tell who to check. The one who would filter such cases would be [_refresh_inventory_blocked](https://github.com/OCA/stock-logistics-warehouse/pull/2589/changes/BASE..9722fe0da40eda07d48c756ea6a0cf6db86be882#diff-77e9abce4103c3800bfd690ec1da502e41c23016b49ed9701b47fbe2086243b5R22)

We cannot leave the option but "don't lock the location".
So the 2 options are:
- lock it anyway
- remove the option to lock on shared location

And **we could allow to choose between both** with a server wide setting "Allow lock on shared locations"

stock_location_lockdown/models/stock_location.py
divad1196: @grindtildeath This behavior has changed since previous version.
I found [this comment](https://www.reddit.com/r/Odoo/comments/1k2wafz/how_to_override_apidepends_and_remove_a/) that mentions the behavior change:
> The "depends" decorator has become cumulative in the latest versions

There are still many occurences for this pattern in the OCA and Odoo source code.

There are still benefits of this approach (e.g. remove options, use dynamic triggers, ...) then main advantage would be retro-compatibility IMO: Do we care ?
If not, then I will do the change.

divad1196:
@grindtildeath 
> * Materialize the new reserved qty through a stock move to a dedicated virtual location for location blocking.

There are many reasons why this won't work:
- Side effect from other modules like dynamic routing
- Recomputation on unreserve
- What source/destination to use
- Heavy inserts and recomputations
- ...

This is why this idea got rejected.
On the other hand, `stock.quant` are not the reflection of moves and from a functional point of view, moves are not the only reason you want to reserve some stock. Odoo does not rely much on stock.move either.
Reserved quantity is just how much quantities must not be used for a different purpose.

TDu:
This is a big change from the original module: 
* adding recursive logic for the setting on locations
* new module dependencies
* previous logic was intended to be used as a permanent configuration (from what I understand)

This is why IMO it would be better suited for a new module. Maybe `stock_location_quarantine` ?

About the implementation choice, I think that staying closer to Odoo mechanism (like proposed by @grindtildeath ) by reserving through stock move / transfer with a dedicated operation type and blocking those moves so they can not be set to done is an interesting alternative.

divad1196:
@TDu Thank you for checking. To respond in order:
- the previous behavior without recursion is still available but with a different name.
- no dependency were added. what are you refering to ?
- not necessarily: you could toggle it on/off as you pleased. The only constrain was that the location had to be empty before locking it.

### For the module name

The previous version of the module was not really a lockdown. It technically allows outgoing traffic and only applied to empty locations => In fact, it ensured that the location stayed empty / didn't contain goods.

This is why I added a field `No physical stock` (`no_physical_stock`) to keep the retro-compatibility.
NOTE: I should add a migration during module update.

### Why not stock.move

I did test `stock.move`, but you need a lot more work for something less stable.
If we were dealing with the `quantity` field, I would 100% agree the we must keep the balance as Odoo does.
But here it's about reservation, there are no balance.

stock_location_lockdown/__manifest__.py
```diff
@@ -4,13 +4,17 @@
...
     "category": "Warehouse",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["stock"],
-    "data": ["views/stock_location.xml"],
+    "pre_init_hook": "pre_init_hook",
+    "depends": ["stock", "stock_inventory"],
```
TDu, line 15: I meant this dependency -> `stock_inventory`

divad1196: @TDu yes, you are right, I missed it. 
In my notes I mentioned that this part of the code should be moved to `stock_inventory_location`.
Similarly, the reservation part is not necessary for lockdown, we should be able to configure it or have it in another module.

i am waiting for functional feedbacks before taking steps in this direction

jbaudoux:
@divad1196 Could have a look to this module https://github.com/OCA/stock-logistics-workflow/pull/1394 that allows to lock a quant on a location. It depends on the context but you don't always want to lock all the content of a location.

I haven't reviewed your changes but in my opinion we should differentiate locking a location and locking some stock. IMO, locking a location means you cannot put stock on it anymore but also you will exclude it from putaway computation.
Locking stock could be a quant or part of a quant, meaning you can go down to a part of a specific lot on a package on a location. That's what the other module is trying to achieve. Having the finest grain level allows you to also lock a complete lot, package, location... It is achieved by making a new reservation with a new move. The destination location of the lock move allows you to control what availability it affects for the reordering rules.

divad1196:
@jbaudoux The need here is to lock the full location, not just some goods in it. That's the goal of a location lockdown.
Scenarios can be fire/bio hazard, flooding, .. there are many reason to want a full location lockdown.
The module `stock_location_lockdown` already targets a full location.

I agree that there are different context, and this is why there are different modules, the one you mentioned and the one in this PR.

PR #2592, opened 2026-06-08
[IMP] sale_stock_on_hand_popup: Show popup in picking

SirPyTech:
Show the quants popup from the picking lines.


SirPyTech:
`pre-commit` will be fixed after https://github.com/OCA/stock-logistics-warehouse/pull/2593.

PR #2594, opened 2026-06-11
[18.0][IMP] stock_inventory: include subcategories in inventory adjustments

IsabelAForgeFlow:
The original _get_domain_category_quants() implementation used child_id.ids to filter by category, which only matched products in the selected category and its direct children — products belonging to grandchild categories were silently excluded from the inventory adjustment. 

This improvement adds a category_include_subcategories boolean field (enabled by default) that switches the filter to use child_of, making category selection fully recursive. Users who want the previous behaviour — direct children only — can uncheck the option on the inventory form.


JordiMForgeFlow (approved):

stock_inventory/models/stock_inventory.py
```diff
@@ -90,6 +90,14 @@ class InventoryAdjustmentsGroup(models.Model):
         readonly=True,
     )
 
+    category_include_subcategories = fields.Boolean(
+        string="Include Subcategories",
+        default=True,
+        readonly=True,
```
IriaAlonso: `readonly` attribute is now declared on views

IriaAlonso (commented):
code review LGTM only minor comment
please upgrade the `version` on the `manifest`

stock_inventory/models/stock_inventory.py
IsabelAForgeFlow: Done!

BhaveshHeliconia (approved):
LGTM!

PR #2603, opened 2026-06-25
[17.0][FIX] stock_inventory: fix move line ordering after filtering

### LINKED ISSUE #2499: [stock_inventory 17.0]: BUG - Wrong stock.move.line is selected on _apply_inventory() method
https://github.com/OCA/stock-logistics-warehouse/blob/df5af0d57afa47626e2c4571d489e266b7736672/stock_inventory/models/stock_quant.py#L35

After yearly inventory i have a lot of mowes with reference to inventory but they are not paired wit it (`inventory_adjustment_id` on `stock.move.lines` is `Null`.

The problem is that filtered changes orders

eg:
```
            moves = record_moves.search(
                [
                    ("product_id", "=", rec.product_id.id),
                    ("lot_id", "=", rec.lot_id.id),
                    "|",
                    ("location_id", "=", rec.location_id.id),
                    ("location_dest_id", "=", rec.location_id.id),
                ],
                order="create_date asc",
            )
```
returns 
```stock.move.line(64837, 483782, 653681, 655690, 661947, 1176881)```

and after filtered it gives 
```stock.move.line(64837, 483782, 1176881, 653681, 655690, 661947)```

move `1176881` is proper one but to write it takes inventory_adjustment_id `661947` which is in some cases it is (but not always) `'Product Quantity Confirmed'` with 0 quantity.

can You modify it and add `.sorted` at the end 
```
                moves = record_moves.search(
                    [
                        ("product_id", "=", rec.product_id.id),
                        ("lot_id", "=", rec.lot_id.id),
                        "|",
                        ("location_id", "=", rec.location_id.id),
                        ("location_dest_id", "=", rec.location_id.id),
                    ],
                    order="create_date asc",
                ).filtered(
                    lambda x, rec=rec: not x.company_id.id
                    or not rec.company_id.id
                    or rec.company_id.id == x.company_id.id
                ).sorted(key=lambda m: (m.create_date or fields.Datetime.now(), m.id))
```

### PR DESCRIPTION
**Pull request by HIDORAKAI002**
Fixes #2499. The \iltered\ method does not guarantee order preservation, causing the wrong stock.move.line to be selected when linking it to an inventory adjustment. This commit enforces proper sorting by \create_date\ and \id\ after filtering.

### DIFF
--- a/stock_inventory/models/stock_quant.py
+++ b/stock_inventory/models/stock_quant.py
@@ -37,7 +37,7 @@ class StockQuant(models.Model):
                 lambda x, rec=rec: not x.company_id.id
                 or not rec.company_id.id
                 or rec.company_id.id == x.company_id.id
-            )
+            ).sorted(key=lambda m: (m.create_date or fields.Datetime.now(), m.id))
             if len(moves) == 0:
                 raise ValueError(_("No move lines have been created"))
             move = moves[len(moves) - 1]

### CODE REVIEW

File: stock_inventory/models/stock_quant.py [hunk 1]
```diff
@@ -37,7 +37,7 @@ def _apply_inventory(self):
                 lambda x, rec=rec: not x.company_id.id
                 or not rec.company_id.id
                 or rec.company_id.id == x.company_id.id
-            )
+            ).sorted(key=lambda m: (m.create_date or fields.Datetime.now(), m.id))
```
rousseldenis, line 40: IMHO, using `fields.Datetime.now()` is not necessary. Using `id` key after is sufficent. This can cause performance problems.

stock_inventory/models/stock_quant.py
```diff
@@ -37,7 +37,7 @@ def _apply_inventory(self):
...
             if len(moves) == 0:
                 raise ValueError(_("No move lines have been created"))
             move = moves[len(moves) - 1]
```
rousseldenis, line 43: @HIDORAKAI002 IMHO, this line could be problematic. What do we want to achieve with this ?

PR #2604, opened 2026-06-28
[FIX] stock_inventory_preparation_filter: exclude virtual locations from domain filter

### LINKED ISSUE #2498: [17.0] stock_inventory_preparation_filter: The filtered product by location is wrong
## Module
stock_inventory_preparation_filter

## The bug
The filtered product is incorrect and displays the products in the virtual locations to adjust.

Image

Image

## To Reproduce
**17.0 or 18.0**:

Steps to reproduce the behaviour:
1. Create a new Inventory Adjustment (Inventory -> Operations -> Inventory Adjustment)
2. Select the Filtered Product - adjust the domain, e.g. using product category
3. Click on Begin Adjustment

**Expected behaviour**
The Virtual location should not be included, and only display the products from the selected location

SaraMohamedlorim:
## Problem

When using **Filtered Products** (domain) selection in Inventory Adjustment,
the results include quants from virtual locations (e.g. Customers, Inventory Adjustments),
which should not appear in a physical inventory count.

Fixes #2498

## Root Cause

`_get_quants()` was searching quants by `product_id` only, without filtering
by `location_id`, causing virtual location quants to be included.

## Solution

Added `location_id` filter to the quant search in `_get_quants()` when
`product_selection == "domain"`, consistent with how other selection
modes handle location filtering.

## Steps to Reproduce

1. Create an Inventory Adjustment
2. Set **Filtered Products** and define a domain (e.g. by category)
3. Click **Begin Adjustment**
4. ❌ Virtual location products appear in the results

## After Fix

- ✅ Only quants from the selected location are returned
- ✅ Virtual locations (Customers, Inventory Adjustments, etc.) are excluded
- ✅ Test added to cover this scenario

### DIFF
--- a/stock_inventory_preparation_filter/models/stock_inventory.py
+++ b/stock_inventory_preparation_filter/models/stock_inventory.py
@@ -42,5 +42,10 @@ class StockInventory(models.Model):
         if self.product_selection == "domain":
             domain = safe_eval(self.product_domain)
             products = self.env["product.product"].search(domain)
-            return self.env["stock.quant"].search([("product_id", "in", products.ids)])
+            return self.env["stock.quant"].search(
+                [
+                    ("product_id", "in", products.ids),
+                    ("location_id", "in", locations.ids),
+                ]
+            )
         return super()._get_quants(locations)
--- a/stock_inventory_preparation_filter/tests/test_stock_inventory_preparation_filter.py
+++ b/stock_inventory_preparation_filter/tests/test_stock_inventory_preparation_filter.py
@@ -146,3 +146,32 @@ class TestStockInventoryPreparationFilterCategories(common.TransactionCase):
         line1 = inventory.stock_quant_ids[0]
         self.assertEqual(line1.product_id, self.product1)
         self.assertEqual(line1.quantity, 2.0)
+
+    def test_inventory_domain_filter_excludes_virtual_locations(self):
+        """Test that domain filter does not include quants from virtual locations."""
+
+        virtual_location = self.env.ref("stock.stock_location_customers")
+        self.env["stock.quant"].create(
+            {
+                "product_id": self.product1.id,
+                "product_uom_id": self.product1.uom_id.id,
+                "location_id": virtual_location.id,
+                "quantity": 5.0,
+            }
+        )
+
+        inventory = self.inventory_model.create(
+            {
+                "name": "Domain inventory excluding virtual locations",
+                "product_selection": "domain",
+                "product_domain": [("id", "=", self.product1.id)],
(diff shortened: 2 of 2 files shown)


rousseldenis:
@SaraMohamedlorim Thanks for this.

You should follow this migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

You need to keep commits history

nurinayati (commented):
Hello @SaraMohamedlorim thank you for fixing this. I can not find the update on Runboat, could you follow the guide, please?

I would love to test and approve it. please let me know once you update and ready to test on Runboat. ;)

SaraMohamedlorim:
@rousseldenis @nurinayati  Thanks for the feedback. You're right — the same bug also exists on 16.0 and 18.0, so I've followed the OCA convention of fixing the oldest maintained branch first and forward-porting from there:

Opened a fix on 16.0 first: #[2604]
Cherry-picked that commit (with -x) into this branch, so the commit history is preserved and traceable back to the original fix: (cherry picked from commit 562e287fd583f7c8919b7a25cc9e78312a8c0183)
I've just force-pushed the updated commit here, so this PR now has a single commit that keeps the history instead of a standalone rewrite. This should also trigger a fresh Runboat build.

Let me know if you'd like me to do the same for 18.0.

nurinayati:
thank you @SaraMohamedlorim. Can we include the sublocations? It's missing for inventory adjustments
Please see example below I think the inventory adjustment should include the stock from WH/Stock/Shelf 2

Inventory Adjustment Current stock
image

WH/Stock: current stock including sublocations
image

nurinayati:
Hi @rousseldenis, are there any other technical points to fix? Could you review again, please? thanks

stock_inventory_preparation_filter/models/stock_inventory.py
```diff
@@ -42,5 +42,10 @@ def _get_quants(self, locations):
         if self.product_selection == "domain":
             domain = safe_eval(self.product_domain)
             products = self.env["product.product"].search(domain)
-            return self.env["stock.quant"].search([("product_id", "in", products.ids)])
+            return self.env["stock.quant"].search(
+                [
+                    ("product_id", "in", products.ids),
+                    ("location_id", "in", locations.ids),
```
Demesmaeker, line 48: We might want to also check if it's not in the `locations.child_ids` here 🙂

Demesmaeker (commented):
Hello @SaraMohamedlorim 🦆 

Thanks for your fix, I just look at it with @nurinayati , and it seems you are missing the child locations / branches of that virtual location.

Have a nice day,

PR #2458, opened 2025-11-09
[18.0][IMP] stock_mts_mto_rule: Several fixes

rousseldenis:
@florian-dacosta I'm not sure about the constraint you put on rules (mts/mto) location_src_id.

I have a case where :

- I have this configuration:

                    WH
                   Stock
      RESERVE          PICK
                     SHOP   MAIN                            

I have a rule that is working without the constraint:

{
    action: 'split_procurement',
    location_src_id: PICK
    location_dest_id: SHOP
    mts_rule_id: {
          action: pull,
          location_src_id: MAIN,
          location_dest_id: SHOP,
},
    mto_rule_id: {
        action : 'pull',
        location_src_id: RESERVE,
       location_dest_id: SHOP,
}

But maybe that something that didn't work before v18. Your advice is welcome. Thanks

PR #2609, opened 2026-07-06
[18.0][IMP] stock_mts_mto_rule: Several improvements

rousseldenis:


rousseldenis:
@florian-dacosta I've merged the migration PR: https://github.com/OCA/stock-logistics-warehouse/pull/2489


This in order to gather some improvements in this PR with better readability.

As discussed in https://github.com/OCA/stock-logistics-warehouse/pull/2458, I've removed an unwanted constraint.

@florian-dacosta @ypapouin I've refactored the way this behaves as I've observed great performances problems if procurements runs are not grouped.

Do you see anything that I should pay attention to with that change ?

Many thanks

rousseldenis:
FYI, I passed from 500 seconds to 55 to run the scheduler with this change. :smile:

ypapouin:
@rousseldenis, I agree about the unneeded constraint , it's already removed on my side since it breaks existing workflows on our production database.

About the procurement grouping, I didn't made any tests and I may be wrong about it, but I believe that it would lead to different results (that initial implementation): since all evaluation is done first and all processing is done after with your change, that means that when the first procurement run it will change stock quantities, quantities that would normally have an impact on the 2nd procurement (same product, etc.) that would adapt its split logic.

If I'm right, this optimization could still be possible if products are not the same.

rousseldenis:
> @rousseldenis, I agree about the unneeded constraint , it's already removed on my side since it breaks existing workflows on our production database.
> 
> About the procurement grouping, I didn't made any tests and I may be wrong about it, but I believe that it would lead to different results (that initial implementation): since all evaluation is done first and all processing is done after with your change, that means that when the first procurement run it will change stock quantities, quantities that would normally have an impact on the 2nd procurement (same product, etc.) that would adapt its split logic.
> 
> If I'm right, this optimization could still be possible if products are not the same.

@ypapouin Thanks for this.

Indeed, I think I need to refine the logic in order to keep the procured quantity for each product by location. And do tests on that :smile:

PR #2611, opened 2026-07-10
[18.0][FIX] stock_vertical_lift: Fix shuttle's inventory-mode filter

divad1196:
# Context

The shuttle view was filtering based on `inventory_quantity_set`.
This field indicates that the value was **entered by the operator**, not that it has been applied.

# Fix 
Remove `inventory_quantity_set` from the filter and use the `inventory_date` instead,
like odoo's `to_count` filter.

Also fix the `vertical_lift_done` when `inventory_date` changes


sebalix (approved):

simahawk (approved):
LGTM

PR #2612, opened 2026-07-14
[18.0][ADD] stock_inventory_hide_apply_all

smorita7749:
Some organizations want counts to be applied and reviewed line by line, using the per-line Apply action only. This module removes "Apply All" button from the view to enforce that workflow.

@qrtl QT7006


stock_inventory_hide_apply_all/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "summary": "Hide the 'Apply All' button on the inventory adjustment list",
+    "version": "18.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "Quartile, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["stock"],
+    "data": ["views/stock_quant_views.xml"],
```
yostashiro: @smorita7749 Please add yourself as a maintainer.

smorita7749: Done!

yostashiro (approved):
Code review. LGTM.

Saran440:
I have the same issue, and I fixed by adding access groups.

Would it make sense to make this module more flexible by adding two separate access groups?

1. Apply Inventory Adjustments
2. Apply All Inventory Adjustments

Users with only the first group could use **Apply**,
while users with the second group could also use **Apply All**.
If a user has neither group, both buttons would be hidden.

What do you think?

nobuQuartile (approved):
Code review: LGTM

smorita7749:
> I have the same issue, and I fixed by adding access groups.
> 
> Would it make sense to make this module more flexible by adding two separate access groups?
> 
> 1. Apply Inventory Adjustments
> 2. Apply All Inventory Adjustments
> 
> Users with only the first group could use **Apply**, while users with the second group could also use **Apply All**. If a user has neither group, both buttons would be hidden.
> 
> What do you think?

Thank you for your comment.
I can understand your point, but I think the scope is a bit broad.

This module intends to just hide the "Apply All" button to force users to check a box and apply, so I would like it to be simple.
I'm concerned about it even being complicated if it controls the access groups.

Please let me know if you have any points.

IriaAlonso (approved):
LGTM code review

stock_inventory_hide_apply_all/views/stock_quant_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="stock_quant_view_list_inventory_editable" model="ir.ui.view">
+        <field name="name">stock.quant.inventory.list.editable.hide.apply.all</field>
+        <field name="model">stock.quant</field>
+        <field name="inherit_id" ref="stock.view_stock_quant_tree_inventory_editable" />
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='action_apply_all']" position="attributes">
+                <attribute name="invisible">1</attribute>
```
rousseldenis: Maybe a group could be more flexible ?

rousseldenis: :smile: Already suggested

rousseldenis:
> > I have the same issue, and I fixed by adding access groups.
> > Would it make sense to make this module more flexible by adding two separate access groups?
> > 
> > 1. Apply Inventory Adjustments
> > 2. Apply All Inventory Adjustments
> > 
> > Users with only the first group could use **Apply**, while users with the second group could also use **Apply All**. If a user has neither group, both buttons would be hidden.
> > What do you think?
> 
> Thank you for your comment. I can understand your point, but I think the scope is a bit broad.
> 
> This module intends to just hide the "Apply All" button to force users to check a box and apply, so I would like it to be simple. I'm concerned about it even being complicated if it controls the access groups.
> 
> Please let me know if you have any points.

As suggested, installing this will hide the action until a user is added to the group.

IMHO, this is more flexible than putting an `invisible` that needs the entire module to be uninstalled if some advanced users want to do it;

stock_inventory_hide_apply_all/views/stock_quant_views.xml
smorita7749: Thank you for your feedback!
I added a group to control visibility.
Could you please review again when you have time?

nobuQuartile (approved):
Code review: LGTM

rousseldenis (approved):

rousseldenis:
@sbidoul

sbidoul:
name reserved

PR #2618, opened 2026-08-06
[19.0][MIG] stock_location_empty: Migration to 19.0

apami0111:
Migración de "stock_location_empty" de 18.0 a 19.0

- Adapt "_search_location_amount" to the new "Domain" search API.
- Fix a test assertion
- Create test products instead of relying on `product` demo data
- Automated tests pass.
- Manually verified "Empty stock" menu and filter in the UI.


BhaveshHeliconia (approved):
Code review LGTM!

stock_location_empty/readme/CREDITS.md
celm1990, line 1: Please remove this file during the migration, according to the migration guide.

Image

celm1990 (changes requested):
I think this module is no longer necessary from v18 onward.

In this PR https://github.com/odoo/odoo/pull/155215, Odoo added a similar feature (without a dedicated menu) in the search view. With this module installed, there are two similar filters, which could confuse users about which one to use:

Image

stock_location_empty/models/stock_location.py
```diff
@@ -0,0 +1,48 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+from odoo.fields import Domain
+from odoo.tools import SQL
+
+
+class StockLocation(models.Model):
+    _inherit = "stock.location"
+
+    stock_amount = fields.Float(
```
rousseldenis, line 11: @celm1990 IMHO, the only thing that could be kept is this field.

BUT, IMHO, this could be inconsistent and has no valuable information if there are mixed products with different UOMs...

PR #2619, opened 2026-08-06
[FIX] stock_vertical_lift: apply the count even when it matches

divad1196:
The Kardex shuttle skipped _apply_inventory() when the counted quantity equalled the on-hand one
-> the scheduled date survived
-> the count left no trace

=> Always apply the result, like Odoo's Physical Inventory does.


divad1196:
@jbaudoux Short change, can you have a look?

TDu:
@divad1196  Could you check the pre-commit and tests error

TDu (approved):

PR #2621, opened 2026-08-17
[ADD] stock_package_autoload

SirPyTech:
> This module allows to automatically add the content of a package to each Transfer line.

We already proposed this module in `14.0` (https://github.com/OCA/stock-logistics-warehouse/pull/2027) and `16.0` (https://github.com/OCA/stock-logistics-warehouse/pull/2218).

This PR supersedes the migration https://github.com/OCA/stock-logistics-warehouse/pull/2495 to resolve https://github.com/OCA/stock-logistics-warehouse/pull/2495#pullrequestreview-4160605306.


stock_package_autoload/models/stock_move.py
```diff
@@ -0,0 +1,54 @@
...
+        The same serials cannot be selected more than once.
+        """
+        self.ensure_one()
+        # Lines added in the view do not have a Lot,
+        # but they have a quant to avoid duplicates.
+        # See `stock.move.line.quant_id`
+        view_lots = self.move_line_ids.quant_id.lot_id
+        current_lots = self.move_line_ids.mapped("lot_id") | view_lots
+        product_quants = self.load_products_from_package_id.quant_ids.filtered(
+            lambda q, lots=current_lots: q.lot_id not in lots
```
HekkiMelody: **issue**: after a test from the interface, it looks like packages can contain various products. So here we should also filter on `q.product_id == self.product_id`

HekkiMelody (changes requested):
Discovered a small issue during functional testing

stock_package_autoload/models/stock_move.py
SirPyTech: Thanks for catching this! It should be fixed now, please check

HekkiMelody (approved):
Code and functional review, LGTM

anusriNPS (commented):
**Praise:** PR is handled with the best approach in this version. 
Functionality of package contents addition works well when selecting package and also duplication of lot number within the package if already selected in move line is not added back.

Minor comments:
1. We can add version 18.0 in the title of PR for readability.
2. When a lot number is selected in stock move line with a quantity 1.0 from other package/lot, then if we try to add package contents later it adds the entire package quantity instead of adding only the necessary quantities apart from already selected one. However, if we try to select manually from package content it adds only necessary quantities.  Should we need to consider this scenario? Please let me know your views.

SirPyTech:
> **Praise:** PR is handled with the best approach in this version. Functionality of package contents addition works well when selecting package and also duplication of lot number within the package if already selected in move line is not added back.

:pray: 

> Minor comments:
> 
>   1. We can add version 18.0 in the title of PR for readability.

The version is already in the branch and in a label, I think that would be redundant

>   2. When a lot number is selected in stock move line with a quantity 1.0 from other package/lot, then if we try to add package contents later it adds the entire package quantity instead of adding only the necessary quantities apart from already selected one. However, if we try to select manually from package content it adds only necessary quantities.  Should we need to consider this scenario? Please let me know your views.

Good catch, I have worked on this in the last commit, please check

anusriNPS:
> > **Praise:** PR is handled with the best approach in this version. Functionality of package contents addition works well when selecting package and also duplication of lot number within the package if already selected in move line is not added back.
> 
> 🙏
> 
> > Minor comments:
> > 
> > 1. We can add version 18.0 in the title of PR for readability.
> 
> The version is already in the branch and in a label, I think that would be redundant
> 
> > 2. When a lot number is selected in stock move line with a quantity 1.0 from other package/lot, then if we try to add package contents later it adds the entire package quantity instead of adding only the necessary quantities apart from already selected one. However, if we try to select manually from package content it adds only necessary quantities.  Should we need to consider this scenario? Please let me know your views.
> 
> Good catch, I have worked on this in the last commit, please check

Thank you for working on it. Products which uses lots when loaded from package uses correct quantities when other quantities for same product if any already selected. Verified latest changes which works as expected. Nice work as testcase addition is also taken care for the same.

anusriNPS (approved):
Code and Functional Review: LGTM

HekkiMelody:
@OCA/logistics-maintainers Hello, this PR is approved with two reviews, could you please have a look? Thanks!

PR #2623, opened 2026-08-18
[18.0][FIX] stock_secondary_unit: don't copy the secondary qty on partial returns

CarlosRoca13:
`secondary_uom_qty` is a stored writable computed field, so Odoo copies it. The return wizard creates the return move with `move.copy(vals)` forcing a different `product_uom_qty`, and as both quantities are then given in the create values, neither compute method is triggered. The return move ends up keeping the secondary quantity of the whole original move.

Drop the copied value whenever the caller overrides `product_uom_qty` without providing a secondary quantity, so it gets computed from the new quantity. Independent secondary units are left untouched, as they aren't derived from the quantity.

cc @Tecnativa TT64233

ping @victoralmau @sergio-teruel


sergio-teruel (approved):
LGTM!

pedrobaeza (commented):
Why not putting attribute `copy=False`?

CarlosRoca13:
I've done it this way because, in the case where a secondary unit is of type "Independent", I'm not sure what the best way to handle it would be... I think transferring it is fine, but I also think leaving it at 0 is fine. If you think it's better to leave it at 0, I'll change it.

pedrobaeza:
I think it's not correct to leave the previous data, as you may change the returned quantity in the wizard, so the old value in secondary qty will be anyway incorrect. You may want to compute the proportional quantity though.

CarlosRoca13:
Changed and added a migration script to update the values. I have grouped the ids by the quantity to be set to reduce the launched updates

stock_secondary_unit/migrations/18.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+from openupgradelib import openupgrade
+
+from odoo.tools import float_compare
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Fix the secondary quantities left inconsistent by ``copy``."""
+    precision = env["decimal.precision"].precision_get("Product Unit of Measure")
+    moves = env["stock.move"].search(
```
pedrobaeza, line 12: This is a huge computation. Reduce the scope to just returned moves.

CarlosRoca13: Done

pedrobaeza (approved):

carlos-lopez-tecnativa (commented):
I tested the module without these changes, and it works as expected, so I don't understand if I'm missing some configuration.

I tested it in v18:
http://oca-stock-logistics-warehouse-18-0-fca3a01f4d8a.runboat.odoo-community.org/odoo/receipts/20/stock.picking/stock.picking/22

And in v19:
http://oca-stock-logistics-warehouse-19-0-pr2622-ecc300392613.runboat.odoo-community.org/odoo/receipts/13/stock.picking/14

I have a receipt for 10 SU = 1,000 units. I return 50 units, and the secondary unit is 0.5, but it doesn't take the original 10 SU into account.

carlos-lopez-tecnativa (approved):
I tested this with deliveries, and it fails as described in the PR. However, it works fine with receipts, which is strange.

Anyway, LGTM.

http://oca-stock-logistics-warehouse-18-0-fca3a01f4d8a.runboat.odoo-community.org/odoo/deliveries/23/stock.picking/stock.picking/29

PR #2625, opened 2026-08-26
[17.0][MIG] stock_vlm_mgmt: Migration to 17.0

adasatorres-tecnativa:
@Tecnativa TT64277
@pedrobaeza @christian-ramos-tecnativa


stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -28,13 +28,14 @@ def _compute_has_vlm_operation(self):
...
 
-    @api.depends("qty_done")
+    @api.depends("quantity")
     def _compute_vlm_pending_quantity(self):
         """Pair the reported quantities to the done ones"""
         # TODO: Maybe get rid of these and rely on operation task once they're created
         self.vlm_pending_quantity = 0
         for line in self.filtered("has_vlm_operation"):
-            line.vlm_pending_quantity = line.qty_done
+            if line.picked:
```
pedrobaeza: I'm not sure you should add this condition, as this is only marked AFAIK when you manually change quantities, but in all changes the vlm quantities should be added.

stock_vlm_mgmt/models/vlm_tray_cell_position_mixin.py
```diff
@@ -26,12 +26,12 @@ class VlmTrayCellPositionMixin(models.AbstractModel):
         readonly=False,
     )
 
-    @api.depends("pos_x")
+    @api.depends("pos_x", "tray_id")
```
pedrobaeza: Why adding this dependency?

stock_vlm_mgmt/models/vlm_tray_cell_position_mixin.py
```diff
@@ -51,13 +51,13 @@ def _compute_pos(self):
     @api.depends("pos_x", "pos_y", "tray_type_id", "tray_id")
     def _compute_tray_matrix(self):
         self.tray_matrix = {}
-        for record in self.filtered("tray_type_id"):
+        for record in self:
```
pedrobaeza: Why removing the filtered (at least `tray_id`)?

adasatorres-tecnativa: I made this change while looking into an error I had encountered during the migration. In the Vertical Lift Module Quants view, it was possible to create new records without them coming from a tray. Since the filtered field was being set in the computed method, it didn’t create a default structure for the matrix, which caused an error when rendering the widget.

In the end, though, I decided to disable the option to create records directly from that view, since it didn’t make much sense to do so.

I’ll add it back, since there’s no need to remove it anymore.

stock_vlm_mgmt/models/stock_move_line.py
adasatorres-tecnativa: The picked field is computed based on the state of the stock.move.line. This line was added because it was also added in 18.0. I assume it was added to ensure that stock.move.line tasks are only created for records that are done.

However, I’ve been testing it and I can’t find any case where it’s actually necessary either.

stock_vlm_mgmt/models/vlm_tray_cell_position_mixin.py
adasatorres-tecnativa: Sorry, I was testing a few things to see how the matrix could be updated when changing the tray_id, and I forgot to remove that dependency.

stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -28,13 +28,13 @@ def _compute_has_vlm_operation(self):
...
 
-    @api.depends("qty_done")
+    @api.depends("quantity")
     def _compute_vlm_pending_quantity(self):
         """Pair the reported quantities to the done ones"""
         # TODO: Maybe get rid of these and rely on operation task once they're created
         self.vlm_pending_quantity = 0
         for line in self.filtered("has_vlm_operation"):
-            line.vlm_pending_quantity = line.qty_done
+            line.vlm_pending_quantity = line.quantity
```
christian-ramos-tecnativa, line 37: After the change in stock.move.line: 
`qty_done != quantity`
`qty_done == quantity + picked`
You have an example in 18.0 https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_vlm_mgmt/models/stock_move_line.py#L32-L39

stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -52,7 +52,7 @@ def _prepare_vlm_put_task(self, key):
         location_trays = location.vlm_tray_ids.filtered(lambda x: not x.is_full)
         # This is our demand for this product we'll try to fulfill it in one shot.
         # Anyway we could use other trays if one gets full
-        quantity_fulfilled = sum(self.mapped("qty_done"))
+        quantity_fulfilled = sum(self.mapped("quantity"))
```
christian-ramos-tecnativa: Same here https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_vlm_mgmt/models/stock_move_line.py#L57

christian-ramos-tecnativa (changes requested):
Check how the quantity is handle in 18.0 as it should be the same behavior here

stock_vlm_mgmt/models/stock_move_line.py
adasatorres-tecnativa: @christian-ramos-tecnativa, If we need to add picked, I understand that this is simply due to how stock.move.line currently works. I don't think adding picked will affect the addon's logic, but I understand the reason for including it.

That said, wouldn't it be better to add it to the filtered using a lambda? That way, we can minimize the number of SMLs that the loop has to iterate over.

```suggestion
 for line in self.filtered(lambda sml: sml.has_vlm_operation and sml.picked):
         line.vlm_pending_quantity = line.quantity
```

christian-ramos-tecnativa: Yes, you could add it directly to the filtered I also think is better

stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -88,7 +88,7 @@ def _prepare_vlm_get_tasks(self, key):
         # TODO: Sort them according to their strategy
         # This is our demand for this product we'll try to fulfill it in one shot.
         # Anyway we could use other trays if one gets full
-        pending_quantity = sum(self.mapped("qty_done"))
+        pending_quantity = sum(self.mapped("quantity"))
```
christian-ramos-tecnativa: This needs to be adapted too

stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -129,7 +129,7 @@ def _action_done(self):
             return res
         sml = self.exists()
         # 1. Prepare put operations
-        put_in_vlm = sml.filtered(lambda x: x.qty_done and x.location_dest_id.is_vlm)
+        put_in_vlm = sml.filtered(lambda x: x.quantity and x.location_dest_id.is_vlm)
```
christian-ramos-tecnativa: This too

stock_vlm_mgmt/models/stock_move_line.py
```diff
@@ -140,7 +140,7 @@ def _action_done(self):
             put_task_values.append(lines._prepare_vlm_put_task(key))
         self.env["stock.vlm.task"].create(put_task_values)
         # 2. Prepare get operations
-        get_from_vlm = sml.filtered(lambda x: x.qty_done and x.location_id.is_vlm)
+        get_from_vlm = sml.filtered(lambda x: x.quantity and x.location_id.is_vlm)
```
christian-ramos-tecnativa: This too

christian-ramos-tecnativa (changes requested):

christian-ramos-tecnativa (approved):

pedrobaeza (approved):
If you have assured about the `picked` thing, then let's continue.

/ocabot merge nobump

PR #2632, opened 2026-08-27
[FIX] stock_vertical_lift: propose the most urgent pick operation first

divad1196:
`select_next_move_line` sorted `stock.move.line` records with the `_order` of `vertical.lift.operation.pick`.  
Sorting a model with another model's `_order` is a latent bug.

Order on `picking_id` instead, which sorts by the transfer's priority and scheduled date.
Skipped lines still come first, and both terms can be turned off per company.

### DIFF
--- a/stock_vertical_lift/models/__init__.py
+++ b/stock_vertical_lift/models/__init__.py
@@ -8,4 +8,5 @@ from . import stock_move
 from . import stock_move_line
 from . import stock_quant
 from . import vertical_lift_command
+from . import res_company
 from . import res_config_settings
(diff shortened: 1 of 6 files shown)


TDu (approved):
LG, the new parameters could be mentioned somewhere in the readme files

simahawk (approved):

divad1196:
@TDu ~~I just realized that the "skipped" settings probably makes no sense
Without this option, the skip button might not have any impact anymore~~

~~I will check that~~ All good in fact

divad1196:
The `Pick skipped lines last` work.

## Worth mentionning

When we process a line, the `current_move_line_id` is not changed
&Rarr; It restart the evaluation from the start instead of picking the next one in line.

It only happens when:
- a new line is added
- all lines have been skipped once

It's not really a bug, it's more of an uncontrolled behavior.

BhaveshHeliconia (approved):

PR #2633, opened 2026-09-01
[18.0][FIX] stock_vlm_mgmt: duplicate key conflicts when registering the widget.

adasatorres-tecnativa:
@Tecnativa TT57303
@christian-ramos-tecnativa @pedrobaeza

The widget registration key has been updated, as it was conflicting with the widget created by `stock_location_tray`. I checked and confirmed that `stock_location_tray` has existed since older versions, so I believe we should use a different key to register the widget in this addon.

This issue was already mentioned in the migration PR #2474


pedrobaeza (approved):

christian-ramos-tecnativa (approved):
Thanks! 😄

PR #2640, opened 2026-09-07
[18.0][FIX] stock_secondary_unit: Keep the count exact across the whole route 

carlosdauden:
Partner sells fish by weight but some orders are placed by piece
count instead (e.g. "40 pieces" for 40 dinner guests) - the count must
survive exactly through every hop of the route and every backorder,
never be recomputed from the weight actually processed.

Adds a new "secondary_priority" dependency_type (the weight is still
estimated from the count on entry, like "dependent", but the count is
never itself recomputed back from the weight, like "independent") and
makes every quantity-touching path respect it for both this type and
the existing "independent" one, gated behind a shared
COUNT_PRESERVING_DEPENDENCY_TYPES constant:

- `_prepare_move_split_vals()`: the backorder created on a partial
  validation gets exactly demand-minus-done pieces, never a
  weight-proportional estimate; the "done" move's own count is synced
  to what was actually processed instead of staying at the stale
  original demand. Uses `_write()` + `invalidate_recordset()` to avoid
  retriggering the primary-quantity compute and clobbering the just-set
  real weight with a factor-based guess.
- `StockMoveLine._compute_quantity()`: once a line's count has been
  measured, a later write to a dependency no longer overwrites it -
  "secondary_priority" needs this same protection "independent" already
  had, since on a move LINE the target field is what was actually
  counted, not a demand estimate.
- `StockRule._push_prepare_move_copy_values()` (new): a route mixing
  pull and push rules (e.g. pick_ship) creates its second leg via
  `copy()` only after the first leg validates - core copies the DONE
  quantity forward, never the demand, so the pushed move must do the
  same for the secondary unit instead of losing the count to
  `copy=False` and falling back to a weight-derived guess.
- `_merge_moves_fields()`: sums the count across merging moves (e.g. a
  backorder's pushed leg merging into an existing move) instead of
  silently keeping whichever move happened to survive the merge.
- `secondary_uom_qty_done` (new stored field): core has both a demand
  (`product_uom_qty`) and a done aggregate (`quantity`, computed from
  `move_line_ids.quantity`) on stock.move: the secondary unit only ever
  had the demand half. Mirrors the done half the same way, so the
  above no longer need to re-derive "how many pieces were actually
  done" inline every time.

Tests cover the backorder split for both dependency types, the
move-line protection, and the pick_ship push/merge scenario with and
without a backorder.

Depends on:
 - https://github.com/OCA/product-attribute/pull/2398

@Tecnativa

ping @sergio-teruel @carlos-lopez-tecnativa @CarlosRoca13


carlos-lopez-tecnativa (changes requested):
Please add the external PR to test-requirements.txt in a temporary commit so it can be tested correctly on Runboat.

carlosdauden:
Changes done

stock_secondary_unit/models/stock_move.py
```diff
@@ -19,22 +26,191 @@ class StockMove(models.Model):
...
+    # Core has both a demand (product_uom_qty) and a done aggregate
+    # (quantity, computed from move_line_ids.quantity) on stock.move. The
+    # secondary unit only ever got the demand half - this mirrors the done
+    # half, the same way quantity does, so code that needs "how many pieces
+    # were actually done" doesn't have to re-derive it ad hoc every time.
+    secondary_uom_qty_done = fields.Float(
+        compute="_compute_secondary_uom_qty_done",
+        store=True,
+        digits="Product Unit of Measure",
+    )
```
carlos-lopez-tecnativa, line 38: I think this new field, being stored in a transactional model such as stock.move, should have a migration script to create and compute its value using SQL or a similar approach. In large databases, this recomputation can take a lot of time.

carlosdauden: Nice catch.
Changes done

CarlosRoca13 (approved):

carlos-lopez-tecnativa (changes requested):
I'm not sure about this change. Let me give you an example; I hope it's not too crazy, but I think it could occur in the real world.

Imagine I buy **2 secondary units = 20 units.**

I receive 1 secondary unit = 10 units, and a backorder is created for the remaining 10 units.

The next step is created correctly: 1 secondary unit = 10 units.

However, the backorder still shows **2 secondary units = 10 units**, which seems incorrect to me.

But if I don't process the next step and instead validate the backorder first (imagine my vendor finally has the full stock and sends it before I process the next picking):

When I validate the backorder, which has 2 secondary units = 10 units, the next picking now has **3 secondary units = 20 units**, whereas it initially had 2 secondary units.

Could you please check this and let me know if this is the expected behavior?

https://github.com/user-attachments/assets/7c1aa1f5-cc3e-400b-931b-f1c4fb85c89e

carlosdauden:
Changes done

carlos-lopez-tecnativa (approved):

PR #2641, opened 2026-09-08
[19.0][FIX] stock_picking_type_user_restriction: imply base.group_user

alan196:
A user having **only** the `User: Assigned Operation Types Only` group is not an internal user (`share=True`): the group does not imply `base.group_user`, unlike `stock.group_stock_user`. Such a user cannot read `ir.ui.menu` at all, so the Inventory app never shows up even though `stock.menu_stock_root` is granted to the group.

This PR makes the group imply `base.group_user` and adds a test checking that a user with only this group is internal and sees the Inventory root menu.

Found on a 19.0 production database where users get this group through `base_user_role` roles without any other group.

Assisted-by: Claude Fable 5.1

### DIFF
--- a/stock_picking_type_user_restriction/__manifest__.py
+++ b/stock_picking_type_user_restriction/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Stock picking type - Restrict users",
     "summary": "Restrict some users to see and use only certain picking types",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
     "category": "Warehouse",
     "website": "https://github.com/OCA/stock-logistics-warehouse",
     "author": "Xtendoo, TAKOBI, Odoo Community Association (OCA)",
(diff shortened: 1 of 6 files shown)


stock_picking_type_user_restriction/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Stock picking type - Restrict users",
     "summary": "Restrict some users to see and use only certain picking types",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
BhaveshHeliconia: No need to bump version manually

BhaveshHeliconia:
You shouldn't add the AI agent as a co-author. It should be mentioned as assisted by AI according to the OCA AI policy: https://github.com/OCA/.github/blob/master/AI_POLICY.md

alan196:
Thanks for the review. I've replaced the `Co-authored-by` trailer with `Assisted-by` per the OCA AI policy and reverted the manual version bump.

BhaveshHeliconia (approved):

HectorMerazChavez (approved):

PR #2642, opened 2026-09-09
[18.0][FIX] stock_vlm_mgmt: Fix widget not recalculating when field value changes.

adasatorres-tecnativa:
@Tecnativa TT64419
@pedrobaeza @christian-ramos-tecnativa 

The matrix generated by the widget is not recomputed when the data used to build it changes.
As a result, the matrix is only updated after the DOM is reloaded.
This PR fixes the issue by recomputing the matrix when its underlying data changes.

PR #2643, opened 2026-09-09
[19.0][MIG] stock_vlm_mgmt: Migration to 19.0.

adasatorres-tecnativa:
@Tecnativa TT64419
@pedrobaeza @christian-ramos-tecnativa 

Standard migration.
include #2642.


stock_vlm_mgmt/models/stock_location_vlm_tray.py
```diff
@@ -50,7 +50,7 @@ def action_tray_content(self, pos_x=None, pos_y=None):
         self.ensure_one()
         domain = [("tray_id", "=", self.id)]
```
pedrobaeza, line 51: Put everything as `Domain`:

```suggestion
        domain = Domain([("tray_id", "=", self.id)])
```

and so on with the rest.

pedrobaeza (approved):
Code review

stock_vlm_mgmt/models/stock_location_vlm_tray.py
christian-ramos-tecnativa: This can also be defined like `Domain("tray_id", "=", self.id)`

christian-ramos-tecnativa (approved):

